PR #54, opened 2019-04-05
Ux-408 Date picker component

allison-c:
### 🛠 Purpose

Date picker design doc: https://paper.dropbox.com/doc/Date-picker-and-Date-range-picker--AexDn0aJ2EN4Ia_x5G46DRewAg-sXyXSaCA60H08bFCSahsL

#### Next step:

- [ ] Adding short cut panel is the next step, this date picker is functional right now
- [x] Add tests

### 🖥 Screenshots
gif:
Jietu20190610-160751-HD-min

with locale: 
image

calendar itself:
image


allison-c:
> `react-date` from airbnb use css in js, so we wouldn't need to import `css` into our code.

They use `react-with-styles-interface-css` interface. If we want to use the same interface, we need to rebuild the CSS files for the calendar. Currently, I'm using the way they recommend in their doc,I feel the overrides are easier to understand.

packages/DatePicker/src/DatePicker.js
```diff
@@ -0,0 +1,166 @@
...
+import DateInput from "./components/DateInput/DateInput";
+
+import { CalendarStyles } from "./DatePicker.styles";
+import { getExtendedInputProps } from "./helpers";
+
+const propTypes = {
+  children: PropTypes.node,
+
+  /** Selected date in moment object. */
+  date: momentPropTypes.momentObj,
```
nahumzs: does this need to be a moment? cannot be a string? could you received a string and then parse with the format? so the consumer doesn't need to install moment in their side?

packages/DatePicker/src/components/DateInput/DateInput.js
```diff
@@ -0,0 +1,32 @@
+import PropTypes from "prop-types";
+
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+
+const propTypes = {
+  /** a11yText on the input. */
+  a11yText: PropTypes.string,
```
nahumzs: why do we need this @allison-c  couldn't we just extract the value from the input?  cc:
@mikrotron

packages/DatePicker/src/helpers.js
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+export function getExtendedInputProps(children) {
```
nahumzs: I'm calling this `extractChildren` in the SidePanel and I think Alex suggested the same name in the ListBox PR so lets keep the same name?

packages/Icon/src/svg/Arrow-Left.svg
```diff
@@ -0,0 +1,17 @@
...
+    <desc>Created with Sketch.</desc>
+    <defs>
+        <path d="M10.5352592,17.313764 C11.3557957,18.1341953 12.6860293,18.1341953 13.5065658,17.313764 C16.4022373,14.4179878 22.5209125,8.29941846 22.5209125,8.29941846 L20.2924062,6.0709125 L12.020965,14.3424577 L3.74941876,6.0709125 L1.5209125,8.29941846 C1.5209125,8.29941846 7.63958768,14.4179878 10.5352592,17.313764 L10.5352592,17.313764 Z" id="path-1"></path>
+    </defs>
+    <g id="Symbols" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Arrow-Left">
+            <mask id="mask-2" fill="white">
+                <use xlink:href="#path-1"></use>
+            </mask>
+            <use id="🌈-Color" fill="#717171" transform="translate(12.020912, 12.000000) rotate(-270.000000) translate(-12.020912, -12.000000) " xlink:href="#path-1"></use>
```
nahumzs: 🌈-color?

nahumzs:
Please include at least basic test for Cypress and React-test-Library 🙏

packages/Icon/src/svg/Arrow-Left.svg
allison-c: Not sure why, but all icons' svg in `src` folder have this emoji id..

packages/DatePicker/src/DatePicker.js
allison-c: Sounds good! Does it mean we'll also output `string` instead of `momentObj` in `onChange` callback? or maybe output both?

nahumzs: yes! lets response with a string. I think they could parse to moment again, the consumer is passing the format from the beginning, so they might be able to do it by themselves without our help.

packages/Icon/src/svg/Arrow-Left.svg
nahumzs: 🤔 @mikrotron do you know something about this?

packages/DatePicker/src/DatePicker.js
alexzherdev: `react-dates` expects to be passed a moment object though. So we'll need to parse the date into moment, do we know how to do this correctly? (i.e. accounting for locale/format)

packages/DatePicker/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@paprika/popover": "^0.1.0",
+    "@paprika/stylers": "^0.1.0",
+    "@paprika/tokens": "^0.1.0",
+    "aphrodite": "^2.3.1",
+    "moment": "^2.24.0",
+    "prop-types": "^15.7.2",
+    "react-dates": "^20.1.0",
+    "react-moment-proptypes": "^1.6.0",
+    "react-with-styles": "^3.2.1",
+    "react-with-styles-interface-aphrodite": "^5.0.1"
```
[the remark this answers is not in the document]
allison-c: oh, thanks! I forgot to remove them

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -0,0 +1,123 @@
+import React from "react";
+import PropTypes from "prop-types";
+import moment from "moment";
+import momentPropTypes from "react-moment-proptypes";
+
+import "react-dates/initialize";
+import { DayPickerSingleDateController as SDPController } from "react-dates";
+import "react-dates/lib/css/_datepicker.css";
```
alexzherdev: @nahumzs what do we do about the CSS dependency?

packages/DatePicker/DatePicker.cypress.js
```diff
@@ -0,0 +1,49 @@
+import moment from "moment";
+
+describe("<DatePicker />", () => {
+  const today = moment();
+  const targetDate = moment(today)
+    .add(1, "month")
+    .date(1);
+  it("should display calendar after clicking input", () => {
+    cy.visitStorybook("datepicker-cypress--datepicker-test");
+    cy.get('[data-qa-anchor="datepicker-input"]').should("not.have.value");
```
allison-c: I saw there's a helper function in side panel PR, will replace it later.

packages/DatePicker/src/DatePicker.js
allison-c: @alexzherdev I think @nahumzs 's idea is based on the format passed in or by the default format.

@nahumzs Sometimes the date format to display is different than the format used in the data flow. For example, the server requires date string in `YYYY-MM-DD` format, but UI requirement is displaying `DD/MM/YYYY` in the input. As the developer, I'll pass in `DD/MM/YYYY` in `format` prop like this: `<DatePicker format="DD/MM/YYYY">`. If the output in `onChange` function is string, I have to convert it to moment obj, and convert it back to server format string, that's two steps. So I still prefer return 2 types of date data in the `onChange` function.

packages/DatePicker/src/components/Calendar/Calendar.js
allison-c: If we want to get rid of any css files, I can move this 3rd party code into our `styles.js`

packages/DatePicker/src/DatePicker.js
nahumzs: One of the problem when using `momentPropTypes.momentObj` is that will yield is an invalid prop value if you pass `null` or `""` which is really annoying because something you can't control that. That's why I'm asking if it's possible to find a different way to avoid using momentPropTypes.momentObj at all. 

How would you mark that you want a `<DatePicker />` without a date?

nahumzs: I talked with @allison-c let's keep moment propTypes even if it's annoying that you can pass `data=""` as an empty string.

packages/DatePicker/src/components/Calendar/Calendar.js
nahumzs: we are not supporting it, let's move the css into a styles.js file. Let's keep the library 100% javascript friendly.

packages/DatePicker/src/components/DateInput/DateInput.js
allison-c: It's nice to provide the ability of customized aria-label.

packages/Icon/src/svg/Arrow-Left.svg
allison-c: How about we keep all of them for now, and clean up SVG files in a separate pr?

packages/DatePicker/src/components/DateInput/DateInput.js
mikrotron: I'm not sure it's intended to be used since this comonent isn't fully implemented yet, but since it will render an `<input>` and all inputs need a label, either a `<label>` (which we'd have to visually hide) or an `aria-label` attribute (which is simpler), this `a11yText` prop seems like an easy way to handle that.

packages/Icon/src/svg/Arrow-Left.svg
mikrotron: This looks like @pxsmith's handiwork.  It's coming from the exported Sketch files anyway.

packages/DatePicker/src/components/Calendar/Calendar.styles.js
```diff
@@ -0,0 +1,141 @@
...
+    top: 46px;
+
+    small {
+      ${stylers.fontSize(-1)};
+    }
+  }
+
+  .CalendarMonth_caption {
+    background-color: ${tokens.color.blackLighten80};
+    height: ${getNumber(tokens.space) * 4 + getNumber(tokens.space)}px;
```
mikrotron: Why not use `parseInt()`? 
Better yet, why not use `stylers.spacer(5)`?

packages/DatePicker/src/components/Calendar/Calendar.styles.js
```diff
@@ -0,0 +1,141 @@
...
+  .DayPickerNavigation_button:focus,
+  .DayPickerNavigation_button:active {
+    box-shadow: ${tokens.highlight.active.withBorder.boxShadow};
+    outline: none;
+  }
+
+  .CalendarMonth_table {
+    border: 0;
+    margin-bottom: 0;
+    margin-top: ${getNumber(tokens.space) * 4}px;
```
mikrotron: `stylers.spacer(4)`

packages/DatePicker/src/components/Calendar/Calendar.styles.js
```diff
@@ -0,0 +1,141 @@
...
+      font-weight: bold;
+    }
+
+    ${props.isSelected ? SelectedStyle : ""}
+    ${props.isToday ? TodayStyle : ""}
+  `;
+};
+
+export const CalendarHeaderStyled = styled.span`
+  font-size: ${stylers.spacer(2)};
```
mikrotron: 👎 `stylers.fontSize()`, please.

packages/DatePicker/src/components/Calendar/Calendar.styles.js
allison-c: I try to make the calculation more clear, but looks like should be ok just use `spacer(5)`, will remove `getNumber`!

packages/DatePicker/src/components/DateInput/DateInput.js
allison-c: `a11yText` will be passed into `<Input />` component. I can actually remove it here, and if developers still want it, they can still pass it into `<DatePicker.Input />`, it will still work.

mikrotron:
I have some trouble getting this to work as I would expect with only the keyboard.  Also, I'm not sure the formatting is being applied as described in the Pitch... shouldn't it show a `mm/dd/yyyy` format while in focus, but after blurring, apply formatting so it could look like `mmmm dd, yyyy`?

packages/DatePicker/stories/DatePicker.stories.js
```diff
@@ -0,0 +1,21 @@
...
+import { storiesOf } from "@storybook/react";
+import L10n from "@paprika/l10n";
+import Example from "./DatePickerExample";
+import DatePicker from "../src/DatePicker";
+
+storiesOf("DatePicker", module)
+  .add("DatePicker", () => <Example locale="en" />)
+  .add("DatePicker with locale", () => <Example locale="ja" />)
+  .add("DatePicker with initialDate", () => <Example initialDate={moment("2019-01-01")} />)
+  .add("DatePicker with DatePicker.Input", () => (
```
mikrotron: "DatePicker with format and DatePicker.Input" ?
(or maybe make two separate stories 🤷‍♂️ not a big deal for now though).

allison-c:
> I have some trouble getting this to work as I would expect with only the keyboard. Also, I'm not sure the formatting is being applied as described in the Pitch... shouldn't it show a `mm/dd/yyyy` format while in focus, but after blurring, apply formatting so it could look like `mmmm dd, yyyy`?

Thank you! I missed `enter` and `calendar should respond to typing in real time`. Will add `enter` in this PR. Predicting part is a little bit complex, I'd like to talk to designers see if they want the same behaviors as Google calendar(which doesn't error out anything).
I think we only formatting like `mmmm dd, yyyy` if the user typed somthing, if the user clicked a date on the calendar, we will skip formatting step, let me talk to designers as well.
Did you try only with keyboard? You should be able to type in though, or did you mean something else?

packages/DatePicker/stories/DatePicker.stories.js
allison-c: I can create a story and put format into knobs!

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -0,0 +1,134 @@
...
+        hideKeyboardShortcutsPanel
+        daySize={30}
+        verticalBorderSpacing={0}
+        transitionDuration={0}
+        horizontalMonthPadding={0}
+        navPrev={renderArrowLeft()}
+        navNext={renderArrowRight()}
+        onPrevMonthClick={handleClickPrevMonth}
+        onNextMonthClick={handleClickNextMonth}
+        renderDayContents={renderDayContents}
```
mikrotron: This is a pretty long list of props where alphabetization might really help to find something specific or see if something is missing.

packages/DatePicker/stories/DatePicker.stories.js
mikrotron: In the next PR, I guess?

packages/DatePicker/src/components/Calendar/Calendar.js
allison-c: Talked in person, followed the order on `react-dates` document

packages/helpers/src/dom/isElementContainsFocus.js
```diff
@@ -0,0 +1,4 @@
+export default function isElementContainsFocus(element) {
```
mikrotron: `isElementFocussed` would read better.  But it's an internal name, not affecting the API, so let's update in the next PR.

PR #59, opened 2019-04-25
Ux 460 migrate text area component

oscarkwan:
### 🛠 Purpose
---
To migrate the `<Textarea />` component from `acl-ui`

### ✏️ Notes to Reviewer
---
- created show case story
- created a11y and screener stories

- added some tests

### 🖥 Screenshots
---
Screen Shot 2019-05-16 at 1 57 27 PM


mikrotron:
@oscarkwan, can you add some description + a screenshot to the top of this PR to summarize briefly, please?

packages/Textarea/src/Textarea.js
```diff
@@ -0,0 +1,107 @@
...
+  componentDidMount() {
+    if (this.props.canExpand) {
+      this.resize();
+      window.addEventListener("resize", this.resize);
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.canExpand) {
+      if (prevProps.canExpand && this.props.value !== prevProps.value) {
```
mikrotron: I don't think we care about `prevProps.canExpand`?
So these two lines can be collapsed to:
`if (this.props.canExpand && this.props.value !== prevProps.value) {`

packages/Textarea/tests/Textarea.spec.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import { render } from "react-testing-library";
+import Textarea from "../src/Textarea";
+
+describe("Textarea", () => {
+  it("should be visible", () => {
+    const { getByText } = render(<Textarea value="placeholder value" />);
+    expect(getByText(/placeholder value/i)).toBeInTheDocument();
+  });
```
mikrotron: Is it possible to test the behaviour of the `canExpand` prop? Although, maybe it would be easier/better to add that to the Screener story test?

packages/Textarea/src/Textarea.styles.js
```diff
@@ -0,0 +1,83 @@
...
+  &.form-textarea--is-readonly .form-textarea__textarea {
+    ${stylers.readOnlyFormStyles};
+    cursor: text;
+  }
+
+  // Disabled
+
+  &.form-textarea--is-disabled, &[disabled] {
+    textarea.form-textarea__textarea {
+      ${stylers.disabledFormStyles}
```
mikrotron: I just noticed that I _think_ there's something wrong with this `include`... can you check with a designer if the `isDisabled` state should have a grey background? And double check the grey background style for the `isReadOnly` state?

packages/Textarea/src/Textarea.js
mikrotron: @alexlaj, in the `acl-ui` version there was an event listener for window resizing... we didn't think it was still necessary though.  Do you remember why we had it originally?
https://github.com/acl-services/acl-ui/blob/master/src/components/__beta__/Textarea/Textarea.js#L11

oscarkwan: I asked him already, its needed when window is resized horizontally

packages/Textarea/tests/Textarea.spec.js
oscarkwan: I'll include in screener test, its kinda hard to test here

packages/Textarea/src/Textarea.styles.js
oscarkwan: @mikrotron , talked with Sarah. she's okay with it, except that the disabled text color probably doesn't pass contrast. But for now we can leave it as is until they change it

packages/Textarea/stories/examples/Showcase.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Story } from "storybook/assets/styles/common.styles";
+import { boolean, number, select, text } from "@storybook/addon-knobs";
+import Textarea from "../../Textarea";
```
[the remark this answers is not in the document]
nahumzs: Shouldn't this have defaultValue and value, that how react handle the input component isn't?

packages/Textarea/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/textarea"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/stylers": "^0.1.0",
+    "@paprika/tokens": "^0.1.0",
+    "class-names": "^1.0.0",
```
alexzherdev: Should be `classnames`.

oscarkwan: Oh i'll fix it next PR

alexzherdev: We should turn `import/no-extraneous-dependencies` back on

PR #68, opened 2019-04-30
UX-464 🦄🐴 ✂ chop chop packages

nahumzs:
### 🛠 Purpose

removing helpers from being published. 

---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---


packages/Button/src/Button.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from "prop-types";
 import RawButton from "@paprika/raw-button";
 import RefreshIcon from "@paprika/icon/lib/Refresh";
 import DownIcon from "@paprika/icon/lib/CaretDown";
-import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import { ShirtSizes } from "../../helpers/src/customPropTypes";
```
alexzherdev: Imagine `Button.js` is installed at `projects/client/node_modules/@paprika/button/lib/Button.js`.
```js
path.join('projects/client/node_modules/@paprika/button/lib', '../../helpers/src/customPropTypes');
// => 'projects/client/node_modules/@paprika/helpers/src/customPropTypes'
```
Doesn't work unless you also publish and install `@paprika/helpers`.

packages/Popover/stories/Popover.cypress.stories.js
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { storiesOf } from "@storybook/react";
-import windowHandles, { Input } from "@paprika/testing-helpers/lib/windowHandles/";
+import windowHandles, { Input } from "../../testingHelpers/src/windowHandles";
```
alexzherdev: This is fine since this is only ever executed within the paprika repo, not when it's installed in an app.

PR #69, opened 2019-04-30
UX-464 🦄🐴 removing testingHelpers as package

nahumzs:
### 🛠 Purpose
removing testingHelpers as a package
---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---


nahumzs:
I don't see where testingHelper was required as a dependency I think was working because yarn bootstrap was creating the package.

Issue #62, opened 2019-04-26
*.a11y.js test are not running.


nahumzs:
# Help wanted
## Please write your question.
Currently, there couple component with a11y.js test in place but are not running because the extension should be spec.js and as well axe is not connected correctly.

Also worth to investigate if using jest-a11y is the best way to test accessibility because is not checking each time the test is interacting with the UI, maybe would be to use something like cypress axe where we can test the interaction and test that accessibility is maintained correctly. cc @oscarkwan 

## Additional context

https://github.com/acl-services/paprika/pull/73#issuecomment-490275068

oscarkwan:
Closing this, done here https://github.com/acl-services/paprika/pull/979

PR #73, opened 2019-05-04
UX-437 Add react-testing-library tests for RawButton 🐐

mikrotron:
### 🛠 Purpose
Add `react-testing-library` "unit" tests for `<RawButton>`.

### ✏️ Notes to Reviewer
✓ Renders with default props
✓ Renders with custom props
✓ Fires onClick callback when clicked
✓ Fires onClick callback when keypressed
✓ Does not fire onClick callback when disabled
✓ Propagates click events by default
✓ Does not propagate click events when canPropagate=false
✓ Is focussable via ref
✓ Is focussable via callback ref

### DIFF
--- a/packages/RawButton/test/RawButton.spec.js
+++ b/packages/RawButton/test/RawButton.spec.js
@@ -25,10 +25,10 @@ describe("RawButton", () => {
   it("should render with default props", () => {
     const { getByText, getByRole, container } = renderComponent();
 
-    expect(getByText(/happy button/i)).toBeVisible();
-    expect(getByRole("button")).toBeVisible();
-    expect(container.querySelector('[tabindex="0"]')).toBeVisible();
-    expect(container.querySelector('[aria-disabled="false"]')).toBeVisible();
+    expect(getByText(/happy button/i)).toBeInTheDocument();
+    expect(getByRole("button")).toBeInTheDocument();
+    expect(container.querySelector('[tabindex="0"]')).toBeInTheDocument();
+    expect(container.querySelector('[aria-disabled="false"]')).toBeInTheDocument();
   });
 
   it("should render with custom props", () => {
@@ -41,11 +41,11 @@ describe("RawButton", () => {
     };
     const { getByRole, container } = renderComponent(customProps);
 
-    expect(container.querySelector('[aria-label="include me"]')).toBeVisible();
-    expect(container.querySelector('[class*="shiny-button"]')).toBeVisible();
-    expect(container.querySelector('[aria-disabled="true"]')).toBeVisible();
-    expect(getByRole("listitem")).toBeVisible();
-    expect(container.querySelector('[tabindex="-1"]')).toBeVisible();
+    expect(container.querySelector('[aria-label="include me"]')).toBeInTheDocument();
+    expect(container.querySelector('[class*="shiny-button"]')).toBeInTheDocument();
+    expect(container.querySelector('[aria-disabled="true"]')).toBeInTheDocument();
+    expect(getByRole("listitem")).toBeInTheDocument();
+    expect(container.querySelector('[tabindex="-1"]')).toBeInTheDocument();
   });
 
   it("should fire onClick callback when clicked", () => {
--- a/packages/RawButton/test/RawButton.spec.js
+++ b/packages/RawButton/test/RawButton.spec.js
@@ -22,7 +22,7 @@ function renderComponent(props = {}) {
 }
 
 describe("RawButton", () => {
-  it("should render with default props", () => {
+  it("Renders with default props", () => {
     const { getByText, getByRole, container } = renderComponent();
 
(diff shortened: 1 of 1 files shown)


packages/RawButton/test/RawButton.spec.js
```diff
@@ -0,0 +1,128 @@
...
+      {props.children || "happy button"}
+    </RawButton>
+  );
+}
+
+describe("RawButton", () => {
+  it("should render with default props", () => {
+    const { getByText, getByRole, container } = renderComponent();
+
+    expect(getByText(/happy button/i)).toBeVisible();
```
alexzherdev: So with styled-components `toBeVisible` is actually able to handle visibility because it's CSS-in-JS right?

mikrotron: These tests pass, so... yes?
I'm not really sure that `toBeVisible()` is the right assertion here.  Really, I want to know that it exists in the DOM... is there a different function I should use instead?

alexzherdev: `toBeInTheDocument`.
That was my question here: `toBeVisible` looks at `visibility: hidden` etc. to check actual visibility. This usually doesn't work with SCSS since we mock those styles out in tests. But with CSS-in-JS it sounds like it should work out

nahumzs: https://github.com/testing-library/jest-dom you can see all the assertions.

oscarkwan: I also suggest to use `toBeInTheDocument`

oscarkwan:
@mikrotron How are we doing a11y testing right now?

oscarkwan:
@mikrotron Overall looks pretty good to me

packages/RawButton/test/RawButton.spec.js
```diff
@@ -0,0 +1,126 @@
...
+
+  return render(
+    <RawButton {...defaultProps} {...props}>
+      {props.children || "happy button"}
+    </RawButton>
+  );
+}
+
+describe("RawButton", () => {
+  it("should render with default props", () => {
```
alexzherdev: On the rails side we say `it('renders with default props', ...)` — I think I prefer that. Any opinions?

oscarkwan: Fine with me, pick one and stick with it 👍

packages/RawButton/test/RawButton.spec.js
```diff
@@ -0,0 +1,126 @@
...
+  it("should render with custom props", () => {
+    const customProps = {
+      a11yText: "include me",
+      isDisabled: true,
+      role: "listitem",
+      tabIndex: -1,
+    };
+    const { getByRole, container } = renderComponent(customProps);
+
+    expect(container.querySelector('[aria-label="include me"]')).toBeInTheDocument();
```
allison-c: Maybe can use `getByLabelText ` ? Their doc said it can find by `aria-label`

nahumzs:
There is a ticket to connect a11y test, the test we were doing with jest was just running only one time per render, which only deal with the initial state of the component, missing a11y issues once you start interacting with it. 

I was thinking of doing something that evaluates a11y as you interact with the component, cypress has a good one, but there is none implementation so far. :/

https://github.com/acl-services/paprika/issues/62 cc @oscarkwan

packages/RawButton/test/RawButton.spec.js
mikrotron: In general, I favour fewer words when possible, so dropping "should" in all test cases is fine with me.  Do we have consensus to make this a standard for the repo?

packages/RawButton/test/RawButton.spec.js
oscarkwan: `getByLabelText` won't find by `aria-label` by by `<label>`'s text.

Example:
```<label for="username-input">Username</label>```

Comment (mikrotron) [reply packages/RawButton/test/RawButton.spec.js]: Allison appears to be correct, it _does_ select based on the `aria-label` too! Pro tip!

**Comment by alexzherdev** [comment]
Do you mean this? https://github.com/avanslaars/cypress-axe

**Comment by alexzherdev** [comment]
@nahumzs do you mean this? https://github.com/avanslaars/cypress-axe

**Comment by nahumzs** [comment]
yeah that one @alexzherdev but have work with it, neither sure if it's the best alternative. If we could do something similar with jest I think would be a win-win.

**Comment by alexzherdev** [comment]
@nahumzs There's a chapter in testing javascript on this, not sure if we watched it. But should be fairly easy to do à la https://github.com/nickcolley/jest-axe#with-react?
```js
const { container, doSomethingElse } = render(<MyComponent />);
let results = await axe(container.innerHTML);
expect(results).toHaveNoViolations();
doSomethingElse();
results = await axe(container.innerHTML);
expect(results).toHaveNoViolations();
```

PR #75, opened 2019-05-10
UX-437 Add react-testing-library tests for Button 🐐

mikrotron:
### 🛠 Purpose
Add `react-testing-library` "unit" tests for `<Button>`.

### ✏️ Notes to Reviewer
- Very similar approach to tests for `<RawButton>` (https://github.com/acl-services/paprika/pull/73)
- Adjusted `<RawButton>` tests to better cover `isDisabled` scenario.

✓ Renders with default props
✓ Renders a span when isSemantic=false
✓ Renders with custom props
✓ Fires onClick callback when clicked
✓ Does not fire onClick callback when disabled
✓ Propagates click events by default
✓ Does not propagate click events when canPropagate=false
✓ Is focussable via ref
✓ Is focussable via callback ref

### DIFF
--- a/packages/Button/test/Button.spec.js
+++ b/packages/Button/test/Button.spec.js
@@ -1,0 +1,118 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */
+
+import React from "react";
+import { render, fireEvent } from "react-testing-library";
+import Button from "../src";
+
+const noop = () => {};
+
+function renderComponent(props = {}) {
+  const defaultProps = {
+    onClick: noop,
+  };
+
+  return render(
+    <Button {...defaultProps} {...props}>
+      {props.children || "happy button"}
+    </Button>
+  );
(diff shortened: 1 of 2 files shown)


packages/Button/test/Button.spec.js
```diff
@@ -0,0 +1,118 @@
+/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */
+
```
oscarkwan: ✂️

mikrotron: Oh, actually I need this for these parts:
https://github.com/acl-services/paprika/blob/5d4d70aea056924b7e4bd9ffbfa80527d584bdc6/packages/Button/test/Button.spec.js#L74
https://github.com/acl-services/paprika/blob/5d4d70aea056924b7e4bd9ffbfa80527d584bdc6/packages/Button/test/Button.spec.js#L88

PR #77, opened 2019-05-10
Ux 434 multiselect

nahumzs:
### 🛠 Purpose
NOTES:

I have moved all test, docs and storybook to *[UX-434-Multiselect-support]* branch [PR](https://github.com/acl-services/paprika/pull/82), that PR has a lot of line but 60% of those lines are part of a fixture for the lazy story. 

So if you want to see and experience the ListBox please check out that branch, since this one is just the `src` and package folder. 

ListBox API

The ListBox component behaves similarly to an html `<select />` tag, and as a first version is an uncontrolled component, which use a useReducer to manage its own state. 

```js
<ListBox>
  <ListBox.Option>Option 1</ListBox>
  <ListBox.Option>Option 2</ListBox>
</ListBox>
```

To observer the changes of the<ListBox /> you add the *onChange* prop `<LisBox onChange={()=>{}} />`.
The ListBox also have the following props:

- children
- isDisabled
- isMulti
- isInline 
- isOpen
- placeholder
- height

This component has also in-built components that you can use to compose it.

- `<ListBox.Option />`
- `<ListBox.RawItem />`
- `<ListBox.Trigger />`
- `<ListBox.Filter />`
- `<ListBox.Footer />`
- `<ListBox.Popover />`
- `<ListBox.Divider />`

With this component, you can modify or extend the behaviour of the ListBox as an example to display a ListBox with a filter can be composed like:

```js
<ListBox>
  <ListBox.Filter />
  <ListBox.Option>Option 1</ListBox>
  <ListBox.Option>Option 2</ListBox>
<ListBox>
```

Take in consideration that some components don't render a visual element though modify the behaviour of the component as an example:

Preventing the Popover from removing the focus from the Trigger:

```js
<ListBox>
  <ListBox.Popover shouldKeepFocus />
  <ListBox.Option>Option 1</ListBox>
  <ListBox.Option>Option 2</ListBox>
<ListBox>
```

Some component take a function as a prop.children to render custom content example of this are the <ListBox.Trigger />, <ListBox.Option /> in the future we should support as well <ListBox.Footer /> and <ListBox.Filter />.

ex for `<ListBox.Option />`

```js
<ListBox>
  {["one","two"].map(option => <ListBox.Option>{({isSelected}) => {
    return isSelected ? `✅ ${option}` : `🙅‍ ${option}` 
  }}</ListBox>)}
<ListBox>
```

In a previous iteration, I present a case for having a Group component which would automatically create sections depending on the group. 

*[deprecated]*
```js
<ListBox.Group label="some option">
  <ListBox.Option />
</ListBox.Group>  
```

I think this introduces extra complexity for maintaining and don't give us a lot of benefit from DX (developer experience) standpoint. 

Instead, I preferred to move into a more explicit direction with the `<ListBox.Divider />`:

```js
<ListBox>
  <ListBox.Divider>Group 1</ListBox.Divider>
  <ListBox.Option>Option 1</ListBox>
  <ListBox.Option>Option 2</ListBox>

  <ListBox.Divider>Group 2</ListBox.Divider>
  <ListBox.Option>Option 3</ListBox>
  <ListBox.Option>Option 4</ListBox>  
<ListBox>
```

Finally, I'm introducing a new Component name `<ListBox.RawItem />` which allowed the consumer of this component to render an option which is not selectable, this is handy for cases when you want to render buttons or links like "Load more", "go to ...", etc. 

### ✏️ Notes to Reviewer

To run test and storybook: https://github.com/acl-services/paprika/pull/82 storybook

- Screen reader is not implemented
- Tokens have not implemented the look and feel will be part of another PR, Please don't pay to much attention to those details.

#### Mini Retrospective
Things that went well:
- Simple API for complex component
- Keyboard accessible
- Extensible is possible to compose it and extend the component
- The consumer doesn't need to be aware of the component is controlled or uncontrolled components. the component will sync itself internally to reflect the props if changes, I do this via adding effect to the props. 
- While is not implemented, later we will be able to override the component reducer to incorporate custom behaviours using a Reducer Pattern [blog](https://kentcdodds.com/blog/the-state-reducer-pattern-with-react-hooks) and [youtube](https://www.youtube.com/watch?v=RENGNzZ0dIs) 

#### Things to improve:
- Complex logic stuff inside of the component
- Big PR hard to review
- Is not a completely controlled component, I'm syncing the props with the inner state when an effect occurs from the props. 
- We have so many cases for this component
- No implementation of [react-window](https://github.com/bvaughn/react-window) though wonders how will behave with the current implementation of the keyboard. 
- CSS and Style need improvement, I didn't focus too much on that, as well might worth change the approach from style component to use `css` prop, lack of using token for this PR. 

### 🖥 Screenshots

---


packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,235 @@
...
+
+  /** When true the ListBox will try to focus to the options container asap the
+  popover is open */
+  isPopoverEager: PropTypes.bool,
+
+  /** Indicates if the popover is visible */
+  isPopoverOpen: PropTypes.bool,
+
+  /** Message to be display once the filtering process doesn't find a match */
+  hasNotResultsMessage: PropTypes.node,
```
jamiek-acl: `hasNotResultsMessage` should be `hasNoResultsMessage`.
Even better, I prefer the name `noResultsMessage`, without the `has` — as that makes me think its boolean.

oscarkwan:
I think we should have a PR party for this.

.eslintrc
```diff
@@ -20,31 +20,40 @@
     "camelcase": "off",
     "class-methods-use-this": "off",
     "consistent-return": "off",
+    "import/no-cycle": "off",
```
mikrotron: Do we still need to disable this? I resolved some issues and merged them recently, this is the commit:
https://github.com/acl-services/paprika/pull/80/commits/5c4397d155cbdd17508e315665a1b80e11a09061

And a little more context:
https://github.com/acl-services/paprika/pull/80#pullrequestreview-241432722

packages/ListBox/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "name": "@paprika/listbox",
+  "version": "0.0.1",
+  "description": "",
+  "main": "lib/index.js",
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/button": "^0.1.0",
+    "@paprika/popover": "^0.1.0",
+    "@paprika/raw-button": "^0.1.0",
+    "@paprika/tokens": "^0.1.0",
```
mikrotron: You probably need `"@paprika/stylers": "^0.1.0"` too...?

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+  isOpen: PropTypes.bool,
+
+  /** Callback returning the current selection on the ListBox */
+  onChange: PropTypes.func,
+
+  /** Defaults label to display when the ListBox has not option selected */
+  placeholder: PropTypes.string,
+
+  /** Indicate which is the height for the options container */
+  height: PropTypes.number,
```
mikrotron: Is this actually more of a `maxHeight`?

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+};
+
+export const defaultProps = {
+  children: null,
+  Filter: null, // eslint-disable-line
```
mikrotron: I have never seen this kind of `defaultProp` before... what is it doing?

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+      </Content>
+    </React.Fragment>
+  );
+}
+
+ListBox.propTypes = {
+  ...propTypes,
+  children: PropTypes.node.isRequired,
+  height: PropTypes.number.isRequired,
+  placeholder: PropTypes.string.isRequired,
```
mikrotron: Why are these three repeated here? Why can't they just be `isRequired` up above?

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+};
+
+ListBox.defaultProps = {};
```
mikrotron: TODO?

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+  if (isInline) {
+    return listBox;
+  }
+
+  if (PopoverWithProps) {
+    const PopoverClone = React.cloneElement(PopoverWithProps, {
+      ...PopoverWithProps.props,
+      children: listBox,
+    });
+    return PopoverClone;
```
mikrotron: I like the naming of this better than how you were doing a similar thing with `Trigger` up above.  I think it's clearer.

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+  const ListBoxProps = {
+    children,
+    Filter,
+    Footer,
+    height,
+    placeholder,
+    Trigger,
+  };
+
+  const listBox = <ListBox {...ListBoxProps}>{children}</ListBox>;
```
mikrotron: `listBox` is a component, so I think it should have a capital first letter.  `ListBox` is already being used though, so I'm not sure what else to call it... 😬

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+  const handleEffectOptionSelected = effects.handleEffectOptionSelected(state, dispatch);
+
+  React.useEffect(handleEffectIsDisabledChange, [props.isDisabled]);
+  React.useEffect(handleEffectListBoxWidth, [state.refTriggerContainer.current]);
+  React.useEffect(handleEffectOptionSelected, [state.selectedOptions]);
+  React.useEffect(handleEffectChildren, [props.children]);
+  React.useLayoutEffect(handleEffectIsPopOverOpen, [state.isOpen]);
+  React.useLayoutEffect(handleEffectListBoxScrolled, [state.activeOption]);
+
+  const ListBoxProps = {
```
mikrotron: `listBoxProps`

packages/ListBox/src/components/Box/Box.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { BoxContainerStyled } from "./Box.styles";
+import useListBox from "../../useListBox";
+
+const propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+const defaultProps = {};
```
mikrotron: Is this really needed? Can't you just omit it?

packages/ListBox/src/components/Content/Content.js
```diff
@@ -0,0 +1,74 @@
...
+        onKeyDown={handleKeyboardKeys(state, dispatch)}
+        ref={refListBoxContainer}
+      >
+        {props.children}
+      </ContentInlineStyled>
+    );
+  }
+
+  return (
+    <Popover.Content
```
mikrotron: This can probably be DRYed up by using this technique:
https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.js#L148

packages/ListBox/src/components/Divider/Divider.js
```diff
@@ -0,0 +1,17 @@
...
+import PropTypes from "prop-types";
+import { dividerCSS } from "./Divider.styles";
+
+const propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+export default function Divider(props) {
+  return (
+    <li css={dividerCSS} aria-hidden="true">
```
mikrotron: Why `aria-hidden`? I know you aren't too concerned with AT for this PR, but I am curious about this.

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -0,0 +1,131 @@
+import React from "react";
+import PropTypes from "prop-types";
+import useListBox from "../../useListBox";
+import * as effects from "./effects";
+import { filter, applyFilter } from "./helpers";
+import { FilterContainerStyled, FilterInputStyled, FilterSearchIconStyled } from "./Filter.styles";
+
+const propTypes = {
+  filter: PropTypes.func,
+  filterExcludeSelectedOptions: PropTypes.bool,
```
mikrotron: I forget which prefixes we decided to allow... `should`? `will`?  Anyway, as a boolean flag prop I'd suggest something like `shouldExcludeSelections`.

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+const defaultProps = {
+  filter: null,
+  filterExcludeSelectedOptions: false,
+  hasSearchIcon: true,
+  onChangeFilter: null,
+  onKeyDown: null,
+  placeholder: "Filter...",
+  noResultsMessage: "Your search did not match any options.",
```
mikrotron: // TODO: use L10n

packages/ListBox/src/components/NoResults/NoResults.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+
+const NoResultsStyled = styled.div`
+  padding: 8px;
+  margin-top: -4px;
```
mikrotron: 😒

packages/ListBox/src/components/Option/Option.js
```diff
@@ -0,0 +1,80 @@
...
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+
+  /** Describe if the option is selected or not */
+  isSelected: PropTypes.bool,
+
+  /** Describe if the option is enable or not */
+  isDisabled: PropTypes.bool,
+
+  /** Describe if the option is hidden or not */
+  isHidden: PropTypes.bool,
```
mikrotron: Do we still need this? Wouldn't the option simply be filtered out from the app if it wasn't needed?

packages/ListBox/src/components/Option/Option.js
```diff
@@ -0,0 +1,80 @@
...
+
+  /** Value of your option this can be any data structure  */
+  value: PropTypes.any, // eslint-disable-line
+};
+
+const defaultProps = {
+  isDisabled: false,
+  isHidden: false,
+  // internal prop, which shouldn't be documented
+  preventDefaultOnSelect: false, // eslint-disable-line
```
mikrotron: It's not in `propTypes`... can it actually be removed now?

packages/ListBox/src/components/Option/Option.js
```diff
@@ -0,0 +1,80 @@
...
+
+  /* eslint-disable react/prop-types */
+  const isSelected = isOptionSelected(state, index);
+  const isDisabled = isDisabledState || props.isDisabled;
+  const id = state.options[index].id;
+  return (
+    <React.Fragment>
+      <OptionStyled
+        {...getA11yAttributesForOption(state.options[index].isSelected)}
+        hasPreventDefaultOnSelect={props.preventDefaultOnSelect}
```
mikrotron: ✂️ snip? I think this is why the `RawItem` was implemented?

packages/ListBox/src/components/Option/helpers/optionState.js
```diff
@@ -0,0 +1,44 @@
...
+      (child.type.componentType === "ListBox.Option" || child.type.componentType === "ListBox.RawItem")
+    ) {
+      options[index] = createOption({ index, child, groupId: child.props.groupId });
+      index += 1;
+    }
+
+    return child;
+  });
+
+  return options;
```
mikrotron: Why isn't this an array?

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -0,0 +1,131 @@
...
+    const { renderFilter, placeholder, value, onChangeFilter, filter, ...moreProps } = props;
+    if (renderFilter) {
+      return props.renderFilter(props);
+    }
+
+    return (
+      <FilterContainerStyled data-qa-anchor="list-filter">
+        {props.hasSearchIcon ? <FilterSearchIconStyled /> : null}
+        <FilterInputStyled
+          data-qa-anchor="list-filter-input"
```
mikrotron: A BEM-style `listbox-filter__input` might be a better name?

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+  } = props;
+
+  const shouldTriggerKeepFocus = PopoverWithProps && PopoverWithProps.props.shouldKeepFocus;
+
+  // IMPERATIVE API
+  const imperativeHandle = handleImperative(state, dispatch);
+  React.useImperativeHandle(ref, imperativeHandle);
+
+  // EFFECTS
+  const handleEffectChildren = effects.handleEffectChildren(props, state, dispatch);
```
rauldudu: Just a thought: shouldn't these effects be renamed to start with the `use` prefix?

eg: `useEffectChildren(props, state, dispatch, [props.children])`

Something like: https://reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook

packages/ListBox/src/effects/effects.js
```diff
@@ -0,0 +1,110 @@
...
+export const handleEffectIsPopOverOpen = (state, dispatch, shouldKeepTriggerFocus) => () => {
+  if (state.isInline) return;
+  if (!state.refListBoxContainer.current) return;
+
+  const filterInput = state.refFilterInput.current;
+  const listBoxContainer = state.refListBoxContainer.current;
+  const trigger = state.refTrigger.current;
+  if (state.isOpen) {
+    if (state.hasFilter && !shouldKeepTriggerFocus) {
+      waitForPopoverAnimation(() => filterInput.focus());
```
jamiek-acl: This is begging to be a promise
```
function waitForPopoverAnimation() {
  return new Promise(resolve => {
    setTimeout(resolve, 300);
  });
}

waitForPopoverAnimation()
  .then(filterInput.focus())
```

packages/ListBox/src/components/Filter/effects.js
```diff
@@ -0,0 +1,24 @@
+export const handleEffectValue = (props, applyFilter) => () => {
+  if (!props.value) {
+    applyFilter({ filteredOptions: [], noResultsFound: false });
+  }
+};
+
+export const handleEffectIsPopOverOpen = (state, setTextSearch) => () => {
```
mikrotron: nit: `PopOver` → `Popover`

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -0,0 +1,131 @@
...
+          },
+        });
+      }
+    });
+  };
+
+  const handleEffectValue = effects.handleEffectValue(props, applyFilter(dispatch, applyFilterType));
+  const handleEffectIsPopOverOpen = effects.handleEffectIsPopOverOpen(state, setTextSearch);
+  const handleEffectTextSearch = effects.handleEffectTextSearch(textSearch, applyFilter(dispatch, applyFilterType));
+  const handleEffectHasFilter = effects.handleEffectHasFilter(dispatch, useListBox.types.hasFilter);
```
mikrotron: Hmm, I'm not so sure this layer of indirection is valuable, I think I'd prefer to see them inline below.

packages/ListBox/src/components/Filter/effects.js
```diff
@@ -0,0 +1,24 @@
+export const handleEffectValue = (props, applyFilter) => () => {
```
mikrotron: If all we care about is `value`, wouldn't it be better to accept that as the argument instead of the entire `props` object?

packages/ListBox/src/components/Footer/Footer.js
```diff
@@ -0,0 +1,132 @@
...
+import Button from "@paprika/button";
+import PropTypes from "prop-types";
+import { FooterContainerStyled } from "./Footer.styles";
+import useListBox from "../../useListBox";
+import applyOnChange from "../../helpers/applyOnChange";
+
+const propTypes = {
+  isAcceptVisible: PropTypes.bool,
+  isCancelVisible: PropTypes.bool,
+  isClearVisible: PropTypes.bool,
```
mikrotron: I thought that as of yesterday we were adopting the single prop pattern...?

packages/ListBox/src/components/Footer/Footer.js
```diff
@@ -0,0 +1,132 @@
...
+  labelCancel: "Cancel",
+  labelClear: "Clear",
+  onClickAccept: null,
+  onClickCancel: null,
+  onClickClear: null,
+  size: "small",
+};
+
+export function FooterComponent(props, ref) {
+  const [, dispatch] = useListBox();
```
mikrotron: I forgot about this shortcut! 
My gut says it's wrong, but my brain says "shut up + deal with it".

packages/ListBox/src/components/Footer/Footer.js
```diff
@@ -0,0 +1,132 @@
...
+    kindClear,
+    size,
+  } = props;
+
+  const handleClickAccept = event => {
+    event.stopPropagation();
+
+    dispatch({
+      type: useListBox.types.accept,
+      payload: { onChangeFn: applyOnChange(props.onClickAccept, "listbox:footer:accept") },
```
mikrotron: I'm not yet sure what these string literals are needed for... but shouldn't they be defined as consts or enums somewhere?

packages/ListBox/src/components/Footer/Footer.styles.js
```diff
@@ -0,0 +1,20 @@
+import styled from "styled-components";
+
+export const FooterContainerStyled = styled.div`
+  display: flex;
+  border-top: 1px solid #d7d7d7;
+  padding: 8px;
+  justify-content: flex-start;
+  width: calc(100% - 16px);
+
+  & > div {
```
mikrotron: No me gusta. But this isn't the PR for nitpicking CSS, so we can fix it later.

packages/ListBox/src/components/Footer/Footer.js
```diff
@@ -0,0 +1,132 @@
...
+}
+
+const Footer = React.forwardRef(FooterComponent);
+
+Footer.propTypes = propTypes;
+Footer.defaultProps = defaultProps;
+
+// forwardRef component are struggling with react-docgen
+// https://github.com/pedronauck/docz/issues/334 once we move to Storybook doc we might be able to
+// remove this
```
mikrotron: Can't you just chain them?
`Footer.propTypes = FooterComponent.propTypes = propTypes;`

packages/ListBox/src/components/Label/Label.js
```diff
@@ -0,0 +1,83 @@
...
+  const [state] = useListBox();
+  const { selectedOptions, isMulti, options } = state;
+
+  if (!isMulti && selectedOptions.length && state.options[selectedOptions[0]].preventDefaultOnSelect) {
+    if (!lastKnownLabel) return placeholder;
+
+    return lastKnownLabel;
+  }
+
+  function getListboxLabelForMulti() {
```
mikrotron: Can this function take `selectedOptions` as an argument and be extracted from the `Label` component to make it more focussed on rendering? I find it disjointed to have an `if` then a `function` then another couple `ifs`.

packages/ListBox/src/components/Label/Label.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    console.log(
+      "Warning: Your Option required a label prop, <ListBox.Option label='your label' /><MyCoolContent /></ListBox.Option>"
+    );
+    return option.content;
+  }
+
+  lastKnownLabel = placeholder;
+  return lastKnownLabel;
```
mikrotron: `return lastKnownLabel || placeholder;`?

packages/ListBox/src/components/List/List.styles.js
```diff
@@ -0,0 +1,23 @@
...
+
+  li {
+    box-sizing: border-box;
+  }
+
+  overflow: auto;
+  ${props => {
+    const paddingNoResults = props.noResultsFound ? "padding: 0" : "";
+    return `
+        max-height: ${props.height}px
```
mikrotron: I knew it! It _is_ a `maxHeight`!

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -0,0 +1,234 @@
+import useListBox from "../../../useListBox";
```
mikrotron: I think I prefer the way you've organized the `Filter` folder — it's flat.  But maybe it would be annoying to have a bunch of `helpers.js` files in your editor 🤔

What about flat, but prefixed?  `filterHelpers.js`, `optionsHelpers.js`, etc?

packages/ListBox/src/components/Option/Option.js
```diff
@@ -0,0 +1,80 @@
...
+  if (!isOptionVisible(state, index)) {
+    return null;
+  }
+
+  /* eslint-disable react/prop-types */
+  const isSelected = isOptionSelected(state, index);
+  const isDisabled = isDisabledState || props.isDisabled;
+  const id = state.options[index].id;
+  return (
+    <React.Fragment>
```
alexzherdev: Don't need this fragment

packages/ListBox/src/components/Option/helpers/optionState.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import uuidv4 from "uuid/v4";
+
+export function createOption({ index, child, groupLabel = null, groupId = null }) {
```
alexzherdev: Don't need to export this

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -0,0 +1,234 @@
...
+      if ((filteredOptions.length && key - 1 < 0) || key - 1 < 0) {
+        keepIterating = false;
+        return null;
+      }
+
+      key--;
+    }
+
+    if (isOptionVisible(state, key) && !isDisabled(state, key)) {
+      index = key;
```
mikrotron: Can you just `return key` here, then eliminate the return on line 99 (as well as the `keepIterating` variable)?

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -0,0 +1,234 @@
...
+export function getNextOptionActiveIndex(state, isAscending = true) {
+  const { activeOption, filteredOptions, options } = state;
+  const optionsKeys = Object.keys(state.options);
+
+  if (state.noResultsFound) return null;
+
+  if (state.activeOption === null) return 0;
+
+  if (optionsKeys.length === 1) {
+    return 0;
```
mikrotron: I prefer the more compact single line style you've used above...

packages/ListBox/src/components/Label/Label.js
```diff
@@ -0,0 +1,83 @@
...
+        const option = options[index];
+        if (typeof option.content === "string") {
+          return option.content;
+        }
+
+        if (option.label) {
+          return option.label;
+        }
+
+        throw Error(
```
alexzherdev:
```suggestion
        throw new Error(
```
Is supposed to work either way, but I think `new` is better to be explicit

packages/ListBox/src/components/Option/helpers/optionState.js
```diff
@@ -0,0 +1,44 @@
...
+    isSelected,
+    label: label || child.props.children, // we will try to extract the label from the children if doesn't have label
+    onClick,
+    preventDefaultOnSelect,
+    value: value || undefined,
+  };
+}
+
+export function getDataOptions(children) {
+  if (!children) throw Error("Listbox.Options is a required prop, please check you are passing correctly the data");
```
alexzherdev:
```suggestion
  if (!children) throw new Error("Listbox.Options is a required prop, please check you are passing correctly the data");
```

packages/ListBox/src/store/reducer.js
```diff
@@ -0,0 +1,236 @@
...
+      return {
+        ...state,
+        options: payload,
+        selectedOptions,
+      };
+    }
+
+    case useListBox.types.unselectOptions: {
+      if (!Array.isArray(payload)) {
+        throw Error("unselectOptions action expect an array as a payload");
```
alexzherdev:
```suggestion
        throw new Error("unselectOptions action expect an array as a payload");
```

packages/ListBox/src/store/reducer.js
```diff
@@ -0,0 +1,236 @@
...
+
+      return {
+        ...state,
+        selectedOptions: state.selectedOptions.filter(index => !payload.includes(index)),
+      };
+    }
+
+    case useListBox.types.hideOptions: {
+      if (!Array.isArray(payload)) {
+        throw Error("hideOptions action expect an array as a payload");
```
alexzherdev:
```suggestion
        throw new Error("hideOptions action expect an array as a payload");
```

packages/ListBox/src/store/reducer.js
```diff
@@ -0,0 +1,236 @@
...
+
+      return {
+        ...state,
+        options: optionsClone,
+      };
+    }
+
+    case useListBox.types.unhideOptions: {
+      if (!Array.isArray(payload)) {
+        throw Error("unhideOptions action expect an array as a payload");
```
alexzherdev:
```suggestion
        throw new Error("unhideOptions action expect an array as a payload");
```

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -0,0 +1,234 @@
...
+  return index || 0;
+}
+
+export function getNextOptionActiveIndexLooping(state) {
+  let activeIndex = getNextOptionActiveIndex(state);
+  if (!activeIndex) {
+    activeIndex = getNextOptionActiveIndex(state, false);
+  }
+
+  return activeIndex;
```
mikrotron:
```js
return getNextOptionActiveIndex(state) || getNextOptionActiveIndex(state, false);
```

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -0,0 +1,234 @@
...
+}
+
+export function handleArrowKeys({ event, state, dispatch, isArrowDown = null }) {
+  if (!state.isOpen) {
+    dispatch({ type: useListBox.types.openPopover });
+    return;
+  }
+
+  event.preventDefault();
+  const next = isArrowDown ? getNextOptionActiveIndex(state) : getNextOptionActiveIndex(state, false);
```
mikrotron:
```js
const next = getNextOptionActiveIndex(state, isArrowDown);
```

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -0,0 +1,234 @@
...
+
+export const handleClickOption = ({ props, state, dispatch }) => event => {
+  const { index } = props;
+  const { options, hasFilter, isMulti, refFilterInput } = state;
+  const hasPreventDefaultOnSelect = options[index].preventDefaultOnSelect;
+  if (state.isDisabled || props.isDisabled) {
+    return;
+  }
+
+  if (state.refListBox.current.contains(event.target) && document.activeElement === document.body && !hasFilter) {
```
mikrotron: Wha? Maybe...

```js
const isTheImportantThingTrue = state.refListBox.current.contains(event.target) && document.activeElement === document.body && !hasFilter;
if (isTheImportantThingTrue) state.refListBoxContainer.current.focus();
```

...would give you the opportunity to self-document this condition.

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    // for single select the option is set when the user interact with up and down arrows
+    // no need to notify which option is selected just close the popover
+    dispatch({ type: useListBox.types.closePopover });
+    state.refTrigger.current.focus();
+  } else {
+    dispatch({ type: useListBox.types.openPopover });
+  }
+}
```
mikrotron: smells

The complexity in here is worrisome, but I suppose this is what you warned us about in your disclaimer:
> Complex logic stuff inside of the component

Hopefully it can be iterated on and improved without any fundamental design changes.

packages/ListBox/src/components/Popover/Popover.js
```diff
@@ -0,0 +1,29 @@
...
+import PopoverStyled from "./Popover.styles";
+import useListBox from "../../useListBox";
+
+const propTypes = {
+  children: PropTypes.node, // eslint-disable-line
+  zIndex: PropTypes.number,
+};
+
+const defaultProps = {
+  zIndex: 100,
```
mikrotron: There's a `z()` helper that is meant to limit our z-indexes to a max of `7`.
https://github.com/acl-services/paprika/blob/master/packages/Stylers/src/helpers.js#L30-L34

packages/ListBox/src/components/RawItem/RawItem.js
```diff
@@ -0,0 +1,25 @@
...
+import Option from "../Option";
+/**
+
+This component let the consumer to write `<ListBox.Option preventDefaultOnSelect />`
+in a more semantic way:
+
+<ListBox>
+  <ListBox.Option>Option 1</ListBox.Option>
+  <ListBox.RawItem> this is not interactive and can be anything </ListBox.RawItem>
+</ListBox>
```
mikrotron: I don't think inline comments are the best place for this kind of documentation.

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,154 @@
...
+
+    if (state.hasFooter) {
+      dispatch({
+        type: useListBox.types.clear,
+        payload: {
+          isOpen: false,
+          onChangeFn: applyOnChange(onFooterClickAccept, "listbox:footer:clear"),
+        },
+      });
+      return;
```
mikrotron: Why do you hate `else` so much? 😄

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,154 @@
...
+const propTypes = {
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+  hasClearButton: PropTypes.bool,
+  onClickClear: PropTypes.func,
+  onFooterClickAccept: PropTypes.func,
+  placeholder: PropTypes.string,
+};
+
+const defaultProps = {
+  children: <React.Fragment />,
```
mikrotron: Why not `null`?

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,154 @@
...
+  RawButtonClearButtonStyled,
+  TriggerActionIconsContainer,
+} from "./Trigger.styles";
+import { getDOMAttributesForListBoxButton } from "../../helpers/DOMAttributes";
+
+const propTypes = {
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
+  hasClearButton: PropTypes.bool,
+  onClickClear: PropTypes.func,
+  onFooterClickAccept: PropTypes.func,
```
mikrotron: `onClickFooterAccept`

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,154 @@
...
+      />
+    ) : (
+      <RawButton
+        type="button"
+        onClick={handleClick}
+        ref={refTrigger}
+        onKeyDown={handleKeyboardKeys(state, dispatch)}
+        onKeyUp={() => {}}
+        isDisabled={isDisabled}
+        data-raw-button="trigger"
```
mikrotron: `data-qa-anchor='listbox-trigger'`?

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,154 @@
...
+        refTrigger: state.refTrigger,
+      });
+    }, [props.children]);
+  }
+
+  return (
+    <ListBoxTriggerStyled isInline={state.isInline} isDisabled={isDisabled} ref={refTriggerContainer}>
+      {hasRenderTrigger ? renderChildrenProps : renderLabel()}
+      <TriggerActionIconsContainer>
+        {hasClearButton ? (
```
mikrotron: Could you just add `|| state.selectedOptions.length` to this condition instead of using a `hasSelectedOptions` prop?

packages/ListBox/src/components/Trigger/Trigger.styles.js
```diff
@@ -0,0 +1,140 @@
...
+  position: absolute;
+  right: 0;
+  top: 0;
+  height: 100%;
+  pointer-events: none;
+`;
+
+export const RawButtonClearButtonStyled = styled(RawButton)`
+  ${RawButtonStyled}
+  && {
```
mikrotron: What does the double ampersand do??

packages/ListBox/src/effects/effects.js
```diff
@@ -0,0 +1,110 @@
...
+  if (state.hasPopupOpened) {
+    listBoxContainer.focus();
+    if (!state.isInline && !state.isOpen) {
+      trigger.focus();
+    }
+  }
+};
+
+export const handleEffectListBoxWidth = (state, dispatch) => () => {
+  const $triggerContainer = state.refTriggerContainer.current;
```
mikrotron: I personally like the `$` prefix for DOM elements... but there are many more opportunities to use this convention consistently.  Lines 25-27 in this very file, for example.

packages/ListBox/src/effects/effects.js
```diff
@@ -0,0 +1,110 @@
...
+      if (state.activeOption === 0) {
+        offsetTop = 0;
+      }
+
+      state.refListBox.current.scrollTo(0, offsetTop - 10);
+    }
+  }
+};
+
+export const handleEffectIsDisabledChange = (props, dispatch) => () => {
```
mikrotron: `handleEffectToggleDisabled` ?

packages/ListBox/src/effects/effects.js
```diff
@@ -0,0 +1,110 @@
...
+  });
+};
+
+export const handleEffectOptionSelected = (state, dispatch) => () => {
+  if (state.onChangeFn) {
+    state.onChangeFn(state, dispatch);
+  }
+};
+
+export const handleEffectHasFooter = (Footer, dispatch) => () => {
```
mikrotron: `handleEffectSetHasFooter`?

packages/ListBox/src/helpers/DOMAttributes.js
```diff
@@ -0,0 +1,41 @@
+import uuidv4 from "uuid/v4";
+
+const listBoxId = `listBoxId_${uuidv4()}`;
+
+export function getDOMAttributesForListBoxContainer() {
+  return { tabIndex: "-1" };
+}
+
+export function getDOMAttributesForListBox(state) {
+  // NOTE: replace this with L10n component
```
mikrotron: Why do you need the L10n here?

packages/ListBox/src/helpers/DOMAttributes.js
```diff
@@ -0,0 +1,41 @@
...
+
+export const getA11yAttributesForOption = isSelected => {
+  const a11yIsSelected = { "aria-selected": `${isSelected ? "true" : "false"}` };
+  return { role: "option", ...a11yIsSelected };
+};
+
+export function getDOMAttributesForListBoxButton() {
+  return {
+    "aria-haspopup": "popup-picker",
+    "aria-labelledby": "ccc listBox-button",
```
mikrotron: `ccc`? Typo? Also, this needs to be an `id` and this does not look safely unique.

packages/ListBox/src/helpers/DOMAttributes.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const getA11yAttributesForOption = isSelected => {
+  const a11yIsSelected = { "aria-selected": `${isSelected ? "true" : "false"}` };
+  return { role: "option", ...a11yIsSelected };
+};
+
+export function getDOMAttributesForListBoxButton() {
+  return {
+    "aria-haspopup": "popup-picker",
```
mikrotron: `aria-haspopup` should be a boolean.  Maybe you want to use `aria-controls` here too?

packages/ListBox/src/helpers/applyOnChange.js
```diff
@@ -0,0 +1,49 @@
+import useListBox from "../useListBox";
```
mikrotron: Why is it `helpers/applyOnChange.js` when you export a bunch of other helpers? I am not convinced about the folder structure for these `/helpers` or the accompanying `/effects`, `/imperative`.  Also, why is there a `ListBox/store` and a `ListBox/src/store`?

packages/ListBox/src/helpers/applyOnChange.js
```diff
@@ -0,0 +1,49 @@
...
+
+export function cleanActionTypes() {
```
mikrotron: What makes them `clean`?

packages/ListBox/src/store/actionTypes.js
```diff
@@ -0,0 +1,24 @@
+export const accept = "ACCEPT";
+export const applyFilter = "APPLY_FILTER";
+export const cancel = "CANCEL";
+export const clear = "CLEAR";
+export const closePopover = "CLOSE_POPOVER";
+export const hideOptions = "HIDE_OPTION";
+export const openPopover = "OPEN_POPOVER";
+export const reset = "RESET";
+export const selectMultipleOption = "SELECT_MULTIPLE_OPTION";
```
mikrotron: Was the choice to use camelCase instead of ALL_CAPS a deliberate one? I believe that in Redux it's typical to use ALL_CAPS for actions.

jamiek-acl: I was expecting all caps, too.

packages/ListBox/src/ListBox.js
alexzherdev: Yeah. Custom hooks are the way to go here. All this trickery to pass in state and dispatch would go away.
I think the lesson here is when using a new thing we should read the docs for said thing.
I was pretty surprised to learn that prism didn't know about the eslint plugin for hooks when it's right there in the documentation.

packages/ListBox/src/store/reducer.js
```diff
@@ -0,0 +1,236 @@
...
+        options: optionsClone,
+      };
+    }
+
+    case useListBox.types.clear: {
+      return {
+        ...state,
+        activeOption: null,
+        isOpen: (payload && payload.isOpen) || false,
+        onChangeFn: payload && payload.onChangeFn ? payload.onChangeFn : null,
```
alexzherdev: Why do we need to pass around the `onChangeFn`? I'm not able to follow all the cases it's used in and it seems to be a different thing all the time.

packages/ListBox/src/store/Provider.js
```diff
@@ -0,0 +1,101 @@
...
+      refFooterContainer,
+      refListBox,
+      refListBoxContainer,
+      refTrigger,
+      refTriggerContainer,
+    },
+    initializeState
+  );
+
+  const value = { state, dispatch };
```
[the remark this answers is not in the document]
alexzherdev: Actually, I'm recalling now that the recommended pattern is to have two different contexts for state and dispatch: https://reactjs.org/docs/hooks-faq.html#how-to-avoid-passing-callbacks-down
Then you only need to memoize the start part because dispatch is guaranteed to be stable.

.eslintrc
```diff
@@ -20,31 +20,40 @@
...
     "no-plusplus": "off",
-    "no-restricted-syntax": ["error", {
-      "selector": "ImportDeclaration[source.value=\"prop-types\"] :matches(ImportSpecifier, ImportNamespaceSpecifier)",
-      "message": "Prefer `import PropTypes from \"prop-types\";`"
-    }, {
-      "selector": "ImportDeclaration[source.value=\"react\"] :matches(ImportSpecifier, ImportNamespaceSpecifier)",
-      "message": "Prefer accessing properties of `React` import, e.g. `React.Component`"
-    }],
+    "no-shadow": "off",
+    "no-console": "off",
```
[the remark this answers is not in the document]
nahumzs: We are allowing console.log is just warning.  

When running the tests we received lots of warning which is kind of annoying, any reason why we should keep it?

packages/ListBox/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/listbox"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "author": "@paprika",
+  "license": "MIT",
+  "devDependencies": {
+    "react-json-tree": "^0.11.2"
```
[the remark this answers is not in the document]
nahumzs: is used on the support branch... we can keep it here, I forgot to remove it in this branch.

packages/ListBox/src/store/reducer.js
nahumzs: there are couple cases that invoke it, the onAccept clicking when has a footer, the click event and the onKeyEvent. 

At the beggining, I was manually calling the onChange in each case. But then the complexity increase and was easier to point all calls to the same function, so the user always gets the same callback answer no matter which event is occurring. 

So for the user is transparent no matter how the user interacts the consumer will always receive and nice structure answer in the onChange event.

packages/ListBox/src/components/Footer/Footer.js
nahumzs: yeap but that would blend the issue with react-docgen, I think this way is more explicit.

packages/ListBox/.npmignore
```diff
@@ -0,0 +1,3 @@
+*
```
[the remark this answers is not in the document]
nahumzs: Maybe this is something that we can address later, not necessarily in this PR, I add an issue for this https://github.com/acl-services/paprika/issues/88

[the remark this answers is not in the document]
nahumzs: All components are using this approach right now.

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: Yes, a lot of complexity for managing keyboard. We can keep moving and improving for sure.

packages/ListBox/src/components/Popover/Popover.js
nahumzs: do we want to limit it? or you mean as a defaultProp?

packages/ListBox/src/components/RawItem/RawItem.js
nahumzs: mmm this is not documentation for our consumers, this is a note for the maintainers, this is a shell component that allowed the consumer to use the `<ListBox.RawItem />` instead of `<ListBox.Option preventDefaultOneSelect />` I think should live here for future reference if someone needs to modify this component. thought if you think there is a better place for this comment I could move it from here. 

Im not attached to the comment, I just want to be clear why this component exists.

packages/ListBox/src/components/Trigger/Trigger.js
nahumzs: Was breaking the rendering. that's the reason why I changed it to React.Fragment. But don't remember exactly the reason.

packages/ListBox/src/components/Trigger/Trigger.styles.js
nahumzs: add specificity to the style so no need to use important https://github.com/styled-components/styled-components-website/issues/187#issuecomment-352727285

packages/ListBox/src/helpers/DOMAttributes.js
nahumzs: Screen-reader is a whole other feature in this component, that's why I didn't even bother in this PR, I focused on keyboard accessibility specifically. 

Though I put some function in place and for this specifically and can support the implementation later. I have created an issue to address ListBox screen reader support. 

https://github.com/acl-services/paprika/issues/89

packages/ListBox/src/helpers/applyOnChange.js
nahumzs: `ListBox/store` might be there because was before we move everything to `src` folder maybe I forgot to delete it. 

I'm not seeing it in atom. 🤔
Screen Shot 2019-06-11 at 5 47 47 PM

About folder structure, I'm not sure if this PR should be the one specifically addressing that issue. Is quite complex merging two branches when changing to many folder structures (files and stories, test, depend on it), 

what about if you have a folder structure that you want to see you can add it here, and I can address it in another PR?

packages/ListBox/src/helpers/applyOnChange.js
nahumzs: maybe is cleaningActionTypes the name, I'm removing some actions types that are internal and the consumer shouldn't use or might not even interested in using.

packages/ListBox/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+export default ListBoxWithProvider;
+
+ListBoxWithProvider.propTypes = propTypes;
+ListBoxWithProvider.defaultProps = defaultProps;
+ListBoxWithProvider.Divider = Divider;
+ListBoxWithProvider.Filter = Filter;
+ListBoxWithProvider.Footer = Footer;
+ListBoxWithProvider.Option = Option;
+ListBoxWithProvider.Popover = Popover;
```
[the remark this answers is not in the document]
nahumzs: Well, the thing is not a body, is actually a Popover, and if you as a consumer learned how to use the Popover you can transfer your knowledge in using it here, the same props are forwarded to the original Popover. If I change the name we lose the ability to transfer that intuitiveness. 

Also changing the name to` ListBox.Body` will be misleading because when the Popover  `is inline` there is not a Popover rendered then there will not be a "Body".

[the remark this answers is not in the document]
alexzherdev: Hmm but for the inline case, it doesn't make sense to call it Popover either as a popover is by definition not inline?

packages/ListBox/src/components/Trigger/Trigger.js
nahumzs: I don't find it useful in most cases if you have the power of return :P. Totally a preference thing.

packages/ListBox/src/helpers/DOMAttributes.js
nahumzs: I don't think we need it might be a comment when I started building the component

packages/ListBox/src/helpers/DOMAttributes.js
nahumzs: yeap, I think I will remove this for now, and we can address those changes we work on screen reader implementation.

nahumzs: https://github.com/acl-services/paprika/issues/89

packages/ListBox/src/store/actionTypes.js
nahumzs: how big is your giveAShitMeter in this? is something that might be addressed in an incoming PR? this will not affect the ListBox is more an "author experience"

packages/ListBox/src/index.js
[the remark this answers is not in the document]
nahumzs: `when the Popover is inline there is not a` 
should read 
`when the **LISTBOX**  is inline there is not a Popover`

packages/ListBox/src/helpers/applyOnChange.js
```diff
@@ -0,0 +1,49 @@
...
+  if (!onChangeCallBack) {
+    // This could mean that the consumer might decide to use onClickAccept in the footer without a onChange or
+    // added a footer without onClickAccept.
+    return;
+  }
+
+  if (state.isMulti) {
+    onChangeCallBack.apply(null, [
+      ...getSelectedOptionsMulti(state),
+      { dispatch, actionTypes: cleanActionTypes(), eventType, ...event },
```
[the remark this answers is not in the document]
nahumzs: I started this component using useState (that the reason why I had that crazy handleEffect functions) and soon I faced some problem running updates one after another. The Reducer pattern and sharing actions is an approach uses by [Kent Dodds](https://www.youtube.com/watch?v=RENGNzZ0dIs), which I think is convenient for this Components. 

There are two things still that I need to do with this component and the reducer.

1.- Apply [good practices](https://redux.js.org/recipes/structuring-reducers/splitting-reducer-logic), organization, name convention as REDUX suggested. 
2.- Implemented the REDUCER override that Kent Dodds explains around [minutes 16](https://youtu.be/RENGNzZ0dIs?t=982) in the previous video

packages/ListBox/src/helpers/applyOnChange.js
```diff
@@ -0,0 +1,49 @@
...
+
+const applyOnChange = (onChangeCallBack = () => {}, eventType = "", event) => (state, dispatch) => {
+  if (!onChangeCallBack) {
+    // This could mean that the consumer might decide to use onClickAccept in the footer without a onChange or
+    // added a footer without onClickAccept.
+    return;
+  }
+
+  if (state.isMulti) {
+    onChangeCallBack.apply(null, [
```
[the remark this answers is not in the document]
nahumzs: if I don't use apply the callback will receive one array as an argument something like:
```js
function callback([allstuff]) {

}
```
but with apply you will get:
```js
function callback(stuff1, stuff2, stuff3) {
  // which gives a better experiences.
}
```

packages/ListBox/src/helpers/applyOnChange.js
mikrotron: Sounds reasonable. After you merge please make sure to follow up on this with me.

packages/ListBox/src/helpers/applyOnChange.js
mikrotron: Ah. Okay, so probably something like `getCleanActionTypes` would be more appropriate as it returns some actionTypes, but isn't a method to mutate the actionTypes.

packages/ListBox/src/store/actionTypes.js
mikrotron: I'm just interested in setting a convention before it propagates as a best practice to other components.  I'm okay with either choice, just don't want to have to change _more_ code later because we didn't give enough shits now.

packages/ListBox/src/store/reducer.js
```diff
@@ -0,0 +1,236 @@
...
+        selectedOptions: [],
+      };
+    }
+
+    case useListBox.types.reset: {
+      return {
+        ...state.originalState,
```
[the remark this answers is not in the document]
nahumzs: I don't want to take the changes on consideration, there are three actions: clear, cancel and reset.

`cancel`: revert you to the latest changes after you have accepted some selections, is like an undo while you are selecting.

`reset`: take you to the original state, the first load, and doesn't take into consideration any change, just a snapshot of the first time.

`clear`, only remove the items from selectedItems, similar to cancel but will not close the popover if it's open. 

Is something that I'm missing?

packages/ListBox/src/helpers/applyOnChange.js
nahumzs: I called sanitizeActionTypes is merged

packages/ListBox/src/store/Provider.js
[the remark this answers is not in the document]
nahumzs: Sound to me like a major refactor having two contexts for this PR.  

I have a list of improvements for this component. Should I add this suggestion to it? 

(Note: I'm down to do it but don't know if this PR is the best option)

packages/ListBox/src/store/actionTypes.js
nahumzs: This is part of an answer that I post about the reducer pattern answering a question from alex. I have a list of things I want to improve for this component, but I don't think this is something that matter for V1. 

- Following best practices for reducer accordingly redux
- Allowed override of reducers
- And Alex suggest separate dispatch and state into two contexts

nahumzs: this is the full list I have so far and is on my ux-status page. I can convert it into an issue if you feel more comfortable:

- [ ] Implementation  Advance Reducer Pattern
- [ ]<Trigger.Tag>
- [ ] Refactor reducer with Redux suggestions and implement override reducer as Kent dods approach for reducer
- [ ] tokens cleaning process
- [ ] Create diagram about Listbox structure
- [ ] Add two context one for dispatch one for the state

packages/ListBox/src/ListBox.js
nahumzs: indeed, my mistake, I have refactored all this.

packages/ListBox/src/components/Filter/Filter.js
nahumzs: is not css IMO and haven't seen using BEM for this in any project at ecosystem level so far. But if we want to go with the BEM for paprika, I'm not really opposed. 

But I think this is more a standard decision at repo level than in this PR

packages/ListBox/src/components/Content/Content.js
nahumzs: Didn't understand how this technique applies here.

packages/ListBox/src/components/Divider/Divider.js
nahumzs: remove for this PR. later we will have to figure out how to link groups from option with the screen reader :/ complex thing.

packages/Popover/src/components/Content/Content.js
```diff
@@ -41,59 +47,63 @@ class Content extends React.Component {
...
-      <PopoverContext.Consumer>
-        {({ content, isEager, isOpen, onClose, onDelayedClose, onDelayedOpen, portalElement, refContent }) => {
-          const contentStyles = {
-            left: content.x,
-            maxWidth: this.isNumber(content.maxWidth) ? `${content.maxWidth}px` : content.maxWidth,
-            top: content.y,
-            width: content.width,
-          };
+  const handleRef = _ref => {
+    refContent(_ref);
```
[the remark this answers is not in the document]
nahumzs: I have to change as well the context because is receiving a function, not feel 💯 confident that the popover has enough testing currently for doing the refactor in this branch.

packages/ListBox/src/components/Content/Content.js
```diff
@@ -0,0 +1,74 @@
...
+
+  if (state.isDisabled) {
+    return;
+  }
+
+  // requestAnimationFrame give time to process
+  // the element that has received the click event
+  // via document.activeElement instead of returning
+  // the body element automatically
+  window.requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
nahumzs: For this specific case I think is better RAF. 

If we use settimeout I will not be sure that the onblur has occurred and the browser had time to digest the event, using RAF for sure will fire after the screen has an update and the dcoument.activeElement will be the actual element and no the body.

packages/ListBox/src/components/Content/Content.js
```diff
@@ -0,0 +1,74 @@
...
+    // the trigger should handle the close and open not the onBlur event
+    if (state.refTriggerContainer.current && state.refTriggerContainer.current.contains(document.activeElement)) {
+      return;
+    }
+
+    if (
+      refListBoxContainer &&
```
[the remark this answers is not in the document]
nahumzs: the trigger is always present, the trigger is the "button"

packages/ListBox/src/components/Filter/Filter.js
nahumzs: this props has been removed is not even part of the component anymore.

packages/ListBox/src/components/Filter/Filter.js
nahumzs: ^ what do you mean? 🤔

packages/ListBox/src/components/Option/Option.js
nahumzs: is not in propTypes because is an internal prop. this makes possible to have `<ListBox.RawItem />` which pretty much is an alias of `<ListBox.Option preventDefaultOnSelect />`.

packages/ListBox/src/components/Option/Option.js
nahumzs: the RawItem is an empty shell. This prop is still in play, was easier to just add the component to the consumer and keep the prop internally didn't have to modify the logic.

packages/ListBox/src/components/Option/helpers/optionState.js
nahumzs: because is easier to access options[index] than `options.find()`

packages/ListBox/src/components/Label/Label.js
```diff
@@ -0,0 +1,83 @@
...
+          `The trigger label on the ListBox needs that either <ListBox.Option> children are typeof string
+        or a label prop is add to the <ListBox.Option label='my description'> component`
+        );
+      })
+      .join(", ");
+
+    const lastKnownLabel =
+      optionsLength > 1 ? (
+        <span css={labelStyles}>
```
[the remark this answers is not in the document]
nahumzs: is because we start using css after the styled.div syntax, I haven't refactored styles to keep with the new standard which is `css`

packages/ListBox/src/components/Option/Option.styles.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  ${props => {
+    const isDisabled = props.isDisabled
+      ? `color: ${tokens.color.blackLighten60}; border: 0; background: transparent;`
+      : "";
+
+    const isActived = props.isActive
+      ? `
+        border: 3px solid Highlight;
```
tristanjasper: Highlighted option content is displaying content in a slightly different position to non highlighted options content. Suggest using `outline` instead of `border` to prevent this affect the layout
image

.eslintrc
nahumzs: yes, I still have problems with it (just test it), I will resolve the conversation.

packages/ListBox/src/ListBox.js
nahumzs: I would prefer to keep it as height even internally is max-height.

.eslintrc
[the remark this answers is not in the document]
nahumzs: also when running `yarn test` is floating the terminal with warnings.

.semaphore/semaphore.yml
```diff
@@ -27,6 +27,7 @@ blocks:
       jobs:
         - name: Lint && prettier
           commands:
+            - yarn add eslint-plugin-react-hooks -D -W
```
nahumzs: CI complaining about not having eslint-plugin-react-hooks installed didn't dig deeper into the issue.

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,164 @@
...
+          {Footer ? React.cloneElement(Footer, { ref: state.refFooterContainer }) : null}
+        </Box>
+      </Content>
+    </React.Fragment>
+  );
+}
+
+ListBox.propTypes = {
+  ...propTypes,
+  children: PropTypes.node.isRequired,
```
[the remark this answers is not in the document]
nahumzs: I have a ListBox and ListBoxContainer. I'm exporting the props for ListBoxContainer. 

The ListBox is only the skeleton without popover. The ListBoxContainer is the one which decides to render the ListBox with or without popover depending on the props.

packages/ListBox/src/ListBox.js
nahumzs: In the trigger case is not only clonning is also spreading so the name doesn't apply exactly there, but I changed it to `const trigger`

packages/ListBox/src/components/Content/Content.js
```diff
@@ -0,0 +1,74 @@
...
+
+export default function Content(props) {
+  const [state, dispatch] = useListBox();
+  const { refListBoxContainer } = state;
+
+  if (state.isInline) {
+    return (
+      <ContentInlineStyled
+        {...getDOMAttributesForListBoxContainer()}
+        onKeyDown={handleKeyboardKeys(state, dispatch)}
```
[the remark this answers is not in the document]
nahumzs: no following here. 🤔

packages/ListBox/src/components/Option/Option.styles.js
mikrotron: I don't think this is the right focus style at all, but this PR is not so concerned with the styling.  There will be a follow up PR to get it to ✨

tristanjasper:
Out of interest should it be the responsibility of the consuming app to prevent the outline on focus for the multiselect? I would have thought this would be handled in the component. wdyt?
image

packages/ListBox/src/components/Filter/Filter.js
mikrotron: Yeah, you probably have to set the default to `null`, then use the L10nized string when rendering, if it's still `null`.

packages/ListBox/src/components/Option/Option.styles.js
```diff
@@ -0,0 +1,50 @@
...
+import tokens from "@paprika/tokens";
+
+const blueSelected = "#e5f1fe";
+
+export const OptionStyled = styled.li`
+  padding: 4px;
+  margin-bottom: 4px;
+  font-size: 13px;
+  border-radius: 3px;
+
```
tristanjasper: `cursor:pointer`, or is there a reason you leave it out?

nahumzs: thank you for the input Tristan, I added a small disclaimer at the top of this PR that I'm not focusing on styling at this moment, it will be done in a different PR.

packages/ListBox/src/components/Option/Option.js
nahumzs: Mmmm... no entirely sure if I want to remove it or keep it, I think for 0.5.0 version I will keep.

packages/ListBox/src/components/Option/helpers/optionState.js
nahumzs: Mmmm I have to think about the nuances of transforming the option to an array, not sure yet it would work better, but this is something that doesn't affect the API and is an internal we can revisit as we keep improving this component.

packages/ListBox/src/components/Footer/Footer.js
nahumzs: We will revisit this component when we have access to the `<ActionDialog />` for now I didn't want to go deeper in fixing this as well, I just copy the one that we had previously in our UI library.

packages/ListBox/src/components/Footer/Footer.js
nahumzs: these are the kind of triggered event, is just notifying the consumer, nothing fancy is happening with them we can remove them later if we feel like it.

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: agree is part of the road map for improving this component.

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,161 @@
...
+  filter: null, // eslint-disable-line
+  footer: null, // eslint-disable-line
+  height: 200,
+  isDisabled: false,
+  isInline: false,
+  isMulti: false,
+  isOpen: null,
+  onChange: () => {},
+  placeholder: "Select...",
+  trigger: null, // eslint-disable-line
```
mikrotron: So `filter`, `footer`, and `trigger` are not in `PropTypes` so that they won't be included in the documentation (via `react-docgen`)? Is that why?

packages/ListBox/src/ListBox.js
```diff
@@ -0,0 +1,161 @@
...
+
+  const trigger = _trigger ? (
+    React.cloneElement(_trigger, { ...propsForTrigger, ..._trigger.props })
+  ) : (
+    <Trigger {...propsForTrigger} />
+  );
+
+  return (
+    <React.Fragment>
+      {trigger}
```
mikrotron: Is there no option to use the ListBox _without_ a trigger yet?

packages/ListBox/src/ListBox.js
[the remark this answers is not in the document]
mikrotron: I'd probably name it a bit differently... Like `ListBoxBody` as the _skeleton_ and `ListBox` as the exported component.

packages/ListBox/src/components/Popover/Popover.js
mikrotron: I think the default prop should be `z(7)` (or lower, preferably).

packages/ListBox/src/store/actionTypes.js
mikrotron: This is a good list, I think adding it as an issue probably makes the most sense... but what is your "ux-status page"?

packages/ListBox/src/ListBox.js
nahumzs: exactly they are internals, so I'm trying to avoid to publish them.

packages/ListBox/src/ListBox.js
nahumzs: Not implemented yet, but should be quite simple to add in a new version.

nahumzs:
🙈 😈

PR #80, opened 2019-05-18
UX-470 Upgrade Storybook to v5 🐐

mikrotron:
### 🛠 Purpose

Upgrade Storybook to `v5`.

### ✏️ Notes to Reviewer

- Needed to resolve some dependency cycles to appease a recently fixed lint rule (`import/no-cycle`).
- Needed to update the Cypress `visitStorybook()` command since URL structure has changed.
- Needed to update stories so that knobs are created at the same time as the story is rendered (probably due to [this known issue](https://github.com/storybooks/storybook/issues/5455)).
- Added the `wait-on` package so Cypress tests can wait for Storybook to be up before running (which must take a bit longer now). 

### 🖥 Screenshots

Screen Shot 2019-05-21 at 10 12 39 PM


packages/Button/stories/Button.stories.js
```diff
@@ -2,15 +2,15 @@ import React from "react";
...
+import Showcase, { showcaseProps } from "./examples/Showcase";
 import Basic from "./examples/Basic";
 import NewRef from "./examples/NewRef";
 import OldRef from "./examples/OldRef";
 import CloseButtonRef from "./examples/CloseButtonRef";
 
 storiesOf("Button", module)
   .addDecorator(withKnobs)
-  .add("Showcase", () => <Showcase />)
+  .add("Showcase", () => <Showcase {...showcaseProps()} />)
```
alexzherdev: I may be missing something - but both `Showcase` and `showcaseProps` are exported from the same file. Why not have that file export just the component that renders the story with the props baked in?

packages/L10n/src/L10nContext.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
```
alexzherdev: Curious why this change is necessary?

mikrotron: A new lint rule has been introduced, `import/no-cycle`, that must have been applied after a fresh `yarn`.  There were a bunch of dependencies I had to fix as a result, but actually unrelated to the Storybook upgrade.

packages/Button/stories/Button.stories.js
mikrotron: I may also be missing something, but this was the simplest way I could find to make the knobs work again (they had all stopped).  It seemed like the knob functions needed to be called outside of the example component...?? But I am not really sure _why_ they stopped working in this new version.

packages/L10n/src/L10nContext.js
mikrotron: Sorry, not a new rule, I saw a post from `May` but just noticed it's 2018, not 2019.  Not sure why this wasn't a complaint before now actually.

alexzherdev: It looks like there was a PR to fix some issues with cycle detection. Bugfixes are not considered breaking changes so we started getting those warnings now.

PR #81, opened 2019-05-24
UX-471 🐙😈 change purple to galvanize/lighten50 to lighten44

patrickm4:
### 🛠 Purpose

Change purple tokens

### ✏️ Notes to Reviewer

Changed purple tokens to galvanize
Changed lighten50 to lighten44 as galvanizeLighten50 was essentially white(#ffffff)

### 🖥 Screenshots

---


packages/Tokens/src/tokens.json
```diff
@@ -45,14 +45,14 @@
...
     "orangeHighlight": "#e1624b",
-    "purple": "#53529f",
-    "purpleDarken10": "#41417d",
-    "purpleDarken20": "#302f5c",
-    "purpleLighten10": "#7575af",
-    "purpleLighten20": "#9695c2",
-    "purpleLighten30": "#b6b5d5",
-    "purpleLighten40": "#d6d6e7",
-    "purpleLighten50": "#f6f6fa",
+    "galvanize": "#785cba",
```
oscarkwan: is there a reason why `galvanize` is the name for the variable?
@mikrotron @nahumzs

johnpeele: I think we should have a quick discussion about this. Eventually we will be replace all purple color hex values, but keeping the same token names, hence mine and Mike P's decision to use `$color--purple-new` as the token name for the new Galvanize purple.

packages/Tokens/src/tokens.yaml
```diff
@@ -50,14 +50,14 @@ color:
...
-  purple-lighten-40: desaturate(lighten($color--purple, 40), 5)
-  purple-lighten-50: desaturate(lighten($color--purple, 50), 5)
+  galvanize: '#785cba'
+  galvanize-darken-10: darken($color--galvanize, 10)
+  galvanize-darken-20: darken($color--galvanize, 20)
+  galvanize-lighten-10: desaturate(lighten($color--galvanize, 10), 5)
+  galvanize-lighten-20: desaturate(lighten($color--galvanize, 20), 5)
+  galvanize-lighten-30: desaturate(lighten($color--galvanize, 30), 5)
+  galvanize-lighten-40: desaturate(lighten($color--galvanize, 40), 5)
+  galvanize-lighten-44: desaturate(lighten($color--galvanize, 44), 5)
```
johnpeele: Also, 44 is a very arbitrary value. We need to treat this as a design system conversation and adjust the colors if absolutely necessary.

packages/Tokens/src/tokens.json
nahumzs: I'm not entirely sure, to me, we shouldn't use the brand name for naming our tokens, they should be agnostic with their name? any specific to use this name @mikrotron ?

packages/Tokens/src/tokens.yaml
nahumzs: @johnpeele wouldn't be more efficient if you provided us with the exact number of colours that we need to support? to me seems too many colours, I'm not entirely sure the need for supporting all these colours, but that's just my opinion.

johnpeele: @nahumzs I agree 100%, but unfortunately we are not ready to redo our color palettes.

packages/Tokens/src/tokens.json
patrickm4: ahh my mistake, i believe i was to leave the purple tokens as "purple" not "galvanize". Should i rename them as #color-purple-new or keep $color-purple?

mikrotron: The approach in the [`acl-ui` PR](https://github.com/acl-services/acl-ui/pull/1182) is what we should go with here too.  Let's just update what `$purple` means.

mikrotron: @nahumzs, @johnpeele are you both more comfortable with this new approach (same approach in `acl-ui` too - https://github.com/acl-services/acl-ui/pull/1182)?

nahumzs: :/ not thrill about this approach but I guess $purple is ok

johnpeele: I'm just not sure what is going on, what is the purpose of this ticket/PR?

mikrotron: @johnpeele, this PR will update all of the `$purple` tokens to be the new Galvanize branded purple tones.  We won't need to change CSS throughout our apps*, just update `acl-ui` (well, _this_ PR is more to keep `paprika` in sync with that change).  

* This isn't 100% true.  Only about 86% true.

johnpeele: @mikrotron First, love the percentage humor, well done.

Second, If we are only updating hex values and not token names then I am 110% on board with this. 👍🏼

Third, now Nahum's last comment in this thread is even more confusing to me. Why are you not thrilled about this approach @nahumzs ?

PR #82, opened 2019-05-30
Ux 434 multiselect support

nahumzs:
### 🛠 Purpose

This branch includes all test and documentation for the <ListBox />

---

### ✏️ Notes to Reviewer

Should be merge after the <ListBox /> Branch is merge

---

### 🖥 Screenshots

---


packages/ListBox/src/components/Label/Label.js
```diff
@@ -0,0 +1,83 @@
...
+
+export default function Label(props) {
+  const { placeholder } = props;
+  const [state] = useListBox();
+  const { selectedOptions, isMulti, options } = state;
+
+  if (!isMulti && selectedOptions.length && state.options[selectedOptions[0]].preventDefaultOnSelect) {
+    if (!lastKnownLabel) return placeholder;
+
+    return lastKnownLabel;
```
jamiek-acl:
```
if (!isMulti && se...) {
  return lastKnownLabel || placeholder;
}
```

packages/ListBox/src/components/Label/Label.js
```diff
@@ -0,0 +1,83 @@
...
+        throw Error(
+          `The trigger label on the ListBox needs that either <ListBox.Option> children are typeof string
+        or a label prop is add to the <ListBox.Option label='my description'> component`
+        );
+      })
+      .join(", ");
+
+    const lastKnownLabel =
+      optionsLength > 1 ? (
+        <span css={labelStyles}>
```
jamiek-acl: I'm noticing here you say:
```
<span css={...}>
  some children
</span>
```
but in other cases you do it like
```
<StyledSpan>
  some children
</StyledSpan>


//StyledSpan.style.js
export const StyledSpan = styled.span`
  ...
`
```

I guess these are equivalent, just different syntaxes? And wondering when to use which?

packages/ListBox/src/components/List/List.styles.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  overflow: auto;
+  ${props => {
+    const paddingNoResults = props.noResultsFound ? "padding: 0" : "";
+    return `
+        max-height: ${props.height}px
+        ${paddingNoResults}
+      `;
+  }}
```
jamiek-acl: Whoa, a function in css.  Trippy.

packages/ListBox/src/components/Option/Option.styles.js
```diff
@@ -0,0 +1,50 @@
+import styled from "styled-components";
+import tokens from "@paprika/tokens";
+
+const blueSelected = "#e5f1fe";
```
jamiek-acl: token!

packages/ListBox/src/components/Option/Option.js
```diff
@@ -0,0 +1,80 @@
...
+import { OptionStyled } from "./Option.styles";
+
+const propTypes = {
+  /** String, number or JSX content */
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+
+  /** Describe if the option is selected or not */
+  isSelected: PropTypes.bool,
+
+  /** Describe if the option is enable or not */
```
jamiek-acl: Not sure how much value some of these comments are adding.
If your response is `They are for auto-documentation` then my response to that is `then how come more components like the Filter and Footer don't have them?` :)

packages/ListBox/ListBox.js
```diff
@@ -0,0 +1,169 @@
...
+
+  /** Message to be display once the filtering process doesn't find a match */
+  hasNotResultsMessage: PropTypes.node,
+};
+
+export const defaultProps = {
+  children: null,
+  Filter: null, // eslint-disable-line
+  Footer: null, // eslint-disable-line
+  hasNotResultsMessage: "Your search did not match any options.",
```
jamiek-acl: Can you make this component be supported by `L10n`?

packages/ListBox/stories/moreExamples/Colums/Colums.stories.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import Columns from "./Columns";
+
+storiesOf("ListBox / more examples", module).add("Colums select", () => <Columns />);
```
mikrotron: This story name, this file, and its directory all have the same typo: `colums` → `columns`.

packages/Popover/stories/Popover.screener.stories.js
```diff
@@ -17,23 +17,23 @@ storiesOf("Popover/Automation Tests/Screener", module).add("basic", () => (
         <Popover.Card>{shortContent}</Popover.Card>
       </Popover.Content>
     </Popover>
-    <div css={gapStyles} />
+    <div style={gapStyles} />
```
mikrotron: Can we add a `Gap` component to the `common.styles.js` instead and avoid inline styling, please.

https://github.com/acl-services/paprika/blob/master/.storybook/assets/styles/common.styles.js

packages/ListBox/README.md
```diff
@@ -0,0 +1,80 @@
...
+    </ListBox>
+  );
+}
+```
+
+### Props
+
+```js
+/** Child of type <ListBox.Option /> */
+children: PropTypes.node,
```
mikrotron: Do we really want to duplicate the amount of maintenance we'll have to do when we update this component's props by copying these into the `README.md`?

nahumzs: good point! those shouldn't be here, I guess a prop table should care about this! I will remove it.

PR #84, opened 2019-06-04
Ux 469 <SidePanel />

### LINKED ISSUE #79: RFC <SidePanel />
# New Component
## Prerequisites
**This is just a checklist, please delete this section after**
Please answer the following questions for yourself before suggesting a new component.
- [x] I checked to make sure that this new component has not already been suggested
- [x] Are there any components in our repo that can be reused?
- [x] Please label as, low / medium / high priority

*Please fill in as much detail as possible below*
From #4 issue 

## Overview
This RFC addressed the <SidePanel /> refactoring. 

[Figma](https://www.figma.com/file/dGl8bgX4FF81ixllzAjRFzDO/Sidepanel?node-id=0%3A1)

We have reduced the use cases for the sidepanel in three basic cases and two different location for the SidePanel (left and right).

Extra cases is a SidePanel inside of a SidePanel also could be seen as two sidepanel one next to the other one. 

## Visual representation and Propose API:

Screen Shot 2019-05-15 at 3 18 59 PM

#### description

The most common SidePanel. Takeover the entire screen without letting the user to interact with the body content. Might have or not overlay and the body content might or might not be scrollable

#### API

```js
<SidePanel isOpen={bool} onClose={fn}>
   <SidePanel.Header>[node]</SidePanel.Header>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.Footer />
</SidePanel>
```

---

Screen Shot 2019-05-15 at 3 19 13 PM

The second most common option, unlinke the (A) SidePanel option, this one has an top offset which allows it to have a different height and be more integrated with the Interface. This SidePanel let the user interact with the entire UI. 

#### API

```js
<SidePanel isOpen={bool} onClose={fn} offset={number}>
   <SidePanel.Header>[node]</SidePanel.Header>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.Footer />
</SidePanel>
```

> The offset will indicate how much space the sidepanel should include from the top. The offset should be passed from the consumer is their responsability. 

---
 
Screen Shot 2019-05-15 at 3 19 18 PM

Exactly as the SidePanel (B) except that includes a second SidePanel letting the consumer to have another area where "Extra" data can be displayed.

#### API

```js
<SidePanel isOpen={bool} onClose={fn} offset={number}>
   <SidePanel.Header>[node]</SidePanel.Header>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.Footer />
</SidePanel>
```

We could inspect the children and detect the number or `SidePanel.Content` and then nested the content depending of the position of the content.

*- or -*

```js
<SidePanel isOpen={bool} onClose={fn} offsetY={number}>
   <SidePanel.Header>[node]</SidePanel.Header>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.Footer />
</SidePanel>
<SidePanel isOpen={bool} onClose={fn} offset={number} offsetX={number}>
   <SidePanel.Header>[node]</SidePanel.Header>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.Footer />
</SidePanel>
```

We could render two sidepanel one each of the other one and just affect the offset position.

*- or -*

```js
<SidePanel isOpen={bool} onClose={fn} offsetY={number}>
   <SidePanel.Header>[node]</SidePanel.Header>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.SubContent>[node]</SidePanel.SubContent>
   <SidePanel.Footer />
</SidePanel>
```

Create a specific subcomponent to deal with the second sidepanel

*Note* I'm `no too` sure about how to deal with this requirement.

---

Screen Shot 2019-05-15 at 3 19 22 PM

SidePanels should be able to render from left or right direction.

*Enum:* `Left` || `Right`

```js
<SidePanel slideIn={Enum} isOpen={bool} onClose={fn}>
   <SidePanel.Header>[node]</SidePanel.Header>
   <SidePanel.Content>[node]</SidePanel.Content>
   <SidePanel.Footer />
</SidePanel>
```

## Questions, wondering, not sure. 

- [ ] This component should be Controlled
- [ ] Should not render content when is close
- [ ] Should be agnostic about the width, the consumer should provide the width which can be defined by tokens cc @krismckinnonacl @mikrotron 
- [ ] Shuld we have a `<Footer />` popover, sidepanel and multiselect ? All of them migth benefit for having the same behaviour. cc @krismckinnonacl @johnpeele @jayde-acl 
- [ ] not sure about the API for the second sidepanel content. 

## A11y considerations and Responsivenes
Let's discuss only about the layout we can have another ISSUE for those specific topics.

nahumzs:
### 🛠 Purpose

close #79 

*The SidePanel is a Controlled Component*

## Props:

```js
const propTypes = {
  /** The content for the SidePanel. */

  /** The width of the open panel. */
  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),

  /** Callback triggered when the side panel needs to be close */
  onClose: PropTypes.func,

  /** Callback once the sidepanel has been opened event */
  onAfterOpen: PropTypes.func,

  /** Callback once the sidepanel has been closed event */
  onAfterClose: PropTypes.func,

  /** Control the visibility of the side panel. This prop makes the side panel */
  isOpen: PropTypes.bool.isRequired,

  /** Control the z position of the sidepanel */
  zIndex: PropTypes.number,

  /** Control y offset of the sidepanel */
  offsetY: PropTypes.number,

  /** Modify the look of the SidePanel */
  kind: PropTypes.oneOf(["default", "child"]),

  /** Disable scroll overflow on document.body when SidePanel is open. */
  disableBodyOverflow: PropTypes.bool,
};
```

## Components
The sidepanel has the following component you can use to compose it:
```js
<SidePanel.Header />
<SidePanel.Footer />
<SidePanel.Overlay />
<SidePanel.FocusTrap />
<SidePanel.Trigger />
```

## The architecture

By default, the SidePanel render whatever you pass as children, but will filter 
`{ Header, Footer, Overlay, Trigger  and FocusTrap}` to properly render those components accordingly 
their optimal position, as an example the Trigger will be rendered as inline since is expected that 
99% of the cases will be a *Button*, while the `<List.Overlay />` and the `<SidePanel />` will be rendered under a `React.Portal`. 

Lastly `<ListBox.Header />` and `<ListBox.Footer />` will be at the top and bottom respectively. 

--- 

## The Ex.

```js
<SidePanel isOpen={isOpen} onEscKey={handleEvent}>
  <SidePanel.Trigger onClick={()=> handleEvent}>Open Sidepanel</SidePanel.Trigger>
   <SidePanel.Header onCloseClick={()=> handleEvent}>New SidePanel</SidePanel.Header>
   [...your_content...]
  <SidePanel.Footer isSticky onClickAccept={handleAccept}>
</SidePanel>
```

### With header and Overlay 

```js
<SidePanel isOpen={isOpen} onEscKey={handleEvent}>
  <SidePanel.Header onCloseClick={handleEvent} />
  <SidePanel.Overlay onClick={handleEvent} />
  [...your_content...]
</SidePanel>
```

---

### Sidepanel Groups
```js
import Sidepanel, { SidePanelGroup } from "@paprika/sidepanel"
<SidePanelGroup>
  <SidePanel></SidePanel>
  <SidePanel></SidePanel>
</SidePanelGroup>
```

## Accessibility

- Focus to the first focusable element accordingly to [wai](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal)
> - In all circumstances, focus moves to an element contained in the dialog.
> - Unless a condition where doing otherwise is advisable, focus is initially set on the first focusable element.
- The focus is trap under the side panel until is dismissed, I'm using [focus-trap-react](https://github.com/davidtheclark/focus-trap-react)
- wai `role=modal`
- If there is not Focusable element on the SidePanel focusable-react has been config to target the `<ListBox.Header />`

## Todo
- [x] *<ListBox.Trigger />*
- [x] *<ListBox.Header />*
- [x] ~<ListBox.Content />~ no need it.
- [x] *<ListBox.Footer />* `isSticky`
- [x] *<ListBox.FocusTrap />* let pass properties to `focus-trap-react`
- [x] *`<SidePanelStoryGroup />`* multiple sidepanel support
- [x] Header `kind={}` prop to allowed having different styles on the header
- [x] Testing SidePanel Jest
- [x] Testing SidePanel Cypress
- [ ] ~Imperative API~ `wont do` in this PR
- [x] SidePanel <SidePanel OffsetY={number} />  
- [x] Block the body from being scrollable

## NOTE 

`Fixed couple test annoying warnings after rebasing master.`

Screen Shot 2019-06-10 at 1 03 20 PM
Screen Shot 2019-06-10 at 1 05 09 PM


AndreyChernykh:
@nahumzs 
1. What `<SidePanel.Trigger />` is for? And why it is a child of a side panel? 

2. Not exactly clear in what circumstances `onOpen` and `onClose` are called. 

3. Why does Side panel has `onEscKey` handler? I seems to me, that we can have a special component for that: 

```
state = {
 isOpen: true
}

return (
  <div> 
    <KeyListener key={Key.ESC} onPress={() => state.isOpen = false } />
    <SidePanel isOpen={state.isOpen} />
  </div>
)
```

nahumzs:
@AndreyChernykh 

- `<Trigger />` takes care of the focus automatically when you close the Sidepanel. So you don't have to worry about that. If you don't want to use the trigger (which is possible) the consumer will need to implement focus management manually. 

- `onOpen` and `onClose` are callback called once the animation of the side panel has finished. pretty much when `<div onAnimationEnd={} />` is triggered. 

- The reason for the onEscKey is `a11y`  requirements, modal required to implemented the `esc` button, while I like your idea of KeyListner, I'm not sure how we can "enforce" the consumer to implement the `esc` behaviour if it's not part of the Component. 

^ adding to your suggestion I think maybe for another version we can think about how to handle the `esc` key across several component `Modal, SidePanels, Takeover and Popover` and try to solve the issue where all "dialog" components are dismissed when the `esc key` event is fired, instead of only dismissed the latest one that was opened.

packages/SidePanel/SidePanel.stories.js
```diff
@@ -0,0 +1,102 @@
...
+  const toggle = () => {
+    setIsOpen(state => !state);
+  };
+
+  return (
+    <React.Fragment>
+      hello lets open a sidepanel
+      <SidePanel isOpen={isOpen}>
+        <SidePanel.Trigger onClick={toggle}>{isOpen ? "close" : "open"}</SidePanel.Trigger>
+        <SidePanel.Overlay onClick={toggle} />
```
KirillKayumov: I have some concerns about the `SidePanel.Overlay` component:
1. An overlay is visually displayed not inside a `SidePanel`, but around it. I think it's not so obvious for a consumer to place `SidePanel.Overlay` as a child of `SidePanel` component.
2. If a consumer should render `SidePanel.Overlay` component manually, it automatically becomes "not required". Although looking at Projects module, there is only one `SidePanel` without an overlay (new To-do SidePanel). All the other side panels have it. Also, I believe that almost always a side panel should be closed if a user clicks on an overlay. Now a consumer should specify it twice: `<SidePanel onClose={this.handleClose}>` and `<SidePanel.Overlay onClick={this.handleClose}`. Maybe it's better to display an overlay by default and call `onClose` on overlay click. Then props for overlays can be passed like so:
```js
<SidePanel isOpen={isOpen} showOverlay={false}>...</SidePanel>
<SidePanel isOpen={isOpen} overlayBackground={...}>...</SidePanel>
OR
<SidePanel isOpen={isOpen} overlay={{ show: false }}>...</SidePanel>
<SidePanel isOpen={isOpen} overlay={{ background: "..." }}>...</SidePanel>
```
WDYT?

nahumzs:
@KirillKayumov 

### 1. I just implemented that if you click on the overlay fire the onClose, so no more specifying onClick on the `<SidePanel.Overlay />`. 💯 for the suggestion

### 2. About `Projects module` having more Sidepanel with overlay. 
While this is true, in incoming designs seems that designers are starting to use the sidepanels without overlay, also a lot of other modules like the ones handling data tent to use the one without the overlay. 

### 3. Agree, visually hierarchically the Overlay should be outside, the reason I extracted inside of the <Sidepanel /> is to remove a lot of properties from that level. I think that's is a good idea because let the `<SidePanel />` lean and with just a small amount of properties. And lets us add any number of props to the overlay without worrying about impacting the `<SidePanel />`

ex.
```js
<SidePanel isOpen={true} onClose={() => setState(isOpen: false)}>
    <SidePanel.Overlay />
    [...content]
</SidePanel>
```

alexzherdev:
$0.02:
- Having a separate `onEscKey` prop means accessibility is not baked in, we should avoid that. Even if we make the prop required, I've noticed folks are lazy to look into the console to check for errors.
- Related, I should only specify the close handler once somewhere (probably on the sidepanel itself), which should then be automatically propagated as the escape key handler, the overlay click handler, and the header's close button handler. Looks like you're almost there except the header.
- Speaking of closing, the current `onClose` on the sidepanel is more like an `onAfterClose`. When I see `onClose` I think of it as the callback where I'm supposed to make the (controlled) sidepanel close.
- `<Trigger>` doesn't seem like it's necessary? If it's just for focus handling then it looks like `focus-trap-react` [should be able to handle it](https://github.com/davidtheclark/focus-trap-react/blob/69c14950c427eddbee14550358ec516459cb0ba1/src/focus-trap-react.js#L64-L70)? And then we're not coupled to always having `<Button>` as the trigger.
- `<Content>` is not clear to me. The only thing it adds is padding. If you don't use it, you will get no padding. Are there use cases for both of these?
- Will we support sticky footers like we had in group permissions?
- Kind of conflicted between setting overlay as a prop vs as an element. I'm not sure if we'll ever need props on the overlay? And in our system in general as of now having overlays is more prevalent than not having them

nahumzs:
> Having a separate onEscKey prop means accessibility is not baked in, we should avoid that. Even if we make the prop required, I've noticed folks are lazy to look into the console to check for errors.

ohh I see, so what about if I change `onClose` to be the default handler when anything that can close the side panel is fire? Esc, click on overlay, header close button, etc, will call onClose. 

and then I will rename the current onClose to `onAfterClose`.

> <Trigger> doesn't seem like it's necessary? If it's just for focus handling then it looks like focus-trap-react should be able to handle it? And then we're not coupled to always having <Button> as the trigger.

Not too sure if react-trap returns the focus to the button, if it does it I will remove the Trigger. The Trigger is as well an optional Component, so you can use anything to open the sidePanel since the isOpen is controlled. 

> <Content> is not clear to me. The only thing it adds is padding. If you don't use it, you will get no padding. Are there use cases for both of these?

Just double check and seems none designer remember no having padding on the sidepanel so I guess we can remove this component.

> Will we support sticky footers like we had in group permissions?
Yes, will be supported, haven't started working on this.

> Kind of conflicted between setting overlay as a prop vs as an element. I'm not sure if we'll ever need props on the overlay? And in our system in general as of now having overlays is more prevalent than not having them

I'm ok with both alternatives if people feel like is better to have it on the props I'm ok 😃with it. Let's @mikrotron make the call on it.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,174 @@
...
+  const refSidePanelContent = React.useRef(null);
+  const refHeader = React.useRef(null);
+
+  // Extracts
+  const {
+    "SidePanel.Trigger": TriggerExtracted,
+    "SidePanel.Overlay": OverlayExtracted,
+    "SidePanel.Header": HeaderExtracted,
+    "SidePanel.FocusTrap": FocusTrapExtracted,
+    children,
```
tristanjasper: is this `children` meant to be here? 🤔

nahumzs: yes, will include everything that has been pass as children except the extracted components.

packages/Heading/stories/examples/HeadingLevels.js
```diff
@@ -80,12 +80,6 @@ const HeadingLevelsStory = () => {
       <Heading level={6} isLight>
         Heading Six with isLight
       </Heading>
-      <br />
```
nahumzs: This is adding propTypes validation errors. and no adding any benefit to the stories.

packages/SidePanel/SidePanel.cypress.js
```diff
@@ -0,0 +1,81 @@
...
+    sidePanelChild.should("exist");
+    sidePanelChild.should("be.visible");
+
+    cy.contains(/Toggle Child/i).click();
+
+    cy.getByTestId("sidepanel-child").should("not.exist");
+  });
+
+  let count = 1;
+  it("should call onAfterOpen and onAfterClose", () => {
```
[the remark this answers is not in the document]
nahumzs: this can't because depends on `onAnimatedEnd` which is fired by css, so will not run in RTL.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,193 @@
...
+
+  /** Modify the look of the SidePanel */
+  kind: PropTypes.oneOf(["default", "child"]),
+
+  /** Disable scroll overflow on document.body when SidePanel is open. */
+  disableBodyOverflow: PropTypes.bool,
+};
+
+const defaultProps = {
+  isInline: false, // this is an internal prop use by SidePanelGroup (Group.js)
```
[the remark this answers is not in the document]
nahumzs: yes is not support it right now, I didn't want to stretch this PR to cover all cases like slideIn=[left and right], those can be add in incoming version. I tried to keep this PR smaller as possible for a first version.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,193 @@
...
+  const offsetScroll = useOffsetScroll(props.offsetY);
+
+  // Refs
+  const refTrigger = React.useRef(null);
+  const refSidePanelContent = React.useRef(null);
+  const refHeader = React.useRef(null);
+
+  // Extracts
+  const {
+    "SidePanel.FocusTrap": FocusTrapExtracted,
```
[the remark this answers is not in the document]
nahumzs: yes, I added this one before the review you did in the ListBox I will fix it.

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -0,0 +1,85 @@
...
+    border-top-left-radius: ${space / 2}px;
+    border-bottom-left-radius: ${space / 2}px;
+  `;
+
+export function slideIn(width) {
+  return keyframes`
+  from {
+    opacity: 0;
+    right: -${width};
+    transform: trasnlateX(100%);
```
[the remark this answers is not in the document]
nahumzs: 😮

packages/Tokens/src/tokens.js
```diff
@@ -48,10 +48,10 @@ const tokens = {
     "purple": "#785cba",
     "purpleDarken10": "#5f449f",
     "purpleDarken20": "#4a347c",
-    "purpleLighten10": "#9680c9",
-    "purpleLighten20": "#b3a4d8",
-    "purpleLighten30": "#d1c7e8",
-    "purpleLighten40": "#efebf7",
+    "purpleLighten10": "#9884c5",
```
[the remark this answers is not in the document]
nahumzs: I think got versioned at some point and are installed when doing yarn because Lerna `pretrasnpile` create them. I will remove them from git history so the `.gitignore` can do their job.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,193 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import PropTypes from "prop-types";
+import { default as FocusTrapLibrary } from "focus-trap-react/dist/focus-trap-react"; // eslint-disable-line
```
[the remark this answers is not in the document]
nahumzs: https://github.com/acl-services/paprika/pull/84/files/decac2d7b3b1fea09cb5ecfdafff325938a422cd#diff-a70e4ad809affbf01bae992a555dc39eR10

I'm importing the shell component which has all the props with FocusTrap name, the shell component will let us build the prop table for documentation. You could argue that this one is the one that I should rename, but is minor IMO.

packages/SidePanel/src/SidePanel.js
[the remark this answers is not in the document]
nahumzs: why would want to expose the prop? Wouldn't that just increase the prop table with props that the consumer will not need? I'm doing it to improve readability.

cypress/support/commands.js
```diff
@@ -9,3 +9,7 @@
 Cypress.Commands.add("visitStorybook", story => {
   cy.visit(`http://localhost:9009/iframe.html?id=${story}`);
 });
+
+Cypress.Commands.add("getByTestId", str => {
```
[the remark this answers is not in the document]
nahumzs: seems like this is the correct approach. I don't see how to override getTestById and use our own `data-qa-anchor` apart from this

https://spectrum.chat/testing-library/general/data-test-id-vs-data-testid~4ba906b1-7390-4c6a-a22a-925011babe67?m=MTUyOTMzNTQ2OTAxNA==

https://github.com/testing-library/dom-testing-library/issues/136#issuecomment-432515243

I will add cypress-testing-library taking advantage of this conversation.

packages/SidePanel/SidePanel.cypress.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  it("should SidePanel.FocusTrap have a different initial focus", () => {
+    cy.visitStorybook("sidepanel-focustrap--sidepanel-focustrap");
+    const input = cy.getByTestId("sidepanel.focustrap.input");
+    const testString = "hello world";
+    input.type(testString).should("have.value", testString);
+  });
+
+  it("should be possible to interact with multiple sidepanels in a group", () => {
```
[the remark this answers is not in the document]
nahumzs: I'm adding this as an integration test, just be 💯 that works in a browser scenario. that's all.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,193 @@
...
+  disableBodyOverflow: true,
+};
+
+function SidePanel(props) {
+  // Props
+  const { onAfterClose, onAfterOpen, onClose, width, isInline, kind, offsetY, ...moreProps } = props;
+
+  // Hooks
+  const [isSidePanelMounted, setMount] = React.useState(props.isOpen);
+  const offsetScroll = useOffsetScroll(props.offsetY);
```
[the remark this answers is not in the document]
nahumzs: offsetY is the initial `top` position when the body can be scrollable the NAV will disappear so the SidePanel has to move up to fill the gap that the NAV has left.

you can see the example gif named: `Scrolling NAV` for a visual example.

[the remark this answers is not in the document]
nahumzs: I can change the name of the variable if you think there is a more suitable name for it.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,193 @@
...
+  zIndex: 100,
+  disableBodyOverflow: true,
+};
+
+function SidePanel(props) {
+  // Props
+  const { onAfterClose, onAfterOpen, onClose, width, isInline, kind, offsetY, ...moreProps } = props;
+
+  // Hooks
+  const [isSidePanelMounted, setMount] = React.useState(props.isOpen);
```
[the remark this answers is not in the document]
nahumzs: more than for the initial mount is for when Closing it and the `onAnimationEnds`, I have to wait until the animation ends to be able to remove it, I have changed the name of the variable to `isSidePanelVisible` to be more clear in communicating the intention of the variable.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,193 @@
...
+            <div>{dialog}</div>
+          </FocusTrapLibrary>
+          {OverlayExtracted ? React.cloneElement(OverlayExtracted, { onClose }) : null}
+        </React.Fragment>,
+        document.body
+      );
+    }
+  }
+
+  const trigger = TriggerExtracted ? React.cloneElement(TriggerExtracted, { ref: refTrigger }) : null;
```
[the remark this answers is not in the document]
nahumzs: I think for that specific case you can create your own `<Button />` and controlled everything with the `<SidePanel.FocusTrap />`. The intention of the Trigger is for a more "casual" easy way to consume this component.  If there is a need eventually I can figure out how to pass the ref eventually.

packages/SidePanel/src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,74 @@
...
+      css={dialogStyles}
+      onAnimationEnd={handleAnimationEnd}
+      role="dialog"
+      width={width}
+      isInline={isInline}
+      offsetY={props.offsetY}
+      ref={refSidePanel}
+      {...moreProps}
+    >
+      {header ? React.cloneElement(header, { ...header.props, ref: refHeader, onClose }) : null}
```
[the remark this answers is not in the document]
nahumzs: 🤔 why the would need a ref for the header? just curious to know if this is an edge case or is something that we really need to take in consideration?

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -0,0 +1,85 @@
+import { css, keyframes } from "styled-components";
+import tokens from "@paprika/tokens/lib/tokens";
+
+const space = Number.parseInt(tokens.space, 10);
+
+const gapChildPanel = space * 7;
+const childPanel = `
+    height:calc(100% - ${gapChildPanel + 80}px);
```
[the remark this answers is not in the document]
nahumzs: I have added an eslint to fix all this issue.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,193 @@
...
+
+    onAfterOpen();
+  };
+
+  // Effects
+  React.useEffect(() => {
+    document.addEventListener("keydown", handleEscKey, false);
+
+    if (props.isOpen) {
+      setMount(true);
```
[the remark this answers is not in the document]
nahumzs: https://github.com/acl-services/paprika/pull/84#discussion_r296409909

packages/SidePanel/src/components/Group/Group.js
```diff
@@ -0,0 +1,54 @@
...
+            ...sidePanel.props,
+            isInline: true,
+          })
+        )}
+      </div>
+    </React.Fragment>,
+    document.body
+  );
+
+  return [sidePanelsGroup];
```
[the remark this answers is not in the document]
nahumzs: 🧠💨

packages/SidePanel/src/components/FocusTrap/FocusTrap.js
```diff
@@ -0,0 +1,37 @@
...
+
+  /** Default: true. If false, when the trap is deactivated, focus will not return to the element that had focus before activation. */
+  returnFocusOnDeactivate: PropTypes.bool,
+};
+
+const defaultProps = FocusTrapLibrary.defaultProps;
+
+// shell component for enhancing UXDeveloper
+export default function FocusTrap() {
+  return null;
```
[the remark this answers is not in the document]
nahumzs: I move the library into this one and now I'm just rendering <FocusTrap /> in SidePanel.js, leaving it out of Dialog, I can imagine that the Dialog could potentially be rendered as a regular sidebar which doesn't need a FocusTrap

packages/SidePanel/src/SidePanel.js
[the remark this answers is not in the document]
rauldudu: should this be?

```
useEffect(() => setMount(props.isOpen), [props.isOpen]);
```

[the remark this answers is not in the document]
rauldudu: Sorry I think the code I posted won't work but I wonder if you should consider `useEffect` anyway...

packages/SidePanel/src/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as SidePanelGroup } from "./components/Group";
```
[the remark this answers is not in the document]
nahumzs: SidePanelGroup has multiple side panels and is not part of the SidePanel, but I guess the overlay is neither part of it. 

This actually was my first approach and then I didn't like it, but now that you are suggesting it maybe is the right approach, I will change it. 

<SidePanel.Group>
  <SidePanel></SidePanel>
  <SidePanel></SidePanel>
</SidePanel.Group>

packages/SidePanel/src/components/Header/Header.js
```diff
@@ -0,0 +1,57 @@
...
+  } = props;
+
+  return (
+    <div ref={ref} css={headerCSS} kind={kind} {...moreProps}>
+      <div tabIndex="-1">{props.children}</div>
+      <div>
+        {hasCloseButton ? (
+          <Button.Close
+            data-qa-anchor="sidepanel-header-close"
+            isSemantic={false}
```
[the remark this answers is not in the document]
nahumzs: because if I do it semantic some app in our platform will experience a "funny" bad style button because of the global styles.

packages/SidePanel/src/SidePanel.js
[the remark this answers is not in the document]
nahumzs: @rauldudu  https://github.com/acl-services/paprika/pull/84/files#diff-a70e4ad809affbf01bae992a555dc39eR116 <-- I have one or do you mean something different?

packages/SidePanel/src/hooks/useOffsetScroll.js
```diff
@@ -0,0 +1,25 @@
...
+  const handleScroll = () => {
+    const top = window.pageYOffset || document.documentElement.scrollTop;
+    if (top <= offsetY) {
+      setOffsetScroll(offsetY - top);
+    } else {
+      setOffsetScroll(0);
+    }
+  };
+
+  React.useLayoutEffect(() => {
```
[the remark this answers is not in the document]
nahumzs: During the advance hooks workshop, kent dodds demonstrated an [example](https://github.com/kentcdodds/advanced-react-hooks/blob/master/src/exercises/05.js#L8) pretty similar to this, he advice was if the effect will affect the visual element while interacting with the browser API, use Layout otherwise useEffect. 

That's the reason I'm using *useLayout*, also he even demonstrated that because react is really fast you can't perceive really a difference between useEffect vs useLayoutEffect in this case, that's the reason why he added a sleep function in [this line](https://github.com/kentcdodds/advanced-react-hooks/blob/master/src/exercises/05.js#L49) so the difference is more perceivable.

packages/SidePanel/src/hooks/useOffsetScrollFooter.js
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import debounce from "lodash.debounce";
+
+const useOffsetScrollFooter = (height, ref) => {
+  const [offsetY, setOffsetY] = React.useState(0);
+  const [opacity, setOpacity] = React.useState(1);
+
+  const handleOffset = debounce(() => {
```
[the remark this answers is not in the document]
nahumzs: The Sticky Footer definitely need more refinement to improve interaction experience, this debounce prevent to see the flickering happening all the time. I added a ticket so we can improve it later. https://github.com/acl-services/paprika/issues/97 I think for this first version the Footer is good enough.

packages/SidePanel/src/components/Header/Header.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return {};
+}
+
+const Header = React.forwardRef((props, ref) => {
+  const {
+    hasCloseButton,
+    kind,
+    // injected by Dialog.js
+    onClose, // eslint-disable-line
```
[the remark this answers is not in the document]
nahumzs: I don't think the consumer will even care about this prop will be overridden by the component isn't?

packages/SidePanel/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/SidePanel"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/heading": "^0.1.0",
```
mikrotron: `<SidePanel.Trigger>` and `<SidePanel.Header>` also depend on `<Button>`.

packages/SidePanel/readme.md
```diff
@@ -0,0 +1,59 @@
+# SidePanel
+
+## Props:
+
+```js
+const propTypes = {
```
mikrotron: Duplicating the props here in the `README.md` seems like too much work... although a quick listing might be handy, with a link to the source code, [like this](https://github.com/acl-services/paprika/blob/master/packages/Button/README.md)? I'm not so sure yet what the best approach will be.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,190 @@
...
+import useOffsetScroll from "./hooks/useOffsetScroll";
+import useBodyOverflow from "./hooks/useBodyOverflow";
+import useEscapeKey from "./hooks/useEscapeKey";
+
+const propTypes = {
+  /** The content for the SidePanel. */
+  children: PropTypes.node.isRequired,
+
+  /** The width of the open panel. */
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
mikrotron: Seems like alphabetic order will be the easiest standard to enforce, even if it's not always as logical.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,190 @@
...
+import Footer from "./components/Footer";
+import Header from "./components/Header";
+import Overlay from "./components/Overlay";
+import Trigger from "./components/Trigger";
+import Group from "./components/Group";
+import FocusTrap from "./components/FocusTrap";
+import { extractChildren } from "./helpers";
+import useOffsetScroll from "./hooks/useOffsetScroll";
+import useBodyOverflow from "./hooks/useBodyOverflow";
+import useEscapeKey from "./hooks/useEscapeKey";
```
mikrotron: Is one hook per file a rule?

packages/SidePanel/src/SidePanel.js
mikrotron: Also, I think we should definitely use semantic enums for `width`.  I think we should impose a limited number of options to enforce consistency with our design system.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,190 @@
...
+  zIndex: PropTypes.number,
+
+  /** Control y offset of the sidepanel */
+  offsetY: PropTypes.number,
+
+  /** Modify the look of the SidePanel */
+  kind: PropTypes.oneOf(["default", "child"]),
+
+  /** Disable scroll overflow on document.body when SidePanel is open. */
+  disableBodyOverflow: PropTypes.bool,
```
mikrotron: Is this necessary? Can we just _always_ hide the document scrollbars when an overlay is used?

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,190 @@
...
+    width,
+    isInline, // eslint-disable-line
+    kind,
+    offsetY,
+    isOpen,
+    ...moreProps
+  } = props;
+
+  // Hooks
+  const [isSidePanelVisible, setIsSidePanelVisible] = React.useState(props.isOpen);
```
mikrotron: Maybe just `isVisible`?

packages/SidePanel/src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,78 @@
...
+    refHeader,
+    refSidePanelContent,
+    width,
+    isOpen,
+    ...moreProps
+  } = props;
+
+  return (
+    <div
+      aria-modal="true"
```
mikrotron: Should this be dependent on `isInline` too? If the user can interact with the content below, it's not really a _modal_.

packages/SidePanel/src/SidePanel.js
```diff
@@ -0,0 +1,190 @@
...
+      </Dialog>
+    );
+
+    if (isInline) {
+      sidePanel = dialog;
+    } else {
+      sidePanel = ReactDOM.createPortal(
+        <React.Fragment>
+          <FocusTrap focusTrapOptions={focusTrapOptions}>
+            <div>{dialog}</div>
```
mikrotron: Does it really need the `<div>`?

packages/SidePanel/src/components/Dialog/Dialog.js
```diff
@@ -0,0 +1,78 @@
...
+const propTypes = {
+  children: PropTypes.node.isRequired,
+  footer: PropTypes.node,
+  header: PropTypes.node,
+  isInline: PropTypes.bool,
+  offsetY: PropTypes.number,
+  onAnimationEnd: PropTypes.func.isRequired,
+  onClose: PropTypes.func,
+  refHeader: PropTypes.shape({ current: PropTypes.instanceOf(Element) }).isRequired,
+  refSidePanelContent: PropTypes.shape({ current: PropTypes.instanceOf(Element) }).isRequired,
```
mikrotron: `refContent`? In general, including `SidePanel` in this component's names seems redundant, and shorter names will be easier to read.

packages/SidePanel/src/hooks/useBodyOverflow.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+
+export default function useBodyOverflow(disableBodyOverflow) {
+  React.useEffect(() => {
+    if (disableBodyOverflow) {
+      document.body.style.overflow = "hidden";
+      return;
+    }
+
+    document.body.style.overflow = "auto";
```
mikrotron: I think I got into some trouble with the old SidePanel over this... We had a SidePanel in a Takeover... the Takeover already set the overflow to hidden, but when the SidePanel closed it would restore the overflow to auto and screw up the Takeover.  It's important to restore the overflow as it was, not assume it should be `auto`.

mikrotron: https://github.com/acl-services/acl-ui/blob/master/src/components/SidePanel/SidePanel.js#L178

packages/SidePanel/src/components/Overlay/Overlay.js
```diff
@@ -0,0 +1,45 @@
...
+  onClose: PropTypes.func,
+  /** Control the z position of the sidepanel overlay */
+  zIndex: PropTypes.number,
+};
+
+const defaultProps = {
+  background: "#000",
+  hasOutsideClick: true,
+  onClose: null,
+  zIndex: 99,
```
mikrotron: `z(n)` where `n <= 7`.

packages/SidePanel/src/components/Header/Header.styles.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+const space = Number.parseInt(tokens.space, 10);
+export const headerCSS = css`
+  border-bottom: 1px solid ${tokens.border.color};
+  box-sizing: border-box;
+  color: ;
+  display: flex;
+  justify-content: space-between;
+  padding: 12px ${space}px; // creates a 48px height
```
mikrotron: How about:
```css
min-height: ${stylers.spacer(6)};
height: ${stylers.spacer(6)};
```

packages/SidePanel/src/components/Header/Header.styles.js
```diff
@@ -0,0 +1,40 @@
...
+const kind = {
+  default: `background: ${tokens.color.white}; color: ${tokens.color.black};`,
+  primary: `background: ${tokens.color.purple}; color: ${tokens.color.white};`,
+};
+
+const space = Number.parseInt(tokens.space, 10);
+export const headerCSS = css`
+  border-bottom: 1px solid ${tokens.border.color};
+  box-sizing: border-box;
+  color: ;
```
mikrotron: ✂️

packages/SidePanel/src/components/Header/Header.styles.js
```diff
@@ -0,0 +1,40 @@
...
+    outline: 0;
+  }
+
+  .heading--level-1,
+  .heading--level-2,
+  .heading--level-3,
+  .heading--level-4,
+  .heading--level-5 {
+    ${fontSize()};
+    font-weight: 400;
```
mikrotron: This concerns me. If we want to enforce the styling of the heading, maybe we can do it by just taking the content as a `string` and putting the `<Heading displayLevel={5} isLight>` component in ourselves.

I'm also starting to regret that we haven't implemented some kind of selector on the root node for each type of component... maybe not a `className`, but something would be convenient.

packages/SidePanel/src/hooks/useOffsetScroll.js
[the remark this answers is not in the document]
alexzherdev: That example is different—it really does "affect the visual element" directly, while yours only sets a listener which has no (immediate) visual implications. In that case, `useEffect` is recommended since it allows React to schedule it better and reduce the workload in the commit phase.

packages/SidePanel/src/components/Header/Header.js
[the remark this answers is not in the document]
alexzherdev: Right, but for the developers working on this component we still want to make sure that the props are wired correctly

packages/SidePanel/src/SidePanel.js
nahumzs: I thought about that, but I think the component shouldn't enforce the width from this side. Would make more sense to provide a specific token that the developer can use and pass them to this prop. `<SidePanel width={tokens.sidepanel.width.half} />` as an example. What do you think? 

We can discuss this maybe in on our slack channel?

packages/SidePanel/src/SidePanel.js
nahumzs: nop you can use multiple hooks in a file like:

const [user, setUser] = React.useState(null);
const [security, setSecurity] = React.useState(null);

packages/SidePanel/src/SidePanel.js
nahumzs: 🤔 good point. Not 100% about this though, can you confirm if you want this or just a suggestion?

packages/SidePanel/src/SidePanel.js
nahumzs: yes it does, the FocusTrap needs to have only 1 element as children can't have two neither a `<Fragment />`

packages/SidePanel/src/components/Header/Header.styles.js
nahumzs: thank you! you found the damn null that I was seeing in the dev console and couldn't found! this makes my soul rest...

packages/SidePanel/src/components/Header/Header.styles.js
nahumzs: That was my suggestion for `[data-component-type="listbox.header"]`

To me, that would super helpful you could target each component via jss, css or scss. as well in testing. even theme the components would be easier

packages/SidePanel/src/SidePanel.js
mikrotron: I'd like to try it without.  It's easier to add a new prop later, than deprecate one.

packages/SidePanel/src/SidePanel.js
mikrotron: No, I mean the way `useEscapeKey` is the only hook exported from `hooks/useEscapeKey`.  This is preferred to `import { hook1, hook2 } from "./hooks/radicalHooks"`?

packages/SidePanel/src/components/Header/Header.styles.js
mikrotron: Yeah.  Let's make an issue to add these to all our components maybe? Then if there's a bigger conversation, it'll be separate from your PR.

packages/SidePanel/readme.md
nahumzs: npm provide a direct link from the npm landing page of the component, as long we use the url property under repository on the package.json file.

packages/SidePanel/src/SidePanel.js
nahumzs: We can merge and then modify this after we agree in a pattern?

mikrotron: Agreed.  This shouldn't block the whole component.  I think it's important to follow up and decide on a pattern soon though.

packages/SidePanel/src/components/Header/Header.styles.js
nahumzs: https://github.com/acl-services/paprika/issues/101 minor for this PR for now.

nahumzs: btw about your previous comment? can we assume `displayLevel={5}` on the sidepanel? 🤔

packages/SidePanel/src/components/Dialog/Dialog.js
nahumzs: mmm actually is the ref for the SidePanel inside of the Dialog.js component. Makes sense to specify the name.

packages/SidePanel/src/components/Overlay/Overlay.js
nahumzs: this doesn't work because z returns a full string `z-index: 1` I have added a new function under z named `z.toNumber(6)` which returns only the number that we need.

packages/SidePanel/src/helpers.js
```diff
@@ -0,0 +1,27 @@
...
+    React.Children.toArray(children).forEach(child => {
+      if (child.type && types.includes(child.type.componentType)) {
+        if (Object.prototype.hasOwnProperty.call(components, child.type.componentType)) {
+          const childs = Array.isArray(components[child.type.componentType])
+            ? [child, ...components[child.type.componentType]]
+            : [child, components[child.type.componentType]];
+
+          components[child.type.componentType] = childs;
+        } else {
+          components[child.type.componentType] = child;
```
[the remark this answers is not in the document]
nahumzs: don't think so because I'm checking if the value of components['somevalue'] is an array. In case is not I want to keep just 1 component there instead of converting it into an array.

packages/SidePanel/src/hooks/useOffsetScrollFooter.js
```diff
@@ -0,0 +1,35 @@
...
+    setOffsetY(-top);
+  }, 350);
+
+  const handleScrolling = () => {
+    setOpacity(0);
+    handleOffset();
+  };
+
+  React.useEffect(() => {
+    if (ref.current) {
```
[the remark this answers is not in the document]
nahumzs: I tried this, but since ref.current is null at the beginning the effect will not get executed, passing ref.current allowed to listen for current and when change applies the effect.

packages/SidePanel/src/hooks/useOffsetScroll.js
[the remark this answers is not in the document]
nahumzs: `useEffect`
Screen Shot 2019-06-26 at 2 29 15 PM

`useLayoutEffect`
Screen Shot 2019-06-26 at 2 29 44 PM

see how the Sidepanel is not a `0px` when scrolling? and in the second example when scrolling is actually at `0px`. 

Changing to useEffect will never update cypress on the scrolling event, useLayoutEffect occurs before useEffect in the cycle and is synchronous so will update the element and cypress can observe the changes.

packages/SidePanel/src/components/Header/Header.js
```diff
@@ -0,0 +1,59 @@
...
+    hasCloseButton,
+    kind,
+    // injected by Dialog.js
+    onClose,
+    ...moreProps
+  } = props;
+
+  return (
+    <div ref={ref} css={headerCSS} kind={kind} {...moreProps}>
+      <div tabIndex="-1">{props.children}</div>
```
mikrotron: I'm not going to suggest we change this now, but I might want to try extending this to be easier to consume.  `children` can remain in a raw container, but if props like `heading`, `onNext`, `onPrev` are provided, I think it would be nice to actually render with a `<Heading>` or `<PrevNextArrows>`.  What do you think?

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -0,0 +1,79 @@
+import { css, keyframes } from "styled-components";
+import tokens from "@paprika/tokens/lib/tokens";
+
+const space = Number.parseInt(tokens.space, 10);
+
+const gapChildPanel = space * 7;
+const childPanel = css`
+  border-bottom-left-radius: ${space / 2}px;
+  border-top-left-radius: ${space / 2}px;
+  height: calc(100% - ${gapChildPanel + 80}px);
```
mikrotron: What is `80`?

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -0,0 +1,79 @@
+import { css, keyframes } from "styled-components";
+import tokens from "@paprika/tokens/lib/tokens";
+
+const space = Number.parseInt(tokens.space, 10);
+
+const gapChildPanel = space * 7;
+const childPanel = css`
+  border-bottom-left-radius: ${space / 2}px;
```
mikrotron: `spaceSm`

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -0,0 +1,79 @@
...
+      z-index: ${props.zIndex};
+      ${props.isInline ? "position: relative;" : "position: fixed;"}
+      ${props.isOpen ? "opacity: 1" : "opacity: 0"};
+      ${childSidePanel}
+    `;
+  }}
+`;
+
+export const dialogContentStyles = css`
+  padding: ${space * 2}px;
```
mikrotron: `stylers.spacer(2)`

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -0,0 +1,79 @@
+import { css, keyframes } from "styled-components";
+import tokens from "@paprika/tokens/lib/tokens";
+
+const space = Number.parseInt(tokens.space, 10);
```
mikrotron: Shouldn't need this.

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  to {
+    opacity: 0;
+    transform: translateX(100%);
+  }
+  `;
+}
+
+export const dialogStyles = css`
+  background: #fff;
```
mikrotron: token

packages/Stylers/src/helpers.js
```diff
@@ -27,10 +27,17 @@ export const lineHeight = (scale = 0) => `
...
   const validLevels = ["auto", "initial", "inherit", "-1", "0", "1", "2", "3", "4", "5", "6", "7"];
+  return validLevels.includes(level);
+}
+
+export const z = level => {
+  return zIsValid(level) ? `z-index: ${level};` : `z-index: 1;`;
+};
 
-  return validLevels.includes(level) ? `z-index: ${level};` : `z-index: 1;`;
+z.toNumber = level => {
```
mikrotron: Clever, but do you think it's better to be more consistent with other helpers and go with `z()` and `zValue()`?

https://github.com/acl-services/paprika/blob/master/packages/Stylers/src/helpers.js#L9-L13
https://github.com/acl-services/paprika/blob/master/packages/Stylers/src/helpers.js#L18-L22

packages/SidePanel/src/components/Header/Header.js
nahumzs: 🤔 ohh you mean if I do something like

```
<SidePanel>
  <SidePanel.Header onNext onPrev>some stuff</SidePanel.Header>
</SidePanel> 
```

will render the `<PrevNextArrow />` that's a cool idea! yea totally out of scope for this PR

packages/SidePanel/src/components/Dialog/Dialog.styles.js
nahumzs: To adjust the position of the child sidepanel so is not exactly at the same level of the Parent, I totally picked by `eye` there is not a hard guideline for this, later we can remove/add create a token if we need it.

packages/SidePanel/src/hooks/useFooterOffset.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import debounce from "lodash.debounce";
+
+const useFooterOffset = (height, ref = null, isSticky) => {
```
mikrotron: This approach seems way more complicated than simply using a different (flexbox) layout when the `isSticky` prop is true.  No handlers, no debouncing... just flex-shrink, flex-grow, and overflow.

mikrotron: Also... it still looks janky.  It floats and then reappears.  😕

packages/SidePanel/src/components/Dialog/Dialog.styles.js
mikrotron: Oh, I see, theres a "gap" at the top AND bottom... and they're different... 🤨
Still, odd that one has a descriptive variable (`gapChildPanel `) and the other is a literal.

packages/Stylers/src/helpers.js
nahumzs: 🤔 Value doesn't describe that you are getting a number you could be getting `1px` `1vh` `1%` `1` or `{1}` right?

packages/SidePanel/src/hooks/useFooterOffset.js
nahumzs: I added a comment about that the footer needs an upgrade, I added it as the last feature, this will not affect the API and is easier to refactor it in a new PR. 

I don't think we should stop the whole SidePanel only for the footer.

nahumzs: I created an issue for this already https://github.com/acl-services/paprika/issues/97

mikrotron: Okay, I agree.

alexzherdev: Huh... I thought there was a specific reason we're not using flexbox here.

packages/SidePanel/src/components/Dialog/Dialog.styles.js
nahumzs: 🤷‍♂

package.json
```diff
@@ -59,11 +59,12 @@
     "@paprika/stylers": "^0.1.0",
     "@paprika/textarea": "^0.1.1",
     "@paprika/tokens": "^0.1.0",
-    "@storybook/addon-actions": "^5.1.9",
-    "@storybook/addon-knobs": "^5.1.9",
-    "@storybook/addon-links": "^5.1.9",
-    "@storybook/addons": "^5.1.9",
-    "@storybook/react": "^5.1.9",
+    "@storybook/addon-actions": "^5.0.11",
```
alexzherdev: Hmm 😕

packages/SidePanel/src/hooks/useOffsetScroll.js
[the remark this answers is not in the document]
alexzherdev: I'd say that's probably a cypress issue then. But not going to block on this for sure

PR #90, opened 2019-06-13
UX-489 Tabs Component

oscarkwan:
### 🛠 Purpose

Tabs Component

## Components
The tabs has the following components:
```js
<Tabs.List />
<Tabs.Tab />
<Tabs.Panels />
<Tabs.Panel />
```

## The architecture
--- 

## The Ex.

```js
<Tabs>
  <Tabs.List>
    <Tabs.Tab label="Hello" />
    <Tabs.Tab label="World" />
  </Tabs.List>
  <Tabs.Panels>
    <Tabs.Panel>Hello</Tabs.Panel>
    <Tabs.Panel>World</Tabs.Panel>
  </Tabs.Panels>
</Tabs>
```

To use tabs as links, simply pass the `linkUrl` prop to the `<Tabs.Tab />` component.

## Accessibility
- aria-selected to show what tab is selected
- When a tab is selected the following states are toggled: aria-selected in both tab elements and aria-hidden on both tab panels.
- The tabs can also toggle using the left and right arrow keys 
- When clicking Home, it will go to the first available tab
- When clicking End, it will go to the last available tab


nahumzs:
What about removing the word `Tab` from **TabList** and `List` from **PanelList**
```js
<Tabs>
  <Tabs.List>
    <Tabs.Tab label="Hello" />
    <Tabs.Tab label="World" />
  </Tabs.List>
  <Tabs.Panels>
    <Tabs.Panel>Hello</Tabs.Panel>
    <Tabs.Panel>World</Tabs.Panel>
  </Tabs.Panels>
</Tabs>
```

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,100 @@
...
+        tabIndex={tabIndex}
+        {...moreProps}
+      >
+        {label}
+      </a>
+    );
+  }
+
+  return (
+    <RawButton
```
nahumzs: We have starting to use `<Button isSemantic={false} />` to describe the `<RawButton />` so instead of importing RawButton let's import Button.

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,100 @@
...
+      aria-selected={isSelected}
+      css={tabStyles}
+      className={cn}
```
nahumzs: 🤔 wouldn't more sense to passdown isSelected and _isDisabled to the `css={tabSyles}`? instead of using a className?  to dictate how they should look like?

packages/Tabs/src/components/List/List.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const childrenWithProps = React.Children.map(children, (tab, index) => {
+    const isSelected = activeIndex === index;
+
+    setNumberOfTabs(React.Children.count(children));
+
+    return React.cloneElement(tab, {
+      isSelected,
+      onClick: e => handleTabClick(e, index),
+      onKeyDownArrows: e => handleKeyDown(e),
```
tristanjasper: can this just be `onKeyDownArrows: handleKeyDown`?

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const handleKeyDown = index => {
+    let tabIndex = index;
+    const totalTabs = numberOfTabs;
+
+    if (index < 0) {
+      tabIndex = numberOfTabs - 1;
+    } else if (index >= totalTabs) {
+      tabIndex = 0;
+    }
```
AndreyChernykh: Why this check is necessary? Is it possible to get index `-1` or bugger than `numberOfTabs` ?

packages/Tabs/src/components/List/List.js
```diff
@@ -0,0 +1,45 @@
...
+    });
+  });
+
+  if (a11yText) moreProps["aria-label"] = a11yText;
```
AndreyChernykh: Changing `props` is not a good pattern. 

```
return (
    <div css={tabsListStyles} 
  role="tablist" 
  ref={ref => setTabListRef(ref)} 
  {...moreProps} 
  {...(a11yText ? {"aria-label": a11yText} : null)}>
      {childrenWithProps}
    </div>
  )
```

AndreyChernykh: and why can't we just do: 

```js
<div aria-label={a11yText || undefined} /> 
```
or something?

packages/Tabs/src/components/Panel/Panel.js
```diff
@@ -0,0 +1,33 @@
...
+
+const Panel = props => {
+  const { isSelected, children, ...moreProps } = props;
+
+  if (!isSelected) {
+    return null;
+  }
+
+  return (
+    <div {...moreProps} css={panelStyles} role="tabpanel">
```
AndreyChernykh: @alexzherdev @nahumzs @oscarkwan 

I also don't really understand all that thing with `{...moreProps}`. For some components we use them, for the other - not. I don't see the pattern. 

And I feel kinda iffy about this approach. Isn't it making the component API more wobbly?

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,70 @@
...
+
+    tabListRef.querySelectorAll(".tab")[tabIndex].focus();
```
allison-c: maybe `[role="tab"]` ?

packages/Tabs/src/components/List/List.js
alexzherdev: Mutating `moreProps` does not mutate the actual `props` object tho

packages/Tabs/src/components/List/List.styles.js
```diff
@@ -0,0 +1,11 @@
+import tokens from "@paprika/tokens";
+
+const tabsListStyles = () => `
```
alexzherdev: Can we use `css` here?

packages/Tabs/src/components/Panel/Panel.js
alexzherdev: I mean... we're supposed to use it wherever it makes sense for the consumer to put additional props on the element? If we're not doing it somewhere, that's probably because we haven't found that need yet.

packages/Tabs/src/components/Panel/Panel.styles.js
```diff
@@ -0,0 +1,7 @@
+import stylers from "@paprika/stylers";
+
+const panelStyles = () => `
```
alexzherdev: Here as well

packages/Tabs/src/components/Panel/Panel.styles.js
```diff
@@ -0,0 +1,7 @@
...
+  padding-top: ${stylers.spacer(2)}
```
alexzherdev:
```suggestion
  padding-top: ${stylers.spacer(2)};
```

packages/Tabs/stories/examples/Showcase.js
```diff
@@ -0,0 +1,34 @@
...
+
+const ExampleStory = props => (
+  <Story>
+    <Heading level={1} displayLevel={2} isLight>
+      Tabs Showcase
+    </Heading>
+    <Tagline>Use the knobs to tinker with the props.</Tagline>
+    <Rule />
+    <Tabs {...props}>
+      <Tabs.List>{renderTabs(false)}</Tabs.List>
```
alexzherdev: Shouldn't the showcase story also include panels?

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,70 @@
...
+  children: PropTypes.node.isRequired,
+  isDisabled: PropTypes.node,
+};
+
+const defaultProps = {
+  isDisabled: false,
+};
+
+const Tabs = props => {
+  const [activeIndex, setActiveIndex] = React.useState(null);
```
alexzherdev: Hmm. It looks like it's impossible for me to render tabs with a specific tab pre-selected? It always starts out with no tab selected because it's uncontrolled and this default value is `null`.

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,100 @@
...
+    isSelected,
+    label,
+    linkUrl,
+    onClick,
+    onKeyDown,
+    onKeyDownArrows,
+    ...moreProps
+  } = props;
+
+  const _isDisabled = context.isDisabled || props.isDisabled;
```
alexzherdev: I'm not a fan of dangling underscores but up to others. I'd just call this `isDisabled`.

packages/Tabs/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/raw-button": "^0.1.0",
+    "@paprika/stylers": "^0.1.0",
+    "@paprika/tokens": "^0.1.0",
+    "classnames": "^2.2.6",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "react": "^16.4.0",
```
alexzherdev:
```suggestion
    "react": "^16.8.0",
```
bump for hooks?

packages/Tabs/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "react-dom": "^16.4.2",
```
alexzherdev:
```suggestion
    "react-dom": "^16.8.0",
```

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,70 @@
...
+    const totalTabs = numberOfTabs;
+
+    if (index < 0) {
+      tabIndex = numberOfTabs - 1;
+    } else if (index >= totalTabs) {
+      tabIndex = 0;
+    }
+
+    tabListRef.querySelectorAll(".tab")[tabIndex].focus();
+    setActiveIndex(tabIndex);
```
alexzherdev: This is currently not taking into account disabled tabs - you can keyboard-navigate to them.

packages/Tabs/src/components/Panels/Panels.js
```diff
@@ -0,0 +1,25 @@
...
+
+  const childrenWithProps = React.Children.map(children, (panel, index) => {
+    const isSelected = context.activeIndex === index;
+
+    return React.cloneElement(panel, {
+      isSelected,
+    });
+  });
+
+  return <div>{childrenWithProps}</div>;
```
alexzherdev: Do we actually need this div? Or could it be a fragment?

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,100 @@
...
+    const leftArrowKey = 37;
+    const rightArrowKey = 39;
+
+    if (event.which === leftArrowKey) {
+      onKeyDownArrows(context.activeIndex - 1);
+    } else if (event.which === rightArrowKey) {
+      onKeyDownArrows(context.activeIndex + 1);
+    }
+
+    onKeyDown(event);
```
alexzherdev: I don't think this is used

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,100 @@
...
+import RawButton from "@paprika/raw-button";
+import TabsContext from "../../TabsContext";
+import tabStyles from "./Tab.styles";
+
+const propTypes = {
+  className: PropTypes.string,
+  label: PropTypes.node,
+  isSelected: PropTypes.bool,
+  isDisabled: PropTypes.bool,
+  linkUrl: PropTypes.string,
```
alexzherdev: Should we just call it `href`?

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+const defaultProps = {
+  isDisabled: false,
+};
+
+const Tabs = props => {
+  const [activeIndex, setActiveIndex] = React.useState(null);
+  const [numberOfTabs, setNumberOfTabs] = React.useState(null);
+  let tabListRef = null;
```
alexzherdev: Why not `useRef`?

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,70 @@
...
+  const handleTabClick = (event, index) => {
+    event.preventDefault();
+    if (activeIndex !== index) {
+      setActiveIndex(index);
+    }
+  };
+
+  const handleKeyDown = index => {
+    let tabIndex = index;
+    const totalTabs = numberOfTabs;
```
alexzherdev: Why reassign to a new variable?

packages/Tabs/src/components/List/List.js
AndreyChernykh: Oh, you're right! My bad)

packages/Tabs/src/components/List/List.js
```diff
@@ -0,0 +1,45 @@
...
+const List = props => {
+  const context = React.useContext(TabsContext);
+
+  const { a11yText, children, ...moreProps } = props;
+  const { activeIndex, handleKeyDown, handleTabClick, setNumberOfTabs, setTabListRef } = context;
+
+  const childrenWithProps = React.Children.map(children, (tab, index) => {
+    const isSelected = activeIndex === index;
+
+    setNumberOfTabs(React.Children.count(children));
```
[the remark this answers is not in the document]
AndreyChernykh: Good catch!

packages/Tabs/src/components/Tab/Tab.js
oscarkwan: It gives a lot of weird styling, that i don't need if i use `Button`

packages/Tabs/src/Tabs.js
oscarkwan: because i am just incrementing tabIndex, therefore it could be bigger than total tabs, so that means we want to loop back to the first tab.

packages/Tabs/src/components/Panels/Panels.js
oscarkwan: Can use fragment :) thanks for the suggestion!

packages/Tabs/src/Tabs.js
oscarkwan: To be flexible, it may not always be a role tab. For example, if the tabs are used for navigation purposes then the consumer will pass in a role manually, so the role is not a unique value

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,87 @@
...
+  const [activeIndex, setActiveIndex] = React.useState(props.defaultIndex);
+  const [currentFocusIndex, setFocusIndex] = React.useState(null);
+
+  let tabListRef = React.useRef(null);
+
+  const { isDisabled } = props;
+
+  function focusAndSetIndex(index) {
+    tabListRef.querySelectorAll(".tab")[index].focus();
+    setFocusIndex(index);
```
oscarkwan: This is for manual activation, We don't want to use automatic activation for tabs

packages/Tabs/tests/Tabs.cypress.js
```diff
@@ -0,0 +1,23 @@
+const baseStory = "tabs-automation-tests";
+const testStory = "cypress";
+
+describe("<Tabs />", () => {
+  beforeEach(() => {
+    cy.visitStorybook(`${baseStory}--${testStory}`);
+  });
+
+  // TODO: Fix test
+  it("should be navigitable with tabbing", () => {
```
alexzherdev: _navigitable_

oscarkwan:
@alexzherdev Through some discussion, we decided that it will not focus on the Content of the tab (the whole wrapper div). Instead we will just focus on the next focusable element inside of the panel.

Rationale: We don't want screen reader to read out all the contents inside a tab, especially in our use case where tabs will contain a lot of information.

packages/Tabs/README.md
```diff
@@ -0,0 +1,70 @@
...
+`> yarn add @paprika/tabs`
+
+### Usage
+
+```js
+import Tabs from "@paprika/tabs";
+
+<Tabs>
+  <Tabs.List>
+    <Tabs.Tab label="Hello" />
```
nahumzs: @oscarkwan I believe this should be on the children prop and no as label prop, if we use it as a children is clear that you can pass whatever you want if you use it as a label gives the impression that only received a string, 

Also in need of passing a complex component over the label prop will be end looking strange, for some developer:

`<Tabs.Tab label={<span><Icon /><Text /></span>} />`

oscarkwan: Done, thanks!

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,78 @@
...
+  children: null,
+  className: null,
+  isSelected: false,
+  isDisabled: false,
+  href: null,
+  onClick: () => {},
+  onKeyDownArrows: () => {},
+};
+
+const Tab = props => {
```
nahumzs: `export default function Tab(props) {}` please.

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,76 @@
...
+        {...moreProps}
+      >
+        {children}
+      </a>
+    );
+  }
+
+  return (
+    <RawButton
+      aria-disabled={isDisabled}
```
mikrotron: `<RawButton>` should already take care of `aria-disabled` and `tabIndex` for you.  The `<a role='tab'>` case would probably benefit from this property though... but since `<a>` doesn't really support that attribute, maybe when it's disabled it should be rendered as a `<span>` instead?

packages/Tabs/src/Tabs.js
```diff
@@ -0,0 +1,88 @@
...
+};
+
+const defaultProps = {
+  defaultIndex: 0,
+  isDisabled: false,
+};
+
+const Tabs = props => {
+  const [activeIndex, setActiveIndex] = React.useState(props.defaultIndex);
+  const [currentFocusIndex, setFocusIndex] = React.useState(props.defaultIndex);
```
mikrotron: Something is a little off with the `currentFocusIndex`... when your focus leaves the tabs, then comes back to the tabs, the focus should be reset to this value (currently it seems to reset to the first tab again, but then when you press an arrow key, it moves the focus relative the saved state of `currentFocusIndex`).

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  const isDisabled = context.isDisabled || props.isDisabled;
+  const handleClick = isDisabled ? () => {} : onClick;
+  const tabIndex = isSelected ? 0 : -1;
+
+  if (href) {
+    return (
+      <a
+        css={tabStyles}
```
mikrotron: Since the focus ring styling is coming from `<RawButton>`, when the tab is a link, the focus ring styling is inconsistent (default browser styling).

PR #92, opened 2019-06-18
PD-12484 Introduce Sortable component 🐐

mikrotron:
### 🛠 Purpose
Add a new `<Sortable>` component for use with re-orderable lists.

### ✏️ Notes
- Implements Atlassian's `react-beautiful-dnd`:
  https://github.com/atlassian/react-beautiful-dnd/blob/master/README.md
- There are two "Item" subcomponents:
  - `<SortableItem>` (sortableItem.js) – the implementation of the Draggable from react-beautiful-dnd
  - `<Sortable.Item>` (item.js) – a basic container to hold (future) props and enforce the desired API
- Storybook: added a Showcase story with deliberate knobs, Examples (to help with usage), a Screener story and a Cypress story (for later). 
- More info in the [README.md](https://github.com/acl-services/paprika/blob/411e53933ca7c08ef9bde52dfed42f873195b1f2/packages/Sortable/README.md)

### 🌟 Bonus
- Updates Storybook to `v5.1`.
- Minor class / style tweaks to `<Input>` and `<Textarea>`.

### 🚧 To Do
- [ ] Tests (in a follow-up PR)
- [ ] Assess performance

### 📽 Screencap
sortables


packages/Sortable/src/components/Item/Item.js
```diff
@@ -0,0 +1,55 @@
...
+        return (
+          <li
+            {...provided.draggableProps}
+            {...provided.dragHandleProps}
+            css={itemStyles}
+            data-is-dragging={snapshot.isDragging ? true : undefined}
+            isDragging={snapshot.isDragging}
+            ref={provided.innerRef}
+          >
+            <div css={itemHandleStyles} {...{}}>
```
mikrotron: This is gross.  But it seems necessary due to our Babel preset... I believe that the `transform-react-constant-elements` plugin conflicts with the `styled-components` plugin that utilizes the `css` prop.

https://github.com/styled-components/babel-plugin-styled-components/issues/221

And unless we configure our own plugins and skip the preset, it seems we either don't use the `css` prop in these kinds of situations... or we insert this weird (but easily searched + removed) hack.

https://stackoverflow.com/questions/33707189/is-there-an-option-to-disable-a-plugin-in-babel-when-using-a-preset

mikrotron: cc @tristanjasper, @alexzherdev

packages/Icon/src/DragHandle.js
```diff
@@ -0,0 +1,26 @@
...
+
+const SvgDragHandle = ({ title, ...props }) => (
+  <svg
+    width="1em"
+    height="1em"
+    viewBox="0 0 24 24"
+    css={`
+      color: ${props.color};
+      width: ${props.size};
+      height: ${props.size};
```
jamiek-acl: I dont see any defaults, and you arent passing in these props... are they just omitted if null?

packages/Sortable/README.md
```diff
@@ -0,0 +1 @@
+Sortable items are rad.
```
jamiek-acl: TLDR

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,70 @@
+import React from "react";
+import PropTypes from "prop-types";
+import uuidv4 from "uuid/v4";
+import { DragDropContext, Droppable } from "react-beautiful-dnd";
+import SortableContext from "./SortableContext";
+import Item from "./components/Item/Item";
```
jamiek-acl: Should the `Item` component live under `Sortable`? It wouldnt be used without it.

packages/Input/src/Input.styles.js
```diff
@@ -19,7 +19,7 @@ const iconStyles = {
   `,
 };
 
-const inputStyles = props => `
+const inputStyles = ({ hasClearButton, size }) => `
```
nahumzs: this seems a little off, wouldn't make more sense to use 
```js
const inputStyles = css`
   ${props => {
        props.hasClearButon; props.size;
     }}
`
```

packages/Sortable/README.md
nahumzs: can you add a simple example of how to use the component, this will appear on npm page.

packages/Sortable/src/components/Item/Item.js
jamiek-acl: Remove `{...{}}` in a few places

packages/Sortable/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/icon": "^0.1.0",
+    "@paprika/stylers": "^0.1.0",
+    "@paprika/tokens": "^0.1.0",
+    "classnames": "^2.2.6",
```
nahumzs: why we need className?

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,70 @@
...
+  onRemove: PropTypes.func,
+};
+
+const defaultProps = {
+  children: null,
+  hasIndexes: true,
+  onRemove: null,
+};
+
+function augmentChildren(children) {
```
nahumzs: we have been using `extractChildren` or maybe this is more like cloneChidren?.

packages/Sortable/README.md
alexzherdev: Or at least a bad "sorta" pun.

packages/Sortable/stories/examples/Basic.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import { Story } from "storybook/assets/styles/common.styles";
+import { storyStyles } from "../Sortable.stories.helpers";
+import Sortable from "../../src";
+
+const Example = () => {
+  return (
+    <Story css={storyStyles}>
+      <Sortable onChange={() => {}} onRemove={() => {}}>
+        <div>Item One</div>
```
jamiek-acl: I wonder if you'd ever want to have some children that are not sortable...

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,70 @@
...
+  const handleDragEnd = result => {
+    const { source, destination } = result;
+    onChange({
+      source: source.index,
+      destination: destination ? destination.index : null,
+    });
+  };
+
+  return (
+    <SortableContext.Provider value={{ hasIndexes, onRemove }}>
```
nahumzs: const value = {hasIndexes, onRemove};
<SortableContext.Provider value={value}>

https://reactjs.org/docs/context.html#caveats
> assigning a value directly might create an undesirable rerender.

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,70 @@
...
+        <Droppable droppableId={`droppable-${uuidv4()}`}>
+          {(provided, snapshot) => (
+            <ul
+              {...provided.droppableProps}
+              css={sortableStyles}
+              data-is-dragging-over={snapshot.isDraggingOver ? true : undefined}
+              isDraggingOver={snapshot.isDraggingOver}
+              ref={provided.innerRef}
+            >
+              {augmentedChildren &&
```
nahumzs: I don't think you need `augmentedChildren &&` map will return an empty array if can't do `React.Children.map`  so is safely do augmentedChildren.lenght.

packages/Sortable/src/components/Item/Item.styles.js
```diff
@@ -0,0 +1,65 @@
...
+const draggingStyles = `
+  background: ${tokens.color.white};
+  box-shadow: ${tokens.shadow}, 0 0 1px ${tokens.border.color};
+
+  &:focus {
+    box-shadow: ${tokens.shadow}, ${tokens.highlight.active.withBorder.boxShadow};
+  }
+`;
+
+export const itemStyles = props => `
```
nahumzs: wouldn't be more "styled-components" approach to use **css``** instead of pass a function with props? the previous can be confused with a definition of a Functional component.

packages/Sortable/stories/examples/Basic.js
nahumzs: what would happen if I do:
```js
<Sortable>
  <MyComponent>one</MyComponent>
  <MyComponent>two</MyComponent>
  <MyComponent>thre</MyComponent>
</Sortable>
```

would that work?

packages/Sortable/src/components/Item/Item.js
```diff
@@ -0,0 +1,55 @@
...
+    <Draggable draggableId={`draggable-${child.props["data-drag-id"]}`} index={index}>
+      {(provided, snapshot) => {
+        return (
+          <li
+            {...provided.draggableProps}
+            {...provided.dragHandleProps}
+            css={itemStyles}
+            data-is-dragging={snapshot.isDragging ? true : undefined}
+            isDragging={snapshot.isDragging}
+            ref={provided.innerRef}
```
nahumzs: I can't find where the reference for  `ref={provide.innerRef}` is being used can you explain to me a little bit about why we are using it :)

packages/Sortable/package.json
nahumzs: seem like this package should be add to TextArea and input but not in Sortable?

packages/Sortable/src/SortableContext.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+
+const SortableContext = React.createContext();
```
nahumzs: 🤔 no sure if this worth to have their own file for 1 line of code. My `giveAFu`  thermometer is low on this.

packages/Sortable/stories/Sortable.stories.helpers.js
```diff
@@ -0,0 +1,103 @@
...
+  "Sustainable succulents",
+];
+
+const selectOptions = [<option value="Dr">Dr</option>, <option value="Mr">Mr</option>, <option value="Ms">Ms</option>];
+
+const MultiBox = styled.div`
+  display: flex;
+  flex-grow: 1;
+
+  > div + div {
```
nahumzs: the power of selectors 💪 lol!!! `div > div + div + span * { }`

packages/Sortable/stories/examples/Basic.js
nahumzs: To me sound like if we have a <Sortable.Item> subcomponent will be clearer to the consumer what to do:

if the component accepts children then a consumer might do:
```js
<Sortable>
 { ["one", "two", "three"].map(i => i) }
</Sortable>
```
___ or ___

```js
<Sortable>
 { ["one", "two", "three"].map(i => <span>i</span>) }
</Sortable>
```

both cases are inline elements don't know if the drag and drop will work on those situations without being wrapped in a div.

To me is more idiomatic to have:

```js
<Sortable>
 { ["one", "two", "three"].map(i => <Sortable.Item>{i}</Sortable.Item>) }
</Sortable>
```

^ where we can control exactly the tag to render and the consumer don't have to worry about what tag to render. 

my `giveAFu` thermometer ★★★★ 4/5

packages/Sortable/src/components/Item/Item.js
alexzherdev: FYI the plugin is from storybook's babel config, not paprika's. So essentially this fix should only be necessary for storybook to run which makes it even more sucky. Let's add a comment for this?

packages/Sortable/src/components/Item/Item.js
```diff
@@ -0,0 +1,55 @@
...
+import SortableContext from "../../SortableContext";
+import { itemStyles, itemIndexStyles, itemHandleStyles, itemBodyStyles, itemCloseStyles } from "./Item.styles";
+
+const propTypes = {
+  child: PropTypes.node.isRequired,
+  index: PropTypes.number.isRequired,
+};
+
+const Item = ({ child, index }) => {
+  const { hasIndexes, onRemove } = React.useContext(SortableContext);
```
rauldudu: Just curious: Why do we need context? Couldn't `Sortable` pass these as prop to `Item`?

nahumzs:
`hasIndex` to me sound like you will do something internally more than actually display the numbers next to the items, maybe a more descriptive name would be beneficial? `giveAFu` thermometer ★ 1/5

packages/Sortable/src/components/Item/Item.js
nahumzs: ^ agree maybe drilling props might be a better solution for this component.

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,70 @@
...
+const defaultProps = {
+  children: null,
+  hasIndexes: true,
+  onRemove: null,
+};
+
+function augmentChildren(children) {
+  return React.Children.map(children, (child, index) => {
+    return child.type
+      ? React.cloneElement(child, { ...child.props, "data-drag-id": index })
```
alexzherdev:
```suggestion
      ? React.cloneElement(child, { "data-drag-id": index })
```

packages/Sortable/src/components/Item/Item.js
nahumzs: Mike put a comment about this, seems to be a problem with babel parsing the code, and this helps to fix it.

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,70 @@
...
+      : React.createElement("span", { ...child.props, "data-drag-id": index }, child);
```
alexzherdev: Want to just use JSX here?

packages/Sortable/README.md
nahumzs: can we add a simple example about how to use the component, this is our basic documentation that will appear on the npm page

packages/Sortable/src/components/Item/Item.js
alexzherdev: +1. Could make things easier

packages/Sortable/src/components/Item/Item.js
nahumzs: https://github.com/storybookjs/storybook/issues/6928#issuecomment-500933394 Is this a scape for this?

packages/Sortable/src/Sortable.js
alexzherdev: Heh, just extracting the object literal into a variable changes nothing, it's still recreated on every render call. You would want to `useMemo` this object (assuming we still want to use context here)

packages/Sortable/src/components/Item/Item.js
alexzherdev: oh, I forgot you also had issues with the `css` prop. It's great if just upgrading storybook fixes this 👍

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,70 @@
...
+              {...provided.droppableProps}
+              css={sortableStyles}
+              data-is-dragging-over={snapshot.isDraggingOver ? true : undefined}
+              isDraggingOver={snapshot.isDraggingOver}
+              ref={provided.innerRef}
+            >
+              {augmentedChildren &&
+                augmentedChildren.length > 0 &&
+                augmentedChildren.map((child, index) => (
+                  <Item child={child} key={child.props["data-drag-id"]} index={index} />
```
alexzherdev: Pretty minor, but `child` could go into the children position here?

packages/Sortable/src/components/Item/Item.js
```diff
@@ -0,0 +1,55 @@
...
+
+  function handleRemove() {
```
alexzherdev: Elsewhere you're using arrow functions for this

packages/Sortable/stories/Sortable.stories.helpers.js
```diff
@@ -0,0 +1,103 @@
+import React from "react";
+import styled from "styled-components";
+import Input from "../../Input/stories/examples/InputExample";
+import Select from "../../Select/stories/examples/SelectExample";
+
+export const storyStyles = `
+  max-width: 360px
```
alexzherdev:
```suggestion
  max-width: 360px;
```
?

packages/Sortable/stories/examples/Basic.js
alexzherdev: We are controlling the elements already though because `Item` wraps the children

packages/Input/src/Input.styles.js
mikrotron: I don't quite follow your example, but find the way I have it keeps the subsequent conditions cleaner (no need for the `props =>` or `props.` anywhere).  Also, kinda nice to see what props these styles depend on right at the top in a convenient location.

packages/Sortable/src/components/Item/Item.js
mikrotron: I will investigate whether an update for Storybook and/or styled-components will help.

packages/Sortable/README.md
mikrotron: 🤦‍♂️ such a squandered opportunity...

packages/Sortable/package.json
mikrotron: Yeah... whoops.

packages/Sortable/src/Sortable.js
mikrotron: Hmm... it's not really extracting in the same way you've used it.  And `clone` is a bit vague... it's really just "adding to" the children, which is why I think `augment` is still the best choice.

packages/Icon/src/DragHandle.js
mikrotron: Yeah, this is how the `svgr` script generates the component.  It would be nicer if it generated `propTypes` too... but I didn't want to get into AST Templates if I could avoid it (with a simple little eslint exception).

packages/Sortable/src/Sortable.js
mikrotron: What do you mean by "live under"?  It's not a component the app consumer will need to know about, so I didn't bother exposing it as `<Sortable.Item>`.

jamiek-acl:
```
import Item from "./Sortable/Item/Item";
```

jamiek-acl: oh nevermind, its good

packages/Sortable/stories/examples/Basic.js
mikrotron: An array of Inline `<spans>` or strings, even `null`, `undefined`, or `[]` are handled.  Strings are handled by wrapping them in `<spans>` in `augmentChildren()`:

https://github.com/acl-services/paprika/blob/0e07cf3d93d346d3737d96efe0f36092880def91/packages/Sortable/src/Sortable.js#L24-L26

So there aren't any issues. Imposing this restriction is not technically necessary, and the implementation will appear redundant. The `giveAFu` rating is high though, so maybe others can weigh in, and/or we can discuss in person.

mikrotron: @jamiek-acl, there's a way to disable individual items with this library.  Depending on the use case, we could implement it in a visual (greyed out), or subtle way (it just doesn't move).  But currently, we don't have this use case, so I'm not implementing it at this time (but I did experiment a little to make sure it would be easy to add later).

packages/Sortable/src/Sortable.js
mikrotron: It seems that `React.Children.map()` doesn't always return an array.  When children is `null` or `undefined`, it will maintain that for a return value.  But I can add `|| []` in `augmentChildren` instead, I think that's a bit cleaner.

packages/Sortable/src/components/Item/Item.styles.js
mikrotron: Hm, I've been using this technique in a few components now, and like I mentioned in a previous comment, like how it cleans up the conditions inline with the reset of the css.  I doubt it will be confused with a functional component, given the context of it being found in a `styles.js` file, and it's body being a template literal.

packages/Sortable/stories/examples/Basic.js
mikrotron: It _is_ working, in the story helpers I generate lists of `<Inputs>` that are being used in various stories.

https://github.com/acl-services/paprika/blob/0e07cf3d93d346d3737d96efe0f36092880def91/packages/Sortable/stories/Sortable.stories.helpers.js#L76-L84

packages/Sortable/src/components/Item/Item.js
mikrotron: > `<Draggable />` and `<Droppable />` components both require a `HTMLElement` to be provided to them. This is done using the `innerRef` property on the `DraggableProvided` and `DroppableProvided` objects.

Source: https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/using-inner-ref.md

packages/Sortable/src/SortableContext.js
mikrotron: This pattern is needed because otherwise there will be a dependency cycle that violates the `import/no-cycle` eslint rule.

Sortable exports context, and imports Item.  Item imports context and exports itself for Sortable. 
Sortable depends on Item and Item depends on Sortable.  And on and on it goes, like:

packages/Sortable/src/components/Item/Item.js
mikrotron: That's a lot of consensus.  Perhaps this hammer was too big for this nail.

packages/Sortable/src/components/Item/Item.js
mikrotron: Success! No more weird `{...{}}` needed!

packages/Sortable/stories/examples/Basic.js
nahumzs: just playing `devil advocate`:

how would you mark `isDisabled` in a specific child with this API while using divs and spans?

// to me this is cleaner:
<Sortable>
  <Sortable.Item isDisable><MyComponent /></Sortable.item>
</Sortable>

// current api
// seems odd to pass isDisabled over a div/span/p. 
<Sortable>
   <div isDisabled><MyComponent /></div>
   <div isDisabled><MyComponent /></div>
</Sortable>

packages/Sortable/src/SortableContext.js
nahumzs: ohhhh now I understand why I hate that rule `import/no-cycle`

packages/Sortable/stories/examples/Basic.js
jamiek-acl: Exactly @nahumzs , that's what I was getting at. I think it is more clear for the children to be `<Sortable.Item>`s.

packages/Sortable/src/components/Item/Item.styles.js
alexzherdev: Do we get syntax highlighting here?

packages/Input/src/Input.styles.js
nahumzs: One possible downside of using as a function that creates css is that will block any static analyzers to detect css code, like this [plugin](https://github.com/siffogh/eslint-plugin-better-styled-components).

See example: 



While I like and preferer your approach or using a function instead of **css``**, for the benefit of the entire repo is better to use **css``** IMHO.

What do you think?

packages/Sortable/stories/examples/Basic.js
mikrotron: Okay, that makes some sense. It also means I can rename `data-drag-id` → `dragId`.

packages/Sortable/src/components/Item/Item.styles.js
mikrotron: Nopes. I don't get proper syntax highlighting ever with `styled-components`.

packages/Input/src/Input.styles.js
mikrotron: While I haven't been able to get syntax highlighting to work properly with the `css**` notation either, if that's something I _could_ and this is the blocker, I don't mind compromising.  I'll probably keep destructuring `props` though, I think it's still a little better:
```js
${({ isDragging }) => (isDragging ? draggingStyles : null)}
```

nahumzs: destructuring is better agree, and at least in atom unless I follow the pattern of `styled.div''` or `css''` highlighting css will not work.

packages/Sortable/src/components/Item/Item.styles.js
nahumzs: ref: https://github.com/acl-services/paprika/pull/92#discussion_r296359081

packages/Input/src/Input.styles.js
mikrotron: Yep. We should probably create an issue to refactor components to use this pattern consistently.

nahumzs: FYI https://github.com/acl-services/paprika/issues/98 @mikrotron

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,93 @@
...
+        I18n.t("sortable.aria_dropped", { source: source.index + 1, destination: destination.index + 1 })
+      );
+    } else {
+      provided.announce(I18n.t("sortable.aria_dropped_outside", { source: source.index + 1 }));
+    }
+  };
+
+  return (
+    <DragDropContext onDragEnd={handleDragEnd} onDragStart={handleDragStart} onDragUpdate={handleDragUpdate}>
+      <Droppable droppableId={`droppable-${uuidv4()}`}>
```
rauldudu: should the `uuid` be consistent between re-renders?

alexzherdev: nice catch. 🎣 
Also, while not relevant now, but most of these techniques we use for generating ids are incompatible with server rendering. Just a note.

packages/Sortable/src/components/SortableItem/SortableItem.styles.js
```diff
@@ -0,0 +1,66 @@
...
+
+  &:focus {
+    outline: none;
+    box-shadow: ${tokens.highlight.active.withBorder.boxShadow};
+  }
+
+  ${({ isDragging }) => (isDragging ? draggingStyles : null)}
+`;
+
+export const itemHandleStyles = `
```
rauldudu: Q: are we only using the `css` helper when there are functions inside?

packages/Sortable/src/Sortable.js
mikrotron: From the React docs:
> However, useRef() is useful for more than the ref attribute. It’s handy for keeping any mutable value around similar to how you’d use instance fields in classes.

So is that the hook to use?

packages/Sortable/src/components/SortableItem/SortableItem.styles.js
nahumzs: css let you receive the props inside the function but also helps IDEs/linters to highlight css and prettier fix issues. 

I guess we should use css in all cases unless someone finds a significant performance issue by using the css method, in that case, using the string assignment should be fine.

packages/Sortable/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "@paprika/icon": "^0.1.0",
+    "@paprika/l10n": "^0.1.0",
+    "@paprika/stylers": "^0.1.0",
+    "@paprika/tokens": "^0.1.0",
+    "prop-types": "^15.7.2",
+    "react-beautiful-dnd": "^11.0.4",
+    "uuid": "^3.3.2"
+  },
+  "peerDependencies": {
+    "react": "^16.4.0",
```
alexzherdev:
```suggestion
    "react": "^16.8.0",
```
mind upping this for hooks?

packages/Sortable/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "react-dom": "^16.4.2",
```
alexzherdev:
```suggestion
    "react-dom": "^16.8.0",
```

packages/Sortable/src/Sortable.js
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const Sortable = ({ children, onChange, hasNumbers, onRemove }) => {
+  const I18n = useI18n();
+  const dropId = React.useRef(uuid());
+  const validChildren = filterChildren(children);
+
+  const handleDragStart = (start, provided) => {
+    const { source } = start;
+    provided.announce(I18n.t("sortable.aria_lift", { source: source.index + 1 }));
```
alexzherdev: 🎉 so they were not lying huh? _beautiful_ 😮

packages/Sortable/stories/Sortable.stories.helpers.js
```diff
@@ -0,0 +1,121 @@
...
+    width: 100%;
+  }
+`;
+
+export function basicChildren(numChildren) {
+  const children = [];
+
+  for (let index = 1; index <= numChildren; index++) {
+    children.push(
+      <Sortable.Item key={index} sortId={index}>
```
alexzherdev: It'd be nice if we could just reuse `key` for `sortId` but apparently not possible as we won't have access to it in the parent component

PR #93, opened 2019-06-18
PD-12468 - Add missing stylers dependency

rauldudu:
### 🛠 Purpose
---
Add missing dependency

### ✏️ Notes to Reviewer
---

### 🖥 Screenshots

---


nahumzs:
the problem with caching yarn in semaphore is that we can't see these problems when some dependency is missed :/

PR #94, opened 2019-06-20
UX-408 Date picker with shortcut panel

allison-c:
### 🛠 Purpose

Adding shortcut panel to date picker component.

### ✏️ Notes to Reviewer

- Added shortcut panel for calendar
- The calendar is bigger
- The initial displayed date will be in spelled out format
- Added prop `humanFormat` for customize spelled out format
- Added a new component `DatePicker.Popover`
- Added knobs in storybook with showcase story
- Fixed popover bug in Edge

### 🖥 Screenshots

GIF
ezgif com-optimize (17)

Popover bug in Edge:
Jietu20190620-103911-HD
It was because `getBoundingClientRect` won't return `x` `y` in Edge and IE11


packages/Popover/src/Popover.js
```diff
@@ -206,18 +206,47 @@ class Popover extends React.Component {
...
+      top: rect.top,
+      right: rect.right,
+      bottom: rect.bottom,
+      x,
+      y,
+      width: rect.width,
+      height: rect.height,
+    };
+  }
+
```
allison-c: @nahumzs could you review this part? Found a bug in Microsoft Edge and fixed it.

packages/DatePicker/src/DatePicker.js
```diff
@@ -82,18 +95,21 @@ function DatePicker(props) {
...
   function handleClosePopover() {
-    window.requestAnimationFrame(() => {
+    setTimeout(() => {
       if (!isElementContainsFocus(calendarRef.current) && !isElementContainsFocus(inputRef.current)) {
+        setConfirmationResult(date ? moment.utc(date).format(humanFormat) : "");
+        setInputtedString(date ? moment.utc(date).format(dataFormat) : "");
         hideCalendar();
       }
-    });
+    }, 10);
```
allison-c: `requestAnimationFrame ` doesn't work anymore if I need to wait for newly added component to render. `setTimeout(, 0)` sometimes works, but not always, don't know what's the best way to do that. `setTimeOut(, 10)` will works better.

packages/DatePicker/stories/screener.stories.js
```diff
@@ -0,0 +1,20 @@
...
+
+storiesOf("DatePicker / screener", module)
+  .add("Calendar", () => (
+    <L10n>
+      <DatePicker.Calendar date={moment("2019-01-01", "YYYY-MM-DD")} isOpen onSelect={noop} />
+    </L10n>
+  ))
+  .add("Shortcut panel", () => (
+    <L10n>
+      <ShortCutPanel date={moment("2019-01-01", "YYYY-MM-DD")} onCancel={noop} onConfirm={noop} />
```
nahumzs: Can you add an image of what is ShortCutPanel? is part of the DatePicker?  might be a better developer experience in the sense of API for using the component?

packages/DatePicker/src/DatePicker.js
alexzherdev: When `setTimeout(, 0)` doesn't work, what does that look like?

packages/DatePicker/src/DatePicker.js
```diff
@@ -157,8 +174,11 @@ function DatePicker(props) {
     }
   }
 
-  function handleClick() {
-    if (!isReadOnly) showCalendar();
+  function handleSelect(seletedDate) {
```
alexzherdev:
```suggestion
  function handleSelect(selectedDate) {
```
and below

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -10,17 +10,20 @@ import ArrowLeft from "@paprika/icon/lib/ArrowLeft";
...
   date: momentPropTypes.momentObj,
 
-  /** Possible date might be selected in moment object */
-  possibleDate: momentPropTypes.momentObj,
+  isOpen: PropTypes.bool.isRequired,
 
   /** Callback to fire when user select date */
   onSelect: PropTypes.func.isRequired,
+
+  /** Possible date might be selected in moment object */
```
alexzherdev: Not sure what this means.

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -35,15 +38,55 @@ function Calendar(props) {
...
+  const [currentMonth, setCurrentMonth] = React.useState(null);
 
   // Ref
   const nextButtonRef = React.useRef(null);
   const prevButtonRef = React.useRef(null);
+  const calendarRef = React.useRef(null);
+
+  // Effect
+  React.useEffect(() => {
+    if (shouldShowShortcut === false && calendarRef.current) {
```
alexzherdev:
```suggestion
    if (!shouldShowShortcut && calendarRef.current) {
```
should work?

packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.js
```diff
@@ -0,0 +1,153 @@
...
+        />
+        <LabelStyled htmlFor={monthIndex} isSelected={monthIndex === selectedMonth}>
+          {month}
+        </LabelStyled>
+      </OptionWrapperStyled>
+    ));
+  }
+
+  function renderYearList() {
+    const SELECTED_YEAR_POSITION = 5;
```
alexzherdev: What does this 5 mean?

allison-c: The selected year's position in shortcut panel
image

packages/DatePicker/src/components/Calendar/Calendar.js
allison-c: We want to rerender `react-date` calendar when the user is typing in the input. This prop is the parsing result(in moment object) from user's input, and I use it to rerender calendar

packages/DatePicker/src/DatePicker.js
```diff
@@ -37,8 +40,9 @@ const propTypes = {
 
 const defaultProps = {
   children: null,
-  date: null,
   dataFormat: "MM/DD/YYYY",
+  date: null,
+  humanFormat: undefined,
```
mikrotron: Can this be `null` instead?

mikrotron: Or better, yet, `"MMMM D, YYYY"`, so that you don't have issues with line 69 below...?

packages/DatePicker/src/DatePicker.js
```diff
@@ -140,14 +154,18 @@ function DatePicker(props) {
...
   }
 
+  function handleInputChange(e) {
+    setInputtedString(e.target.value);
+  }
+
   function handleKeyUp(event) {
     if (event.key === "Enter") {
       handleInputConfirm();
+    } else if (event.key === "Escape") {
```
mikrotron: This only works if the focus is on the Input, since the handler is attached to that component. If you add the handler to the Popover, I think it might work better for those users that are using the mouse, but want the esc shortcut to work anyway.

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -10,17 +10,28 @@ import ArrowLeft from "@paprika/icon/lib/ArrowLeft";
...
 import useI18n from "@paprika/l10n/lib/useI18n";
 
-import CalendarStyled, { DayTriggerStyle, CalendarHeaderStyled } from "./Calendar.styles";
+import ShortcutPanel from "../ShortcutPanel";
+import { isMomentObjectOrNull } from "../../helpers";
+import CalendarStyled, {
+  CalendarWrapperStyled,
+  DayTriggerStyle,
+  MonthHeaderButtonStyled,
+  ArrowIconStyles,
```
mikrotron: I find the mix of `Styles` and `Styled` a bit odd... are you planning to refactor all to `Styles`?

packages/DatePicker/src/components/Calendar/Calendar.js
mikrotron: What makes it _possible_ though? I probably don't mind the name, but this comment is the opportunity to explain that name choice.

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -68,20 +128,32 @@ function Calendar(props) {
...
+        {month.format(I18n.t("datePicker.calendar_header_format"))}
+      </MonthHeaderButtonStyled>
+    );
   }
 
   function renderArrowLeft() {
     return (
-      <span ref={prevButtonRef} data-qa-anchor="datepicker-prev-month">
+      <span css={ArrowIconStyles} ref={prevButtonRef} data-qa-anchor="datepicker-prev-month">
         <ArrowLeft role="presentation" size="14px" />
```
mikrotron: Incidental, but the `size` prop can be omitted, and `fontSize(-1)` can be used instead in the `ArrowIconStyles`.

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -99,30 +171,42 @@ function Calendar(props) {
...
-        verticalBorderSpacing={0}
-        transitionDuration={0}
-        horizontalMonthPadding={0}
-        navPrev={renderArrowLeft()}
-        navNext={renderArrowRight()}
-        onPrevMonthClick={handleClickPrevMonth}
-        onNextMonthClick={handleClickNextMonth}
-        renderDayContents={renderDayContents}
+    <CalendarWrapperStyled tabIndex={-1} ref={calendarRef}>
+      <CalendarStyled shouldHidden={shouldShowShortcut || !isOpen}>
```
mikrotron: How about `isVisible` instead of `shouldHidden`?

packages/DatePicker/src/components/Calendar/Calendar.styles.js
```diff
@@ -1,17 +1,57 @@
...
+    display: block;
+  }
+`;
+
+const IconButtonStyles = `
+  transition: background-color 0.2s ease-out;
+  
+
+  &:hover {
+    background-color: ${stylers.alpha(tokens.color.black, 0.1)};
```
mikrotron: You found the `alpha()`!  🙌

packages/DatePicker/src/components/Calendar/Calendar.styles.js
```diff
@@ -1,17 +1,57 @@
...
+    shouldHidden &&
+    `
+    width: 0;
+    height: 0;
+    overflow: hidden;
+  `}
+
   * {
     box-sizing: border-box;
     font-family: ${tokens.fontFamily.default};
```
mikrotron: Incidental again... but would `inherit` work here instead, so this component remains agnostic of the app's typeface?

packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.styles.js
```diff
@@ -0,0 +1,140 @@
...
+  box-sizing: border-box;
+  color: ${tokens.color.black};
+  font-family: ${tokens.fontFamily.default};
+  ${stylers.fontSize(-1)}
+  height: 245px;
+  overflow: hidden;
+  position: relative;
+  width: 257px;
+
+  ${({ isOpen }) =>
```
mikrotron: I'd be tempted to define the two lines of css as a variable so you can do this in one clean line... but it's not _so_ bad as it is.  Your call.

packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.styles.js
```diff
@@ -0,0 +1,140 @@
...
+  justify-content: center;
+  width: 46px;
+
+  &:hover {
+    ${HoveredItemStyles}
+  }
+
+  ${({ isSelected }) =>
+    isSelected &&
+    css`
```
mikrotron: I think this can just be:
`${({ isSelected }) => isSelected && SelectedItemStyles}`

No need for the `css` here, but you'll need to add it in the `SelectedItemStyles` definition.

packages/DatePicker/src/shared.styles.js
```diff
@@ -0,0 +1,11 @@
+import tokens from "@paprika/tokens";
+
+export const HoveredItemStyles = `
+  background-color: ${tokens.color.blackLighten70};
+  font-weight: bold;
```
mikrotron: I don't like the bold on hover.  I know it's probably that way in the mocks, but it makes the text shift a little.  Small detail, but my `GiveAFudge` scale is disproportionately high for this kind of thing.

packages/Popover/src/Popover.js
```diff
@@ -198,7 +198,7 @@ class Popover extends React.Component {
     $shadowContent.style.maxWidth = this.props.maxWidth;
 
     document.body.appendChild($shadowContent);
-    const contentWidth = $shadowContent.getBoundingClientRect().width;
+    const contentWidth = this.getBoundingClientRect($shadowContent).width;
```
mikrotron: Nice job! 🕵️‍♀️

packages/DatePicker/src/DatePicker.js
allison-c: The actual default value is `"MMMM D, YYYY".  However, it depends on translation because this format still needs to be localized. My thought is if the consumer didn't pass in, I will use 
`I18n.t("datePicker.confirmation_format")`, I did it here:
https://github.com/acl-services/paprika/pull/94/files/e1fc2399b1b667f06cadc0a38e6b0aa2c8e9f533#diff-29403a20e6a7b79ffb881da8212b736cR58
I didn't use `null` because I need to use `I18n.t` in the functional component to set the default value. If I use null I will have another check like:
```
if (!humanFormat) {
    humanFormat = "MMMM D, YYYY"
}
```

packages/DatePicker/src/components/Calendar/Calendar.js
allison-c: Yes, I already finish it in my local branch, will push the changes today

packages/DatePicker/src/DatePicker.js
mikrotron: I see.  🤔
I'm not sure if anyone else is bothered by setting the default prop to `undefined`, but maybe an alternative would be to keep `null` as the default prop and adjust formatDateProp() like :
```js
function formatDateProp(format) {
  return date && date.isValid() 
    ? moment.utc(date).format(format || I18n.t("datePicker.confirmation_format")) 
    : "";
}
```

packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.styles.js
allison-c: I think that makes sense to me, I will refactor it 😄

packages/DatePicker/src/components/Calendar/Calendar.js
allison-c: maybe `predictedDate` ? or `prediction` ?

packages/DatePicker/src/components/Calendar/Calendar.styles.js
allison-c: I moved ` font-family: ${tokens.fontFamily.default};` to `DatePicker.styles.js`. I think it's good to set the default `font-family` otherwise I need to set it in the date picker example in storybook.

packages/DatePicker/src/shared.styles.js
allison-c: Talked with @sunminsarahlee , let's keep `bold` on hover.

packages/DatePicker/src/components/Calendar/Calendar.styles.js
mikrotron: Does the component _need_ to be a specific font? Ideally when we switch our apps to use `IBM Plex`, this component shouldn't have to be changed.  In fact we could use it in different apps with different UI fonts, and the component should adapt.  Ideally.

packages/DatePicker/src/shared.styles.js
mikrotron: Okay.

packages/DatePicker/src/components/Calendar/Calendar.styles.js
allison-c: 🤔Sure! The only thing is some font doesn't works well with `bold` on hover, for example `Helvetica Neue`. It will shift a little bit as you said before. `IBM Plex` and `Lato` doesn't have this issue. I will just use inherit then. I can add the default font in storybook if that looks bad.

packages/DatePicker/src/DatePicker.js
```diff
@@ -62,38 +71,51 @@ function DatePicker(props) {
 
   // Effect
   React.useEffect(() => {
-    setInputtedString(date ? moment.utc(date).format(dataFormat) : "");
+    setInputtedString(formatDateProp(dataFormat));
+    setConfirmationResult(formatDateProp(humanFormat));
   }, [date]);
```
nahumzs: some dependencies are missed:
```
  // Effect
  React.useEffect(() => {
    setInputtedString(formatDateProp(dataFormat));
    setConfirmationResult(formatDateProp(humanFormat));
  }, [dataFormat, date, formatDateProp, humanFormat]);
```

packages/DatePicker/src/DatePicker.js
```diff
@@ -47,12 +52,16 @@ function DatePicker(props) {
   const I18n = useI18n();
 
   // Props
-  const { children, date, dataFormat, isDisabled, isReadOnly, onChange } = props;
+  const { children, dataFormat, date, humanFormat, isDisabled, isReadOnly, onChange } = props;
+
+  function formatDateProp(format) {
```
nahumzs: React.useCallback can help in this function.

packages/DatePicker/src/helpers.js
```diff
@@ -1,13 +1,20 @@
...
   });
 
   return overrides;
 }
+
+export function isMomentObjectOrNull(date) {
+  if (date === null) return true;
+
+  return momentPropTypes.momentObj(date);
+}
```
mikrotron: I think this means we need to bump the version for this `helpers` package in its `package.json`... right @nahumzs?

packages/Popover/src/Popover.js
```diff
@@ -240,9 +269,9 @@ class Popover extends React.Component {
...
     if (this.$tip) {
       tipCoords = getTipCoordinates({
-        tipRect: this.$tip.getBoundingClientRect(),
+        tipRect: this.getBoundingClientRect(this.$tip),
         targetRect,
-        contentRect: this.$content.getBoundingClientRect(),
+        contentRect: this.getBoundingClientRect(this.$content),
         contentCoords,
         align,
       });
```
mikrotron: I think this means we need to bump the Popover's version its `package.json`... right @nahumzs?

allison-c: No bump needed
see: https://github.com/acl-services/paprika/pull/110#discussion_r300758329

nahumzs:
I just noticed that moment is a dependency, but shouldn't be a peer-dependency? cc @alexzherdev

packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -35,15 +49,64 @@ function Calendar(props) {
...
+  }
+
+  // Effect
+  React.useEffect(() => {
+    if (!isVisible) {
+      setShouldShowShortcut(false);
+      setCurrentMonth(null);
+    }
+    keepFocus();
+  }, [isVisible]);
```
nahumzs: I think the linter should be warning you that you need to add keepFocus as part of the arrayDependency isn't?

packages/Popover/src/Popover.js
```diff
@@ -220,18 +220,47 @@ class Popover extends React.Component {
     this.updateVisibilityAndPositionState(isOpening);
   }
 
+  getBoundingClientRect(el) {
```
nahumzs: this could be a hook, but that scope would be totally out of this PR. seem ok to me.

alexzherdev:
> I just noticed that moment is a dependency, but shouldn't be a peer-dependency? cc @alexzherdev

Hmm did some research. TL;DR yeah let's make it a peerDependency. Unless we're really using some super new features, let's depend on a broad range like `^2.0.0` to avoid "unmet peer dependency" warnings in the apps.

---
I had thought that in this case peerDep would only help ensure one copy of moment was in the tree for bundle size purposes (and not break anything, unlike with React, where two copies could break the app). But turns out that two copies of moment can also cause breakage when you do checks on whether something is a moment object:
```js
import moment from 'moment';
moment.isMoment(momentObj); // or momentObj instanceof moment
```
If `momentObj` is created by a different copy of moment than `moment`, this check will fail.

allison-c:
> > I just noticed that moment is a dependency, but shouldn't be a peer-dependency? cc @alexzherdev
> 
> Hmm did some research. TL;DR yeah let's make it a peerDependency. Unless we're really using some super new features, let's depend on a broad range like `^2.0.0` to avoid "unmet peer dependency" warnings in the apps.
> 
> I had thought that in this case peerDep would only help ensure one copy of moment was in the tree for bundle size purposes (and not break anything, unlike with React, where two copies could break the app). But turns out that two copies of moment can also cause breakage when you do checks on whether something is a moment object:
> 
> ```js
> import moment from 'moment';
> moment.isMoment(momentObj); // or momentObj instanceof moment
> ```
> 
> If `momentObj` is created by a different copy of moment than `moment`, this check will fail.

Thanks Alex! Nahum and I also had a small conversion in person. I was asking why not put `@paprika/raw-button` in `peerDep` as well. Then we end up with we don't want to force consumers to upgrade `peerDep` so frequently, so paprika components should be `dependency` but `@paprika/icon` might be valuable to be a `peerDep` because we won't change it much once it's stable

alexzherdev:
> I was asking why not put @paprika/raw-button in peerDep as well. Then we end up with we don't want to force consumers to upgrade peerDep so frequently, so paprika components should be dependency but @paprika/icon might be valuable to be a peerDep because we won't change it much once it's stable

I'd say that's overdoing it. Part of the reason we switched to a monorepo was the ability to have multiple versions of the same component in the tree if we need to—peerDep undoes that.
I don't think the frequency of updates to a package is an indicator of a peerDep. The indicator is usually "I need one and only one copy of this package in the tree otherwise my code breaks".
Additionally, having peerDeps is a hassle because increasing a peerDep's version is a breaking change for your package.

PR #99, opened 2019-06-25
UX-494 Migrate switch component

allison-c:
### 🛠 Purpose

Migrate switch component

Plus
- Fixed raw button css issue
- Disabled raw button was still focusable by click, so @mikrotron and I decided just assign null to `tabIndex` when it's disabled

#### Usage

 For a basic switch

 ```js
import Switch from "@paprika/switch";
 <Switch isChecked={switchState} onChange={changeHandler} />;
```

 #### Props

   Descriptive a11y text for assistive technologies. Typically required.
 ` a11yText: PropTypes.string`

   If click events are allowed to propagate up the DOM tree.
  `canPropagate: PropTypes.bool`

   If the switch is on.
 `isChecked: PropTypes.bool`

  If the switch is disabled.
  `sDisabled: PropTypes.bool`

Callback to be executed when the switch is toggled on or off. Typically required. 
  `onChange: PropTypes.func`

  Size of the switch.
  `size: PropTypes.oneOf(ShirtSizes.DEFAULT)`

#### Default props

```
  a11yText: null,
  canPropagate: true,
  isChecked: false,
  isDisabled: false,
  onChange: () => {},
  size: "medium",
```

### 🖥 Screenshots

image

image

### CODE REVIEW

File: packages/RawButton/src/RawButton.js [hunk 1]
```diff
@@ -86,7 +86,7 @@ const RawButton = React.forwardRef((props, ref) => {
       onKeyDown={handleKeyDown}
       onKeyUp={handleKeyUp}
       ref={rawButtonRef}
-      tabIndex={isDisabled ? -1 : tabIndex}
+      tabIndex={isDisabled ? null : tabIndex}
```
nahumzs: shouldn't this be undefined so the tabIndex don't even render on the component?

mikrotron: We tested it and it doesn't add the attribute with `null` either! Maybe this is a difference between React 15 and 16?

allison-c: It won't be rendered when using `null` as well

packages/Switch/src/Switch.styles.js
```diff
@@ -0,0 +1,158 @@
...
+    height:  ${stylers.spacer(3)};
+    width: 44px;
+  `,
+  large: `
+    height:${stylers.spacer(4)};
+    width: 60px;
+  `,
+};
+
+export const KnobStyled = styled.span`
```
rauldudu: should we just export the raw styles here and apply them to a span defined in the component itself? Just curious because I'm not sure if we standardize this

packages/Switch/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/switch"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/helpers": "^0.1.0",
+    "@paprika/stylers": "^0.1.0",
```
mikrotron: You'll need `@paprika/raw-button` too.

packages/Switch/src/Switch.styles.js
allison-c: Good question, I'm not sure as well 😄 
From my understanding, if we want to read `props` is easier to use `css` prop. If it's just plain css, it's cleaner to just import `ElementStyled`

mikrotron: Yeah, the approach we've been trying out has been to export the raw styles and use the `css` prop to apply them to an element / component. Maybe @allison-c had a reason for this approach though?

packages/Switch/src/Switch.js
```diff
@@ -0,0 +1,63 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import RawButton from "@paprika/raw-button";
+
+import SwitchStyles, { KnobStyled, UnderlayStyled } from "./Switch.styles";
```
mikrotron: `switchStyles`, with the lower case first letter is the convention we've been going with.

packages/Switch/test/Switch.spec.js
```diff
@@ -0,0 +1,49 @@
...
+  };
+  const renderedComponent = render(<Switch {...defaultProps} {...props} />);
+
+  return {
+    ...renderedComponent,
+    switchElement: renderedComponent.getByRole("switch"),
+  };
+}
+
+describe("RawButton", () => {
```
alexzherdev:
```suggestion
describe("Switch", () => {
```

packages/Switch/src/Switch.styles.js
```diff
@@ -0,0 +1,158 @@
...
+  transition: 0.25s ease-in-out;
+  user-select: none;
+  -moz-user-select: none;
+  -khtml-user-select: none;
+  -webkit-user-select: none;
+  -ms-user-select: none;
+  vertical-align: middle;
+`;
+
+const SwitchStyles = ({ isDisabled, size }) => `
```
mikrotron: This came up recently in my PR, and we settled on a different approach here for dealing with props, even though this _IS_ a nice clean way of implementing.  Instead we will go with something like:

```js
const switchStyles = css`
  ${({isDisabled}) => isDisabled ? disabledStyles : null}
`;
```

In fact, it helps with syntax highlighting and linting if we just use the `css` helper for all our js styles.

Reference: https://github.com/acl-services/paprika/pull/92#pullrequestreview-252441383

packages/Switch/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@paprika/switch",
+  "version": "0.0.1",
```
mikrotron: Oops, no big deal, but we've been launching components as `0.1.0` to start with.

PR #100, opened 2019-06-26
Fix git hooks

AndreyChernykh:
### 🛠 Purpose

Move `prettier` and `eslint` checks to the pre-commit phase. 

---

### ✏️ Notes to Reviewer

1. Bumped `prettier` package version
2. `tokens` were updated automatically

---

### 🖥 Screenshots
image
image
image

---


nahumzs:
I think everything is fine except the tokens files, we have to removing those from the index so we don't track them at all, tokens files are automatically generated in the `pretranspile` phase by lerna, which mean that will get generated during `npx lerna bootstrap`.

Could you remove them from the git cache? I think those are already part of `.gitignore` 🙏

AndreyChernykh:
> I think those are already part of .gitignore 
Not yet.

> Could you remove them from the git cache?
Do you mean just remove these 4 files from repository, commit it, and then add it to `.gitignore`?

nahumzs:
you are correct they are not removed:
```sh
# Tokens
packages/Tokens/src/tokens.js
packages/Tokens/src/tokens.json
packages/Tokens/src/tokens.mdx
packages/Tokens/src/tokens.scss
```

> Do you mean just remove these 4 files from the repository, commit it, and then add it to .gitignore?

I don't think just removing them and adding the `.gitignore` is enough, we will need to stop tracking changes on those files because they were already committed. 

https://stackoverflow.com/a/1274447/196038

If you can help us with this would be fantastic.

AndreyChernykh:
@nahumzs Done

package.json
```diff
@@ -72,15 +72,16 @@
...
     "jest": "^24.1.0",
     "jest-cli": "^24.1.0",
     "jest-dom": "^3.1.3",
     "js-yaml": "^3.13.1",
     "lerna": "^3.13.0",
+    "lint-staged": "^8.2.1",
     "npm-run-all": "^4.1.5",
     "pa11y": "^5.0.4",
-    "prettier": "^1.16.3",
+    "prettier": "^1.18.2",
```
alexzherdev: Can you pin this to follow their recommendation?

Issue #98, opened 2019-06-22
Refactor css styled-components prop


nahumzs:
# Task requested

Follow up for the discussion happened on https://github.com/acl-services/paprika/pull/92#discussion_r296375783 during Sortable component PR. 

In short, we need to adjust all our current component to follow the following patterns:

## 1.- All definition of css styles should use the `css` tag function from styled-components

```js
import { css } from "styled-components";

const myStyle = css`
   color: rebeccapurple;
`
```

no using the previous syntax or `styled.div` tag function from styled component will lead a bad static analysis from the linter.

## 2.- when using a prop function inside of the `css` tag function please destructured the probs in the arrow function parameters ex.

```js
const myStyle = css`
 ${({ aProp, anotherProp}) => {
    return `
        color: ${aProp};
        background: ${anotherProp};
    `
 }}
`
```

PR #103, opened 2019-06-27
UX-346 Fix Button styling 🐐

mikrotron:
### 🛠 Purpose
Resolve styling regression with `<Button.Icon>` and `<Button.Close>`.

### ✍ Notes
- Refactored some `.styles.js` files to use the styled-components `css` helper, to contribute to https://github.com/acl-services/paprika/issues/98

### 🖥 Screenshots
Screener regression that was fixed:
Screen Shot 2019-06-26 at 8 42 30 PM

### DIFF
--- a/packages/Button/src/IconButton.styles.js
+++ b/packages/Button/src/IconButton.styles.js
@@ -1,22 +1,23 @@
+import { css } from "styled-components";
 import tokens from "@paprika/tokens";
 import stylers from "@paprika/stylers";
 
 const iconButtonSizes = {
-  small: `
+  small: css`
     ${stylers.fontSize(-2)};
     height: ${stylers.spacer(3)}; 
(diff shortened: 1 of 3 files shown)


packages/Button/src/IconButton.styles.js
```diff
@@ -25,21 +26,21 @@ const iconButtonSizes = {
...
 };
 
-const minorStyles = `
+const minorStyles = css`
   transition: background-color 0.2s ease-out;
-  
+
   &:hover {
-    background-color: ${stylers.alpha(tokens.color.black, 0.1)}; 
+    background-color: ${stylers.alpha(tokens.color.black, 0.1)};
```
nahumzs: no action need it but we should decide soon if we want to use: { alpha } or stylers.alpha ...

Issue #101, opened 2019-06-26
RFC Adding an anchor attribute for each component.


nahumzs:
https://github.com/acl-services/paprika/pull/84/files#r296462713

While implementing the SidePanel there was a small discussion about the benefit that would bring to add a data anchor attribute to our components. This would help in targeting them easier in CSS/SCSS/JSS and while testing.

Example:
Let's put a scenario where we want to build a `SuperDupper` component the return "render" would look like:

```js
export default function SuperDupper(props) {
  return (<div {...props} data-display-name="SuperDupper">Super Dupper</div>)
} 
```

so in a consumer app if you need to modify this component by any specific requirement or accessing it by testing could do:

```css
/* css/jss/scss */
[data-display-name="SuperDupper"] { color: red; }
```
___or 

```js
  // testing
  expect(cy.queryBy('[data-component-type="listbox.filter"]')).toBeInTheDocument()
```

Thoughts?

mikrotron:
This can be achieved by adding a `className` to the component, and then using that selector to target instead, but: 

1) When composing Paprika components, adding `classNames` is being avoided in favour of a purely `styled-components` approach. 

2) In an app where, for example, there are a number of buttons you want to apply spacing to for your layout, it's cumbersome to apply a custom className to all of them.  It's more convenient to do:
```scss
.myComponent {
  [data-component-type='button'] { 
    margin-right: 16px; 
  }
}
```

alexzherdev:
I find it illogical to use display name for this 🤔The purpose of this anchor is completely different from setting how the component is called in the dev tools (which is what display name is)? When someone sees a `data-display-name`, there's a chance they will mix up the two and think this is some alternative way for setting a display name for the component rendered.

alexzherdev:
As for targeting items for styling, I would warn against giving too much freedom—makes it harder to enforce consistency of the design system. Especially for a complex component, like a listbox for example, you don't want to be able to override, say, the paddings within the body. In that regard, `className` is particularly troubling because it screams "please override some styles on me".

mikrotron:
After some follow-up discussion offline, there was consensus that a generic handle should be added to the root node and can optionally be added to internal elements.  The same attribute `data-pka-anchor` will serve for both situations.  

Examples:
```jsx
<div data-pka-anchor="my-component">
  <span data-pka-anchor="my-component.some-element" />
</div>
```

nahumzs:
👍 closing this as a decision has been reached.

PR #110, opened 2019-07-04
PD-12484 Cypress tests for Sortable component 🐐

mikrotron:
### 🛠 Purpose
Add test coverage for the new `<Sortable>` component.

### ✏️ Notes to Reviewer
Test cases:
✓ it should be re-orderable by mouse
✓ it should ignore out-of-bounds drops
✓ it should be re-orderable by keyboard
✓ it should allow items to be removed

Was unable to get tests working with `react-testing-library`, despite [one person's claim](https://github.com/atlassian/react-beautiful-dnd/issues/162#issuecomment-422461580)

### 🌟 Bonus
- Added propagation of `moreProps` to root-level `<ul>` element.
- Added `provided.placeholder` to `<Droppable>` so that its height is maintained while dragging.
- Added ~~`data-component-type`~~ `data-pka-anchor` attributes all over as per issue https://github.com/acl-services/paprika/issues/101

### DIFF
--- a/packages/Sortable/package.json
+++ b/packages/Sortable/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/sortable",
-  "version": "0.1.1",
+  "version": "0.1.0",
   "description": "Sortable component",
   "author": "@paprika",
   "license": "MIT",
--- a/packages/Sortable/package.json
+++ b/packages/Sortable/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/sortable",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "Sortable component",
   "author": "@paprika",
   "license": "MIT",
--- a/packages/Sortable/src/Sortable.js
+++ b/packages/Sortable/src/Sortable.js
@@ -68,7 +68,7 @@ const Sortable = ({ children, onChange, hasNumbers, onRemove, ...moreProps }) =>
             {...provided.droppableProps}
             {...moreProps}
             css={sortableStyles}
-            data-display-name="sortable"
+            data-pka-anchor="sortable"
             data-is-dragging-over={snapshot.isDraggingOver ? true : undefined}
             isDraggingOver={snapshot.isDraggingOver}
             ref={provided.innerRef}
--- a/packages/Sortable/src/Sortable.js
+++ b/packages/Sortable/src/Sortable.js
@@ -24,7 +24,7 @@ function filterChildren(children) {
   return React.Children.toArray(children).filter(child => child.type.displayName === "Sortable.Item");
 }
 
-const Sortable = ({ children, onChange, hasNumbers, onRemove }) => {
+const Sortable = ({ children, onChange, hasNumbers, onRemove, ...moreProps }) => {
   const I18n = useI18n();
   const dropId = React.useRef(uuid());
   const validChildren = filterChildren(children);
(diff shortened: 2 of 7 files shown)


packages/Sortable/src/components/SortableItem/SortableItem.js
```diff
@@ -26,30 +26,35 @@ const SortableItem = ({ children, index, hasNumbers, onRemove }) => {
...
-            )}
-          </li>
-        );
-      }}
+          )}
+          <div css={itemBodyStyles} data-component-type="sortable.item.body">
+            {children}
+          </div>
+          {onRemove && (
+            <div css={itemCloseStyles} data-component-type="sortable.item.remove">
```
mikrotron: Too much to add `data-component-type` throughout the DOM to allow precise control over styling? Or maybe just a different attribute name would be more appropriate since these aren't really _components_?

nahumzs: maybe just `data-type`?

mikrotron: Hm... `data-type` makes me think of arrays and hashmaps... maybe `data-element`? Or even `data-element-type` to maintain a similar naming pattern?

nahumzs: ohhh I suggested data-type because I thought *data-component-type* was too long, but on that note, I would suggested `data-react-display-name=""` maybe? or just `data-display-name`?

mikrotron: Oh. Yeah, both `component-type` and `display-name` are used in the issue https://github.com/acl-services/paprika/issues/101 so it's not clear what is actually being proposed.  

I like `component-type` better, but since we already set the `.displayName` for our components as a convention (https://github.com/acl-services/paprika/blob/master/packages/Sortable/src/Sortable.js#L87) maybe it's best to keep with the same name to reduce cognitive load? 

I was worried this might derail this PR, and it's not really that important at this point, but it's also a decision I would rather force now than just kick down the road.

nahumzs: yes, let's keep it with `data-display-name` then.

packages/Sortable/src/Sortable.js
```diff
@@ -66,7 +66,9 @@ const Sortable = ({ children, onChange, hasNumbers, onRemove }) => {
         {(provided, snapshot) => (
           <ul
             {...provided.droppableProps}
+            {...moreProps}
             css={sortableStyles}
+            data-display-name="sortable"
```
alexzherdev: In addition to https://github.com/acl-services/paprika/issues/101#issuecomment-508814582, this is even more confusing to me because it's on a DOM component. To an outside observer, this looks like some magical way of setting `displayName` on this `<ul>`, which it isn't.

packages/Sortable/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/sortable",
-  "version": "0.1.0",
+  "version": "0.1.1",
```
alexzherdev: I don't think we want to update versions manually in PRs. It should be handled by lerna when you publish (at which point you determine whether it's a major/minor/patch bump). cc @nahumzs

allison-c: How about the devdependencies in `paprika/package.json` file? Do we need to manually update it to keep it updated?

mikrotron: Good question, I suspect it _should_ be part of our publishing procedure though.

nahumzs: Lerna allowed the option of manually bumping the version vs automatically handle. Maybe we should discuss this a little further, but right now you don't need to bump the version mike

mikrotron: Yep, reverted in a more recent commit.

PR #111, opened 2019-07-05
Fix dev depencies babel runtime

oscarkwan:
### 🛠 Purpose

- babel-runtime should be a dependency, not a dev dependency.
---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

### DIFF
--- a/packages/Tokens/package.json
+++ b/packages/Tokens/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/tokens",
-  "version": "0.1.2",
+  "version": "0.1.1",
   "description": "Design tokens generated as scss + js",
   "author": "@paprika",
   "main": "lib/index.js",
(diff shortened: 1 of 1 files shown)


packages/Tokens/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/tokens",
-  "version": "0.1.1",
+  "version": "0.1.2",
```
alexzherdev: See https://github.com/acl-services/paprika/pull/110#discussion_r300758329

packages/Tokens/package.json
```diff
@@ -16,10 +16,8 @@
     "pretranspile": "node processTokens.js"
   },
   "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
```
alexzherdev: 👍

packages/Tokens/package.json
```diff
@@ -16,10 +16,8 @@
...
     "node-sass": "^4.9.2",
```
alexzherdev: Do you mind moving these two to devDeps?

oscarkwan: okiie

PR #113, opened 2019-07-09
BOT-1825 DropdownMenu component

tristanjasper:
### 🛠 Purpose

Migrate DropdownMenu component 

Reason:
Primarily to fix accessibility issues with the component, namely opening and closing the dropdown via keyboard commands, which is a required accessibility feature for robots.

Usage:

 ```js

import DropdownMenu from "@paprika/dropdown-menu";

<DropdownMenu
        renderTrigger={({ isOpen, handleOpenMenu }) => (
          <DropdownMenu.Trigger isOpen={isOpen} handleOpenMenu={handleOpenMenu}>
            Trigger
          </DropdownMenu.Trigger>
  )}>
<DropdownMenu.Item onClick={() => {}}>Edit</DropdownMenu.Item>
<DropdownMenu.Item onClick={() => {}}>Duplicate</DropdownMenu.Item>
</DropdownMenu>

```
 #### Props

**`<DropdownMenu>`**

Alignment of the dropdown menu
`align: oneOf(alignTypes)`

  Children should consist of <Dropdown.Item />
  `children: node.isRequired`

  Render prop for rendering the trigger element that toggles the dropdown
`renderTrigger: func.isRequired`

#### Default Props

 ` align: "bottom"`

### Available sub components

**<DropdownMenu.Item>**
Custom class name for element
`className: string`

HTML for each item
 `children: node.isRequired`

 If the item is destructive.
` isDestructive: bool`

Callback to be executed when button is clicked
`onClick: func`

Callback to be executed when dropdown needs to be closed
`onClose: func`

Callback to be executed when delete item is clicked
`onShowConfirmation: func`

Render prop to render the replacement node
`renderConfirmation: func`

**<DropdownMenu.LinkItem>**
  /** HTML for each LinkItem */
  children: node.isRequired,
  /** The url for the href */
  link: string.isRequired,
/** Should the link open content in a new tab */
  isExternal: bool,

**<DropdownMenu.Confirmation>**
`buttonSize: string
  confirmButtonType: string,
  confirmLabel: string.isRequired,
  description: string,
  onCancel: func.isRequired,
  onConfirm: func.isRequired,
  title: string.isRequired`

**<DropdownMenu.Trigger>**
`isOpen: bool.isRequired,
handleOpenMenu: func.isRequired`

**<DropdownMenu.Divider>**

### ✏️ Notes to Reviewer

**Other issues**

- When `Popover.Content` element is empty it still affects the a confirmation panels positioning.
image

- When aligning trigger to the right the DropDownMenu contents positioning doesn't align with the trigger button

### 🖥 Screenshots

image

image

---

### CODE REVIEW

File: packages/DropdownMenu/src/Confirmation/Confirmation.js [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+const defaultProps = {
+  buttonSize: ShirtSizes.DEFAULT,
+  confirmButtonType: "destructive",
+  description: null,
+};
+
+const Confirmation = ({ title, buttonSize, confirmButtonType, confirmLabel, description, onConfirm, onCancel }) => {
+  const I18n = useI18n();
+  return (
+    <Popover defaultIsOpen>
```
tristanjasper: Needing to use another popover component here as the confirmation panel is a different width to the initial content loaded. Another option would be to re-render the outer popover in DropDownMenu.js

packages/DropdownMenu/src/ContentContainer.styles.js
```diff
@@ -0,0 +1,23 @@
...
+  min-width: 150px;
+  padding: ${tokens.space} 0;
+  text-align: left;
+
+  ${({ isOpen }) => {
+    if (isOpen === false) {
+      return { opacity: 0 };
+    }
+  }}
+`;
```
tristanjasper: I'm setting the opacity to 0 for the container to stop the transition of opacity on the initial popover content, at the point of closing the confirmation panel. The Paprika Popover has a 150ms opacity transition built in.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,102 @@
...
+
+  const handleCloseMenu = () => {
+    setIsOpen(false);
+    setIsConfirming(false);
+    setRenderConfirmation(null);
+  };
+
+  const handleOpenMenu = () => {
+    setIsOpen(true);
+  };
```
tristanjasper: As the Paprika Popover is now handling opening and closing keyboard shortcuts for its content, as well as handling popover closing when clicking outside of the content area, I felt handling dedicated `opening` or `closing` actions would be beneficial over a `toggle` action.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,102 @@
...
+    <Popover
+      align={align}
+      className={className}
+      offset={4}
+      role={!isConfirming ? "menu" : null}
+      isOpen={isOpen}
+      onClose={() => {
+        if (!isConfirming) {
+          handleCloseMenu();
+        }
```
tristanjasper: onClose is being triggered by the Paprika popover component when we want it to render the confirmation panel. This is because of the unmount of rendered content in the popover content causing an `onBlur` event. 
Another solution may be to ensure focus is retained on the popover content so as not to trigger the onBlur, and to stop an `onClose` event, similar to how we do in the Date Picker component - https://github.com/acl-services/paprika/pull/94/files#diff-f2f7e73e1aac4537ad454c8d3e6c3258R63

packages/DropdownMenu/src/ContentContainer.styles.js
```diff
@@ -0,0 +1,23 @@
...
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+const ContentContainerStyles = styled.div`
+  background: ${tokens.color.white};
+  border: 1px solid ${tokens.border.color};
+  border-radius: ${tokens.popover.borderRadius};
+  box-shadow: ${tokens.popover.shadow};
+  margin-left: ${stylers.spacer(2)};
+  max-width: 320px;
```
tristanjasper: Changed this `max-width` to match `max-width` that is the default in Popover content component

packages/DropdownMenu/src/Item/Item.js
```diff
@@ -0,0 +1,87 @@
...
+  className: string,
+
+  /** HTML for each item */
+  children: node.isRequired,
+
+  /** If the item is destructive. */
+  isDestructive: bool,
+
+  /**  If true, will render an <a> tag, otherwise a <span> is generated */
+  isLink: bool,
```
tristanjasper: Instead of creating a separate <Dropdown.Link> as in the acl-ui 3 version, I've elected to pass an `isLink` prop here which renders out the correct element. As the css styles being applied to the elements would be the same for either element i felt this should be sufficient

packages/DropdownMenu/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@paprika/drop-down-menu",
+  "version": "0.0.1",
```
mikrotron: Should actually be `0.1.0`

packages/DropdownMenu/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/button": "^0.1.1",
+    "@paprika/popover": "^0.1.1",
+    "@paprika/raw-button": "^0.1.1",
+    "@paprika/stylers": "^0.1.1",
+    "@paprika/tokens": "^0.1.1",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "react": "^16.4.0",
```
mikrotron: `^16.8.0`

packages/DropdownMenu/stories/examples/Showcase.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import { Rule, Tagline } from "storybook/assets/styles/common.styles";
+import Heading from "@paprika/heading";
+import { DropDownMenuStory } from "../DropDownMenu.stories.styles";
+import DropDownMenuExample from "./DropDownMenuExample";
```
mikrotron: My Storybook fails to load, due to the second `D` being capitalized here, and the line above (but not for the actual file name).

packages/DropdownMenu/src/index.js
```diff
@@ -0,0 +1 @@
+export { default } from "./DropDownMenu";
```
mikrotron: The second `D` should not be capitalized here.  Unless you want to use `DropDownMenu` consistently everywhere (I'd support that adjustment, dropdown isn't really a single word).

tristanjasper: yeah thats why I did it, as we are calling it `drop-down-menu` in the repo

packages/DropdownMenu/stories/examples/Showcase.js
tristanjasper: Hmm, yeah i'm seeing storybook not compiling it correctly for some reason. As those casing errors keep reappearing

packages/DropdownMenu/src/index.js
oscarkwan: Dropdown is one word though..  :(

mikrotron: Never mind, Oscar is correct. It’s one word. We should keep it like it was.

packages/DropdownMenu/src/ContentContainer.styles.js
mikrotron: Interesting... maybe we should make a token?

packages/DropdownMenu/src/index.js
tristanjasper: Oh, yeah thats how I originally had it @mikrotron. Ok i'll change it back to that.

tristanjasper: I'll be reverting the paprika package name to `@paprika/dropdown-menu` againd

packages/DropdownMenu/src/ContentContainer.styles.js
tristanjasper: Sure, that could be a good idea

johnpeele:
One thought... We need to make sure that the font and button sizes in the popover match the font and button size in the trigger. I have noticed that sometimes we'll have a medium button with large text in the popover.

packages/DropdownMenu/src/Item/Item.js
nahumzs: Let's use `<Dropdown.Link />` since is similar to the other apis that we already have in paprika, We are trying to remove props as far we can from the components so are more easy to understand. 

```js
<DropDownMenu>
  <DropDown.Item>Some 1</DropDown.Item>
  <DropDown.Item>Some 2</DropDown.Item>
  <DropDown.Link>Link</DropDown.Link>
</DropDownMenu>
```

packages/DropdownMenu/stories/examples/Showcase.js
tristanjasper: Should be fine now after the component rename

packages/DropdownMenu/src/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,56 @@
...
+  title: string.isRequired,
+};
+
+const defaultProps = {
+  buttonSize: ShirtSizes.DEFAULT,
+  confirmButtonType: "destructive",
+  description: null,
+};
+
+const Confirmation = ({ title, buttonSize, confirmButtonType, confirmLabel, description, onConfirm, onCancel }) => {
```
nahumzs: this is a functional component please use the parameter (prop) and destructured everything in the first line :).

The only place where we destructure props from the parameter is while using style component.

also might be a good idea to rewrite this as:
```js
export default function Confirmation() {
  ... your code 
}
```

^ that will make it clearer that is a component.

packages/DropdownMenu/src/Divider/Divider.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+import DividerStyles from "./Divider.styles";
+
+const Divider = () => <span css={DividerStyles} />;
```
nahumzs:
```js
export default function Divider() {
  return <span css={DividerStyles} /> 
}
```

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+
+const propTypes = {
+  /** Alignment of the dropdown menu */
+  align: oneOf(alignTypes),
+
+  /** Children should consist of <Dropdown.Item /> */
+  children: node.isRequired,
+
+  /** Optional custom classes */
+  className: string,
```
nahumzs: this is not needed it you can be destructured via `...moreProps`. 

```js
  return (
    <Popover
      align={align}
     {...moreProps}
    >
```

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+
+  /** Render prop for rendering the trigger element that toggles the dropdown */
+  renderTrigger: func.isRequired,
+};
+
+const defaultProps = {
+  align: "bottom",
+  className: "",
```
nahumzs: we can omit className from here if we pass down {...moreProps}

packages/DropdownMenu/src/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,56 @@
...
+              </Button>
+            </div>
+          </div>
+        </ContentContainerStyled>
+      </Popover.Content>
+    </Popover>
+  );
+};
+
+Confirmation.componentType = "DropdownMenu.Confirmation";
```
nahumzs: we favour `displayName` instead of componentType, I have filed an issue to replace the current component that use componentType to displayName.

packages/DropdownMenu/src/Confirmation/Confirmation.js
mikrotron: Hm, I wonder if our `<Popover>` should have an imperative API to allow repositioning? I bet this is not the only use case where it would come in handy.  What do you think, @nahumzs?

packages/DropdownMenu/src/Item/Item.js
mikrotron: This doesn't seem to be working quite right in the Storybook example currently.  The Link does get rendered as an `<a>`, but the `onClick` doesn't make sense, it needs an `href` attribute.  And likely a way to launch in a new tab (maybe the introduction of an `isExternal` prop for the `<Dropdown.Link>` would be a good idea?).

packages/DropdownMenu/stories/examples/DropdownMenuExample.js
```diff
@@ -0,0 +1,59 @@
...
+        renderTrigger={({ isOpen, handleOpenMenu }) => (
+          <DropdownMenu.Trigger isOpen={isOpen} handleOpenMenu={handleOpenMenu}>
+            Trigger
+          </DropdownMenu.Trigger>
+        )}
+      >
+        <DropdownMenu.Item onClick={() => {}}>Edit</DropdownMenu.Item>
+        <DropdownMenu.Item onClick={() => {}}>Duplicate</DropdownMenu.Item>
+        <DropdownMenu.Item isDestructive isDisabled onClick={() => {}}>
+          Google
```
mikrotron: If this is for Google, it should be a "link", right?  
Also, right now the styling is not "destructive" or "disabled", it just looks like a regular item (but isn't clickable).  I'd recommend hooking up the `onClick` props to a Storybook `action`, so you can verify if they are actually firing or not, as intended.  Take a look at the Button's story for an example:
https://github.com/acl-services/paprika/blob/master/packages/Button/stories/examples/Basic.js#L19

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+      offset={4}
+      role={!isConfirming ? "menu" : null}
+      isOpen={isOpen}
+      onClose={() => {
+        if (!isConfirming) {
+          handleCloseMenu();
+        }
+      }}
+    >
+      <Popover.Trigger>{props.renderTrigger(getTriggerStateAndHelpers())}</Popover.Trigger>
```
mikrotron: In the storybook you'll notice that the Trigger Button gets the focus twice... because a RawButton is being wrapped around the Trigger.  I think this is a bit of a flaw in our Popover API, but for now, if you pass a function as the `children`, it won't do that, and the `A11y` will be better.  So, simply:
```jsx
<Popover.Trigger>{() => props.renderTrigger(getTriggerStateAndHelpers())}</Popover.Trigger>
```
Will remedy this (for now).

mikrotron: (You could leave a `// TODO:` in the code to fix this later though)

packages/DropdownMenu/src/Confirmation/Confirmation.js
alexzherdev: If it's sufficient to unmount the old popover and mount a new one, we can use the `key` trick to achieve that. (i.e. switching from `<Popover key="menu" />` to `<Popover key="confirmation" />` will unmount/mount as opposed to updating)

packages/DropdownMenu/src/Confirmation/Confirmation.js
alexzherdev: See also: https://github.com/acl-services/workpapers/pull/11888#discussion_r302283896 @nahumzs

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+const DropdownMenu = props => {
+  const { align, children, className } = props;
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [isConfirming, setIsConfirming] = React.useState(false);
+  const [renderConfirmation, setRenderConfirmation] = React.useState(null);
+
+  const handleCloseMenu = () => {
+    setIsOpen(false);
+    setIsConfirming(false);
+    setRenderConfirmation(null);
```
mikrotron: I think this is probably the place to restore the document focus to the Trigger element (currently is lost and reverts to the `body` after closing).

packages/DropdownMenu/src/Confirmation/Confirmation.js
tristanjasper: @alexzherdev So you think it would be better to to the re-render of initial popover, Do you think there is there an advantage to either way? Performance maybe?

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+  };
+
+  const handleOpenMenu = () => {
+    setIsOpen(true);
+  };
+
+  const getTriggerStateAndHelpers = () => {
+    return {
```
mikrotron: Don't really need an explicit `return` here... 🤷‍♂

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+    setRenderConfirmation(prevIsConfirmingState => (prevIsConfirmingState ? null : renderConfirmation));
+  };
+
+  const renderContent = () => {
+    if (isConfirming) {
+      return renderConfirmation(handleCloseMenu);
+    }
+
+    return (
+      <ContentContainerStyled isOpen={isOpen}>
```
mikrotron: If just the actual styles are exported from `ContentContainer.styles.js`, then this could be something like: `<div css={contentStyles}`.

packages/DropdownMenu/src/ContentContainer.styles.js
```diff
@@ -0,0 +1,23 @@
+import styled from "styled-components";
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+const ContentContainerStyles = styled.div`
```
mikrotron: The approach we've been leaning towards is to just export the raw styles, so this could simply be something like:
```js 
export default const contentStyles = css`
```
...but if you have an argument for why this approach is preferred, I'm all 👂👂

packages/DropdownMenu/src/ContentContainer.styles.js
mikrotron: How about a more terse one-liner:
```
${({ isOpen }) => (!isOpen ? "opacity: 0;" : "")}
```

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+        {React.Children.toArray(children).map(child => {
```
mikrotron: Do we need `toArray()` here, or can we just use `React.Children.map()`?

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+          return child;
+        })}
+      </ContentContainerStyled>
+    );
+  };
+
+  return (
+    <Popover
+      align={align}
+      className={className}
```
mikrotron: WE should probably propagate `...moreProps` onto this element, and then could avoid these two lines.

packages/DropdownMenu/src/ContentContainer.styles.js
nahumzs: I don't think you can do `export default const` you might still need to do the `const` but definitely, you can use `css''` instead

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+      role={!isConfirming ? "menu" : null}
+      isOpen={isOpen}
+      onClose={() => {
+        if (!isConfirming) {
+          handleCloseMenu();
+        }
+      }}
+    >
+      <Popover.Trigger>{props.renderTrigger(getTriggerStateAndHelpers())}</Popover.Trigger>
+      <Popover.Content>{renderContent()}</Popover.Content>
```
mikrotron: Since this part gets rendered into a portal, maybe we should have a mechanism to append a custom `className` (and maybe other testing attributes?) to this element? The app consumer will not be able to tag this otherwise which will be a problem for testing I think.  

I ran into this issue when using the `<Popover>` in Results.

packages/DropdownMenu/src/DropdownMenu.js
nahumzs: Can you created an issue about this, so we can address this later on the popover @tristanjasper 🙏

packages/DropdownMenu/stories/examples/DropdownMenuExample.js
nahumzs: Also, can we change Goggle for Galvanize :P ...

packages/DropdownMenu/src/Confirmation/Confirmation.js
nahumzs: We can do the key trick from now, and if we see another case for having an imperative API we can implement mike.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,111 @@
...
+      </ContentContainerStyled>
+    );
+  };
+
+  return (
+    <Popover
+      align={align}
+      className={className}
+      offset={4}
+      role={!isConfirming ? "menu" : null}
```
mikrotron: This guy should actually be on the `Popover.Content`

packages/DropdownMenu/src/DropdownMenu.js
nahumzs: something like: `<Popover morePropsPopover={}` `morePropsTrigger={} />` 🙈

packages/DropdownMenu/src/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,56 @@
...
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import Popover from "@paprika/popover";
+import ContentContainerStyled from "../ContentContainer.styles";
+import ConfirmationStyles from "./Confirmation.styles";
+
+const { func, string } = PropTypes;
+
+const propTypes = {
+  buttonSize: string,
```
mikrotron: This should be constrained to `ShirtSizes`... but actually, to JP's point, I kind of think we should have a size prop for the whole `<DropdownMenu>` that will dictate the sizing of _everything_.  A little more complicated, but easier to use, and enforces consistency.

packages/DropdownMenu/src/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,56 @@
...
+  confirmButtonType: string,
```
mikrotron: This should also be constrained... probably just to `[primary, destructive]`, but maybe there are other use cases to support?

packages/DropdownMenu/src/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,56 @@
...
+  confirmLabel: string.isRequired,
+  description: string,
```
mikrotron: Can we make this a `node`?  At some point someone will probably want to include a help link or something...

packages/DropdownMenu/src/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,56 @@
...
+  onCancel: func.isRequired,
+  onConfirm: func.isRequired,
+  title: string.isRequired,
```
mikrotron: 😬 Can we also rename this to `header` or `heading`? I think as a rule we should avoid `title`, which is a valid html attribute, for a prop name.

packages/DropdownMenu/src/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Button from "@paprika/button";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import Popover from "@paprika/popover";
+import ContentContainerStyled from "../ContentContainer.styles";
+import ConfirmationStyles from "./Confirmation.styles";
+
+const { func, string } = PropTypes;
```
alexzherdev: image

packages/DropdownMenu/src/Item/Item.styles.js
```diff
@@ -0,0 +1,45 @@
...
+
+const ItemStyles = css`
+  cursor: pointer;
+  display: block;
+  ${stylers.fontSize()};
+  font-family: ${tokens.fontFamily.default};
+  font-weight: normal;
+  padding: ${tokens.spaceSm} ${stylers.spacer(2)};
+  text-decoration: none;
+  line-height:1.5;
```
mikrotron: We have a `stylers.lineHeight()` for this now 😄

packages/DropdownMenu/src/Item/Item.styles.js
```diff
@@ -0,0 +1,45 @@
...
+  &:hover{
+    ${({ isDestructive, isDisabled }) => {
+      if (!isDisabled) {
+        const bgColorString = isDestructive ? "colorOrangeLighten40" : "blackLighten70";
+        return { backgroundColor: tokens.color[bgColorString] };
+      }
+    }}
+  }
+
+
```
mikrotron: extra spaces

packages/DropdownMenu/src/Item/Item.styles.js
```diff
@@ -0,0 +1,45 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+const ItemStyles = css`
```
mikrotron: nit: `itemStyles`.  This isn't a (styled) component, so lowerCamelCase is preferred.

packages/DropdownMenu/src/Item/Item.styles.js
```diff
@@ -0,0 +1,45 @@
...
+  font-weight: normal;
+  padding: ${tokens.spaceSm} ${stylers.spacer(2)};
+  text-decoration: none;
+  line-height:1.5;
+
+  &:focus {
+    box-shadow: ${tokens.highlight.active.withBorder}
+    outline: none;
+    position: relative;
+    z-index: stylers.z(1);
```
mikrotron: Don't forget the `${}` or this will just get ignored.

packages/DropdownMenu/src/Item/Item.styles.js
alexzherdev: nit^2: lowerCamelCase is just camelCase. "upper" camelCase is often called PascalCase 😄

packages/DropdownMenu/src/Confirmation/Confirmation.js
nahumzs: `PropTypes.func` and `PropTypes.string` will do the trick.

packages/DropdownMenu/src/DropdownMenu.js
mikrotron: I can create the issue... I noticed it first 😬

packages/DropdownMenu/stories/examples/DropdownMenuExample.js
mikrotron: ...and here's a URL you can use to test it:
https://youtu.be/IdkCEioCp24?t=73

packages/DropdownMenu/src/Confirmation/Confirmation.js
mikrotron: I agree, for _now_ the `key` is a fine solution... but do you think I should I add and imperative API for repositioning to the list in the Popover's github issue...?

tristanjasper: People, I think for and against opinions and reasons would be be helpful here as anyone revisiting this PR should be able to understand why we go a certain route.

mikrotron: Modifying the Popover is a bit of scope creep, changing the `key` will work and keeps you focussed on this component.  We can improve performance / improve the Popover in a later task, but it's not really the concern of this PR.

packages/DropdownMenu/stories/examples/DropdownMenuExample.js
```diff
@@ -0,0 +1,59 @@
...
+        <DropdownMenu.Item isLink onClick={() => {}}>
+          Is Link Item
+        </DropdownMenu.Item>
+        <DropdownMenu.Item isDisabled onClick={() => {}}>
+          Galvanize
+        </DropdownMenu.Item>
+        <DropdownMenu.Divider />
+        <DropdownMenu.Item
+          isDestructive
+          renderConfirmation={handleCloseMenu => {
```
mikrotron: Not a big deal, especially for story code, but I'd prefer `onCloseMenu` (or even `onClose`) for this prop.  It's not defined here, so I wouldn't use a `handle` prefix.  I _would_ use it for `handleConfirm` and `handleCancel`, like you are doing... _except_ I think you can use use this `onClose` prop directly.

packages/DropdownMenu/src/Confirmation/Confirmation.js
tristanjasper: @mikrotron Confused by this comment, My intention was never to modify the Popover, especially not for performance, I am wondering why we should favour the re-rerender of the base Popover instead of using a separate Popover for the confirmation panel. Would it be more performant to do either? My original gut feeling was to re-rerender the Popover but it would be good to discuss the reasons behind it.

packages/DropdownMenu/stories/DropdownMenu.stories.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+import { Story } from "storybook/assets/styles/common.styles";
+
+// Common DropdownMenu story styles
+
+export const DropdownMenuStory = styled(Story)`
+  max-width: 500px;
```
mikrotron: Do you need to confine the story to 500px? If not, you can skip this whole file and just use `Story` directly in your Showcase story.

packages/DropdownMenu/src/Confirmation/Confirmation.js
mikrotron: Oh sorry, I was thinking imperative API vs updating `key`.  I believe @alexzherdev's suggestion was to modify the `key` on the Popover to re-render, instead of including another Popover here.  I'll let him speak to the advantages of that approach though.

nahumzs: I thought the confirmation popover was a popover inside of a popover, but seems you just want to rerender the popover but with the confirmation panel once the user clicked the delete button.

In that scenario I think would be sufficient to use the `key` replacement technique, what the key is doing is just telling react that the component is a "fresh component" so will drop any state and remove the Dom from the tree while starting a new entire react cycle is like having the popover render as it were the first time. 
[key in react](https://reactjs.org/docs/reconciliation.html#keys) and a more digest explanation) [force reset components](https://medium.com/@albertogasparin/forcing-state-reset-on-a-react-component-by-using-the-key-prop-14b36cd7448e)

What could occur is that the popover will show a small "flicker/flash" when replacing the key and loading the component again, But if this flickering is not a big of the deal, just change the key and you will be fine.

Having two popovers will let you have more control in the transition from one popover to the other one because you could control both popover's state, so you can decide whenever to close it or open it each of them.

packages/DropdownMenu/stories/examples/Showcase.js
```diff
@@ -0,0 +1,18 @@
...
+import Heading from "@paprika/heading";
+import { DropdownMenuStory } from "../DropdownMenu.stories.styles";
+import DropdownMenuExample from "./DropdownMenuExample";
+
+const ExampleStory = () => (
+  <DropdownMenuStory>
+    <Heading level={1} displayLevel={2} isLight>
+      Showcase
+    </Heading>
+    <Tagline>Use the knobs to tinker with the props.</Tagline>
```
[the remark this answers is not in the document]
mikrotron: Ha. This is still "experimental".  Though I thought it might be fun if each developer replaced it with their own joke... like:
https://github.com/acl-services/paprika/blob/master/packages/Sortable/stories/examples/Showcase.js#L30

packages/DropdownMenu/src/Confirmation/Confirmation.js
tristanjasper: 😬

packages/DropdownMenu/src/ContentContainer.styles.js
tristanjasper: i've used `${({ isOpen }) => (isOpen === false ? "opacity: 0;" : "")}` as isOpen is not a prop on a content container for the items that don't need a confirmation, so it could also be `undefined`

mikrotron: Even better!

packages/DropdownMenu/src/DropdownMenu.js
tristanjasper: Ok added

packages/DropdownMenu/src/DropdownMenu.js
tristanjasper: hmm not sure i follow this. Do you mean adding additional trigger/content props to the popover component? Which are then spread on the `Trigger` and `Content` components respectively?

packages/DropdownMenu/src/Confirmation/Confirmation.js
tristanjasper: Ok i've restricted it to ShirtSizes.Default

packages/DropdownMenu/src/Confirmation/Confirmation.js
tristanjasper: I think sounds about right for now as this is only meant to be a port

packages/DropdownMenu/src/DropdownMenu.js
mikrotron: The Trigger doesn't matter, it's contained in the root node for this component that already gets `moreProps` added to it, so that's sufficient.  But those `moreProps` will have no impact on the Popover.Content which is rendered in a portal `<div>`.  Not sure about how to best handle this though.... _could_ be more like this?
```jsx
<DropdownMenu>
  <DropdownMenu.Content className="something" data-qa-id="something" />
</DropdownMenu>
```

packages/DropdownMenu/src/Confirmation/Confirmation.js
tristanjasper: gone with heading for now

packages/DropdownMenu/src/DropdownMenu.js
tristanjasper: 😕 Hmm, Let's discuss this. I'm unsure if you mean we should use a permanent class name ie "popOver--content" or want to spread an additional props object here.

packages/DropdownMenu/src/Item/Item.styles.js
tristanjasper: 👌

packages/DropdownMenu/stories/DropdownMenu.stories.styles.js
tristanjasper: I've left this in as it gives us some padding to the edge of the screen, otherwise the popup content will display slightly offset from the side of the screen which looks odd.

mikrotron: Oh, I see! A hard-coded width will only sometimes work though, depending on the size of the window / width of add-on panel... `padding-right` is probably a better approach, no?

packages/DropdownMenu/src/DropdownMenu.js
nahumzs: I like your approach @mikrotron

packages/DropdownMenu/src/Confirmation/Confirmation.js
tristanjasper: I've been running into problems with onBlur/focus when re-rendering one popover and rendering the confirmation panel in this way. I've created a secondary branch - https://github.com/acl-services/paprika/tree/bot-1825-dropdown-paprika-rerender - with those changes if someone want to investigate the solution further, but for iteration of the component I plan to stick with the approach of rendering the a second popover for the confirmation panel as this also is a working solution.

packages/DropdownMenu/src/DropdownMenu.js
tristanjasper: Well we arn't sure what is being passed in a children in this the items could in theory contain more than just text, so I'm not sure if we have a good solution for setting the key as we don't allow to use `index` Ideas?

mikrotron: I mean use doing this instead:
`React.Children.map(children, child => { ... });`
I saves a step (calling `toArray()`) and should be a bit more performant:
https://stackoverflow.com/questions/44721768/react-children-map-vs-react-children-toarray-map

packages/DropdownMenu/src/DropdownMenu.js
tristanjasper: @mikrotron Ok perhaps can be added in next iteration as this PR is meant for port from acl-ui. If there is still alot more todo in this Pr i can address.

mikrotron: Sure. Can you open an issue in GitHub to capture these kinds of "to-do" items.

nahumzs:
Just a comment key/down arrows are not implemented, and will not in this PR, though I wonder if it's something we need to address and support. If the answer is yes then we need to create a ticket for this task. cc @mikrotron

nahumzs:
On the story "dropdownmenu--dropdownmenu-with-longest-story" words don't include ellipsis, what should be the resolutions for those options that are larger than the space provided by the menu? 

Should we need to show an ellipsis or will just jump to a new line creating an item which will have a bigger height? cc @mikrotron

nahumzs:
Let's add a `components` folder and put the components inside there. So follow the rest of our components file structure.

packages/DropdownMenu/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/dropdown-menu"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/button": "^0.1.1",
+    "@paprika/popover": "^0.1.1",
```
mikrotron: Still needs `@paprika/helpers`

mikrotron: ...and `@paprika/l10n`

packages/DropdownMenu/src/Item/Item.js
```diff
@@ -0,0 +1,81 @@
...
+
+const propTypes = {
+  /** HTML for each item */
+  children: node.isRequired,
+
+  /** If the item is destructive. */
+  isDestructive: bool,
+
+  /**  If true, will render an <a> tag, otherwise a <span> is generated */
+  isLink: bool,
```
mikrotron: The `isLink` prop can now be removed, right?

packages/Tokens/src/tokens.yaml
```diff
@@ -147,6 +147,7 @@ card:
 popover:
   border-radius: $card--border-radius
   shadow: 0 2px 4px rgba(0, 0, 0, 0.1)
+  maxWidth: 320
```
mikrotron: `max-width`

mikrotron: Actually, should be `max-width: 320px`.  That means adjusting the code too though.

packages/DropdownMenu/src/Item/Item.js
```diff
@@ -0,0 +1,81 @@
...
+    onClose();
+  };
+
+  const itemProps = {
+    onClick: renderConfirmation !== null ? onShowConfirmation : handleClickItem,
+    role: "menuitem",
+    isDestructive,
+  };
+
+  if (isLink) {
```
mikrotron: ...and this whole block can be removed.

packages/DropdownMenu/src/Item/Item.js
```diff
@@ -0,0 +1,81 @@
...
+    return (
+      <a css={ItemStyles} {...itemProps}>
+        {children}
+      </a>
+    );
+  }
+  return (
+    <RawButton css={ItemStyles} {...itemProps} {...remainingProps}>
```
mikrotron: nit: we settled on `moreProps` some time after this original component was created.  Let's stick with `moreProps` in Paprika.

packages/DropdownMenu/src/LinkItem/LinkItem.js
```diff
@@ -0,0 +1,43 @@
...
+const LinkItem = props => {
+  const { children, isExternal, link, ...remainingProps } = props;
+
+  const linkItemProps = {
+    role: "menuitem",
+    ...remainingProps,
+  };
+
+  if (isExternal) {
+    linkItemProps.target = "_blank";
```
mikrotron: `rel="noopener noreferrer"` should be added to the element too, please.

Reference: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

packages/DropdownMenu/src/Trigger/Trigger.js
```diff
@@ -0,0 +1,35 @@
...
+const defaultProps = { triggerRef: null };
+
+const Trigger = props => {
+  const { isOpen, handleOpenMenu, triggerRef, ...otherProps } = props;
+  return (
+    <Button
+      ref={triggerRef}
+      aria-expanded={isOpen}
+      aria-haspopup="true"
+      onClick={handleOpenMenu}
```
mikrotron: I think we should have `aria-controls` added here too.  It means we probably need to generate an `id` though... 😒 

Reference:
https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html

packages/DropdownMenu/src/LinkItem/LinkItem.js
```diff
@@ -0,0 +1,43 @@
...
+const defaultProps = {
+  isExternal: false,
+};
+
+const LinkItem = props => {
+  const { children, isExternal, link, ...remainingProps } = props;
+
+  const linkItemProps = {
+    role: "menuitem",
+    ...remainingProps,
```
mikrotron: You could add `link` here to keep all the props together.

packages/DropdownMenu/src/LinkItem/LinkItem.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import PropTypes from "prop-types";
+import LinkItemStyles from "../Item.styles";
```
mikrotron: `linkItemStyles` (lower `l`)

packages/DropdownMenu/src/ContentContainer.styles.js
```diff
@@ -0,0 +1,19 @@
...
+  margin-left: ${stylers.spacer(2)};
+  max-width: ${tokens.popover.maxWidth}px;
+  min-width: 150px;
+  padding: ${tokens.space} 0;
+  text-align: left;
+
+  ${({ isOpen }) => (isOpen === false ? "opacity: 0;" : "")}
+`;
+
+export default contentContainerStyles;
```
mikrotron: I think this file should be `DropdownMenu.styles.js`.  

And this export could easily be just `contentStyles`... but that's nitpicking.

packages/DropdownMenu/src/Confirmation/Confirmation.styles.js
```diff
@@ -0,0 +1,27 @@
...
+const ConfirmationStyles = css`
+  padding: ${stylers.spacer(2)};
+
+  .dropdown-menu__confirmation-header {
+    ${stylers.fontSize()};
+    font-weight: bold;
+  }
+
+  .dropdown-menu__confirmation-description {
+    line-height: 1.3;
```
mikrotron: `stylers.lineHeight()`

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -0,0 +1,52 @@
...
+  display: block;
+  ${stylers.fontSize()};
+  font-family: ${tokens.fontFamily.default};
+  font-weight: normal;
+  padding: ${tokens.spaceSm} ${stylers.spacer(2)};
+  text-decoration: none;
+  line-height: ${stylers.lineHeight()};
+
+  &:focus {
+    box-shadow: ${tokens.highlight.active.withBorder}
```
mikrotron: This is broken.  I regret our naming of this token, but you need to use:
`box-shadow: ${tokens.highlight.active.withBorder.boxShadow};`

😬

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -0,0 +1,52 @@
...
+    outline: none;
+    position: relative;
+    z-index: ${stylers.z(1)};
```
mikrotron: Just:
`${stylers.z(1)}`

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -0,0 +1,52 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+const itemStyles = css`
+  cursor: pointer;
+  display: block;
+  ${stylers.fontSize()};
```
mikrotron: This should be first, before `cursor`.

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -0,0 +1,52 @@
...
+  font-family: ${tokens.fontFamily.default};
+  font-weight: normal;
+  padding: ${tokens.spaceSm} ${stylers.spacer(2)};
+  text-decoration: none;
+  line-height: ${stylers.lineHeight()};
```
mikrotron: Just:
${stylers.lineHeight()}

mikrotron:
@nahumzs, the current implementation breaks to multiple lines, and looks fine to me, but it might be up to a designer to make the call.  However, if there's no whitespace to break on, it overflows, which looks bad, we should probably hide that overflow, though it's such an edge case, we probably don't need to worry about ellipsising it.

alexzherdev:
Since this is an edge case and we probably can forgo full cross-browserness, we could try https://css-tricks.com/almanac/properties/o/overflow-wrap/ plus https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens?

packages/Tokens/src/tokens.yaml
tristanjasper: Actually the popover component actually would use this token without the "px" hence why i left it off... wdyt?

packages/DropdownMenu/src/LinkItem/LinkItem.js
tristanjasper: iiiinteresting... 🥇

packages/Tokens/src/tokens.yaml
tristanjasper: Now passing "320px"

packages/DropdownMenu/src/ContentContainer.styles.js
tristanjasper: I have made it "DropdownMenuContentStyles.styles.js" as I think it is specific to the content of the dropdownMenu

packages/DropdownMenu/src/Item.styles.js
tristanjasper: how come? Are we ordering things differently when using stylers?

tristanjasper:
> Since this is an edge case and we probably can forgo full cross-browserness, we could try https://css-tricks.com/almanac/properties/o/overflow-wrap/ plus https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens?

I've added `overflow-wrap` property

packages/DropdownMenu/src/Item.styles.js
tristanjasper: I have reordered it to be at the top though

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,126 @@
...
+    setIsConfirming(false);
+    setRenderConfirmation(null);
+    if (triggerRef.current) triggerRef.current.focus();
+  };
+
+  const handleOpenMenu = () => {
+    setIsOpen(true);
+  };
+
+  const getTriggerStateAndHelpers = () => ({
```
oscarkwan: This, i believe does not need to be an arrow function

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,126 @@
...
+    isOpen,
+    handleOpenMenu,
+  });
+
+  const handleShowConfirmation = renderConfirmation => () => {
+    setIsConfirming(prevIsConfirmingState => !prevIsConfirmingState);
+    setRenderConfirmation(prevIsConfirmingState => (prevIsConfirmingState ? null : renderConfirmation));
+  };
+
+  const randomIdString = Math.random()
```
oscarkwan: is UUID better here?

packages/DropdownMenu/src/DropdownMenu.js
oscarkwan: @tristanjasper Have you addressed this?

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -0,0 +1,52 @@
...
+    if (isDisabled) {
+      colorString = isDestructive ? "orangeDisabled" : "blackDisabled";
+      return { color: tokens.color[colorString] };
+    }
+
+    colorString = isDestructive ? "orangeDarken10" : "colorBlack";
+    return { color: tokens.color[colorString] };
+  }}}
+
+  &:hover{
```
oscarkwan: minor, but i feel like there needs to be a space after hover
`hover {` `active {`

packages/DropdownMenu/src/DropdownMenu.js
alexzherdev: This will change on every render—probably not something we want to do. We can store this value in a ref

packages/DropdownMenu/stories/DropdownMenu.stories.styles.js
oscarkwan: @tristanjasper did you address this?

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -0,0 +1,126 @@
...
+      offset={4}
+      isOpen={isOpen}
+      onClose={() => {
+        if (!isConfirming) {
+          handleCloseMenu();
+        }
+      }}
+      {...moreProps}
+    >
+      <Popover.Trigger>{() => renderTrigger()}</Popover.Trigger>
```
alexzherdev:
```suggestion
      <Popover.Trigger>{renderTrigger}</Popover.Trigger>
```
should work?

packages/DropdownMenu/src/Item.styles.js
alexzherdev:
```suggestion
  &:hover {
```
nit

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -0,0 +1,52 @@
...
+    ${({ isDestructive, isDisabled }) => {
+      if (!isDisabled) {
+        const bgColorString = isDestructive ? "orangeLighten40" : "blackLighten70";
+        return { backgroundColor: tokens.color[bgColorString] };
+      }
+    }}
+  }
+
+  &:active{
```
alexzherdev:
```suggestion
  &:active {
```
nit

packages/DropdownMenu/src/DropdownMenu.js
alexzherdev: What I'm thinking is, we should be able to avoid using `cloneElement` here as well similarly to https://github.com/acl-services/paprika/pull/134/commits/f5063084fb5c90ab0477b913fc0ac8bd7b759b04.
We can put the two functions onto context instead (unconditionally). This would make the component more composable as it wouldn't rely on `DropdownMenu.Item` being its direct child.

packages/DropdownMenu/stories/examples/DropdownMenuDisabledExample.js
```diff
@@ -0,0 +1,24 @@
...
+    <DropdownMenu
+      align="bottom"
+      renderTrigger={({ isOpen, handleOpenMenu }) => (
+        <DropdownMenu.Trigger isOpen={isOpen} handleOpenMenu={handleOpenMenu}>
+          Trigger
+        </DropdownMenu.Trigger>
+      )}
+    >
+      <DropdownMenu.Item isDestructive isDisabled onClick={() => {}}>
+        Galvanize disabled distructive
```
oscarkwan: typo

packages/DropdownMenu/test/dropdownMenu.spec.js
```diff
@@ -0,0 +1,122 @@
...
+      </DropdownMenu>
+    </L10n>
+  );
+
+  return {
+    ...renderedComponent,
+  };
+}
+
+describe("DropdownMenu", () => {
```
oscarkwan: Could test focus here as well, can be another PR.

One example
"should focus on trigger after dropdown menu is closed"

tristanjasper:
@oscarkwan In reply to https://github.com/acl-services/paprika/pull/113#discussion_r305988266
Yes I discussed with Mike, I want to keep it in so that we have some padding to the side of the screen so leaving this in

packages/DropdownMenu/test/dropdownMenu.spec.js
tristanjasper: Ok added but i'm having a few issues with this which im investigating

tristanjasper: I think there are problems related to https://github.com/acl-services/paprika/pull/113#discussion_r302289468

PR #115, opened 2019-07-11
Ux 496 add collapsible

jamiek-acl:
### 🛠 Purpose

Create the `<Collapsible>` component

### 🖥 Screenshots

collapsible


packages/Collapsible/src/Collapsible.js
```diff
@@ -6,16 +6,43 @@ import useI18n from "@paprika/l10n/lib/useI18n";
...
+  isCollapsed: true,
+  isDisabled: false,
+  hasOnlyIconToggle: false,
+  onClick: () => {},
+};
+
 const Collapsible = props => {
   const I18n = useI18n();
-  let hasWarnedForAriaText = false;
+  let hasWarnedFora11yText = false;
```
mikrotron: So nit: `hasWarnedForA11yText`

packages/Collapsible/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+```js
+import Collapsible from "@paprika/Collapsible";
+
+const [isCollapsed, setIsCollapsed] = React.useState(false);
+
+const yourComponent = () => {
+  return (
+    <Collapsible
+      ariaText="collapsible section"
```
mikrotron: I think you updated this to `a11yText` already

packages/Collapsible/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@paprika/collapsible",
+  "version": "0.1.1",
```
mikrotron: Why not `0.1.0`? This is brand new, right?

packages/Collapsible/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/collapsible"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/stylers": "^0.1.1",
```
mikrotron: You'll also need the `raw-button`, `icon` and `l10n`.

packages/Collapsible/src/Collapsible.js
```diff
@@ -0,0 +1,135 @@
...
+import RightArrowIcon from "@paprika/icon/lib/ArrowRight";
+import DownArrowIcon from "@paprika/icon/lib/ArrowDown";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import RawButton from "../../RawButton";
+import collapsibleStyles from "./Collapsible.styles";
+
+const propTypes = {
+  a11yText: PropTypes.string,
+  children: PropTypes.node,
+  className: PropTypes.string,
```
mikrotron: Can omit, it will get passed along in `moreProps`.

packages/Collapsible/src/Collapsible.js
```diff
@@ -0,0 +1,135 @@
...
+        {label}
+      </RawButton>
+    );
+  };
+
+  const renderCollapsibleByIcon = () => {
+    const { a11yText, label, isCollapsed, isDisabled, iconAlign } = props;
+
+    return (
+      <div>
```
mikrotron: Can we use a `Fragment` instead, please?

packages/Collapsible/src/Collapsible.js
```diff
@@ -0,0 +1,135 @@
...
+        {label}
+      </div>
+    );
+  };
+
+  React.useEffect(() => {
+    checkPropsError();
+  });
+
+  /* Destructuring is used to isolate additional props in moreProps and pass them along to
```
mikrotron: This is such a standard practice now, I think we don't need to explain it.  Also, I thought it was needed to explain the disabled lint rule, which I no longer see...?

packages/Collapsible/src/Collapsible.js
```diff
@@ -0,0 +1,135 @@
...
+    iconCollapse,
+    iconExpand,
+    isCollapsed,
+    isDisabled,
+    hasOnlyIconToggle,
+    onClick,
+    label,
+    ...moreProps
+  } = props;
+  /* eslint-enable no-unused-vars */
```
mikrotron: ✂️

packages/Collapsible/src/Collapsible.styles.js
```diff
@@ -0,0 +1,48 @@
...
+
+const collapsibleStyles = props => `
+  box-sizing: border-box;
+  color: ${tokens.color.black};
+  display: block;
+  ${stylers.fontSize()};
+  line-height: 24px;
+  width: 100%;
+
+  > .raw-button {
```
mikrotron: I don't think the `RawButton` has this class... but I think `.collapsible__label` will suit your purpose instead.

packages/Collapsible/src/Collapsible.styles.js
```diff
@@ -0,0 +1,48 @@
...
+    border-radius: ${tokens.borderRadius};
+  }
+
+  .collapsible__body {
+    display: none;
+    width: 100%;
+    ${props.isCollapsed ? "display: none;" : "display: block;"}
+
+    p {
```
mikrotron: ☹️ Can we avoid this? Or at least be more specific (`> p:first-child { margin-top:0; }`)?

packages/Collapsible/src/Collapsible.styles.js
```diff
@@ -0,0 +1,48 @@
...
+
+  .collapsible--icon-left {
+    padding-right: ${tokens.spaceSm};
+  }
+
+  .collapsible--icon-right {
+    float: right;
+  }
+
+  .collapsible__aria-label {
```
mikrotron: No longer used.

packages/Collapsible/stories/examples/Basic.js
```diff
@@ -0,0 +1,27 @@
...
+import { Story } from "storybook/assets/styles/common.styles";
+import Collapsible from "../../src";
+
+const BasicCollapsibleStory = () => {
+  const [isCollapsed, setIsCollapsed] = React.useState(false);
+
+  return (
+    <Story>
+      <Collapsible
+        ariaText="collapsible section"
```
mikrotron: `a11yText`

packages/Collapsible/tests/Collapsible.spec.js
```diff
@@ -0,0 +1,82 @@
+import React from "react";
+import { render, fireEvent } from "react-testing-library";
+import Collapsible from "../src";
+
+const searchableContent = "The body of the collapsible";
+const defaultProps = {
+  ariaText: "collapsible section",
```
mikrotron: `a11yText`

packages/Collapsible/tests/Collapsible.spec.js
```diff
@@ -0,0 +1,82 @@
...
+  it("should not include label in button when hasOnlyIconToggle === true", () => {
+    const label = <span className="custom-label">my label</span>;
+    const { container } = renderComponent({ label, hasOnlyIconToggle: true });
+    expect(container.querySelector("span[role='button'] .custom-label")).not.toBeInTheDocument();
+  });
+
+  it("should include label in button when hasOnlyIconToggle === false", () => {
+    const label = <span className="custom-label">my label</span>;
+    const { container } = renderComponent({ label, hasOnlyIconToggle: false });
+    expect(container.querySelector("span[role='button'] .custom-label")).toBeInTheDocument();
```
mikrotron: `getByRole()`?

packages/Collapsible/src/Collapsible.styles.js
jamiek-acl: Hm why do you want to avoid this?

packages/Collapsible/package.json
jamiek-acl: Screen Shot 2019-07-11 at 2 17 07 PM

packages/Collapsible/src/Collapsible.styles.js
mikrotron: Because it will affect all the `<p>` elements in the Collapsible's content.  This shouldn't be the concern of the Collapsible, and the increased CSS specificity will make it annoying to undo if we want to in some situations.  Essentially, the Collapsible shouldn't know or care about what goes into it.

packages/Collapsible/src/Collapsible.js
```diff
@@ -0,0 +1,126 @@
...
+  const renderDefaultCollapsible = () => {
+    const { a11yText, label, isCollapsed, isDisabled, iconAlign } = props;
+
+    return (
+      <RawButton
+        ariaLabel={a11yText || I18n.t("collapsible.a11yText")}
+        aria-expanded={!isCollapsed}
+        className="collapsible__label"
+        isDisabled={isDisabled}
+        onClick={e => handleClickCollapse(e)}
```
mikrotron: Minor: `onClick={handleClickCollapse}`
(sorry, I think I missed this the first time around)

packages/Collapsible/README.md
```diff
@@ -0,0 +1,53 @@
+## Collapsible
+
+A controlled component that shows/hides content on click.
+
+### Installation
+
+`> npm install --save @paprika/Collapsible`
```
nahumzs: this is an error you can't capitalize package on npm.

packages/Collapsible/README.md
```diff
@@ -0,0 +1,53 @@
...
+or
+`> yarn add @paprika/Collapsible`
```
nahumzs: same here!

packages/Collapsible/README.md
mikrotron: Oops, I missed that too.  It's correct in the `package.json` so I don't think we need to worry too much. I'm going to take a pass at this too, when I use it for my `<ResponseAccordion>` component, so I can fix it then.

packages/Collapsible/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@paprika/collapsible",
+  "version": "1.0.0",
```
alexzherdev: @nahumzs should this be 0.1.0?

packages/Collapsible/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@paprika/icon": "^0.1.2",
+    "@paprika/l10n": "^0.1.2",
+    "@paprika/raw-button": "^0.1.2",
+    "@paprika/stylers": "^0.1.1",
+    "@paprika/tokens": "^0.1.1",
+    "class-names": "^1.0.0",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "react": "^16.4.0",
```
alexzherdev: 16.8.0 since this uses hooks

packages/Collapsible/package.json
nahumzs: yes indeed can you fix this @jamiek-acl

PR #117, opened 2019-07-12
AD-4014 Fix popover triggering by keyboard

allison-c:
### 🛠 Purpose

<Popover.Trigger> cannot triggered by keyboard

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---


nahumzs:
is adding space support right?

alexzherdev:
Btw I don’t think triggering on enter is standard for buttons?

nahumzs:
I think it's space and enter
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Keyboard_interactions

allison-c:
> is adding space support right?

Yes, I saw we use this in raw button as well

PR #118, opened 2019-07-12
UX-500 removing docz

nahumzs:
### 🛠 Purpose

Removing docz we are going to use Storybook mdx support for documentation. 

---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---


cypress/support/commands.js
```diff
@@ -7,7 +7,7 @@
 //
 // -- This is a parent command --
 
-import "@testing-library/cypress/add-commands";
+// import "@testing-library/cypress";
```
alexzherdev: remove?

jest.config.js
```diff
@@ -0,0 +1,4 @@
+// jest.config.js
+module.exports = {
```
alexzherdev: Any reason you prefer a separate file? Not sure which way I like better

nahumzs: I just felt that this doesn't belong to `package.json`, most of our tools configs are outside in their own files, so made sense to me to put it in their own file.

Issue #91, opened 2019-06-14
Repository links are not working on npm


alexzherdev:
# Bug Report

## Expected behavior
Repository links open to an existing page.

## Current behavior
Repository links are incorrect in most, if not all, packages, due to case sensitivity.

## Screenshots / Gifs / Codepens
![image](https://user-images.githubusercontent.com/93752/59532280-1bdd2c00-8e9d-11e9-8552-93ba7b81f54e.png)

## Additional context
Actually, providing a subdirectory within a monorepo doesn't seem to be the correct usage of this field, which is supposed to [refer to the actual cloneable repo](https://docs.npmjs.com/files/package.json#repository). [This example in Storybook](https://github.com/storybookjs/presets/blob/1a4ee04197eb062a7fd56817cd53071253495cdb/packages/preset-typescript/package.json#L11-L15) is the correct one. It would lead to those links just going to the root of the repo for all packages, but we can use the `homepage` field to provide a URL to the appropriate subdirectory.

nahumzs:
has been addressed here https://github.com/acl-services/paprika/pull/364

PR #123, opened 2019-07-15
Pd 12487 collapsible checklist

jamiek-acl:
### 🛠 Purpose

Create 'CollapsibleChecklists' component.
Can have optional headings.
Can disable parent or child.
Clicking checkbox of parent selects all children.
Clicking a parent or child fires a callback, passing in the children to change.
Labels of parent and child can be any node (not just a string).

---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

Screen Shot 2019-07-16 at 2 26 39 PM

---


packages/Collapsible/package.json
```diff
@@ -7,7 +7,7 @@
   "main": "lib/index.js",
   "repository": {
     "type": "git",
-    "url": "https://github.com/acl-services/paprika/tree/master/packages/collapsible"
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/Collapsible"
```
nahumzs: I think we should use homepage for this value and url property should point to the repo itself. https://github.com/acl-services/paprika/issues/91

packages/CollapsibleChecklists/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/collapsible": "^0.1.0",
+    "@paprika/stylers": "^0.1.1",
+    "@paprika/tokens": "^0.1.1",
+    "class-names": "^1.0.0",
```
nahumzs: 🤔 I don't think you are using this package maybe was a copy/pasta issue?

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Heading from "./components/Heading";
```
tristanjasper: should import paprika components instead of relative paths

packages/CollapsibleChecklists/src/components/Heading/Heading.js
```diff
@@ -0,0 +1,19 @@
...
+import PropTypes from "prop-types";
+import headingStyles from "./Heading.styles";
+
+const propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+const defaultProps = {};
+
+const Heading = ({ children }) => {
```
tristanjasper: return on one line

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
jamiek-acl: oh its not a paprika Heading, its my own

packages/CollapsibleChecklists/src/components/Item/Item.js
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    <div css={itemStyles} {...props}>
+      <input
+        aria-label={textContent(children)}
+        type="checkbox"
+        checked={isChecked}
+        disabled={isDisabled}
+        onChange={onChange}
+      />{" "}
```
tristanjasper: {" "} needed? Perhaps use styling?

packages/Collapsible/src/Collapsible.js
```diff
@@ -72,7 +72,7 @@ const Collapsible = props => {
     const { a11yText, label, isCollapsed, isDisabled, iconAlign } = props;
 
     return (
-      <React.Fragment>
+      <div data-pka-anchor="collapsible__heading">
```
mikrotron: `collapsible.heading`

packages/CollapsibleChecklists/README.md
```diff
@@ -0,0 +1,47 @@
+## CollapsibleChecklists
+
+A controlled component that has checkable groups of checkable items.
+Checking the parent checks all of the children.
+
+### Installation
+
+`> npm install --save @paprika/collapsible-list`
+or
+`> yarn add @paprika/collapsible-list`
```
mikrotron: `> ... @paprika/collapsible-checklists`

packages/CollapsibleChecklists/stories/examples/Basic.js
```diff
@@ -0,0 +1,138 @@
...
+
+  const handleExpand = () => {
+    setTimeout(() => {
+      const newSportsData = sportsData.slice(0);
+      newSportsData[1].sports[0].teams = [
+        { isChecked: false, isDisabled: false, name: "Cincinnati Bengals" },
+        { isChecked: false, isDisabled: false, name: "Cleveland Browns" },
+      ];
+      setSportsData(newSportsData);
+    }, 2000);
```
tristanjasper: create a static variable for this number eg `const EXPAND_DURATION=2000` 🤷‍♂

packages/CollapsibleChecklists/stories/examples/Basic.js
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  function handleOnChange(changedItemsArray) {
+    console.log("handleOnChange");
```
tristanjasper: remove console log

packages/CollapsibleChecklists/stories/examples/Basic.js
```diff
@@ -0,0 +1,138 @@
...
+    const newSportsData = sportsData.slice(0);
+
+    newSportsData.forEach(newSportsDatum => {
+      newSportsDatum.sports.forEach(sport => {
+        sport.teams.forEach(team => {
+          const thisTeamWasChanged = changedItemsArray.filter(changedItem => changedItem.key === team.name).length > 0;
```
tristanjasper: `.length` instead of `.length > 0`

packages/CollapsibleChecklists/src/components/Heading/Heading.styles.js
```diff
@@ -0,0 +1,11 @@
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+const headingStyles = () => `
+  background-color: ${tokens.table.header.backgroundColor};
+  border-bottom: 2px solid ${tokens.color.blackLighten60};
```
mikrotron: `tokens.border.color`

packages/CollapsibleChecklists/stories/examples/Basic.js
```diff
@@ -0,0 +1,138 @@
...
+    }, 2000);
+  };
+
+  function handleOnChange(changedItemsArray) {
+    console.log("handleOnChange");
+    const newSportsData = sportsData.slice(0);
+
+    newSportsData.forEach(newSportsDatum => {
+      newSportsDatum.sports.forEach(sport => {
+        sport.teams.forEach(team => {
```
tristanjasper: could these be mapped over instead of forEach?

packages/CollapsibleChecklists/stories/examples/Basic.js
alexzherdev: I beg to differ ¯\\\_(ツ)_/¯

packages/CollapsibleChecklists/stories/examples/Basic.js
```diff
@@ -0,0 +1,138 @@
...
+  function renderTeams(teams) {
+    return teams.map(team => (
+      <CollapsibleChecklists.Item key={team.name} isChecked={team.isChecked} isDisabled={team.isDisabled}>
+        {team.name}
+      </CollapsibleChecklists.Item>
+    ));
+  }
+
+  function renderOhioFootballTeams() {
+    if (sportsData[1].sports[0].teams.length === 0) {
```
tristanjasper: perhaps create variables for the items with namings like `const firstSport = sports[1].sports[0]` so that can avoid setting the array index in multiple places and easy to follow

packages/CollapsibleChecklists/src/components/Heading/Heading.styles.js
```diff
@@ -0,0 +1,11 @@
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+const headingStyles = () => `
```
rauldudu: better use `css`\` in these styles? https://github.com/acl-services/paprika/issues/98

packages/CollapsibleChecklists/stories/examples/Basic.js
mikrotron: I prefer using Storybook's `actions` addon, but it's just Storybook code... 🤷‍♂️.  
Cleaning up / curating the stories can/should be a separate task.

packages/CollapsibleChecklists/src/components/Group/Group.js
```diff
@@ -0,0 +1,120 @@
...
+      childItems.length;
+
+    let childItemsToChange = [];
+    if (allChildItemsAreChecked) {
+      childItemsToChange = childItems.filter(childItem => !childItem.props.isDisabled);
+    } else {
+      childItemsToChange = childItems.filter(childItem => !childItem.props.isChecked && !childItem.props.isDisabled);
+    }
+
+    onChange(childItemsToChange);
```
alexzherdev: I'm interested in hearing others' feedback about this. While discussing the API for `<Sortable>` with @mikrotron I said that an `onChange` API based on React elements doesn't make a lot of sense to me. In this case, given a set of React elements, how would the consumer know what to do? They would need to somehow map those elements to their actual data to determine what is being checked/unchecked. The storybook story doesn't really handle this, so it's not obvious that there's an issue.

We might want to require some kind of `id` prop on the items to handle this. (can't use `key` because that wouldn't be visible to us)

packages/CollapsibleChecklists/src/index.js
```diff
@@ -0,0 +1,3 @@
+import CollapsibleChecklists from "./CollapsibleChecklists";
```
alexzherdev:
```suggestion
export { default } from "./CollapsibleChecklists";
```

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
```diff
@@ -0,0 +1,41 @@
...
+
+const defaultProps = {};
+
+const CollapsibleChecklists = props => {
+  const { children, onChange } = props;
+
+  return (
+    <React.Fragment>
+      {children.map((child, index) => {
+        switch (child.type.displayName) {
```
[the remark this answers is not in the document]
mikrotron: Can we count on the `type` not being changed during transpilation / minification?

packages/CollapsibleChecklists/src/components/Group/Group.js
mikrotron: Since we expect an array, maybe just the `indexes` would be sufficient?

nahumzs: The idea is that you always received the selectedIndexes as the first argument, and we could pass the full list of elements as second argument so you could do:

```js
function handleChange(indexes, list) {
  indexes.forEach(index => console.log(list[index].props.value))
}
```

packages/CollapsibleChecklists/src/components/Group/Group.js
```diff
@@ -0,0 +1,120 @@
...
+    <React.Fragment>
+      <input
+        ref={checkboxRef}
+        checked={allAreChecked}
+        type="checkbox"
+        disabled={isDisabled}
+        onChange={toggleChildren}
+        aria-label={textContent(title)}
+      />
+      {title}
```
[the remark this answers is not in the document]
mikrotron: Well, that's what Jamie's trying to do with `textContent(title)`, but as mentioned, that won't really work without rendering the node. Even the library Jamie is using will probably not work for us in all situations:

https://github.com/rwu823/react-addons-text-content/issues/1#issuecomment-236906494

So yeah, we should probably require a string version from the consumer as an `a11yText` prop.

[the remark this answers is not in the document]
mikrotron: ... or ... maybe if there was a `visuallyHidden` `<span id={someUID}>{title}</span>`, and then we could have an `aria-labelledby={someUID}` on the `input` ... 🤔

packages/CollapsibleChecklists/stories/examples/Basic.js
jamiek-acl: It's just a story; think that is necessary?

packages/CollapsibleChecklists/stories/examples/Basic.js
jamiek-acl: Meh. Just a story and I dont think that helps readability.

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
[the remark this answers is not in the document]
alexzherdev: Yes that identity should be preserved. Especially when there is no hot reloading involved, there should be nothing to worry about AFAIK.

packages/CollapsibleChecklists/tests/CollapsibleChecklists.spec.js
```diff
@@ -0,0 +1,85 @@
...
+  it("after expand, does show group items", () => {
+    const { getByText, getByRole } = renderComponent();
+    fireEvent.click(getByRole(/button/i));
+    expect(getByText(/damask/i)).toBeVisible();
+  });
+
+  it("cannot expand a disabled group", () => {
+    const { getByText, container } = renderComponent();
+    expect(getByText(/easter/i)).not.toBeVisible();
+    fireEvent.click(container.querySelectorAll(".collapsible__label")[2]); // Lilies
```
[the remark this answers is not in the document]
jamiek-acl: Great catch

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
```diff
@@ -0,0 +1,41 @@
...
+          case Heading.displayName:
+            return <Heading key={`heading${index}`} {...child.props} />; // eslint-disable-line react/no-array-index-key
```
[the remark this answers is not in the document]
jamiek-acl: The `key` is necessary for React console warning:
```
Warning: Each child in a list should have a unique "key" prop.
```

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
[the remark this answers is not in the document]
jamiek-acl: I can remove the `.displayName` part

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
alexzherdev: But that wouldn't satisfy the requirement of checking the checkbox via clicking the label? Do we still want that?

packages/CollapsibleChecklists/src/components/Group/Group.js
alexzherdev: We only have a single `onChange` prop for the entire component now, so a single index would not be enough to describe a child checkbox, you'd also need an index of the group. (Alternatively, put a separate `onChange` prop on each group—which might be too repetitive though? Realistically a consumer is more likely to define a single callback handler for this component anyway I think?)

For the snippet above, seems like the list is redundant? The consumer is the one passing in those "values" in the first place, so they don't really need to be given them back. They already have some kind of state that they can iterate over.

I suggest creating a story that would better reflect the real world usage of this component (i.e. resembling the feature that will need it). That should make it easier to design.

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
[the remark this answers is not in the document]
alexzherdev: I'm kind of hesitant to recommend this though since that would require everyone to use the latest `react-hot-loader`... thoughts?

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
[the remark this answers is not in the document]
alexzherdev: See if that's still there if you just use `child` itself?
It should also disappear for the `Group` case if you use `React.cloneElement` since that should preserve keys.

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
mikrotron: Excellent point! In general clicking the label should activate the checkbox since that's such a small click target... it as a big impact on `a11y`!  In this specific case though, I wonder... there's not an easy "undo" to the select all / deselect all operation of this checkbox, and maybe it's better to confine it to the actual checkbox.  It could easily be assumed by the user that the `title` in this case would actually expand the Collapsible, and then they'd be frustrated.  

I think we need a @krismckinnonacl for this one!

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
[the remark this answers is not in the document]
mikrotron: How many apps are using the latest `react-hot-loader` vs the broken one?

Since there are a few components now using this `.displayName` approach, I'm leaning towards keeping it for now, not creating issues in our apps, and capturing this possible refactor in an issue which we can revisit when at least most of our apps have updated their 🔥 loaders.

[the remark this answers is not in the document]
alexzherdev: Some of them are not using `react-hot-loader` at all...

Sounds good. Looks like refactoring this will just be a semver-patch.

packages/CollapsibleChecklists/src/components/Group/Group.js
```diff
@@ -0,0 +1,120 @@
...
+
+  const label = (
+    <React.Fragment>
+      <input
+        ref={checkboxRef}
+        checked={allAreChecked}
+        type="checkbox"
+        disabled={isDisabled}
+        onChange={toggleChildren}
+        aria-label={textContent(title)}
```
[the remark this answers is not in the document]
nahumzs: I suggest to test if the value pass is of type string we reuse the value and assign it to aria-label properly, if we don't detect it we just asked the developer to provide an a11yText, most of the cases the user I think will pass a regular string option.

packages/CollapsibleChecklists/src/components/Group/Group.js
```diff
@@ -0,0 +1,120 @@
...
+
+  if (allAreChecked || noneAreChecked) {
+    checkboxRef.current.indeterminate = false;
+  } else if (!allAreChecked && !noneAreChecked) {
+    checkboxRef.current.indeterminate = true;
+  }
+
+  function toggleChildren() {
+    const childItems =
+      children.length > 0 ? children.filter(childItem => childItem.type.displayName === Item.displayName) : [];
```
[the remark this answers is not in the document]
jamiek-acl: Oh man, this one stumped me for a few hours.

Note that using `React.Children.toArray` (and `React.Children.map`) **will change the key of the child** (prepends with `$.`). Can get around this by using `React.Children.forEach` if you want to maintain keys.

packages/CollapsibleChecklists/stories/examples/Basic.js
tristanjasper: 🤷‍♂

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
tristanjasper: Wondering if perhaps we should try to not call our own components the same as paprika component names in general to avoid that confusion

packages/CollapsibleChecklists/src/components/Group/Group.js
```diff
@@ -0,0 +1,120 @@
...
+  const [, setIsIndeterminate] = React.useState(null);
+  React.useEffect(() => {
+    setIsIndeterminate(checkboxRef.current.indeterminate);
+  }, [checkboxRef.current.indeterminate]);
+}
+
+function Group(props) {
+  const { children, isDisabled, onChange, onExpand, title } = props;
+  const [isCollapsed, setIsCollapsed] = React.useState(true);
+  const checkboxRef = React.useRef({});
```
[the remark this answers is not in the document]
jamiek-acl: @alexzherdev this does not work as I am calling `checkboxRef.current.indeterminate`

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
jamiek-acl: I can just wrap the whole thing in a label:
```
<label>
   <input type="checkbox" />
   {title}
</label>
```

packages/CollapsibleChecklists/stories/examples/Basic.js
```diff
@@ -0,0 +1,138 @@
...
+      return <Spinner />;
+    }
+
+    return renderTeams(sportsData[1].sports[0].teams);
+  }
+
+  return (
+    <CollapsibleChecklists onChange={handleOnChange}>
+      <CollapsibleChecklists.Heading>California Sports Teams</CollapsibleChecklists.Heading>
+      <CollapsibleChecklists.Group title={sportsData[0].sports[0].title}>
```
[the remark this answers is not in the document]
jamiek-acl: I'm trying to keep it as simple as possible; I want to demonstrate to future onlookers that the API looks like this:

```
    <CollapsibleChecklists onChange={() => {}}>
      <CollapsibleChecklists.Heading>Flowers</CollapsibleChecklists.Heading>
      <CollapsibleChecklists.Group title="Roses">
        <CollapsibleChecklists.Item isChecked>Damask</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item isChecked>Eden</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item isChecked>Lady Banks</CollapsibleChecklists.Item>
      </CollapsibleChecklists.Group>

      <CollapsibleChecklists.Group title="Irises">
        <CollapsibleChecklists.Item isChecked>Siberian</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item>Yellow</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item isDisabled>Bearded</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item>Netted</CollapsibleChecklists.Item>
      </CollapsibleChecklists.Group>

      <CollapsibleChecklists.Group title="Lilies" isDisabled>
        <CollapsibleChecklists.Item>Easter</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item>Madonna</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item>Stargazer</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item>Tiger</CollapsibleChecklists.Item>
        <CollapsibleChecklists.Item>Turks Cap</CollapsibleChecklists.Item>
      </CollapsibleChecklists.Group>
    </CollapsibleChecklists>
```

not have them spend time deciphering what all the loops are doing.

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
alexzherdev: Oh thanks for the reminder. Found a [nice explanation](https://github.com/facebook/react/issues/5541#issuecomment-159425628) to this behaviour btw. I think it shouldn't matter for us here?

packages/CollapsibleChecklists/src/components/Group/Group.js
jamiek-acl: The consumer can compare on anything... I chose to compare by `key` (as I set key={team.name} but could have just as easily called it `foo={team.name}`. In fact i'll update the example to use `foobar`, and the comparison checks `props.foobar`

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
mikrotron: Yeah, that's a good solution... except I'm still concerned about the UX in this specific situation, and think it's worth discussing with a designer.

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
jamiek-acl: It did since my example code was comparing based on `key` to see if that is an item that changed.

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
jamiek-acl: Sure. but if it changes that can be later, i want to merge this fella :)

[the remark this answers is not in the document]
nahumzs: yes we can do it in another version not need it for this.

packages/CollapsibleChecklists/src/components/Group/Group.js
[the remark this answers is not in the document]
nahumzs: Agree with all the discussion here, but I think this can be addressed in another version, this is not the part of the API so we could refactor it easily in a bump it.

Can we create an issue with all the followed steps from this review @jamiek-acl and include this one?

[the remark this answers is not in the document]
krismckinnonacl: I would make clicking the label activate that checkbox. For the following reasons:
1. Assuming that the child entries underneath work normally, if we did make that change, we'd be changing the behaviour of an element that looks visually the same, in close proximity to other controls that do the same thing, in a subtle way - I assume that clicking on the titles of the items underneath would check the associated checkbox, right?
2. If the user does in fact assume that clicking the title expands or collapses the section but instead selects all the elements, it's an easy mistake to correct, because they will encounter it BEFORE they select items underneath. It would be atypical for someone to 
1. click the caret, then 
2. select some things, then 
3. click the caret to collapse again then 
4. come back later and click the title instead of where they clicked before to open the elements.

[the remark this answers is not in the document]
jamiek-acl: 1. Yes checking the titles of the children toggles them, too.
2. No it's not easily undone. If some children are selected and i click the parent, that would select all of its children.  If you didn't intend that and wanted to undo, you wouldn't know which ones to de-select.

packages/CollapsibleChecklists/src/components/Group/Group.js
alexzherdev: I see, I'm -1 on this.

alexzherdev:
@nahumzs this was merged too soon IMO. The reason I was asking for a real-world story is that I believe the current API will not work. Here's roughly how you'll have to use it in a loop:
```jsx
<CollapsibleChecklists onChange={handleOnChange}>
  {sections.map(({ heading, items }) => (
    <React.Fragment key={heading}>
      <CollapsibleChecklists.Heading>{heading}</CollapsibleChecklists.Heading>
      <CollapsibleChecklists.Group>...</CollapsibleChecklists.Group>
    </React.Fragment>
  )}
</CollapsibleChecklists>
```
This doesn't work because we are `forEach`ing through `<CollapsibleChecklists>`' children to find `<Group>`s to clone and we won't find any here.

Additionally, since we're fiddling with keys, we'd want a story where a consumer has their own keys on components to be sure we haven't broken anything for that case.

jamiek-acl:
@alexzherdev why is the story that is included not a real-world story?

alexzherdev:
Probably not worded well enough—I think subjectively components like this are more often data-driven than not? Meaning you'll have a collection of state to loop over (perhaps coming from an API), rather than hard code individual data pieces in. Even if we assume it's 50/50, still we want the component to work in those other 50% of cases.

jamiek-acl:
The existing story simulates:

- grabbing some of the data from an API (the Ohio Football group)
- looping over that data and maintaining state of selected items/groups of items
- disabled items
- disabled groups
- headings
- items that are simple strings
- items that are nodes

Check this video of the Storybook story that demonstrates those things, plus keyboard a11y:
use2

alexzherdev:
Ah, I didn't mean the actual end customer usage in the browser—that is perfectly covered in the story like that video demonstrates. I meant the usage of the component by developers consuming it in application code. The story _code_ doesn't reflect the case when headings+groups are in an array (grabbed from an API) that has to be looped over, like in my snippet above. And if it did, you'd find that it doesn't work (unless I'm totally missing something).

packages/CollapsibleChecklists/src/CollapsibleChecklists.js
```diff
@@ -0,0 +1,39 @@
...
+const defaultProps = {};
+
+const CollapsibleChecklists = props => {
+  const { children, onChange } = props;
+  const modifiedChildren = [];
+
+  React.Children.forEach(children, (child, index) => {
+    switch (child.type.displayName) {
+      case Group.displayName:
+        modifiedChildren.push(React.cloneElement(child, { key: `group${index}`, onChange })); // eslint-disable-line react/no-array-index-key
```
alexzherdev: That ESLint rule speaks the truth. Even if you're not explicitly reordering items a la `<Sortable>`, when an item gets removed from the middle of the list things will break. https://reactjs.org/docs/lists-and-keys.html#keys

jamiek-acl:
@alexzherdev I've added another story to demonstrate grabbing headings and groups from an API, that Nahum created the other day: https://github.com/acl-services/paprika/pull/134

PR #124, opened 2019-07-17
UX-502 Form element

allison-c:
### 🛠 Purpose

---

### ✏️ Notes to Reviewer

- [x] Adding tests
- [x] Talk to designers for the size of the info button
- [ ] Fix date picker popover alignment issue

### Installation

`> npm install --save @paprika/form-element`
or
`> yarn add @paprika/form-element`

### Usage

Please use `<L10n />` component to wrap `<FormElement />` or you application.

Using with paprika `<Input />`

```js
import FormElement from "@paprika/form-element";

<FormElement label="Name">
  <Input onChange={changeHandler} />
</FormElement>;
```

Using with paprika `<DatePicker />`

```js
import DatePicker from "@paprika/date-picker";
import FormElement from "@paprika/form-element";

<FormElement label="Name">
  <DatePicker onChange={changeHandler} />
</FormElement>;
```

Using with DOM input

```js
import FormElement from "@paprika/form-element";

<FormElement label="Name">
  <input onChange={changeHandler} />
</FormElement>;
```

Using with your component, you'll get `aria-describedby`, `hasError`, `isDisabled`, `isReadOnly` and `size` as additional props.

```js
import FormElement from "@paprika/form-element";
import MyInput from "./MyInput";

<FormElement label="Name">
  <MyInput />
</FormElement>;
```

#### FormElement

- `children` (required)
- `id`
- `isDisabled`
- `isInline`
- `isLabelVisuallyHidden`
- `isOptional`
- `isReadOnly`
- `isRequired`
- `label` (required)
- `size`

#### FormElement.Description

- `children` (required)
- `ariaDescriptionId`

#### FormElement.Error

- `children`

#### FormElement.Hint

- `children` (required)
- `triggerA11yText`

[More detail about these props](https://github.com/acl-services/paprika/blob/master/packages/FormElement/src/FormElement.js)

### 🖥 Screenshots

Showcase: 
image

Inline form element:
image


packages/DatePicker/src/DatePicker.js
```diff
@@ -29,6 +29,9 @@ const propTypes = {
   /** Date format used while displaying date. It should be human-friendly and spelled out, default is MMMM DD,YYYY */
   humanFormat: PropTypes.string,
 
+  /** Id of the input, you might don't need this if you wrap this component into <FormElement> */
```
mikrotron:
```suggestion
/** ID for the <input>.  Not required when wrapped with <FormElement>. */
```

.storybook/reset.scss
```diff
@@ -7,5 +7,5 @@ body,
 body {
   margin: 0;
   padding: 0;
-  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-family: "Lato", "Helvetica Neue", Helvetica, Arial, sans-serif;
```
mikrotron: This isn't going to work reliably without a link to the Google font somewhere. 
Maybe we should just go straight to the `IBM Plex Sans` though?

packages/FormElement/src/FormElement.js
```diff
@@ -0,0 +1,169 @@
...
+  isLabelVisuallyHidden: PropTypes.bool,
+
+  /** Should show is optional text besides the label or not. */
+  isOptional: PropTypes.bool,
+
+  /** Should be read-only or not, default is false. */
+  isReadOnly: PropTypes.bool,
+
+  /** Should show is required text besides the label or not. */
+  isRequired: PropTypes.bool,
```
nahumzs: To me this feels or can make the impression that the `<FormElement />` will be responsible for validating the field which is not the case. 

I might suggest the use of another prop like kind="optional" || kind="required" instead of isOptional or isRequired

packages/FormElement/src/FormElement.js
```diff
@@ -0,0 +1,169 @@
...
+        isInline={isInline}
+        isVisuallyHidden={isLabelVisuallyHidden}
+        isOptional={isOptional}
+        isRequired={isRequired}
+        label={label}
+      />
+
+      {extratedChildren.children.map(child => {
+        if (React.isValidElement(child)) {
+          const extendedProps = isString(child.type)
```
nahumzs: While this might work, for now, I fear that this might failed in the feature seem more like abusing react api than a proper way though I don't have a solution on how to solved differently. No action just pointing a possible scenario.

nahumzs:
What about changing `<FormElement.Hint />` to `<FormElement.Info />`

packages/FormElement/src/FormElement.js
allison-c: Decided use `hasOptionalLabel` `hasRequiredLabel` instead.

packages/FormElement/src/FormElement.js
allison-c: > What about changing `<FormElement.Hint />` to `<FormElement.Info />`

Decided use ``<FormElement.Help />``

packages/FormElement/src/FormElement.js
```diff
@@ -0,0 +1,164 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import isNil from "lodash/isNil";
+import isString from "lodash/isString";
+import uuid from "uuid/v1";
```
nahumzs: mmmm minor but I think we should use v4 for our case v1 is too much.

packages/FormElement/src/FormElement.js
```diff
@@ -0,0 +1,164 @@
...
+  } = props;
+
+  const extratedChildren = extractChildren(children, [
+    "FormElement.Description",
+    "FormElement.Error",
+    "FormElement.Help",
+  ]);
+  const ariaDescriptionId = uuid();
+  const hasError = !!extratedChildren["FormElement.Error"] && !!extratedChildren["FormElement.Error"].props.children;
+  const uniqueId = isNil(id) || id === "" ? uuid() : id;
```
nahumzs: This will created a new uuid each render.

`const uuid = React.useRef(uuidV4());`

would work better.

PR #125, opened 2019-07-17
PD-12544 Introduce ResponsesAccordion component 🐐

mikrotron:
### 🛠 Purpose
Introduce a new component for displaying questionnaire responses in a collapsible accordion format.

### ✏️ Notes to Reviewer
- **Patch** update to the `<Collapsible>` – add a `data-pka-anchor` and refactor styling.
- **Patch** update to the `<Icon>` – Add the `svg` source and generated component for `Arrow-Up` icon.
- **Patch** update to the `<L10n>` – Add two keys for this component.
- **Patch** update to `Stylers` – Change `inlineBlockStyles` include (which affects the `<Popover>`).
- **Patch** update to `<Button>` – Fix a line-height bug.
- **Patch** update to `<RawButton>` styles remove use of `inlineBlockStyles`.

### 🖥 Screenshot
Screen Shot 2019-07-17 at 3 31 34 PM

### 🛠 To Do (in a follow up PR)
- Document props inline.
- Add less "basic" Storybook stories.
- Testing (likely just Screener).


packages/ResponsesAccordion/src/components/Responses/Responses.styles.js
```diff
@@ -0,0 +1,35 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+const responsesStyles = css`
+  ${stylers.lineHeight()}
+
+  color: ${tokens.textColor.subtle};
+  margin: ${tokens.space} 0;
+  max-height: 420px;
```
alexlaj: Do we want a larger max height since this is going inside a scrolling div and can also be collapsed? Double scrollbar in a sidepanel might not be the greatest. We won't be able to avoid it in all cases but maybe it should be larger to allow more content to be shown before the second scrollbar appears?

mikrotron: That's a good question.... for @krismckinnonacl!
Luckily he already gave me his answer.  He was happy with this height.

alexzherdev: Nice trolling tagging the person that doesn't need to be tagged 😄

mikrotron: Ha. He's told me he doesn't get the notifications anyway. And if he does, :trollface:

packages/ResponsesAccordion/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/ResponsesAccordion"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/collapsible": "^0.1.1",
+    "@paprika/icon": "^0.1.3",
+    "@paprika/l10n": "^0.1.2",
```
mikrotron: I actually depend on a future version of the L10n for the key I added. It's a patch update, so it shouldn't really matter, but will lerna also update this dependency here?

packages/ResponsesAccordion/src/components/Indicator/Indicator.js
```diff
@@ -0,0 +1,34 @@
...
+  isComplete: false,
+  isLast: false,
+};
+
+const Indicator = props => {
+  const { isComplete, isActive } = props;
+
+  return (
+    <div data-pka-anchor="indicator" css={indicatorStyles} {...props}>
+      <span data-pka-anchor="indicator.dot" css={indicatorDotStyles} {...props}>
```
alexzherdev: Seems unlikely we really need to spread all props onto two different elements

packages/ResponsesAccordion/src/ResponsesAccordion.js
```diff
@@ -0,0 +1,82 @@
...
+            isComplete,
+            isActive: activeIndex === index,
+            isLast: index === validChildren.length - 1,
+          };
+
+          return (
+            <div css={itemStyles} key={label}>
+              <Indicator {...indicatorProps} />
+              <Collapsible label={getLabel(label, index)} isComplete={isComplete} {...moreChildProps}>
+                {child}
```
alexzherdev: `child` is `<Item>` which currently just renders its children. Could the surrounding `<Collapsible>` be moved into `<Item>` since that's what it really is?

packages/ResponsesAccordion/src/components/Responses/Responses.js
```diff
@@ -0,0 +1,49 @@
...
+
+const defaultProps = {
+  responses: [],
+};
+
+const Response = ({ heading, body }) => {
+  const I18n = useI18n();
+
+  return (
+    <>
```
alexzherdev: I _think_ you didn't like this syntax before?
(also note it can't be used 100% consistently since you need `<React.Fragment>` if you need a `key`)

packages/ResponsesAccordion/src/components/Responses/Responses.js
```diff
@@ -0,0 +1,49 @@
...
+Response.propTypes = responseShape;
+Response.defaultProps = {
+  body: null,
+};
+
+const Responses = ({ responses }) => {
+  return (
+    <dl css={responsesStyles}>
+      {responses.map(({ heading, body }) => (
+        <Response heading={heading} body={body} key={heading} />
```
alexzherdev: Another case of a node being used for key

packages/Icon/src/svg/Arrow-Up.svg
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
mikrotron: No, but it's okay, `svgr` will optimize it when it generates the component.

packages/ResponsesAccordion/src/ResponsesAccordion.js
```diff
@@ -0,0 +1,82 @@
...
+          const { label, ...moreChildProps } = child.props;
+          const isComplete = activeIndex > index;
+          const indicatorProps = {
+            isComplete,
+            isActive: activeIndex === index,
+            isLast: index === validChildren.length - 1,
+          };
+
+          return (
+            <div css={itemStyles} key={label}>
```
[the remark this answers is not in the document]
mikrotron: I think if we make this a more generic component, we can't really rely on the `label` being unique anyway... so I think we need to introduce a required `id` prop... unless you have an alternative suggestion?

packages/ResponsesAccordion/src/components/Indicator/Indicator.js
mikrotron: I need 3 of 3 for `indicatorStyles` and 2 of 3 for `indicatorDotStyles`.  Seems simplest to just pass them all through, though maybe it would read better to declare explicit `xStylingProps` for each of them...?

packages/ResponsesAccordion/src/ResponsesAccordion.js
mikrotron: 🤯 I will give this a try. Don't see any reason at the moment why it wouldn't work.

packages/ResponsesAccordion/src/components/Responses/Responses.js
mikrotron: I think it was Nahum who didn't like it.  I like it though, I keep misspelling fragment with a lowercase f. 
I'm also not that concerned you need to use the long form with a `key` 🤷‍♂️

packages/ResponsesAccordion/src/components/Responses/Responses.js
mikrotron: Yeah, and probably shouldn't rely on it's uniqueness either, *sigh*.  These are static, so I feel like I should just be able to use the `index`, but I'm afraid of the eslint / zdev cops 👮‍♂️

packages/ResponsesAccordion/src/components/Responses/Responses.js
nahumzs: I would prefer to use `<React.Fragment />` is clearly to the developer that is a component and they can search on google about it and find an answer quicker, and even if we use this syntax we can't avoid use `<React.Fragment />` when keys are involved. 

My `giveAFu` is not really high on this, but I don't see a need to introduce a new "code format" in our code that's all.

packages/ResponsesAccordion/src/components/Responses/Responses.js
nahumzs: you can use an index as long the component is quite static (will not rerender), and seem to me this component is unlikely to rerender the headings or even the component, seems more a presentational component than an input component. So should be fine to use indexes.

packages/Collapsible/src/Collapsible.styles.js
```diff
@@ -2,39 +2,39 @@ import tokens from "@paprika/tokens";
...
-
-  .collapsible__body {
-    display: none;
-    width: 100%;
-    ${props.isCollapsed ? "display: none;" : "display: block;"}
-  }
-
   .collapsible__label {
-    display: inline-block;
+    ${stylers.inlineBlockStyle}
```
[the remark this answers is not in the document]
nahumzs: mmm, me not like it :/,  is obscure, and make me think what the rule is doing.

Maybe `${stylers.display.inlineBlock}` is more descriptive? inlineBlockStyle to me sound like you will get 
`display: inline-block` plus other css that will make it *style*. 

I don't see the real benefit, to me is better just to write *display: inline-block*  everyone understands it.

packages/ProgressAccordion/README.md
```diff
@@ -0,0 +1,53 @@
...
+  {
+    id: "1",
+    heading: "Who is on first?",
+    body: <p>Yes.</p>,
+  },
+];
+
+<ProgressAccordion activeIndex={1} activeStatus="5 days idle" a11yText="Responses">
+  <ProgressAccordion.Item id="0" label="Zero">
+    <ProgressAccordion.Responses responses={responses} />
```
nahumzs: We haven't used an object into our component, this should be a component itself 

`<Response id="" heading="">[node]<Response>`

Then you could abstract the content in an array.

packages/ResponsesAccordion/src/components/Responses/Responses.js
mikrotron: I'm going to leave it, but will follow up on this topic with the team later (probably in a ShopTalk).

packages/Collapsible/src/Collapsible.styles.js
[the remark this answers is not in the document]
mikrotron: Yeah, this was a silly pattern.  I will fix it here, but should remove this and `stylers.boxSizing` everywhere later.

packages/ProgressAccordion/src/ProgressAccordion.js
```diff
@@ -0,0 +1,86 @@
...
+          const isComplete = activeIndex > index;
+          const indicatorProps = {
+            isComplete,
+            isActive: activeIndex === index,
+            isLast: index === validChildren.length - 1,
+          };
+
+          return (
+            // eslint-disable-next-line react/no-array-index-key
+            <div css={itemStyles} key={index} role="listitem">
```
mikrotron: @nahumzs said it's okay to break the rules! 😁 
This component is unlikely to re-render these items, and even if it did occasionally, the impact on performance is not a significant concern. Simplifying the API for the consumer is preferred.

alexzherdev: It's not so much about performance as it is about correctness though? https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318
Assuming we're following the robustness principle, we should be able to re-render the component with a different set of items without breaking 🤷‍♂

mikrotron: > It is true that sometimes generating new ids is redundant and may be avoided. To help you decide, I put together three conditions which these examples have in common:

> - the list and items are static–they are not computed and do not change;
> - the items in the list have no ids;
> - the list is never reordered or filtered.

> When all of them are met, **you may safely use the index as a key**.

So I think since we would never push an item into the list, remove one from the middle, or re-order the list it should be safe?  In fact, the list won't ever change _at all_ in either of the use cases we have lined up for this component.

alexzherdev: Ok, as long as we understand why it is safe to do.
I can still imagine cases where this can be necessary (e.g., you pick something in step 1 that makes step 3 irrelevant and we want to remove it). But let's invoke YAGNI for now.

While not directly related, this reminds me of how people sometimes assume that a certain prop will never change its value during a component's lifecycle, when in reality the component model implies you should be able to accept a new prop value after the initial mount (e.g. https://overreacted.io/writing-resilient-components/#dont-stop-the-data-flow-in-rendering)

packages/ResponsesAccordion/src/components/Responses/Responses.js
alexzherdev: @nahumzs see https://github.com/acl-services/paprika/pull/125#discussion_r305509234
(It will definitely re-render btw; we're not doing anything to prevent that)

PR #134, opened 2019-07-19
PD-12487 Add a basic example story

jamiek-acl:
### 🛠 Purpose

Ended up doing a small refactor as if the children were `React.Fragment`s (like would be the case looping over a couple headings), the onChange wouldnt work (since I was adding onChange to the children of the main component).

### 🖥 Screenshots

Screen Shot 2019-07-19 at 5 03 25 PM


packages/CollapsibleChecklists/stories/examples/Basic.js
```diff
@@ -1,150 +1,56 @@
...
-
-      <CollapsibleChecklists.Heading>Ohio Sports Teams</CollapsibleChecklists.Heading>
-      <CollapsibleChecklists.Group title={ohioFootball.title} onExpand={handleExpand}>
-        {renderOhioFootballTeams()}
-      </CollapsibleChecklists.Group>
-
-      <CollapsibleChecklists.Heading>Saskatchewan Sports Teams</CollapsibleChecklists.Heading>
-      <div style={{ fontStyle: "italic", padding: "4px" }}>None available (this shows that any child is accepted)</div>
+    <CollapsibleChecklists onChange={handleChange}>
+      <CollapsibleChecklists.Heading>Movies</CollapsibleChecklists.Heading>
```
alexzherdev: @nahumzs Compare this with my snippet: https://github.com/acl-services/paprika/pull/123#issuecomment-512918934
So for example, in addition to "Movies" there would be another "TV Shows" section, "Podcasts" etc. that are in an array. Is that too far-fetched of an example? It's hard to tell now (doesn't look like this will be needed for the current feature). @jamiek-acl already has a similar usage in the advanced story with those sections hard-coded. But as soon as we have it in an array, the current API won't work so we'll need to redesign it and force a breaking change. This is what I'm trying to avoid. Let's chat offline if there is any other misunderstanding

alexzherdev: Here's another example that doesn't require an outer loop:
```jsx
const Frameworks = () => (
  <React.Fragment>
    <CollapsibleChecklists.Heading />
    {groups} {/* this can be whatever */}
  </React.Fragment>
);

const Projects = () => (
  <React.Fragment>
    {/* similar as above */}
  </React.Fragment>
);

<CollapsibleChecklists>
  <Frameworks />
  <Projects />
</CollapsibleChecklists>
```

This is a totally reasonable way to organize code but will not work either.

PR #136, opened 2019-07-19
AN-7723 Fix export issues

allison-c:
### 🛠 Purpose

Fix export issue on `Select`, `Switch` and `Textarea`

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

### DIFF
--- a/packages/Select/package.json
+++ b/packages/Select/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.4",
   "description": "Select component",
   "author": "@paprika",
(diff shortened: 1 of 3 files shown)


nahumzs:
@alexzherdev do you know if there any eslint rule that looks into package.json and checked if we have the *main* property?

alexzherdev:
@nahumzs that's not really ESLint's use case, although I suppose you could use it to parse and process JSON.

PR #138, opened 2019-07-22
AN-7742 Extend paprika L10n component so can provide your own transla…

jamiek-acl:
…tions

### 🛠 Purpose

---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

### DIFF
--- a/packages/L10n/src/L10n.js
+++ b/packages/L10n/src/L10n.js
@@ -5,17 +5,17 @@ import L10nContext from "./L10nContext";
 
 const propTypes = {
   locale: PropTypes.string,
-  Locales: PropTypes.node,
+  locales: PropTypes.object /* eslint-disable-line react/forbid-prop-types */,
   children: PropTypes.node.isRequired,
 };
 
 const defaultProps = {
   locale: "en",
-  Locales: null,
+  locales: null,
 };
 
 export default function L10n(props) {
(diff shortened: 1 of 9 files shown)


packages/L10n/src/L10n.js
```diff
@@ -5,15 +5,17 @@ import L10nContext from "./L10nContext";
 
 const propTypes = {
   locale: PropTypes.string,
+  Locales: PropTypes.node,
```
alexzherdev: This is an object though?

alexzherdev: (also probably lowercase `locales` like other props)

packages/L10n/src/L10n.js
```diff
@@ -5,15 +5,17 @@ import L10nContext from "./L10nContext";
...
 };
 
 const defaultProps = {
   locale: "en",
+  Locales: null,
 };
 
 export default function L10n(props) {
-  const value = React.useMemo(() => getI18nObject(props.locale), [props.locale]);
+  const value = React.useMemo(() => getI18nObject(props.locale, props.Locales), [props.locale]);
```
alexzherdev: While unlikely to cause issues, better to add the new prop to the dependencies of `useMemo` as well

Issue #126, opened 2019-07-17
Popover – RawButton wraps Trigger when not needed


mikrotron:
# Bug Report

## Expected behavior
When a `<Button>` is used with the `<Popover.Trigger>`, no `<RawButton>` should wrap it.  However, sometimes if an `<Icon>` or other non-interactive element is used for the trigger, it may still be convenient. 

## Current behavior
When the `<Popover.Trigger>` is given a `node` as `children`, it will always wrap it with a `<RawButton>`.  It's necessary to provide a `function` as `children` to avoid this behaviour.

## Examples
`<DropdownMenu>`
https://github.com/acl-services/paprika/pull/113/files#diff-edd13737a415dcf236bf45816da474cbR103

Results
https://github.com/acl-services/acl-exception/pull/6429/files#r301834271

nahumzs:
possible solution.

in this line https://github.com/acl-services/paprika/blob/master/packages/Popover/src/components/Trigger/Trigger.js#L63

we need to check if children are a function:

```js

const propsForTrigger = {
   a11yText,
   "data-qa-anchor": "popover-trigger",
   onClick={handleTriggerEvent},
   onBlur={shouldKeepFocus ? handleTriggerEvent : null}
}

if (typeof children === "function") {
  return children(propsForTrigger)
}

return  return (
    <RawButton {...propsForTrigger}>{children}</RawButton>
  );
```

mikrotron:
@nahumzs, this is kind of how it's currently implemented.  But upon reflection, the solution might be to just _not_ use a `<Trigger>` when you're using as a controlled component, like:

https://github.com/acl-services/paprika/blob/master/packages/Popover/stories/examples/Controlled.js 

I'd like to test the a11y of that a bit more before I close this issue, but for now, it might not be necessary to "fix" this.

PR #145, opened 2019-07-24
Dropdown tweaks, Tab focus, Buttons as spans, create Cypress test

tristanjasper:
### 🛠 Purpose

Ensure tabbing into trigger button only takes one tab action
Convert failing spec tests (because of above change) to cypress test
Ensure confirmation panel buttons render as <spans> so that consuming application styles doesn't override them.

### ✏️ Notes to Reviewer

This is a minor bugfix on the component - Please bump to `0.1.2`

### 🖥 Screenshots

Before
doubleTabDropdown
After
singleTabDropdown

### DIFF
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -58,10 +58,11 @@ const DropdownMenu = props => {
 
   const renderTrigger = () => {
     const triggerComponent = props.renderTrigger(getTriggerStateAndHelpers);
-    return React.cloneElement(triggerComponent, {
-      triggerRef,
-      menuRefId: menuRefId.current,
-    });
+    return () =>
+      React.cloneElement(triggerComponent, {
+        triggerRef,
+        menuRefId: menuRefId.current,
+      });
   };
 
   const renderContent = () => {
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -44,7 +44,7 @@ const DropdownMenu = props => {
     setIsOpen(true);
   };
 
-  const getTriggerStateAndHelpers = {
+  const triggerProps = {
     isOpen,
     handleOpenMenu,
   };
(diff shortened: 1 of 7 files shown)


packages/DropdownMenu/test/dropdownMenu.cypress.js
```diff
@@ -0,0 +1,26 @@
...
+    cy.getByTestId("popover-content").should("be.visible");
+    cy.wait(100).then(() => {
+      cy.getByTestId("dropdown-menu__trigger").click();
+      cy.getByTestId("popover-content").should("not.be.visible");
+    });
+  });
+
+  it("should retain trigger focus when the dropdown is closed", () => {
+    cy.getByTestId("dropdown-menu__trigger").click();
+    cy.wait(100).then(() => {
```
nahumzs: I think there is an alternative in cypress to do this and avoid wait.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -58,10 +58,11 @@ const DropdownMenu = props => {
 
   const renderTrigger = () => {
     const triggerComponent = props.renderTrigger(getTriggerStateAndHelpers);
-    return React.cloneElement(triggerComponent, {
-      triggerRef,
-      menuRefId: menuRefId.current,
-    });
+    return () =>
```
nahumzs: let's add the issue number for this approach so we know what's happening. 

also might be easier to pass down the props to the consumer so we don't have to clone what they are passing? But that's for other time discussion.

tristanjasper: https://github.com/acl-services/paprika/issues/126

packages/DropdownMenu/test/dropdownMenu.cypress.js
tristanjasper: I have tried using `cy.clock()` with `cy.tick(100)` but currently it isn't working the same way as  `wait` currently, When we have addressed https://github.com/acl-services/paprika/issues/126 we may see a difference in its behaviour.

PR #146, opened 2019-07-24
Add ellipsis icon

tristanjasper:
### 🛠 Purpose

Add Ellipsis icon. Required in Robots for overflow menu. 

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

image


packages/Icon/src/Ellipsis.js
```diff
@@ -0,0 +1,29 @@
...
+    css={`
+      color: ${props.color};
+      width: ${props.size};
+      height: ${props.size};
+      vertical-align: text-top;
+    `}
+    focusable={false}
+    {...props}
+  >
+    <title>Ellipsis</title>
```
jamiek-acl: The other icons have `<title>{title}</title>`. Not sure how important that is

tristanjasper: Yup me neither but I wasn't about to break convention...

jamiek-acl: But you did break convention. The other icons have <title>{title}</title>.

packages/Icon/stories/examples/Test.js
```diff
@@ -44,6 +45,7 @@ const ExampleStory = () => (
         }}
       />
       , photo booth small batch poke. <ArrowIcon />
+      <EllipsisIcon />
```
jamiek-acl: Can you please boy scout this story.
Please dont name it Test.js
Maybe make a table with each icon on a line, the name in the first cell, the icon in the second.

packages/Icon/src/Ellipsis.js
tristanjasper: i see, ok replaced with `title`

packages/Icon/stories/examples/Test.js
tristanjasper: I agree this story file does have an odd approach to demonstrate the icons, I'm not aware why it was decided to build it out like this though. Perhaps @mikrotron knows more. I would like to come back to refactor this file/ icons after some discussion so perhaps I'll address in a follow up PR

mikrotron: Yeah, I made this story as a quick and dirty way to verify the component was working. But deliberately deferred on making a "proper" story until I figure out a consistent approach for our Paprika Storybook (which I did during the HackFest).  I think we need a story that automatically displays all of the icons, rather than something we have to manually update, but that's beyond the scope of this PR.

tristanjasper: Ok, lets add this component now, and we can address the icons story in follow up issue.

PR #147, opened 2019-07-24
PD-12487 Expand children when select group

jamiek-acl:
### 🛠 Purpose

As per AC, when click on a Group's checkbox, it should expand to reveal the children.
It will also select the children (but it already does that).

### ✏️ Notes to Reviewer
After expanding it will also fire`props.onExpand()`.

Note that if `onExpand` makes an API call to update the Group's children, the children will NOT get selected. The consumer will need to click on the checkbox again after the API has completed to select them.

### 🖥 Screenshots
This illustrates what happens when 

clicky

### DIFF
--- a/packages/CollapsibleChecklists/src/components/Group/Group.js
+++ b/packages/CollapsibleChecklists/src/components/Group/Group.js
@@ -15,7 +15,7 @@ const propTypes = {
 const defaultProps = {
   children: [],
   isDisabled: false,
-  onExpand: () => {},
+  onExpand: null,
 };
 
 function useIsIndeterminate(checkboxRef) {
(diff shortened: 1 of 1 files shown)


alexzherdev:
I'm 👍 except it seems a bit odd when you click the parent checkbox and it ultimately doesn't get checked. But up to designers

jamiek-acl:
> I'm 👍 except it seems a bit odd when you click the parent checkbox and it ultimately doesn't get checked. But up to designers

I agree, it is a bit odd.  But if the children are unknown (i.e. still to be retrieved from an API), we can't really check them.  Nahum and I tried and couldn't find a nice solution.

nahumzs:
eventually, we will need to pass on the onExpand what's the state of isCollapsed I can imagine a couple cases where the consumer might need to know that, but no relevant for now.

krismckinnonacl:
Took a look at this - it's not an ideal UX and I could see it tripping up some less technical users, but it is acceptable for now if there's not an easy way to improve it.

PR #148, opened 2019-07-25
PD-12723 propagate zindex to tip

oscarkwan:
### 🛠 Purpose

- we want to propogate the zindex to also the tip, so when someone passes in
`<Popover.Help zIndex={100} />`
- it will work properly for both the tip and the popover content
---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

### DIFF
--- a/packages/FormElement/src/components/Help/Help.js
+++ b/packages/FormElement/src/components/Help/Help.js
@@ -28,7 +28,7 @@ function Help(props) {
         <InfoCircleIcon css={iconStyles} aria-hidden type="exclamation-circle" />
       </StyledTrigger>
       <Popover.Content>
-        <Popover.Tip />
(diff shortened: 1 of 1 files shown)


packages/FormElement/src/components/Help/Help.js
```diff
@@ -28,7 +28,7 @@ function Help(props) {
         <InfoCircleIcon css={iconStyles} aria-hidden type="exclamation-circle" />
       </StyledTrigger>
       <Popover.Content>
-        <Popover.Tip />
+        <Popover.Tip zIndex={moreProps.zIndex ? moreProps.zIndex + 1 : undefined} />
```
oscarkwan: + 1, because tip needs to be 1 index higher than popover or it shows a border

PR #152, opened 2019-07-30
Pd 12487 more props

jamiek-acl:
### 🛠 Purpose

Add a new prop so a group can be checked or indeterminate even if we dont know what its children are yet.

### 🖥 Screenshots

Notice how the group checkboxes are already checked, or indeterminate even though we dont know the children yet.  This is possible because of the prop I added in this PR.

checkboxes

 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/pd-12487-more-props' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/pd-12487-more-props</a>


packages/CollapsibleChecklists/src/components/Group/Group.js
```diff
@@ -25,29 +29,59 @@ function useIsIndeterminate(checkboxRef) {
...
 
-function Group(props) {
-  const { children, isDisabled, onExpand, title } = props;
-  const onChange = React.useContext(CollapsibleChecklistsContext);
-  const [isCollapsed, setIsCollapsed] = React.useState(true);
-  const checkboxRef = React.useRef({});
-  useIsIndeterminate(checkboxRef);
-  let allAreChecked = React.Children.count(props.children) > 0;
-  let noneAreChecked = true;
+function setIsIndeterminateAndGetAllChildItemsAreChecked(
```
nahumzs: there are two function here `setIsIndeterminate` And `GetAllChildItemsAreChecked` let's split them to make it clearer what is happening.
```js
function setIsInderterminate(...params) {

}

function areAllChildItemsAreChecked(...params) {
   if(allChildItemsAreChecked) {
     setIsInderterminate(ref, false)
     return;
   }

   setIsInderterminate(ref, true);
}
```

PR #153, opened 2019-07-31
AN-7724 Add operations prop in Collapsible

allison-c:
### 🛠 Purpose

Add `operations` prop in `<Collapsible />` for some cases like this :
image
Confirmed with @johnpeele this will be a common case

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

image


allison-c:
Won't do it.

nahumzs:
this is no longer need it, we are going with the standard collapsible.

PR #157, opened 2019-08-07
PD-12544 ProgressAccordion tests + docs 🐐

mikrotron:
### 🛠 Purpose
Add test coverage, better Storybook stories, and documentation of props to `<ProgressAccordion>`

### ✏️ Notes
- Utilized Screener's API for interaction to capture a screenshot of Accordion in both collapsed and expanded states.
- No `react-testing-library` or `cypress` tests added, as this is a very presentational component and visual regression testing should suffice.
- Adjusted `screener.conf` to include all stories with "Screener" in the name.
- **[PATCH UPDATE]** Fixed a small styling inconsistency for the height of incomplete items.

### 🖥 Screenshot
Screen Shot 2019-08-07 at 10 53 40 AM
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/PD-12544-responses-accordion-tests-docs' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/PD-12544-responses-accordion-tests-docs</a>


screener.config.js
```diff
@@ -7,5 +7,5 @@ module.exports = {
   storybookConfigDir: ".storybook",
   storybookStaticDir: "./.storybook/assets",
   failureExitCode: 0, // Don't fail CI builds
-  includeRules: [/Automation Tests\/Screener/, /^Stylers/, /^Button: Basic/],
```
oscarkwan: We aren't using Automation Tests folder anymore?

mikrotron: We are, but this condition would only catch stories in a `Screener` subfolder of `Automation Tests`.  Since I only have one test story (called `Screener` in `Automation Tests`) I don't want the extra layer.  And then though, well, any stories with `Screener` in the name, or in a `Screener` subfolder would be reasonable to snapshot.

nahumzs:
Why the storybook link is not working 🤔

mikrotron:
@nahumzs, where do you see that message?

It's working for me:
http://storybooks.highbond-s3.com/paprika/PD-12544-responses-accordion-tests-docs/?path=/story/progressaccordion--showcase

nahumzs:
It should automatically append the storybook link into the description of the PR. you can see a new step on semaphore that does that. @mikrotron

mikrotron:
Hm... Well, I pushed my last commits, which kicked off Semaphore... then a bit later actually opened the PR. I haven't pushed any more commits since then, so I suspect there was no opportunity to append to the PR description.

Issue #156, opened 2019-08-01
Extract DropdownMenu.Confirmation


jamiek-acl:
This issue is not exactly a bug, not exactly a "New feature request".

# Bug Report
## Expected behavior

I want to use the `DropdownMenu.Confirmation` popover without using a `DropdownMenu`.  That is, clicking on a delete icon reveals the `Confirmation` popover, like this:

![delete](https://user-images.githubusercontent.com/23224777/62310496-8bec4500-b43e-11e9-94f9-6d5bf7ff2fd0.gif)

I do not want the popover to appear until I click on the delete icon. To make this possible, I have to use state in the consuming app:

```
{isConfirmationVisible ? (
    <DropdownMenu.Confirmation
      onConfirm={() => {
        props.onClick();
        setIsConfirmationVisible(false);
      }}
      onClose={() => { setIsConfirmationVisible(false); }}
      onCancel={() => { setIsConfirmationVisible(false); }}
    />
) : null}
```

## Current behavior

When I click 'Cancel', or click away from the popover, that un-mounts the component in my consuming app.  Then the component itself tries to set its internal state to `isOpen: false` (Popover.js line 322), resulting in a console warning:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in Popover (created by DropdownMenu.Confirmation)
    in DropdownMenu.Confirmation (created by Trashbin)
```

`DropdownMenu.Confirmation` should be converted to its _own_ component (independent of `DropdownMenu`), then the DropdownMenu should be updated to use _that new_ component.

PR #159, opened 2019-08-08
Add confirmation component

tristanjasper:
### 🛠 Purpose

Create a confirmation component 

Tests to be provided in follow up PR
### Usage

```js
import Confirmation from "@paprika/confirmation";

<Confirmation
  confirmLabel="Delete filter"
  body="Lorem ipsum dolor amet vexillologist tacos selvage narwhal butcher twee ethical hot chicken"
  isOpenByDefault
  onConfirm={() => handleConfirm(onClose)}
  onCancel={() => handleCancel(onClose)}
  heading="Delete filter?"
/>;

or 

<Confirmation
  confirmLabel="Delete filter"
  body="Lorem ipsum dolor amet vexillologist tacos selvage narwhal butcher twee ethical hot chicken"
  onConfirm={() => handleConfirm(onClose)}
  onCancel={() => handleCancel(onClose)}
  onClose={() => handleClose(onClose)}
  heading="Delete filter?"
  renderTrigger={({ isConfirmOpen, handleOpenConfirm }) => (
    <Confirmation.Trigger isConfirmOpen={isConfirmOpen} handleOpenConfirm={handleOpenConfirm}>
      Trigger
    </Confirmation.Trigger>
  )}
/>;
```

### Props

- isOpenByDefault
- buttonSize
- confirmButtonType
- confirmLabel
- body
- onCancel
- onClose
- onConfirm
- heading

### ✏️ Notes to Reviewer

This has been extracted from the DropdownMenu component
In a follow up PR the DropdownMenu component will be refactor to use this component

### 🖥 Screenshots

image

 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-540-confirmation-component' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-540-confirmation-component</a>


packages/Confirmation/test/confirmation.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+  beforeEach(() => {
+    ({ getByText } = renderComponent());
+    confirmationComponent = getByText(/confirm delete/i);
+  });
+
+  it("should show confirmation panel in the document", () => {
+    expect(confirmationComponent).toBeVisible();
+  });
+});
```
tristanjasper: I'm intending to add some tests for the buttons

packages/Confirmation/stories/examples/Showcase.js
```diff
@@ -0,0 +1,39 @@
...
+import L10n from "@paprika/l10n";
+import Heading from "@paprika/heading";
+import Confirmation from "../../src";
+
+const handleConfirm = () => {
+  console.log("confirmed");
+};
+
+const handleCancel = () => {
+  console.log("canceled");
```
tristanjasper: todo convert these to be storybook actions

packages/Confirmation/stories/examples/Showcase.js
```diff
@@ -0,0 +1,39 @@
...
+    <Heading level={1} displayLevel={2} isLight>
+      Confirmation Showcase
+    </Heading>
+    <Tagline>Play with the controls to change the confirmation.</Tagline>
+    <Rule />
+    <L10n locale={select("locale", ["en", "de", "es", "fr", "ja", "pt", "zh"], "en")}>
+      <Confirmation
+        isOpen
+        buttonSize={select("Confirmation Button Size", ["x-small", "small", "medium", "large"], "medium")}
+        confirmLabel="Delete filter"
```
tristanjasper: todo: make `confirmLabel` and `description` editable in the showcase

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,80 @@
...
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import Popover from "@paprika/popover";
+import contentStyles from "./ContentStyles.styles";
+import ConfirmationStyles from "./Confirmation.styles";
+
+const propTypes = {
+  buttonSize: PropTypes.oneOf(ShirtSizes.DEFAULT),
+  confirmButtonType: PropTypes.oneOf(["primary", "destructive"]),
+  confirmLabel: PropTypes.string.isRequired,
```
nahumzs: this must be node, we shouldn't restrict the user to be an only string value.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,80 @@
...
+  description: PropTypes.node,
+  isOpen: PropTypes.bool.isRequired,
+  onCancel: PropTypes.func.isRequired,
+  onClose: PropTypes.func,
+  onConfirm: PropTypes.func.isRequired,
+  heading: PropTypes.string.isRequired,
```
nahumzs: same goes for this one, you should be able to pass whatever you want here.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,80 @@
...
+    buttonSize,
+    confirmButtonType,
+    confirmLabel,
+    description,
+    onConfirm,
+    onCancel,
+    onClose,
+    isOpen,
+  } = props;
+  const confirmRef = React.useRef(null);
```
nahumzs: could you rename the variable so is clear that the ref is targeting the Button Confirm? 

Without the word button, this seems that this ref would be targeting the entire confirm container.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,80 @@
...
+
+  React.useEffect(() => {
+    if (confirmRef.current) confirmRef.current.focus();
+  });
```
nahumzs: should this have `, [confirmRef])` as a dependency? 
```
React.useEffect(() => {
    if (confirmRef.current) confirmRef.current.focus();
  }, [confirmRef.current]);
```

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,80 @@
...
+
+  const I18n = useI18n();
+  return (
+    <Popover isOpen={isOpen} onClose={onClose}>
+      <Popover.Content>
+        <div css={contentStyles}>
+          <div css={ConfirmationStyles}>
+            <div className="dropdown-menu__confirmation-header">{heading}</div>
```
nahumzs: Why there are some `css` and some classNames? Use of className is not encouraged. Maybe now is the time to properly refactor this part?

packages/Confirmation/src/Confirmation.styles.js
```diff
@@ -0,0 +1,27 @@
...
+    ${stylers.fontSize()};
+    font-weight: bold;
+  }
+
+  .dropdown-menu__confirmation-description {
+    ${stylers.lineHeight()}
+    margin: ${tokens.space} 0;
+  }
+
+  .dropdown-menu__confirmation-footer {
```
nahumzs: all these can be their own variable avoiding the use of classNames.

packages/Confirmation/src/Confirmation.js
mikrotron: Since this component should always have an `<h>` heading, by making this a node, you give the consumer more flexibility, but you put more responsibility on them.  So I'm not sure sure about always reaching for `PropTypes.node`.   In fact, I'd consider taking on the additional complexity of supporting both `string` and `node`, and if it is a plain string, automatically including it inside a `<Heading level={2} displayLevel={5} />` (or however we want to code/display this heading to be properly accessible and consistent with the design system).

packages/Button/src/Button.styles.js
```diff
@@ -7,7 +7,7 @@ import { css, keyframes } from "styled-components";
 const commonStyles = `
   ${stylers.alignMiddle}
   ${stylers.lineHeight(-1)};
```
mikrotron: Semi-colon not needed for this line.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,85 @@
...
+const propTypes = {
+  buttonSize: PropTypes.oneOf(ShirtSizes.DEFAULT),
+  confirmButtonType: PropTypes.oneOf(["primary", "destructive"]),
+  confirmLabel: PropTypes.node.isRequired,
+  description: PropTypes.node,
+  isOpen: PropTypes.bool.isRequired,
+  onCancel: PropTypes.func.isRequired,
+  onClose: PropTypes.func,
+  onConfirm: PropTypes.func.isRequired,
+  heading: PropTypes.node.isRequired,
```
mikrotron: Would be nice to have this in alphabetical order.

packages/Confirmation/src/Confirmation.js
mikrotron: This is the label for the button.  I'm not so sure it should be anything besides a plain string.

packages/Confirmation/src/Confirmation.js
mikrotron: In fact... since `description` is a `node`, if this prop was simply not required, then the user could put whatever markup they want there, and opt out of this "shortcut" prop for providing a heading.

packages/Confirmation/src/Confirmation.styles.js
```diff
@@ -0,0 +1,25 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+export const confirmStyles = css`
+  padding: ${stylers.spacer(2)};
+`;
+
+export const confirmHeaderStyles = css`
+  ${stylers.fontSize()};
```
mikrotron: `;` not needed – if the `styler` function generates an entire line, it should include the `;`.  Up on line 6 where it only generates a value, then the `;` _IS_ needed.

packages/Confirmation/stories/cypress.stories.js
```diff
@@ -0,0 +1,6 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+
+import ConfirmationExample from "./examples/ConfirmationExample";
+
+storiesOf("confirmation / cypress", module).add("confirmation test", () => <ConfirmationExample />);
```
mikrotron: I don't think you'll need a Cypress story for this component... I think `react-testing-library` is going to give you enough coverage... but maybe a Screener one would be good to have?

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,85 @@
...
+  } = props;
+  const confirmButtonRef = React.useRef(null);
+
+  React.useEffect(() => {
+    if (confirmButtonRef.current) confirmButtonRef.current.focus();
+  }, [confirmButtonRef]);
+
+  const I18n = useI18n();
+  return (
+    <Popover isOpen={isOpen} onClose={onClose}>
```
mikrotron: In order for this component to be used _without_ a `<DropdownMenu>` component (like https://github.com/acl-services/paprika/issues/156), I think it's going to need a `getPositioningElement` prop for the `<Popover>`... isn't it?

packages/Confirmation/stories/Confirmation.stories.js
```diff
@@ -0,0 +1,16 @@
...
+import L10n from "@paprika/l10n";
+import ShowcaseStory from "./examples/Showcase";
+import ConfirmationExample from "./examples/ConfirmationExample";
+
+storiesOf("Confirmation", module)
+  .addDecorator(withKnobs)
+  .add("Showcase", ShowcaseStory)
+  .add("Confirmation", () => <ConfirmationExample />)
+  .add("ConfirmationExample with locale", () => (
+    <L10n locale="zh">
```
mikrotron: Can you wrap this in a `<Story>` component, like most other stories to, to give it some padding?

packages/Confirmation/stories/examples/Showcase.js
```diff
@@ -0,0 +1,41 @@
...
+const handleConfirm = () => {
+  action("Clicked the confirmation button")();
+};
+
+const handleCancel = () => {
+  action("Clicked the cancel button")();
+};
+
+const ExampleStory = () => (
+  <React.Fragment>
```
mikrotron: Can you wrap this in a `<Story>` component, like most other stories are, to give it some padding?

packages/Confirmation/stories/examples/Showcase.js
```diff
@@ -0,0 +1,41 @@
...
+    <Heading level={1} displayLevel={2} isLight>
+      Confirmation Showcase
+    </Heading>
+    <Tagline>Play with the controls to change the confirmation.</Tagline>
+    <Rule />
+    <L10n locale={select("locale", ["en", "de", "es", "fr", "ja", "pt", "zh"], "en")}>
+      <Confirmation
+        isOpen
+        buttonSize={select("Confirmation Button Size", ["x-small", "small", "medium", "large"], "medium")}
```
mikrotron: `x-small` doesn't work.

packages/Confirmation/stories/Confirmation.stories.js
```diff
@@ -0,0 +1,16 @@
...
+import { withKnobs } from "@storybook/addon-knobs";
+import L10n from "@paprika/l10n";
+import ShowcaseStory from "./examples/Showcase";
+import ConfirmationExample from "./examples/ConfirmationExample";
+
+storiesOf("Confirmation", module)
+  .addDecorator(withKnobs)
+  .add("Showcase", ShowcaseStory)
+  .add("Confirmation", () => <ConfirmationExample />)
+  .add("ConfirmationExample with locale", () => (
```
mikrotron: It would be nice to build a story that supports the use case that drove this task:
https://github.com/acl-services/paprika/issues/156

jamiek-acl:
Can I request the addition of an `isPending` prop.
This prop should get passed down to the primary/destructive button, and if it is pending, disable the `cancel` button.

mikrotron:
@jamiek-acl That's a great idea! It should probably disable both buttons... and probably apply the `isPending` prop to the primary button.

packages/Confirmation/src/Confirmation.js
tristanjasper: Good point I'll look into this

mikrotron: I think this might be a good pattern to follow in this case:
https://github.com/acl-services/paprika/blob/master/packages/DatePicker/src/DatePicker.js#L91

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,143 @@
...
+
+  const focusConfirmButton = () => {
+    if (confirmButtonRef.current) confirmButtonRef.current.focus();
+  };
+
+  const handleOpenConfirm = () => {
+    setIsConfirmOpen(true);
+    setTimeout(() => {
+      focusConfirmButton();
+    }, 200);
```
tristanjasper: this timeout was needed to ensure the confirm button got focus, due to popover changing the focus by default, however, a better solution could be found in future

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,143 @@
...
+  const {
+    heading,
+    buttonSize,
+    confirmButtonType,
+    confirmLabel,
+    description,
+    onConfirm,
+    onCancel,
+    onClose,
+    isOpenByDefault,
```
tristanjasper: The `isOpenByDefault` is important in the case of the `dropDownMenu` component using the confirmation component. In that situation it doesn't require a trigger element by default so we can just set it to show on mount

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,143 @@
...
+      React.cloneElement(triggerComponent, {
+        triggerRef,
+        confirmPanelRefId: confirmPanelRefId.current,
+      });
+  };
+
+  const I18n = useI18n();
+  // need to add moreProps to the popover
+  return (
+    <Popover isOpen={isConfirmOpen} onClose={handleCloseConfirm} {...moreProps}>
```
tristanjasper: I'm still a little unsure if we would need to pass an `getPositioningElement` in the case that a Popover.Trigger isn't used, Currently when using this confirmation component in the context of the `dropDownMenu` component it does seem to align itself somewhat in relation to the `dropDownMenu` trigger element which is interesting. - https://github.com/acl-services/paprika/pull/160

tristanjasper:
@jamiek-acl @mikrotron Please review this approach and if it will work for the certifications usage context

packages/Confirmation/src/Confirmation.js
tristanjasper: Ok i'm now passing 
```
getPositioningElement={() => document.getElementById("triggerElement")}
          offset={4}
``` in this situation

File: packages/Confirmation/stories/examples/ConfirmationExample.js [hunk 22]
```diff
@@ -0,0 +1,31 @@
...
+          align="right"
+          buttonSize="medium"
+          confirmLabel="Confirm button text"
+          description="Description text"
+          onConfirm={() => {}}
+          onCancel={() => {}}
+          onClose={() => {}}
+          heading="Delete filter?"
+          getPositioningElement={() => document.getElementById("triggerElement")}
+          offset={4}
```
tristanjasper: @mikrotron I'm not finding `offset` making a difference in the story example, did you get a chance to look into this usage?

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+  confirmLabel: PropTypes.node.isRequired,
+  description: PropTypes.node,
+  heading: PropTypes.node.isRequired,
+  defaultIsOpen: PropTypes.bool,
+  isPending: PropTypes.bool,
+  onCancel: PropTypes.func.isRequired,
+  onClose: PropTypes.func,
+  onConfirm: PropTypes.func.isRequired,
+  /** Render prop for rendering the trigger element that toggles the render panel */
+  renderTrigger: PropTypes.func,
```
mikrotron: I also like the idea of having this as `children`.  It can still be a render prop, and maybe leave a note that it _could_ support a `node` in the future, but for now a `func` is more important.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+  confirmDescriptionStyles,
+  confirmFooterStyles,
+} from "./Confirmation.styles";
+
+const propTypes = {
+  buttonSize: PropTypes.oneOf(ShirtSizes.DEFAULT),
+  confirmButtonType: PropTypes.oneOf(["primary", "destructive"]),
+  confirmLabel: PropTypes.node.isRequired,
+  description: PropTypes.node,
+  heading: PropTypes.node.isRequired,
```
mikrotron: Still think this should be a `string` and not `isRequired. Then wrapping it in a `<Heading>` can be taken care of by this component.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+  confirmStyles,
+  confirmHeaderStyles,
+  confirmDescriptionStyles,
+  confirmFooterStyles,
+} from "./Confirmation.styles";
+
+const propTypes = {
+  buttonSize: PropTypes.oneOf(ShirtSizes.DEFAULT),
+  confirmButtonType: PropTypes.oneOf(["primary", "destructive"]),
+  confirmLabel: PropTypes.node.isRequired,
```
mikrotron: Also, just I think this should just be a `string`, it's just the label for the primary button, what else would you need?

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+  description: PropTypes.node,
+  heading: PropTypes.node.isRequired,
+  defaultIsOpen: PropTypes.bool,
+  isPending: PropTypes.bool,
+  onCancel: PropTypes.func.isRequired,
+  onClose: PropTypes.func,
```
mikrotron: This can be optional, for cases when this is used as an uncontrolled component.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+    onCancel,
+    onClose,
+    ...moreProps
+  } = props;
+  const confirmButtonRef = React.useRef(null);
+  const triggerRef = React.useRef(null);
+  const confirmPanelRefId = React.useRef(null);
+
+  const focusConfirmButton = () => {
+    if (confirmButtonRef.current) confirmButtonRef.current.focus();
```
mikrotron: I feel like this should be in a `useEffect` hook...? It shouldn't refocus every time it renders... should it?

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+  confirmButtonType: "destructive",
+  description: null,
+  defaultIsOpen: false,
+  isPending: false,
+  renderTrigger: null,
+  onClose: () => {},
+};
+
+const Confirmation = props => {
+  const [isConfirmOpen, setIsConfirmOpen] = React.useState(false);
```
mikrotron: Could you just use `defaultIsOpen` as the default here, instead of having a hook to run `handleOpenConfirm()` later below?

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+
+  const I18n = useI18n();
+  // need to add moreProps to the popover
+  return (
+    <Popover isOpen={isConfirmOpen} onClose={handleCloseConfirm} {...moreProps}>
+      {props.renderTrigger ? <Popover.Trigger>{renderTrigger()}</Popover.Trigger> : null}
+      <Popover.Content id={confirmPanelRefId.current}>
+        <div css={contentStyles}>
+          <div css={confirmStyles}>
+            <div css={confirmHeaderStyles}>{heading}</div>
```
mikrotron: As per my previous suggestion to make `heading` a `string` and wrap it in a `<Header>`... that would mean no need for `confirmHeaderStyles`.

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+    defaultIsOpen,
+    isPending,
+    onConfirm,
+    onCancel,
+    onClose,
+    ...moreProps
+  } = props;
+  const confirmButtonRef = React.useRef(null);
+  const triggerRef = React.useRef(null);
+  const confirmPanelRefId = React.useRef(null);
```
mikrotron: I think you need to generate a `uid` for this somehow? It's being passed around, but remains `null`.

packages/Confirmation/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from "prop-types";
+import Button from "@paprika/button";
+
+const handleKeyDown = e => {
+  e.stopPropagation();
+};
+
+const propTypes = {
+  isConfirmOpen: PropTypes.bool.isRequired,
+  handleOpenConfirm: PropTypes.func.isRequired,
```
mikrotron: As a prop, this callback should have an `on` prefix.

packages/Confirmation/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,38 @@
...
+  triggerRef: PropTypes.shape({ current: PropTypes.instanceOf(Object) }) || null,
+  confirmPanelRefId: PropTypes.string,
```
mikrotron: I'd suggest simply `confirmId` would be a descriptive enough name for this.

packages/Confirmation/stories/examples/ConfirmationExample.js
mikrotron: I see this having an affect in Storybook.  Because `align='right'`, it creates a gap to the right of the trigger button and the confirmation popover.

packages/Confirmation/src/Confirmation.js
mikrotron: Shouldn't really be necessary if you are propagating `moreProps` to the `<Popover>`.

packages/Confirmation/src/Confirmation.js
```diff
@@ -41,18 +48,68 @@ const Confirmation = props => {
...
 
   const I18n = useI18n();
+  // need to add moreProps to the popover
   return (
-    <Popover isOpen={isOpen} onClose={onClose}>
-      <Popover.Content>
+    <Popover isOpen={isConfirmOpen} onClose={handleCloseConfirm} {...moreProps}>
+      {props.renderTrigger ? <Popover.Trigger>{renderTrigger()}</Popover.Trigger> : null}
+      <Popover.Content id={confirmPanelRefId.current}>
         <div css={contentStyles}>
```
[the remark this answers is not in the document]
tristanjasper: Hmm, I'm not seeing much overlap with the `contentStyles` that i am using, which styles specifically is the `card` component setting that i am setting in `contentStyles`

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+
+const propTypes = {
+  buttonSize: PropTypes.oneOf(ShirtSizes.DEFAULT),
+  confirmButtonType: PropTypes.oneOf(["primary", "destructive"]),
+  confirmLabel: PropTypes.node.isRequired,
+  description: PropTypes.node,
+  heading: PropTypes.node.isRequired,
+  defaultIsOpen: PropTypes.bool,
+  isPending: PropTypes.bool,
+  onCancel: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
tristanjasper: Ok, I guess we still have the `onClose` callback

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }, [confirmButtonRef, defaultIsOpen]);
+
+  const triggerProps = {
+    isConfirmOpen,
+    handleOpenConfirm,
+  };
+
+  const handleOnConfirm = () => {
+    setIsConfirmOpen(false);
```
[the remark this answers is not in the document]
tristanjasper: Ok i'm passing a callback to close the popup to the handleConfirm function. I think that will work

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,147 @@
...
+                isPending={isPending}
+                isSemantic={false}
+                ref={confirmButtonRef}
+                kind={confirmButtonType}
+                size={buttonSize}
+                onClick={handleOnConfirm}
+              >
+                {confirmLabel}
+              </Button>
+              <Button disabled={isPending} isSemantic={false} kind="minor" size={buttonSize} onClick={handleOnCancel}>
```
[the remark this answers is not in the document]
tristanjasper: hmm, I would think the button component should make itself disabled if the `isPending` prop is true, I guess it doesn't currently do that? Perhaps thats something we should change...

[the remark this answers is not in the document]
mikrotron: If you prefer to make that change to the `<Button>`, I would approve it. I agree, if a button `isPending`, it should always be `isDisabled` too.

packages/Confirmation/src/Confirmation.js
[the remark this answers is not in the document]
mikrotron: Most of the styles in this:
https://github.com/acl-services/paprika/blob/UX-540-confirmation-component/packages/Confirmation/src/ContentStyles.styles.js

Are already in this:
https://github.com/acl-services/paprika/blob/master/packages/Popover/src/components/Card/Card.styles.js#L5-L12

packages/Confirmation/src/Confirmation.js
tristanjasper: Ok i'm just rendering it as a string, not as html though. IS that what you mean?

packages/Confirmation/src/Confirmation.js
tristanjasper: I'm not sure, I think in the past it has been suggested to be able to pass an icon as well? I think a string is fine for now.

packages/Confirmation/src/Confirmation.js
tristanjasper: yeah it already isn't required

mikrotron: 🤦‍♂

packages/Confirmation/src/Confirmation.js
tristanjasper: It's currently only called via the `handleOpenConfirm` call, which isn't called on every render but is called in a `useEffect` under specific circumstances.

packages/Confirmation/src/Confirmation.js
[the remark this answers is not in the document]
tristanjasper: Ok i'll do the change in this Pr

packages/Confirmation/src/Confirmation.js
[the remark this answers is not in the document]
tristanjasper: got it

packages/Confirmation/src/components/Trigger/Trigger.js
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from "prop-types";
+import Button from "@paprika/button";
+
+const handleKeyDown = e => {
+  e.stopPropagation();
+};
+
+const propTypes = {
+  isConfirmOpen: PropTypes.bool.isRequired,
+  onHandleOpenConfirm: PropTypes.func.isRequired,
```
mikrotron: Whoop, shoulda been more specific.  Our naming conventions specify using `on` _or_ `handle` as a prefix, so just `onOpenConfirm` here will do.

packages/Confirmation/stories/examples/Showcase.js
```diff
@@ -0,0 +1,45 @@
...
+      <L10n locale={select("locale", ["en", "de", "es", "fr", "ja", "pt", "zh"], "en")}>
+        <Confirmation
+          align={select("Alignment", ["bottom", "right", "left", "top"], "bottom")}
+          buttonSize={select("Confirmation and Cancel Button Size", ["x-small", "small", "medium", "large"], "medium")}
+          confirmLabel={text("Confirm button text", "Delete filter")}
+          body={text(
+            "Description text",
+            "Lorem ipsum dolor amet vexillologist tacos selvage narwhal butcher twee ethical hot chicken."
+          )}
+          onConfirm={closeConfirm => handleConfirm(closeConfirm)}
```
mikrotron: Minor, since this is Storybook, but this can be:
`onConfirm={handleConfirm}`

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,154 @@
...
+  const triggerProps = {
+    isConfirmOpen,
+    handleOpenConfirm,
+  };
+
+  const closeConfirm = () => {
+    setIsConfirmOpen(false);
+  };
+
+  const handleOnConfirm = () => {
```
mikrotron: I think this is now unused?

packages/Confirmation/src/Confirmation.styles.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const confirmBodyStyles = css`
+    ${stylers.lineHeight()}
+    padding: ${tokens.spaceSm} 0 0 0;
+`;
+
+export const confirmFooterStyles = css`
+  margin-top: ${stylers.spacer(2)};
+
+  span[role="button"] {
```
mikrotron: Real minor, but I'd suggest this instead:
```scss
[role="button"] + [role="button"] {
  margin-left: ${tokens.space};
}
```

packages/Button/src/Button.js
```diff
@@ -131,7 +131,7 @@ const Button = React.forwardRef((props, ref) => {
     ...moreProps,
   };
   if (isSemantic) {
-    buttonProps.disabled = isDisabled;
+    buttonProps.disabled = isDisabled || isPending;
```
mikrotron: Good, but this only affects semantic buttons.  For `isSemantic={false}` buttons, you'll need to adjust line 126 as well.

tristanjasper: Ok made a few more adjustments in here as well to deal with isPending showing button in disabled state

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,158 @@
...
+    onConfirm(closeConfirm);
```
mikrotron:
```suggestion
    onConfirm(handleCloseConfirm);
```

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,158 @@
...
+        </Popover.Card>
+      </Popover.Content>
+    </Popover>
+  );
+};
+
+Confirmation.displayName = "Confirmation";
+Confirmation.propTypes = propTypes;
+Confirmation.defaultProps = defaultProps;
+Confirmation.Trigger = Trigger;
```
mikrotron: Let's rename to TriggerButton to be more specific.

packages/Confirmation/test/confirmation.spec.js
```diff
@@ -0,0 +1,66 @@
...
+      <Confirmation defaultIsOpen confirmLabel="Confirm Delete" onConfirm={() => {}} heading="Delete Button?" />
+    </L10n>
+  );
+
+  return {
+    ...renderedComponent,
+  };
+}
+
+// function renderComponentWithTrigger() {
```
mikrotron: Can we remove this? If you're saving it for later, a comment that explains why would be helpful.

mikrotron: ... just in case someone else takes it over and writes the additional tests...

packages/Confirmation/src/Confirmation.js
mikrotron: Oh, you'll have to do a little more work after this change to appease the linter... 😬

packages/Confirmation/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@paprika/stylers": "^0.1.1",
+    "@paprika/tokens": "^0.1.1",
+    "prop-types": "^15.7.2",
+    "uuid": "^3.3.2"
+  },
+  "peerDependencies": {
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
+    "styled-components": "^4.2.0",
+    "@paprika/icon": "^0.2.0"
```
alexzherdev: Why make icons a peer dependency?

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,158 @@
...
+  const triggerRef = React.useRef(null);
+  const confirmId = React.useRef(uuid());
+
+  const focusConfirmButton = () => {
+    if (confirmButtonRef.current) confirmButtonRef.current.focus();
+  };
+
+  const handleOpenConfirm = () => {
+    setIsConfirmOpen(true);
+    setTimeout(() => {
```
alexzherdev: nit:
```js
setTimeout(focusConfirmation, 200);
```

packages/Confirmation/src/Confirmation.js
```diff
@@ -0,0 +1,158 @@
...
+
+  const handleOnConfirm = () => {
+    onConfirm(closeConfirm);
+  };
+
+  const handleCloseConfirm = () => {
+    if (isConfirmOpen) {
+      setIsConfirmOpen(false);
+      if (triggerRef.current) triggerRef.current.focus();
+      setTimeout(() => {
```
alexzherdev: Similarly
```
setTimeout(onClose, 0);
```

packages/Confirmation/package.json
tristanjasper: Oh, i've moved to `devDependencies` as that is more appropriate, This was only need for a storybook example.

mikrotron: For Storybook, you can just rely on the `package.json` in the root of the Paprika repo.  As long as its dependencies include this latest version of `icon`, you should be good.

PR #161, opened 2019-08-08
UX-489 Fix package json, add borderColor prop, fix isDisabled

oscarkwan:
### 🛠 Purpose

---

- Fixed: Importing the tab was broken because there was no entry point
- Fixed: Disabled link styling
- Added: prop `borderColor`

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-489-edits' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-489-edits</a>


packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -37,6 +37,14 @@ export default function Tab(props) {
   const handleClick = isDisabled ? () => {} : onClick;
   const tabIndex = isSelected ? 0 : -1;
 
+  if (href && isDisabled) {
```
mikrotron: Instead if the following condition was just modified as `if (href && !isDisabled)` you could eliminate this whole block and let it render the `<RawButton>` below... no?

packages/Tabs/src/Tabs.js
```diff
@@ -1,18 +1,21 @@
...
 import PropTypes from "prop-types";
+import tokens from "@paprika/tokens";
 import TabsContext from "./TabsContext";
 import Panel from "./components/Panel/Panel";
 import Panels from "./components/Panels/Panels";
 import Tab from "./components/Tab/Tab";
 import List from "./components/List/List";
 
 const propTypes = {
+  borderColor: PropTypes.string,
```
mikrotron: How come we need this prop? Is a purple version needed sometimes? 

I would rather have something more controlled like `oneOf(['primary', 'secondary'])` so we maintain control / consistency.

packages/Tabs/src/components/Tab/Tab.styles.js
```diff
@@ -17,11 +24,11 @@ export const tabStyles = css`
   background-color: ${tokens.color.white};
   border: 0;
   border-bottom: ${tokens.spaceSm} solid transparent;
-  border-radius: 0;
+  border-radius: ${tokens.border.radius};
```
mikrotron: I think you could still use the `box-shadow` property (with the appropriate token) in the `&:focus` block to make the link focus ring styling consistent?

oscarkwan: It's using the rawbutton focus styling though, i extracted this to the &:focus anyways though

PR #163, opened 2019-08-09
PD-12544  ProgressAccordion touch ups 🐐

mikrotron:
### 🛠 Purpose
Some styling tweaks + some `data-pka-anchors` added for testing.

### ✏️ Notes to Reviewer
- Font sizes are now more _resilient_ to global CSS (example `body { font-size: 14px; }`)
- **[PATCH UPDATE]**

### DIFF
--- a/packages/ProgressAccordion/src/ProgressAccordion.js
+++ b/packages/ProgressAccordion/src/ProgressAccordion.js
@@ -63,7 +63,7 @@ const ProgressAccordion = props => {
   };
 
   return (
-    <div css={accordionStyles} {...moreProps} role="list" aria-label={a11yText}>
+    <div css={accordionStyles} {...moreProps} role="list" aria-label={a11yText} data-pka-anchor="progress-accordion">
       {validChildren.length > 0 &&
         validChildren.map((child, index) => {
           const isComplete = activeIndex > index;
@@ -75,7 +75,7 @@ const ProgressAccordion = props => {
 
           return (
             // eslint-disable-next-line react/no-array-index-key
-            <div css={itemStyles} key={index} role="listitem">
+            <div css={itemStyles} key={index} role="listitem" data-pka-anchor="progress-accordion.item">
               <Indicator {...indicatorProps} />
               {React.cloneElement(child, { label: getLabel(child.props.label, index), isComplete })}
             </div>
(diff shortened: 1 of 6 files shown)


packages/ProgressAccordion/src/ProgressAccordion.js
```diff
@@ -75,7 +75,7 @@ const ProgressAccordion = props => {
 
           return (
             // eslint-disable-next-line react/no-array-index-key
-            <div css={itemStyles} key={index} role="listitem">
+            <div css={itemStyles} key={index} role="listitem" data-pka-anchor="progress-accordion.item">
```
alexlaj: 🙇

packages/ProgressAccordion/src/components/Indicator/Indicator.styles.js
```diff
@@ -69,7 +69,8 @@ export const indicatorStyles = css`
 `;
 
 export const indicatorDotStyles = css`
-  ${stylers.z(2)};
+  ${stylers.fontSize()}
```
alexzherdev: So do we need semi-colons with these or not? 😄

packages/ProgressAccordion/src/ProgressAccordion.js
```diff
@@ -63,7 +63,7 @@ const ProgressAccordion = props => {
   };
 
   return (
-    <div css={accordionStyles} {...moreProps} role="list" aria-label={a11yText}>
+    <div css={accordionStyles} {...moreProps} role="list" aria-label={a11yText} data-pka-anchor="progress-accordion">
```
alexzherdev: Btw... is this not an `<ul>` for global styles reasons?

nahumzs: I would think that would be the reason.

mikrotron: Right.  The `listitem` is also a `<div>` instead of an `<li>` too.  It's more _resilient_ this way.

packages/ProgressAccordion/src/components/Indicator/Indicator.styles.js
mikrotron: Not. If the `styler` generates the whole line, then it generates the _whole_ line.  I think there's also a `stylers.fontSizeValue()` that does NOT produce the `;`... but you'd have to use it like:

```
font-size: ${stylers.fontSizeValue()};
```

PR #169, opened 2019-08-16
UX-486 Add toast component

allison-c:
### Installation

`> npm install --save @paprika/toast`
or
`> yarn add @paprika/toast`

### Usage

Please use `<L10n />` component to wrap `<Toast />` or you application.

Example:

```js
import Toast from "@paprika/toast";

<Toast>Toast message.</Toast>;
```

#### Toast

- `ariaAlert`
- `autoclose`
- `autocloseTimeout`
- `children`
- `hasCloseButton`
- `isOpen`
- `isSticky`
- `type`
- `zIndex`

[More detail about these props](https://github.com/acl-services/paprika/blob/master/packages/Toast/src/Toast.js)

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---
image

 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-486-Migrate-toast' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-486-Migrate-toast</a>


packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+
+const propTypes = {
+  /**
+   * Will make this an "assertive" alert that will interrupt immediately.
+   * Default behaviour is to be "polite" and wait until the user is idle.
+   */
+  ariaAlert: PropTypes.bool,
+
+  /** Will automatically close after 1500ms (or longer if provided by autocloseTimeout) */
+  autoclose: PropTypes.bool,
```
mikrotron: How about `canAutoClose`?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+
+  /** Duration (in ms) before Toast will automaticall close (if autoclose is true) */
+  autocloseTimeout: PropTypes.number,
```
mikrotron: How about `autoCloseDelay`?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+
+  /** Content of the Toast */
+  children: PropTypes.node,
```
mikrotron: Should it be `isRequired`?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+  isOpen: PropTypes.bool,
+
+  /** Callback that is executed after clicking the 'close' button */
+  onClose: PropTypes.func,
+
+  /** If the Toast is fixed to the top of the viewport */
+  isSticky: PropTypes.bool,
+
+  /** Determines the styling of the Toast */
+  type: PropTypes.oneOf(["success", "warning", "error", "info", "locked", "visually-hidden"]),
```
mikrotron: To align with the `<Button>` can we rename to `kind`?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+  ariaAlert: false,
+  autoclose: false,
+  autocloseTimeout: 5000,
+  children: null,
+  hasCloseButton: true,
+  isOpen: undefined,
+  onClose: () => {},
+  isSticky: false,
+  type: "info",
+  zIndex: 1006,
```
mikrotron: I think we should stick to our 7-layer system for this, but not sure which layer the `<Toast>` should be on by default.  Maybe just `1`?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+import toastStyles, { contentStyles, IconStyled, CloseButtonStyled } from "./Toast.styles";
+
+const propTypes = {
+  /**
+   * Will make this an "assertive" alert that will interrupt immediately.
+   * Default behaviour is to be "polite" and wait until the user is idle.
+   */
+  ariaAlert: PropTypes.bool,
+
+  /** Will automatically close after 1500ms (or longer if provided by autocloseTimeout) */
```
mikrotron: The actual default value is 5000, not 1500.  Not sure if that should be updated at this time? Do any designers have an opinion?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+    isOpen,
+    onClose,
+    isSticky,
+    type,
+    zIndex,
+    ...moreProps
+  } = props;
+
+  const ariaLive = ariaAlert ? "assertive" : "polite";
+  const [shouldShowToast, setShouldShowToast] = React.useState(isOpen === undefined ? true : isOpen);
```
mikrotron: This is just the internal state version of `isOpen`, right? I'd like to better reflect that relationship in the name... 🤔 Maybe `localIsOpen`? I don't like the `l` beside the `I` though... maybe `isToastOpen`?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+function Toast(props) {
+  const {
+    ariaAlert,
+    autoclose,
+    autocloseTimeout,
+    children,
+    hasCloseButton,
+    isOpen,
+    onClose,
+    isSticky,
```
mikrotron: `sticky` suggests a different behaviour to me... where it is inline, but as it scrolls to the top, it will "stick" in place.  We might even want to support that behaviour at some point.  For this implementation maybe `isFixed` would be better?

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+
+const minimumCloseTimeout = 1500;
+
+const icons = {
+  success: CheckIcon,
+  warning: CautionIcon,
+  error: ExclamationCircleIcon,
+  info: InfoCircleIcon,
+  locked: LockIcon,
+  "visually-hidden": InfoCircleIcon,
```
mikrotron: If it's hidden, it doesn't need an icon at all...

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,143 @@
...
+      aria-live={ariaLive}
+      css={toastStyles}
+      hasCloseButton={hasCloseButton}
+      isSticky={isSticky}
+      role="alert"
+      style={{ zIndex }}
+      type={type}
+      {...moreProps}
+    >
+      <IconStyled as={icons[type]} type={type} />
```
mikrotron: How about:
```js
{type !== 'visually-hidden' && <IconStyled as={icons[type]}  />}
```
Note `type={type}` is not needed.

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+    warning: tokens.color.yellowLighten30,
+    error: tokens.color.orangeLighten40,
+    info: tokens.color.cremeLighten5,
+    locked: tokens.color.yellowLighten30,
+  };
+
+  return css`
+    background-color: ${primaryColors[type]};
+    border-color: ${tinycolor(primaryColors[type])
+      .darken(10)
```
mikrotron: Interesting... 🤔 I'm concerned about including a whole library just for this function... but then again, lighten and darken functions would be really nice to have access to since we can use the ones from Sass.  What if we expose both of those via the `stylers` package, so we can hide the implementation details a bit?  There are some nice tools in that `tinycolor` library we might want to use later, like `setAlpha()` and `mostReadable()`.

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+import tinycolor from "tinycolor2";
+
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+import Button from "@paprika/button";
+import { visuallyHidden } from "@paprika/stylers/lib/includes";
+
+function getStylesForType(type) {
+  if (type === "visually-hidden") return "";
+  const primaryColors = {
```
mikrotron: I think this hash (and all other ones) could be defined first in this file.

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+  display: flex;
+  font-weight: normal;
+  line-height: ${AlertLineHeight};
+  margin-bottom: ${stylers.spacer(2)};
+  padding: ${tokens.spaceLg};
+  position: relative;
+  text-align: left;
+  transition: opacity 0.3s ease-out;
+
+  ${stylers.fontSize(0)};
```
mikrotron: Prefer `fontSize()`

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+  return css`
+    background-color: ${primaryColors[type]};
+    border-color: ${tinycolor(primaryColors[type])
+      .darken(10)
+      .toString()};
+    color: ${tinycolor(primaryColors[type]).toHsl().l > 0.7 ? tokens.color.black : tokens.color.white};
+  `;
+}
+
+const AlertLineHeight = "1.3";
```
mikrotron: `stylers.lineHeight()`

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+  if (type === "visually-hidden") return "";
+  const primaryColors = {
+    success: tokens.color.greenLighten50,
+    warning: tokens.color.yellowLighten30,
+    error: tokens.color.orangeLighten40,
+    info: tokens.color.cremeLighten5,
+    locked: tokens.color.yellowLighten30,
+  };
+
+  return css`
```
mikrotron: I would rather conditionally include these 3 rules below if `type !== "visually-hidden"`, maybe even combined with line 97.

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+
+export const IconStyled = styled.div`
+  flex-grow: 0;
+  flex-shrink: 0;
+  height: 20px;
+  padding: 2px ${tokens.space} 2px 0;
+  vertical-align: text-top;
+  width: 20px;
+
+  ${({ type }) => `color: ${iconColors[type]};`}
```
mikrotron: Do you think it's more readable as:
```
color: ${({ type }) => iconColors[type]};
```

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+  flex-shrink: 0;
+  min-height: 0;
+
+  ${({ type }) => closeButtonColors[type] && `color: ${closeButtonColors[type]};`}
+`;
+
+export const IconStyled = styled.div`
+  flex-grow: 0;
+  flex-shrink: 0;
+  height: 20px;
```
mikrotron: Can we use a combination of `font-size` and `line-height` instead of `height` and `width`?

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+
+export const CloseButtonStyled = styled(Button.Close)`
+  flex-grow: 0;
+  flex-shrink: 0;
+  min-height: 0;
+
+  ${({ type }) => closeButtonColors[type] && `color: ${closeButtonColors[type]};`}
+`;
+
+export const IconStyled = styled.div`
```
mikrotron: I assume you chose to export a component instead of styles (to use with the `css` prop) because it's a problem to use that with the `as` prop?

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const closeButtonColors = {
+  success: tokens.color.greenDarken20,
+  warning: tokens.color.yellowDarken40,
+  error: tokens.color.orangeLighten40,
+  locked: tokens.color.yellowLighten30,
+};
+
+export const CloseButtonStyled = styled(Button.Close)`
```
mikrotron: Could this be exported as styles and applied with the `css` prop?

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+`;
+
+const toastStyles = css`
+  align-items: flex-start;
+  border-radius: ${tokens.border.radius};
+  border-style: solid;
+  border-width: 1px;
+  display: flex;
+  font-weight: normal;
+  line-height: ${AlertLineHeight};
```
mikrotron: Not sure why this rule is special enough to warrant its own variable...?

packages/Toast/src/Toast.js
mikrotron: Do you think we could/should use an enum for these values?

packages/Toast/stories/NotControlledToastExample.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+import L10n from "@paprika/l10n";
+import Toast from "../src";
+
+export default function NotControlledToast() {
```
mikrotron: "Uncontrolled" is a more conventional name for this.

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+  warning: tokens.color.yellowDarken40,
+  error: tokens.color.orangeLighten40,
+  locked: tokens.color.yellowLighten30,
+};
+
+export const CloseButtonStyled = styled(Button.Close)`
+  flex-grow: 0;
+  flex-shrink: 0;
+  min-height: 0;
+
```
mikrotron: We should also have a bit of `margin-left` here, maybe a `$spaceLg`?

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+      .toString()};
+    color: ${tinycolor(primaryColors[type]).toHsl().l > 0.7 ? tokens.color.black : tokens.color.white};
+  `;
+}
+
+const AlertLineHeight = "1.3";
+
+const stickyStyles = css`
+  left: 50%;
+  max-width: 675px;
```
mikrotron: Should this be `min-width`? Should `max-width` be more like `calc(100%-32px)`?

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,100 @@
...
+  width: calc(100% - #{${stylers.spacer(4)}});
+`;
+
+const iconColors = {
+  success: tokens.color.green,
+  warning: tokens.color.yellowDarken10,
+  error: tokens.color.orange,
+  info: tokens.color.blue,
+  locked: tokens.color.yellowDarken10,
+  "visually-hidden": "inherit",
```
mikrotron: Also shouldn't be needed.

packages/Toast/src/Toast.js
allison-c: This is the current value we're using in acl-ui, if we change it, some toast might be broken in our apps

packages/Toast/src/Toast.js
allison-c: Oh my bad, I was testing, it should be 1500ms

packages/Toast/src/Toast.js
mikrotron: Only if they refactor to use the Paprika one.  If they do that, they'll have to add their own `zIndex` override... no big deal, in my opinion.

packages/Toast/src/Toast.js
allison-c: Currently it has info icon, maybe for accessibility?

packages/Toast/src/Toast.js
allison-c: I still need it for different colors though

packages/Toast/src/Toast.styles.js
allison-c: We can also hardcode those colors since they're only used in toast...

packages/Toast/src/Toast.js
mikrotron: I don't think it adds anything for accessibility, it's just decoration. I think it's better to not render it.

packages/Toast/src/Toast.styles.js
allison-c: Deleted!

packages/Toast/src/Toast.styles.js
mikrotron: That is the simpler solution! If you are up to it, you could create a "Feature Request" in issues to use this library to provide `lighten`, `darken`, `setAlpha` and any others you think would be particularly useful.

packages/Toast/src/Toast.styles.js
allison-c: It's because styled-components doesn't support dynamic tag rendering, so we have to use `as` prop to render the dynamic component and apply the style

packages/Toast/src/Toast.styles.js
allison-c: It's because  using `css` prop on `Button.Close` will override its default styles

packages/Toast/src/Toast.styles.js
allison-c: Double checked, it's fixed `min-width` in acl-ui. I think the idea was toast could be very long like this:
image

packages/Toast/src/Toast.js
```diff
@@ -0,0 +1,146 @@
...
+
+  function startTimer() {
+    const newTimer = setTimeout(() => {
+      handleClose();
+    }, Math.max(autoCloseDelay, minimumCloseTimeout));
+
+    timerRef.current = newTimer;
+  }
+
+  const memoizedStartTimer = React.useCallback(startTimer, []);
```
alexzherdev: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
> usually you’ll want to declare functions needed by an effect inside of it

Let's do that by inlining `startTimer` and moving `handleClose` into `useCallback`:
```js
const memoizedStartTimer = React.useCallback(() => {
  function handleClose() {
    clearTimeout(timerRef.current);
    if (isOpen === undefined) setIsToastOpen(false);
    onClose();
  }
  
  const newTimer = setTimeout(handleClose, Math.max(autoCloseDelay, minimumCloseTimeout));
  timerRef.current = newTimer;
}, []);
```
Now we can see that we've missed `autoCloseDelay`, `isOpen` and `onClose` as dependencies of `useCallback`

packages/Toast/src/Toast.styles.js
```diff
@@ -0,0 +1,97 @@
...
+  [Kinds.SUCCESS]: tokens.color.greenLighten50,
+  [Kinds.WARNING]: tokens.color.yellowLighten30,
+  [Kinds.ERROR]: tokens.color.orangeLighten40,
+  [Kinds.INFO]: tokens.color.cremeLighten5,
+  [Kinds.LOCKED]: tokens.color.yellowLighten30,
+};
+
+// Calculated by brightness.
+const borderColors = {
+  [Kinds.SUCCESS]: "#c5e2d3",
```
mikrotron: Can we use `greenLighten40` instead? The value is `#c4e3d3`... close enough for me!  I didn't check the others, but if there's a close enough token for each, that would be preferable.

packages/Toast/src/Toast.styles.js
mikrotron: Still verifying with designers?

Issue #127, opened 2019-07-17
Popover – follow-up issues


mikrotron:
#### Problem
`<Popover.Tip>` does not utilize the `zIndex` prop on the `<Popover>` root element, and so inconveniently requires its own `zIndex` prop to display on the same level as the rest of the Popover content.

#### Suggested Solution
If no `zIndex` is specifically provided to the `<Popover.Tip>`, then it should use the `zIndex` prop from `context`.

---

#### ~Problem~
~Sometimes a layout change is applied to the view without re-rendering the Popover, and so the `<Popover.Trigger>` position may move, but if the Popover is open, its content is not repositioned.~

#### ~Suggested Solution~
~Introduce an imperative API method – something like `reposition()` that can be called on the `<Popover>` by a controlling component that is aware of the (potential) view change.~

Re-opened as: https://github.com/acl-services/paprika/issues/240

alexzherdev:
I'd be curious to see an example of the second issue; perhaps we could avoid an imperative API. Additionally, if something like that is needed, maybe a `forceUpdate()` would be enough?

tristanjasper:
New Popover Issue to add to this:

**Problem**
Popover Content alignment is not correctly aligned to the trigger element when rendering the trigger in different positions. This is affecting the dropdown component as well as the datepicker.

![image](https://user-images.githubusercontent.com/4040102/61498124-e2ee1680-a976-11e9-8c27-39275c1943d1.png)
![image](https://user-images.githubusercontent.com/4040102/61498136-eda8ab80-a976-11e9-8581-425053961d0d.png)

**Suggested solution**
Allow an additional prop of "isContentAlignedToTrigger" perhaps to restrict its position

tristanjasper:
**Problem**
When there is no children rendered in the <Popover.Content> component there is still an 18px gap being rendered. Perhaps due to positioning. This is a a problem for the Paprika Dropdown Component as it shows a gab to a second Popover.
![image](https://user-images.githubusercontent.com/4040102/61498424-0e253580-a978-11e9-9184-2d685287852a.png)

**Suggested solution**
Unsure at this time

nahumzs:
@mikrotron are all these issues fixed on UX-548? can we close this issue?

mikrotron:
2/4 are fixed.  
1/4 is maybe not needed... what do you think? (https://github.com/acl-services/paprika/issues/127#issuecomment-512510035)
[...]

PR #170, opened 2019-08-17
UX-548  Popover Enhancements 🐐

mikrotron:
### 🛠 Purpose

After reproducing the original issue (UX-548 – positioning of Popover after Trigger scrolls out of view) it became clear that the behaviour was intentional and the solution was to simply use the `onClose` callback.

So the primary purpose of this PR is to resolve a lingering issue – If no `zIndex` is specifically provided to the `<Popover.Tip>`, then it should use the `zIndex` prop from context (https://github.com/acl-services/paprika/issues/127).

### ✏️ Notes to Reviewer

- Because this update changes the `zIndex` defaultProp for `<Tip>` it may affect existing implementations that rely on the default to layer the tip above the content.
- Recommended usage of `<Popover.Tip>` has also changed (should be a subsequent sibling of `<Popover.Content>` for intended layering since they will share the same `zIndex`). 
- [Patch Update] Also fixed the `z()` helper from the `Stylers` package to accept integers as valid arguments.
- Also added more detail to the `README.md` for the `<Popover>`.
- Added a `breaklines()` helper to the Storybook common styling helpers (was used in a story to reproduce original issue, which was later discarded). 
- Popover stories have been organized to use a single `popover.stories.js` manifest file, with all stories extracted as independent files in `/Examples`. 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-548--popover-positioning-bug' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-548--popover-positioning-bug</a>


packages/Stylers/src/helpers.js
```diff
@@ -28,7 +28,29 @@ export const lineHeight = (scale = 0) => `
...
+    -1,
+    0,
+    1,
+    2,
+    3,
+    4,
+    5,
+    6,
+    7,
+  ];
```
allison-c: Maybe something like `/^[0-7]$|^-1$|inherit/g.test(level)` is shorter

mikrotron: I like it!

PR #173, opened 2019-08-23
UX-541  refactor dropdownmenu confirmation - suggestions 🐐

mikrotron:
### ✏️ Notes to Reviewer

- Add `getPositioningElement` and `offset` to `<Confirmation>`.
- Exclude `null` children.
- Add `AlignTypes` to `customPropTypes` and also use with `<Popover>` and `<DatePickerPopover>`. 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-541--Refactor-dropdownmenu-confirmation--suggestions' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-541--Refactor-dropdownmenu-confirmation--suggestions</a>


packages/DatePicker/src/components/DatePickerPopover/DatePickerPopover.js
```diff
@@ -1,8 +1,9 @@
 import PropTypes from "prop-types";
+import { AlignTypes } from "@paprika/helpers/lib/customPropTypes";
 
 const propTypes = {
   /** Where the calendar is positioned relative to the trigger. */
-  align: PropTypes.oneOf(["top", "right", "bottom", "left"]),
+  align: PropTypes.oneOf(AlignTypes.ALL),
```
alexzherdev: Should we go further and export `PropTypes.oneOf(AlignTypes.ALL)` as e.g. `AlignType` propType?

mikrotron: A reasonable proposal 🤔 

For now I thought it would be reasonable to be consistent with how we are using `ShirtSizes`
https://github.com/acl-services/paprika/blob/master/packages/helpers/src/customPropTypes.js#L3-L11

But I actually have concerns about the approach for both since if we start relying on `react-docgen` to generate prop tables in our documentation, we might not get anything meaningful. Maybe there's a solution to that problem, but it might not be worth the effort and maybe we'll just revert these back, at least for the `propTypes` definition that we parse for the docs.

PR #178, opened 2019-08-30
UX-558 fix listbox styles

allison-c:
### 🛠 Purpose

Listbox styles are wrong

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

image

### DIFF
--- a/packages/ListBox/src/components/Option/Option.styles.js
+++ b/packages/ListBox/src/components/Option/Option.styles.js
@@ -4,6 +4,7 @@ import tokens from "@paprika/tokens";
 const blueSelected = "#e5f1fe";
 
 export const OptionStyled = styled.li`
+  border: 2px solid transparent;
   border-radius: 3px;
   font-size: 13px;
   margin-bottom: 4px;
@@ -12,7 +13,7 @@ export const OptionStyled = styled.li`
   &:hover {
     ${props => {
       const isSelected = props.isSelected ? `background: ${blueSelected}` : `background: #f0f0f0`;
-      const isDisabled = props.isDisabled ? "background: transparent; cursor: not-allowed; border: 0;" : "";
+      const isDisabled = props.isDisabled ? "background: transparent; cursor: not-allowed;" : "";
       const hasPreventDefaultOnSelect = props.hasPreventDefaultOnSelect ? "background: transparent;" : "";
 
       return `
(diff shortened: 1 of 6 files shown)


oscarkwan:
Conflict :P

packages/ListBox/src/components/Option/Option.styles.js
```diff
@@ -30,13 +31,13 @@ export const OptionStyled = styled.li`
 
     const isActived = props.isActive
       ? `
-        border: 3px solid Highlight;
+        border: 2px solid Highlight;
```
mikrotron: I don't think `Highlight` is valid for `border-style`.

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -129,25 +137,17 @@ export default function Trigger(props) {
...
       {hasRenderTrigger ? renderChildrenProps : renderLabel()}
       <TriggerActionIconsContainer>
-        {state.selectedOptions.length && hasClearButton ? (
-          <RawButtonClearButtonStyled
-            data-qa-anchor="clear-button"
-            hasRenderTrigger={hasRenderTrigger}
+        {state.selectedOptions.length && hasClearButton && !shouldHideClearButton ? (
+          <ClearButtonStyled
             isDisabled={isDisabled}
+            data-qa-anchor="clear-button"
```
mikrotron: Can we update to `data-pka-anchor` to align with our new convention, please?

packages/ListBox/src/components/Option/Option.styles.js
```diff
@@ -30,13 +31,13 @@ export const OptionStyled = styled.li`
...
         border-radius: 3px;
         `
       : "";
 
     const hasPreventDefaultOnSelect =
-      props.isActive && props.hasPreventDefaultOnSelect ? `border: 2px solid ${tokens.color.blackLighten60}` : "";
+      props.isActive && props.hasPreventDefaultOnSelect ? `border-color: ${tokens.color.blackLighten60}` : "";
 
     const isSelected = props.isSelected ? `font-weight: 600; background: ${blueSelected}` : "";
```
mikrotron: While we're here, can we remove the `font-weight: 600`? I believe I argued successfully with @sunminsarahlee that we shouldn't change font weights for this kind of "fleeting" state change.

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -1,6 +1,9 @@
 import React from "react";
 import PropTypes from "prop-types";
 import RawButton from "@paprika/raw-button";
+import CaretDownIcon from "@paprika/icon/lib/CaretDown";
+import CaretUpIcon from "@paprika/icon/lib/CaretUp";
+import TimesCircleIcon from "@paprika/icon/lib/TimesCircle";
```
mikrotron: What do you think about importing this icons with names that describe their usage?  Like `ClearIcon` vs `TimesCircleIcon`?  Or `ExpandIcon` vs `CaretDownIcon`?

packages/ListBox/src/components/Option/Option.styles.js
allison-c: I checked it in dev tool, `Highlight` is valid. I don't know why we use `Highlight` though. Can I confirm with @johnpeele and update it later?

packages/ListBox/src/components/Trigger/Trigger.js
allison-c: I think keeping the names same everywhere is easier for other devs who looks into the code and want to find out what icons are used, also can be easily copied and pasted. But I'm ok with both!

PR #181, opened 2019-08-30
UX-559 Fix default z index in toast

allison-c:
### 🛠 Purpose

Default zIndex should only apply to fixed toast

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-559-Fix-initial-zIndex-in-toast' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-559-Fix-initial-zIndex-in-toast</a>


packages/Toast/src/Toast.js
```diff
@@ -56,11 +56,13 @@ const defaultProps = {
   onClose: () => {},
   isFixed: false,
   kind: Kinds.INFO,
-  zIndex: 1,
+  zIndex: undefined,
```
mikrotron: I think it would be better to use `null` instead of `undefined` here.

packages/Toast/src/Toast.js
```diff
@@ -125,7 +128,7 @@ function Toast(props) {
       hasCloseButton={hasCloseButton}
       isFixed={isFixed}
       role="alert"
-      style={{ zIndex }}
+      style={{ zIndex: zIndex !== undefined ? zIndex : deafultZIndex }}
```
mikrotron: Can we take this opportunity to migrate this inline style into the `styles.js`, please?

PR #187, opened 2019-09-13
Ux 487 uploader migration

nahumzs:
### 🛠 Purpose

# UPLOADER

An agnostic UI component that let you upload files and customize the interface.

## Installation

`> npm install --save @paprika/uploader`

or

`> yarn add @paprika/uploader`

## Usage

The `<Uploader />` component make use of a context so you can create you own UI.

```js
import { UploaderContext } from "@paprika/uploader";
function YourUI() {
  const {
    /*provided by context*/
    cancelFile,
    FileInput,
    files,
    isCompleted,
    isDisabled,
    isDraggingOver,
    isDragLeave,
    removeFile,
    upload,
  } = React.useContext(UploaderContext);

  return (
    <>
      <FileInput>
        <div css={yourStyle}>FILE INPUT ZONE</div>
      </FileInput>
      {files.length ? (
        <ul>
          {files.map(file => (
            <li>{file.filename}</li>
          ))}
        </ul>
      ) : null}
    </>
  );
}


function App() {
  return (
     <Uploader>
       <YourUI />
     </Uploader>;
  )
}
```

### The children function

- **children** you can pass a component using the Provided Context.
- **FileInput**
  The input[type="file"] element ready to be consumed
- **files**
  A enhance list of the user selected files
- **isCompleted**
  Is true when all files have been processed
- **isDisabled**
  Describe the status of the component
- **isDragLeave**
  Will be true if a dragOver event occurred and leave the droppable area
- **isDraggingOver**
  Is `True` when detecting the user is dragging files over the droppable area
- **removeFile**
  `removeFile(key)` will remove a file from the files list, this function doesn't work while the file is on _PROCESSING_ state.
- **cancelFile**
  `cancelFile(key)` stop the request cycle keep in mind that if it's on _WAITINGFORSERVER_ state the server might save the file even if - the request has been cancel
- **upload**
  `upload()` give you the option to manually upload the files if the consumer decided to not use _hasAutoUpload_

### FileInput Component

The `<FileInput>{children}</FileInput>` is a ready to use component which is passed down by the `<Uploader />` context.
This component has attached accessibility and attributes required to make it work correctly, you can pass any children to beautify their aspect, keep in mind that is already an input[type='file'] so avoid putting a <Button /> component or a <button /> element as a children, just make it look like one.

### Endpoint

The _Uploader_ provides a `endpoint` prop which you can use to point the url of the service you want to reach, Ex.
`<Uploader endpoint="https://yoururl.com/api/upload" ...`

## Props

| Prop              | Type            | Description                                                                      |
| ----------------- | --------------- | -------------------------------------------------------------------------------- |
| a11yText          | string          | Accessible message for the input[type="file"].                                   |
| okFileTypes       | arrayOf(string) | An array of string describing the allowed file types for the uploader.           |
| canChooseMultiple | bool            | When false the uploader only accept one file per upload.                         |
| children          | function        | children function                                                                |
| defaultIsDisabled | bool            | initial disable state for the uploader                                           |
| endpoint          | string          | The url that will be use to upload the files ex. https://yoursite.com/api/upload |
| hasAutoUpload     | bool            | On true will upload the file as soon they are selected or dropped                |
| isBodyDroppable   | bool            | When true the user will be able to drop files at any part of the document.body   |
| maxFileSize       | number          | Size in Mebibytes                                                                |
| onChange          | func            | This callback fires every time a file has been processed                         |
| onCompleted       | func            | Will be fire once all files has been processed with the files as parameter       |
| headers           | arrayOf(object) | you can pass an array of header objects as need                                  |

[More detail about these props](https://github.com/acl-services/paprika/blob/master/packages/Uploader/src/Uploader.js)

### Accessibility

The Uploader automatically handle the on focus circle when using the `<FileInput />` component on it.

- it should notify when start uploading the files and when finishing processing the files.

### ✏️ Notes to Reviewer

*NOTES For developers*
- Still, need to add testing for the uploader
- Screen reader ccessibility will be implement once we have the finish design for or upload ui [See more in Figma](https://www.figma.com/file/rBaC100TzHUDSIwRmgYCyE/Paprika-Uploader?node-id=0%3A1), remember this is only the component to make the uploader to work. 

*NOTE for designer*
  - let's say you have designated only the FileInput to be a droppable area,
    shouldn't we display an error if you dropped the file outside of the container?
  - for the contrary, shouldn't we display in the other case that they can drop the files wherever they want?
  - when the file has uploaded

---

#### Features on this component
- [x] Accepst a defaultsDisabled prop.
- [x]  Handle and pass down the states for onDragOver onDragLeaves.
- [x] Pass down the latest state for the processing files.
- [x] Pass down function which allow to upload all files at once, delete a file, cancel a file request, remove a file from the list
- [x] Allowed the developer to sent multlipe headers to the server
- [x] Abstract the complexity of uploading files into a hook `useProcessFiles`
- [x] Allow the developer to restrain the user from selecting only specific file extensions
- [x] create stories for the component
- [x] Delete item only works if the file is on file.status === types.IDLE or all files has been processed
- [x] Allowed to handle errors depending if the file has a valid type, size, if the server fail, etc.
- [x] The input type file will adapt their size to whatever custom UI pass as a children on the `<FileInput />` component.
- [x] Allowed the developer to control when to upload the files if auto upload is disabled
- [x] A ProgressBar component for convenience. 
- [x] Blocks any interactions with the uploader if the files are on uploading/progress state.
- [x] Show a blue ring when the user focus on the FileInput component
- [x] Allowed to restraing the drop area to only the `FileInput` which is useful when there are more than one uploader in the same page
- [x] allowMultipleFile or force the user to only upload one file per interaction.
- [x] onFinished callback which gets fired once all files are processed.
- [x] onChange callback which gets fired every time the uploader process a file
- [x] ability to cancel a request
- [x] ability to pass custom headers as example [X-CSRF-Token]

### 🖥 Screenshots

---
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-487-uploader-migration' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-487-uploader-migration</a>


packages/Uploader/server/upload.php
```diff
@@ -0,0 +1,57 @@
+<?php
```
mikrotron: This is just for testing, right? Would that be more obvious if it were in the `tests/` folder?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+const propTypes = {
+  /**
+    Accessible message for the input[type="file"].
+  */
+  a11yText: PropTypes.string,
+  /**
+    An array of string describing the allowed file types for the uploader ex. ["image/*", ".pdf", ".doc", "docx"]
+    reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#Unique_file_type_specifiers.
+  */
+  acceptableFileTypes: PropTypes.arrayOf(PropTypes.string),
```
mikrotron: `okFileTypes`? Make it easier to avoid typos?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+  /**
+    When false the uploader only accept one file per upload.
+   */
+  allowMultipleFile: PropTypes.bool,
```
mikrotron: `canChooseMultiple`? I'd like to limit how many boolean prefixes we use, if possible.  Need to document it, but `['is', 'are', 'has', 'can', 'should']` should be sufficient.

Here's a recommendation I like that establishes this convention:
https://wiki.eclipse.org/Recommenders/CodingConventions#is_prefix_should_be_used_for_boolean_variables_and_methods.

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+
+    FileInput,   // the input[type="file"] element ready to be consumed
+    files,       // a enhance list of the user selected files
+    hasFinished, // is true when all files have been processed
+    isDisabled,  // describe the status of the component
+    isDragLeave, // will be true if a dragOver event occurred and leave the droppable area
+    isDragOver,  // True when detecting the user is draggin files over the droppable area
+    removeFile,  // removeFile(key) will remove a file from the files list, this function doesn't work while the file is on PROCESSING state.
+    cancelFile,  // cancelFile(key) stop the request cycle keep in mind that if it's on WAITINGFORSERVER state the server might save the file even if the request has been cancel
+    upload,      // upload() give you the option to manually upload the files if you are decide to not use hasAutoupload
```
mikrotron: This is a lot for inline prop comments.  Can we just leave it to the README, or even reference the README here for more info?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+  */
+  children: PropTypes.func.isRequired,
+  /**
+    initial disable state for the uploader
+  */
+  defaultIsDisable: PropTypes.bool,
```
mikrotron: `defaultIsDisable` → `defaultIsDisabled`

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+  /**
+    The url that will be use to upload the files ex. https://yourEndPointURL.com
+  */
+  endpoint: PropTypes.string.isRequired,
+  /**
+    On true will upload the file as soon they are selected or dropped
+  */
+  hasAutoupload: PropTypes.bool,
```
mikrotron: `hasAutoupload` → `hasAutoUpload`

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+    When true the user will be able to drop files at any part of the document.body triggering
+    the upload listeners. On false will only received files if they are drop exactly on the FileInput area.
+  */
+  isBodyDroppableArea: PropTypes.bool,
+  /**
+    Size in Mebibytes which is use for comparing each file that will be upload
+    you can make use of Uploader.convertUnitsToMebibytes() for easily convert units to Mebibyes
+    ex.  Uploader.convertUnitsToMebibytes(4) => 4194304 (close to 4MB);
+  */
+  maximumFileSize: PropTypes.number,
```
mikrotron: `maxFileSize`?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+  endpoint: PropTypes.string.isRequired,
+  /**
+    On true will upload the file as soon they are selected or dropped
+  */
+  hasAutoupload: PropTypes.bool,
+  /**
+    When true the user will be able to drop files at any part of the document.body triggering
+    the upload listeners. On false will only received files if they are drop exactly on the FileInput area.
+  */
+  isBodyDroppableArea: PropTypes.bool,
```
mikrotron: `isBodyDroppable`?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+  /**
+    This callback fires every time a file has been processed
+  */
+  onChange: PropTypes.func,
+  /**
+    Will be fire once all files has been processed with the files as parameter, finishing doesn't mean that all
+    files has been correctly uploaded, only means that they were processed, to find about the state of each file
+    iterate over the files list.
+  */
+  onFinished: PropTypes.func,
```
mikrotron: `onCompleted` seems more commonly used?
For example: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onCompleted

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+    isBodyDroppableArea,
+    endpoint,
+    children,
+    defaultIsDisable,
+    headers,
+  } = props;
+
+  const refInput = React.useRef();
+  const refContainer = React.useRef(null);
+  const [refId] = React.useState(uuidv4());
```
mikrotron: `const inputId = React.useRef(uuidv4()).current;` ?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+  const i18n = useI18n();
+  const label = a11yText || i18n.t("uploader.label");
```
mikrotron: Maybe this should be below all the hooks?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+
+  function handleChange(event) {
+    if (isDisabled) return;
+
+    const files = getFiles({ event, maximumFileSize, acceptableFileTypes, endpoint });
+    setFiles(() => {
+      if (refInput.current) {
+        refInput.current.value = "";
+      }
+      return allowMultipleFile ? files : [files[0]]; // in case only allow one file per upload
```
mikrotron: Should there be an error displayed in the case of `allowMultipleFile`, and `file.length > 1`?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+  allowMultipleFile: PropTypes.bool,
+  /**
+    children function received the following parameter which you can use to build you own UI:
+
+    FileInput,   // the input[type="file"] element ready to be consumed
+    files,       // a enhance list of the user selected files
+    hasFinished, // is true when all files have been processed
+    isDisabled,  // describe the status of the component
+    isDragLeave, // will be true if a dragOver event occurred and leave the droppable area
+    isDragOver,  // True when detecting the user is draggin files over the droppable area
```
mikrotron: `isDraggingOver` (this is the same name that `react-beautiful-dnd` uses)

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+   */
+  allowMultipleFile: PropTypes.bool,
+  /**
+    children function received the following parameter which you can use to build you own UI:
+
+    FileInput,   // the input[type="file"] element ready to be consumed
+    files,       // a enhance list of the user selected files
+    hasFinished, // is true when all files have been processed
+    isDisabled,  // describe the status of the component
+    isDragLeave, // will be true if a dragOver event occurred and leave the droppable area
```
mikrotron: Why couldn't `!isDraggingOver` be used for this ?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+      {children({
+        FileInput,
+        files,
+        hasFinished,
+        isDisabled,
+        isDragLeave,
+        isDragOver,
+        removeFile,
+        cancelFile,
+        upload,
```
mikrotron: Should these three callback parameters be prefixed with `on`?

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+      </div>
+    );
+  }
+
+  return (
+    <>
+      {children({
+        FileInput,
+        files,
+        hasFinished,
```
mikrotron: If you rename the prop, then I'd rename this too to `isCompleted`.

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,219 @@
...
+        isDragOver,
+        removeFile,
+        cancelFile,
+        upload,
+      })}
+    </>
+  );
+}
+
+const Uploader = React.forwardRef(UploaderComponent);
```
mikrotron: Seems cleaner to do this inline above, no need for a Uploader AND UploaderComponent that both need propTypes attached to them.

packages/Uploader/src/Uploader.styles.js
```diff
@@ -0,0 +1,27 @@
...
+  display: inline-block;
+  height: 100%;
+  position: absolute;
+  width: 100%;
+`;
+
+export const inputFileStyles = css`
+  border: 0;
+  height: 100%;
+  opacity: 0;
```
mikrotron: Why not simply use `stylers.visuallyHidden`?

packages/Uploader/src/Uploader.styles.js
```diff
@@ -0,0 +1,27 @@
...
+  overflow: hidden;
+  position: absolute;
+  width: 100%;
+  &:focus + label {
```
mikrotron: A bit odd that this bit of label styling is defined in the `inputFileStyles`.  Since the styling requires that `input` and `label` be siblings and are both descendants of the `container`, I'd suggest just combining it all into the `containerStyles`.

packages/Uploader/src/helpers.js
```diff
@@ -0,0 +1,122 @@
...
+
+export function fileSizeUnitsToHumanReadableFormat(size) {
+  const aMultiples = ["KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"];
+  let nMultiple = 0;
+  let sOutput = "";
+  // this calculate Kibebytes, Mebibyes, etc.
+  // But are being label as KB and MB for easier understanding.
+  for (let nApprox = size / 1024; nApprox > 1; nApprox /= 1024, nMultiple++) {
+    // fixed to 2 decimals but we could add more if need it
+    sOutput = `${nApprox.toFixed(2)} ${aMultiples[nMultiple]}`;
```
mikrotron: Seems odd to keep recalculating this.  Maybe a `while` loop could be employed to get the `nApprox` and `nMultiple` you need, then just return this.

packages/Uploader/src/Uploader.js
nahumzs: Like their approach.

packages/Uploader/server/upload.php
nahumzs: I think for this specific component which depended on a server having the folder at that level is ok.

packages/Uploader/src/Uploader.js
nahumzs: preference I guess, I put all the const variables a the top.

packages/Uploader/src/Uploader.js
nahumzs: I'm leaning to think we don't need it 🤔.  That error wouldn't occur if the user select files via the file input, because the input automatically will only let her pick one file. 

This is relevant when a user drop multiple files, in that case, I'm just taking the first one, is a little edge casie, not sure if we want to show/display an error.

packages/Uploader/src/Uploader.js
nahumzs: yes we can, is just a different event, up to the author how wants to use the isDraggingOver prop

packages/Uploader/src/Uploader.js
nahumzs: no, they shouldn't, these are methods which will be invoked, they are not callbacks.

packages/Uploader/src/Uploader.js
mikrotron: If it's not needed, I'd suggest leaving it out of the API, less is more.  If it's really needed though, we should try come up with a name that's a bit clearer. `isDragLeave` sounds a bit awkward.

packages/Uploader/src/helpers.js
nahumzs: no, understanding exactly what you mean, this was a port from our previous code, and we need to calculate for each field uploaded, and recalculated it if the files changes.

packages/Uploader/README.md
```diff
@@ -0,0 +1,95 @@
...
+- **cancelFile**
+  `cancelFile(key)` stop the request cycle keep in mind that if it's on _WAITINGFORSERVER_ state the server might save the file even if - the request has been cancel
+- **upload**
+  `upload()` give you the option to manually upload the files if the consumer decided to not use _hasAutoUpload_
+
+Example:
+
+```js
+<Uploader>
+  {({ FileInput, files, isDisabled, isDraggingOver, isDragLeave, isCompleted, upload, removeFile, cancelFile }) => {
```
alexzherdev: Couple things:
- I agree it can be useful to have a base component with the functionality that you can customize with the UI. But are there plans for Paprika to implement some "blessed" variant of that UI? Just having a skeleton like this is too bare-bones IMO (pun intended). Imagine everyone consuming this having to implement something that correctly makes use of all the `isDraggingOver`/`isDragLeave`/etc. I guess in other words, is this UI something that should be part of the design system and thus be implemented in Paprika?
- I don't really see the need to surface `FileInput` to the consumer. It _would_ be useful if we were expecting any props on it that the consumer could pass, but it only has children. So it seems like it can just be "inlined" into the uploader component.
- Have you considered an API alternative to a render prop? Seems like we could use context plus a custom hook? I.e. (taking into account the point above):
```jsx
<Uploader>
  <YourStylishUI />
</Uploader>
```
Where `Uploader` provides context and a hook to consume it so that
```jsx
import { useUploader } from '@paprika/uploader';

function YourStylishUI(props) {
  const { files, ... } = useUploader();
  ...
}
```

packages/Uploader/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "@paprika/stylers": "^0.1.5",
+    "@paprika/tokens": "^0.1.4",
+    "prop-types": "^15.7.2",
+    "uuid": "^3.3.3"
+  },
+  "peerDependencies": {
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
+    "styled-components": "^4.2.0",
+    "superagent": "^5.1.0"
```
alexzherdev: Why peer? Seems like a regular dependency to me.

packages/Uploader/server/upload.php
alexzherdev: I understand this is not going to be the most high-traffic part of the repo, but requiring PHP to be installed, and also understanding PHP seems a maintenance liability.

packages/Uploader/README.md
nahumzs: > is this UI something that should be part of the design system and thus be implemented in Paprika
Yes, this PR is only the Component which I'm going to use to implement the UI in my next PR. The developers shouldn't need to use this component, shouldn't even care that exist. 

I didn't add the UI here to keep the PR focus and avoid extra code to review.

About the DS, currently, we have two cases for the FileInput. One the common Drag and Drop, and second when the uploader is a plain button. But of then will benefit from this component. 

> Have you considered an API alternative
No, didn't think about it until reading your comment, I like your idea will change the API. 

> I don't really see the need to surface FileInput 
Is need it because the input file should expand or decrease their size depending the content of the UI, for hover event or focus. 
I will try to see if I can avoid passing it down.

packages/Uploader/package.json
nahumzs: what if a consumer has already declare superagent? I trying to avoid similar conflicts that we could run into like the one with `numeraljs` been declare in one app and the need to use a different version in our sriracha library. 

That's why I did it.

packages/Uploader/server/upload.php
nahumzs: 🤔 most mac has PHP installed automatically, so would not be a problem running it on macs, which is most of our cases currently. 
Also, I couldn't find a better alternative to mockup a simple server, handling files, while coding the uploader. 

Open to suggestions.

packages/Uploader/package.json
alexzherdev: (I don't remember/know what it was but) It's possible that `numeraljs` had an issue that required only having one version of it. But unless there is such evidence, we should go with a regular dependency and let the package manager sort out the versions. It would also not be nice to require someone to manually install superagent if they were not already using it.

packages/Uploader/src/helpers.js
```diff
@@ -0,0 +1,122 @@
...
+    isTypeValid: isValidFileType({ file, okFileTypes }),
+  };
+
+  validation.isValid = [validation.isSizeValid, validation.isTypeValid].every(rule => rule);
+
+  return validation;
+}
+
+function createFilesDataStructure({ files, maxFileSize, okFileTypes, endpoint }) {
+  return [...files].map(file => {
```
alexzherdev:
```suggestion
  return files.map(file => {
```
Don't need a new array since you're only mapping over

packages/Uploader/src/helpers.js
```diff
@@ -0,0 +1,122 @@
...
+    // fixed to 2 decimals but we could add more if need it
+    sOutput = `${nApprox.toFixed(2)} ${aMultiples[nMultiple]}`;
+  }
+
+  return sOutput;
+}
+
+function getExtension({ file }) {
+  const filename = file.name;
+  return filename.substr((~-filename.lastIndexOf(".") >>> 0) + 2); // eslint-disable-line
```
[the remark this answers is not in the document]
nahumzs: copy/pasta years ago from a stackoverflow question, I'm just porting our previous implementation. :/

packages/Uploader/stories/Testing.js
```diff
@@ -0,0 +1,146 @@
...
+        <p>
+          After selecting your images, click the button to upload it.
+          <button type="button" onClick={handleUpload}>
+            upload images
+          </button>
+        </p>
+      ) : null}
+      <p>{isDisabled ? "isDisabled is true" : "isDisabled is false"}</p>
+      <p>{isDraggingOver ? "isDraggingOver 🤚" : ""}</p>
+      <p>{isDragLeave ? "isDragLeave 👋" : ""}</p>
```
mikrotron: Apart from logging this event here, none of the stories illustrate how (or why) this prop is used.  I'm still not convinced it's necessary – I think `!isDraggingOver` should suffice?

packages/Uploader/stories/Uploader.stories.js
```diff
@@ -0,0 +1,115 @@
...
+import Uploader from "../src/Uploader";
+import Testing from "./Testing";
+
+storiesOf("Uploader", module)
+  .add("Basic example with all files successfully uploaded", () => (
+    <>
+      <Uploader
+        endpoint="http://localhost:9000/upload.php"
+        onChange={files => {
+          console.log("onChange files:", files);
```
mikrotron: I'd suggest using the `actions` addon instead of a console log.

packages/Uploader/stories/Uploader.stories.js
```diff
@@ -0,0 +1,115 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import Uploader from "../src/Uploader";
+import Testing from "./Testing";
+
+storiesOf("Uploader", module)
+  .add("Basic example with all files successfully uploaded", () => (
+    <>
```
mikrotron: If you use `<Story>` here (`import { Story } from "storybook/assets/styles/common.styles"`) then the story will have a bit of padding.

packages/Uploader/stories/Testing.js
```diff
@@ -0,0 +1,146 @@
...
+    width: 300px;
+  `,
+};
+const propTypes = {
+  hasUploadButton: PropTypes.bool,
+};
+
+const defaultProps = {
+  hasUploadButton: false,
+};
```
mikrotron: No big deal, but I'd suggest _not_ adding `propTypes` to story components. It's just extra maintenance.

packages/Uploader/src/Uploader.js
```diff
@@ -0,0 +1,221 @@
...
+        <div css={containerStyles} ref={refContainer}>
+          <input
+            multiple={canChooseMultiple}
+            id={refId}
+            onChange={handleChange}
+            ref={refInput}
+            type="file"
+            accept={okFileTypes.join(",")}
+          />
+          <label htmlFor={refId}>
```
mikrotron: you could avoid the logic of generating an id by simply using `<input aria-label={label} />` and removing the actual `<label>`.

packages/Uploader/src/components/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,22 @@
...
+  hasError: PropTypes.bool.isRequired,
+  isCompleted: PropTypes.bool.isRequired,
+  progress: PropTypes.number.isRequired,
+};
+
+export default function ProgressBar(props) {
+  const { progress, isCompleted, hasError } = props;
+
+  return (
+    <div css={styles.container} data-type="uploader-progress-bar">
```
mikrotron: `data-pka-anchor='uploader.progress-bar``

packages/Uploader/src/components/ProgressBar/ProgressBar.styles.js
```diff
@@ -0,0 +1,27 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens";
+
+export default {
```
mikrotron: Interesting pattern. Kind of trivial , but I like `styles.container` better than `containerStyles`.  

Something worth striving for consistently, or just an author's choice kind of thing?

packages/Uploader/src/helpers.js
mikrotron: Okay, 🙈

packages/Uploader/src/types.js
```diff
@@ -0,0 +1,8 @@
+export default {
+  IDLE: "idle",
+  PROCESSING: "processing",
+  WAITINGFORSERVER: "waiting-for-server",
+  SUCCESS: "success",
+  ERROR: "error",
+  CANCEL: "cancel",
+};
```
mikrotron: Seems pretty specific to the `<Uploader>`, would this actually be a candidate to migrate to a `types` package?

packages/Uploader/server/upload.php
nahumzs: no at this point. cypress can fake upload, from what I have read, but testing will happen in the next PR where I will be implementing the UI

packages/Uploader/src/components/ProgressBar/ProgressBar.styles.js
nahumzs: lol, this is the pattern that I follow on my side project. 
I guess, I implemented without noticing, to me is slightly better and easier to read and easier to import because you only have to import the styles object.

packages/Uploader/src/types.js
nahumzs: I think this is too specific where status and props are more generic, it might not make sense to move it outside of here.

packages/Uploader/stories/Uploader.stories.js
nahumzs: I will polish stories and everything in my next PR, still need to build the UI, this is just logic handle. :)

PR #188, opened 2019-09-16
UX-570 Introduce CollapsibleText component 🐐

mikrotron:
### 🛠 Purpose
Add a new component that will truncate content to be revealed when user clicks a "show more" button.

### ✏️ Notes to Reviewer
- [PATCH] update to `<Button>` component – added `data-pka-anchors` and fixed a minor styling regression.

### ⚡️To Do
- [x] ~~Consider using "Show all" instead of "Show more".~~
- [x] Rename component.
- [x] Add a test (just Screener).

### 🖥 Screencap
showmore
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-570--create-showmore-component' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-570--create-showmore-component</a>


packages/ShowMore/src/ShowMore.js
```diff
@@ -0,0 +1,94 @@
...
+import PropTypes from "prop-types";
+import truncate from "lodash/truncate";
+import uuid from "uuid/v4";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import Button from "@paprika/button";
+import { toggleStyles } from "./ShowMore.styles";
+
+const propTypes = {
+  /** Additional description for "Show more" link. Should be a "topic" that will be appended to "Show more about [topic]". */
+  a11yText: PropTypes.string,
```
oscarkwan: if we want to make our components accessible out of the box, should this almost be required?

oscarkwan: I find myself in the UX accessibility compliance tickets that I just miss out on some a11yText in components, but if it was required (or somehow strongly encouraged) then i wouldn't forget

packages/ShowMore/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/l10n": "^0.2.1",
+    "@paprika/button": "^0.1.9",
+    "@paprika/tokens": "^0.1.4",
+    "class-names": "^1.0.0",
```
nahumzs: you don't need this one :)

packages/ShowMore/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "lodash": "^4.17.15",
```
nahumzs: you can install only the truncate package https://www.npmjs.com/package/lodash.truncate smaller dependency.

packages/ShowMore/src/ShowMore.js
nahumzs: mmmm what if you only want to be `show more` without a topic oscar?

packages/ShowMore/src/ShowMore.js
```diff
@@ -0,0 +1,94 @@
...
+  /* Full content to be revealed. */
+  children: PropTypes.node.isRequired,
+
+  /* Length, in characters, of truncated preview content. */
+  collapsedLength: PropTypes.number,
+};
+
+const defaultProps = {
+  a11yText: null,
+  collapsedLength: 255,
```
nahumzs: is this number random?

packages/ShowMore/src/ShowMore.js
```diff
@@ -0,0 +1,94 @@
...
+  const { a11yText, children, collapsedLength, ...moreProps } = props;
+
+  const I18n = useI18n();
+  const [isCollapsed, setIsCollapsed] = React.useState(true);
+  const [collapsedContent, setCollapsedContent] = React.useState();
+  const contentId = React.useRef(uuid()).current;
+
+  React.useEffect(() => {
+    setCollapsedContent(`${getTruncatedVersion(ReactDOMServer.renderToStaticMarkup(children), collapsedLength)} `);
+  }, [children, collapsedLength]);
```
nahumzs: move `getTruncatedVersion` inside of the use effect because is a dependency, you can't add it because will make your component rerender all the time. 

you eslint should be warning you about this.

packages/ShowMore/src/ShowMore.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+function getTruncatedVersion(content, length) {
+  return truncate(content, {
+    length,
+    separator: " ",
+  });
+}
+
+const ShowMore = props => {
```
nahumzs: nip:  `function ShowMore(props) {

}`

better than assigning it to a variable.

packages/ShowMore/src/ShowMore.js
alexzherdev: I think in this case it's fine because `getTruncatedVersion` doesn't depend on any props or state and can safely be created outside of the component altogether.
(As an absurd example, the same can be said about `ReactDOMServer.renderToStaticMarkup`—you're not specifying it as a dependency, right?)

alexzherdev: (find "You can try moving that function outside of your component" in this section: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies)

alexzherdev: But ACKCHUALLY I have a couple issues with this effect as is:
- I don't think this needs to be an effect at all? It's not a side effect and it doesn't depend on the DOM being updated. So instead of having collapsed content as state, it can just be inlined into `dangerouslySetInnerHTML` (`__html: ReactDOM.renderToStaticMarkup(...)`)
- What's the current use case for this? `renderToStaticMarkup` assumes server-rendering compatibility of what is passed into it. Meaning you generally can't pass arbitrary component tree and expect it to work. (notably, if a component needs its `componentDidMount` invoked to be able to work properly, that will not fly) So this component will be restricted to simpler use cases (e.g. paragraphs of static content), which I think should be fine?

packages/ShowMore/src/ShowMore.js
mikrotron: Yeah, it's technically accessible without the `a11yText`, it's just _better_ if there's more context for screen readers than simply "show more".  It might be a bit too heavy-handed to make it required in this case.

packages/ShowMore/src/ShowMore.js
mikrotron: This is the "usual" truncation limit. At least in every case I've implemented a similar function in Projects, this was the arbitrary number.

packages/CollapsibleText/README.md
```diff
@@ -0,0 +1,38 @@
+## CollapsibleText
```
alexzherdev: I'm 👍 but would `CollapsibleContent` be a better name?

mikrotron: We could squirrel all day long on a name... but to me `content` could suggest _anything_ whereas `text` at least _suggests_ it's intended only for static elements (plain or rich text).  

I'd also suggested `<CollapsiblePreview>` and `<CollapsibleTeaser>`... but somewhat preferred JP's suggestion or `<CollapsibleText>`, even if it is kinda boring 😆.

PR #193, opened 2019-09-19
UX-572 Fix tooltip focus

allison-c:
### 🛠 Purpose

- Keep current focus when hovering or tabbing through an `isEager` popover
- Using aria-describedby for popover content

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

Jietu20190919-140355

 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-572-Fix-tooltip' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-572-Fix-tooltip</a>


packages/Popover/src/Popover.js
```diff
@@ -123,15 +124,31 @@ class Popover extends React.Component {
         isOpen prop is also provided.`
       );
     }
+
+    this.ariaIdForContent = `popover-content-${uuidv4()}`;
```
nahumzs: nip: shouldn't just be called `elementId`? I don't see the reason to match it only with `aria`.

packages/Popover/src/Popover.js
```diff
@@ -274,7 +291,7 @@ class Popover extends React.Component {
     //       find the first focusable element like button, input, etc?
     //       can focus automatically
     //       should we set focus into the popover content automatically?
```
nahumzs: this comment don't make sense anymore, I think we already know what we should do with the focus right? maybe update it or remove the comments?

packages/Popover/stories/examples/WithTriggers.js
```diff
@@ -46,14 +45,16 @@ const ExampleStory = () => (
       </h5>
       <Popover isEager>
         <Popover.Trigger>
-          {handler => (
+          {(handler, ariaId) => (
```
nahumzs: maybe wrap all attributes in an object and then spread it across the component `{...attributes}` this is how I have seen other API doing it. That way if we want to add more attributes will be simple and transparent for our users

packages/Popover/src/components/Content/Content.js
```diff
@@ -79,6 +79,20 @@ const Content = React.forwardRef((props, ref) => {
...
+          hasCardComponent = true;
+          return React.cloneElement(child, {
+            id: content.ariaId,
+          });
+        }
+
+        return child;
+      })
+    : props.children;
+
```
tristanjasper: Currently in the dropdown component we are adding an id to the <Popover.content> directly in order to set focus on the Dropdown content. 
https://github.com/acl-services/paprika/blob/master/packages/DropdownMenu/src/DropdownMenu.js#L122

I believe with these changes we won't need to do that anymore in the dropdown component if it is done internally in the Popover right?

packages/Popover/src/Popover.js
allison-c: Because the only usage for this id is to accomplish a11y requirement

packages/Popover/src/components/Trigger/Trigger.js
```diff
@@ -41,10 +48,9 @@ function Trigger(props) {
   const { children, a11yText } = props;
 
   if (typeof children === "function") {
-    return <React.Fragment>{children(handleTriggerEvent)}</React.Fragment>;
+    return <React.Fragment>{children(handleTriggerEvent, content.ariaId)}</React.Fragment>;
```
mikrotron: What if you cloned this `children` element and added the `id` to its props, so that the consumer never even needed to know it existed?

I used this technique for the `<DropdownMenu>`:
https://github.com/acl-services/paprika/blob/master/packages/DropdownMenu/src/DropdownMenu.js#L74

packages/Popover/src/components/Content/Content.js
```diff
@@ -94,8 +108,9 @@ const Content = React.forwardRef((props, ref) => {
       style={contentStyles}
       tabIndex={isOpen ? 0 : -1}
       zIndex={content.zIndex}
+      id={isEager && !hasCardComponent ? content.ariaId : null}
```
mikrotron: Why not just always apply the `id` to the `Content`, whether there's a `Card` or not? It would be nice if `Content` didn't have to care about its contents.

allison-c: Because when there's `Card` component, we need to apply the `id` on `Card` component, otherwise we apply the same id on `Content` component. `aria-describedby` should be on the node has text as child.

mikrotron: It's supposed to ONLY contain text? 😳 

I tried hacking an example page and the screen reader still picks it up (Voiceover with Chrome, at least):

Screen Shot 2019-09-20 at 3 50 05 PM

Did you find a guideline or recommendation or something about the id being on the leaf node?

allison-c: Let me do some research, I tested with NVDA it only works with nodes ONLY contain text

mikrotron: I guess it turned out to be okay with JAWS / NVDA?

allison-c: Found a similar example(https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) and tested with browsers. I think I messed up my voiceover on Chrome, but it works fine on Safari. This node structure also works on NVDA, so I remove the code for cloning the child and pass it down. The code looks much cleaner now.

PR #196, opened 2019-09-19
UX-561 Fix nested collapsible

allison-c:
### 🛠 Purpose

Fix nested collapsible

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-561-Fix-nested-collapsible' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-561-Fix-nested-collapsible</a>


packages/Collapsible/src/Collapsible.js
```diff
@@ -115,10 +111,16 @@ const Collapsible = props => {
...
 
+  const hiddenStyles = {
+    display: "none",
+  };
+
   return (
     <div css={collapsibleStyles} {...collapsibleProps} role="group" {...moreProps}>
       {hasOnlyIconToggle ? renderCollapsibleByIcon() : renderDefaultCollapsible()}
-      <div className="collapsible__body">{children}</div>
+      <div className="collapsible__body" style={isCollapsed ? hiddenStyles : null}>
```
mikrotron: Instead of adding inline styles, can we export a `collapsibleBodyStyles` and apply it with the `css` prop? 

Maybe combined with the `width:100%` and get rid of the class in favour of a `pka-anchor`?

allison-c: Using styled-components caused the bug I want to solve. It's because if we use 2 `Collapsible` on the same page, they'll have the same class name of `collapsible_body` and the collapse state was messed up. I did some research on styled-components, there's no solution on their side. So using inline CSS should be a quick and reliable fix.

allison-c: but I'll remove the css using `data-pka` as selector

Issue #182, opened 2019-09-03
Popover – align to the "edge" of a Trigger


mikrotron:
**Problem**
The only Popover alignments available are: top, bottom, right, left. The Popover content will always try to be aligned along the center of the trigger and content. For the `<DropdownMenu>` and `<DatePicker>` it is preferable to align the content to the left edge (if space permits) of the `<Trigger>` element.  

![image](https://user-images.githubusercontent.com/4040102/61498124-e2ee1680-a976-11e9-8c27-39275c1943d1.png)
![image](https://user-images.githubusercontent.com/4040102/61498136-eda8ab80-a976-11e9-8581-425053961d0d.png)

**Suggested solution**
Add a new `edge` alignment option that will open the Popover below the trigger, with left edges aligned.  If space is insufficient, the Popover will adapt automatically to align to the right edge / above the Trigger.

nahumzs:
would the default edge alignment be left/right? or how the popover would align if it's a the center of the screen without constrains?

mikrotron:
So, that's an open question for a designer, but my proposal is the default is "open the Popover below the trigger, with left edges aligned".  It would only adjust to use the right edge, or open above the trigger if it doesn't fit.

jayde-acl:
@nahumzs Sounds like the default `edge` alignment will be left if there are no constraints. 

@mikrotron The proposed solution sounds like it would work well, I do have a question though. There appear to be cases in our designs where most of the dropdown menus are right-aligned to the trigger.

Does the dropdown menu popover also use this component? Is it worthwhile to add in an `edge-right` as well as the `edge-left` instead of just `edge`?

Examples:
- Projects / Assurance Plans - ellipsis to trigger - dropdown menu with edit & delete options
https://acl.invisionapp.com/share/3XO2W7HND7E#/320207010_Delete_Assurance_Plan

- Projects / Project - Certifications - ellipsis to trigger dropdown menu with various options
https://acl.invisionapp.com/share/GXQEQPFNHM8#/351796991_Deleting_Certification_-_1

- Projects / Scheduler / Add new trigger dropdown menu 

I am sure that there are many more of these `edge-right` popover menus throughout our platform, these are just the ones I've seen recently.

I'm also trying to think of the last time I've seen someone use `top` while designing. Is this one necessary? I'm assuming that `top` means for the popover to appear above the trigger.

cc @johnpeele

nahumzs:
[...]

PR #197, opened 2019-09-19
UX-581 Support edge prop

allison-c:
### 🛠 Purpose

Add `edge` prop

### ✏️ Notes to Reviewer

**PATCH UPDATE**
Applications using popover don't need to change anything.

### 🖥 Screenshots

---
Resolve https://github.com/acl-services/paprika/issues/182 

 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/UX-581-Add-edge-prop-for-popover' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/UX-581-Add-edge-prop-for-popover</a>


packages/Popover/src/helpers/getPosition.js
```diff
@@ -61,21 +62,31 @@ const getAnchor = (rect, align) => {
   return positions[align]();
 };
 
-export const getContentCoordinates = ({ rect, targetRect, scrollRect, align, offset = 12 }) => {
+export const getContentCoordinates = ({ rect, targetRect, scrollRect, align, offset = 12, edge }) => {
   const anchor = getAnchor(targetRect, align);
 
+  const getContentXByEdge = edge => {
```
AndreyChernykh: Is there a reason to pass `edge` as a parameter to this function, instead of using is directly (as we do for `rect` or `targetRect`, for instance)?

AndreyChernykh:
It would be nice to add unit-tests for all these calculations.

packages/Popover/src/helpers/getPosition.js
allison-c: Oh no reason, I forgot to remove it when I move around the functions 😿

mikrotron:
Would it make the calculations much more complicated if, instead of adding an `edge` prop, `edge-left` and `edge-right` were added as options for `align`?

Also, I know we still need to add a proper Showcase story for this component, but in the meantime, can we add a knob to one of the stories somewhere so we can see the behaviour in Storybook?

mikrotron:
> It would be nice to add unit-tests for all these calculations.

Agreed. Maybe in a separate PR though?

allison-c:
> Would it make the calculations much more complicated if, instead of adding an `edge` prop, `edge-left` and `edge-right` were added as options for `align`?
> 
> Also, I know we still need to add a proper Showcase story for this component, but in the meantime, can we add a knob to one of the stories somewhere so we can see the behaviour in Storybook?

Hey Mike! I just added the knob for `edge` prop. I think this should be a separate prop, because we have `edge-right align-top` and `edge-right align-bottom`, so if we adding options in `align ` prop, we'll have to introduce 4 options

PR #199, opened 2019-09-20
Ux 580 New Checkbox component

tristanjasper:
### 🛠 Purpose

Initial addition of checkbox component, 

Checkbox component can be set with three values in `checkedState` prop.

`checked`, `unchecked` and `indeterminate`

Todo:

- Test to be added in follow up PR

### ✏️ Notes to Reviewer

- **[BREAKING CHANGE]** made to the `stylers` package (removed two regrettable includes).  Should have no impact in the wild, these includes were almost certainly not being used anywhere outside Paprika.

### 🖥 Screenshots

Current storybook :
image

Intention
image

---
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/ux-580-checkbox-component' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/ux-580-checkbox-component</a>


packages/Stylers/package.json
```diff
@@ -14,5 +14,8 @@
   },
   "dependencies": {
     "@paprika/tokens": "^0.1.4"
+  },
+  "peerDependencies": {
+    "styled-components": "^4.2.0"
```
alexzherdev: FYI technically this is a breaking change for `Stylers`.

packages/Stylers/src/includes.js
```diff
@@ -1,19 +1,20 @@
...
 
 //
 // Layout
 //
 
-export const boxSizingStyle = `
-  box-sizing: border-box;
-`;
-
-export const inlineBlockStyle = `
```
mikrotron: @alexzherdev I'd consider this (and the change to line 7) an even clearer breaking change to `Stylers`.  I don't think anyone's using them outside of Paprika, so I'm keen to break it.

packages/Stylers/package.json
mikrotron: Do you think this might have been causing my build issue with Storybook v5.2?

packages/Stylers/src/includes.js
alexzherdev: haha I didn't notice this.
I can see at least a couple repos depending on stylers. But regardless we should just do the default thing and increase version to 0.2.0 next release and be set

packages/Stylers/package.json
alexzherdev: Extremely unlikely 🤔

packages/Stylers/src/includes.js
mikrotron: Sorry, by "them" I mean these _specific_ includes that I've removed.  So I don't think anything will actually break when people update to `v0.2`.

alexzherdev: ah ok ok 👍

packages/Checkbox/stories/examples/CheckboxExample.js
```diff
@@ -0,0 +1,72 @@
...
+      </Checkbox>
+      <Rule />
+      <div
+        css={`
+          [data-pka-anchor="checkbox"] {
+            margin-bottom: 12px;
+          }
+        `}
+      >
+        <Checkbox {...props} onChange={handleChange(state2)} isChecked={state2[0]}>
```
tristanjasper: @mikrotron Are you addressing these stories?

image

packages/Checkbox/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/helpers": "^0.2.0",
+    "@paprika/icon": "^0.2.0",
+    "@paprika/stylers": "^0.1.5",
+    "@paprika/tokens": "^0.1.4",
+    "classnames": "^2.2.6",
```
mikrotron: ✂️

packages/Checkbox/src/Checkbox.js
```diff
@@ -0,0 +1,69 @@
...
+  children: null,
+  isChecked: false,
+  isDisabled: false,
+  isIndeterminate: false,
+  size: ShirtSizes.MEDIUM,
+};
+
+const Checkbox = props => {
+  const checkboxId = React.useRef(uuid());
+  const inputRef = React.useRef(null);
```
mikrotron: How about adding `.current` on the end of these two lines, then removing that everywhere else to clean it up a bit?

packages/Checkbox/src/Checkbox.js
```diff
@@ -0,0 +1,69 @@
...
+  const { a11yText, children, isChecked, isDisabled, isIndeterminate, size, ...moreProps } = props;
+
+  const styleProps = {
+    hasChildren: !!children,
+    size,
+  };
+
+  React.useEffect(() => {
+    if (inputRef.current) inputRef.current.indeterminate = isIndeterminate;
+  }, [isIndeterminate]);
```
mikrotron: This should also be dependent on `isChecked`.

mikrotron: Actually, on second though, this might be problematic too... I think it might work better if there was a `checkedState: oneOf(["checked", "unchecked", "indeterminate"])` or something like that, since the states are all mutually exclusive.

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,121 @@
...
+
+const topOffset = size => (getSpacingInt(size) - integize(stylers.fontSizeValue()) * stylers.lineHeightValue(-1)) / 2;
+const isCheckboxBigger = size => topOffset(size) > 0;
+const labelPadding = ({ hasChildren, size }) => {
+  const top = isCheckboxBigger(size) ? `${topOffset(size)}px` : 0;
+  const left = hasChildren ? `${getSpacingInt(size) + integize(tokens.space)}px` : getSpacing(size);
+  return `${top} 0 ${top} ${left}`;
+};
+
+const checkerTop = size => (isCheckboxBigger ? "0px" : `${Math.abs(topOffset(size) / 2)}px`);
```
mikrotron: I don't think the `Math.abs()` is needed... the `isCheckboxBigger` condition will ensure it's a positive value, right?

packages/Checkbox/stories/Checkbox.stories.js
```diff
@@ -0,0 +1,16 @@
...
+import CheckboxExample from "./examples/CheckboxExample";
+import ShowcaseStory from "./examples/Showcase";
+
+storiesOf("Checkbox", module)
+  .addDecorator(withKnobs)
+  .add("Showcase", ShowcaseStory);
+
+storiesOf("Checkbox/Automation Tests/Screener", module).add("CheckboxExample", () => (
+  <CheckboxStory>
+    <CheckboxExample size="large" />
```
mikrotron: We can clean up the Storybook stories later, but why make the Screener story "large"?

tristanjasper:
@mikrotron Just waiting on the dash icon now, I don't see any other issues, the hover state doesn't seem to be a problem for me when the component is disabled. Do you see any other issues?

packages/Checkbox/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/helpers": "^0.2.0",
+    "@paprika/icon": "^0.2.0",
+    "@paprika/stylers": "^0.1.5",
+    "@paprika/tokens": "^0.1.4",
+    "prop-types": "^15.7.2",
+    "uuid": "^3.3.2"
+  },
+  "devDependencies": {
```
mikrotron: As long as these are in the main package.json, I don't think they're required here.

packages/Checkbox/src/Checkbox.js
```diff
@@ -0,0 +1,75 @@
...
+  INDETERMINATE: "indeterminate",
+};
+
+const propTypes = {
+  a11yText: PropTypes.string,
+  children: PropTypes.node,
+  isDisabled: PropTypes.bool,
+  onChange: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(ShirtSizes.DEFAULT),
+  checkedState: PropTypes.oneOf([checkboxStates.CHECKED, checkboxStates.UNCHECKED, checkboxStates.INDETERMINATE]),
```
mikrotron: `oneOf(Object.values(checkboxStates))`

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,125 @@
+import { css } from "styled-components";
+import stylers from "@paprika/stylers";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import tokens from "@paprika/tokens";
+
+const integize = token => Number.parseInt(token, 10);
```
mikrotron: Should we add this to `stylers`? Maybe not in this PR, but when you have to figure out how to share styling for the `<Radio>`, I think it might be time to shuffle this out since it's pretty generic, and helps keep the code more concise.

mikrotron: Not sure about the name `integize` though... 🤔

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,125 @@
...
+  return `${top} 0 ${top} ${left}`;
+};
+
+const checkerTop = size => (isCheckboxBigger ? "-1px" : `${Math.abs(topOffset(size) / 2)}px`);
+const checkBoxIconLeft = ({ size }) => `${getSpacingInt(size) / 2}px`;
+
+const checkboxStyles = css`
+  ${stylers.boxSizingStyles};
+  font-size: ${fontSizeValue};
+
```
mikrotron: ✂️

packages/Checkbox/src/Checkbox.styles.js
tristanjasper: 😄 we can change when we refactor it for radio button

packages/DropdownMenu/README.md
```diff
@@ -1,6 +1,6 @@
 ## DropdownMenu
```
mikrotron: Looks like you've edited the wrong README file? This is actually for the DropdownMenu.

packages/DropdownMenu/README.md
```diff
@@ -11,77 +11,20 @@ or
...
-          heading="Delete filter?"
-          onConfirm={handleConfirm(onCloseMenu)}
-        />
-      );
-    }}
-  >
-    Delete filter
-  </DropdownMenu.Item>
-</DropdownMenu>;
+<Checkbox {...props} onChange={handleChange} checkState={checkedStateValue>
```
mikrotron: The `{...props}` here seems unneeded and maybe a bit confusing.

packages/DropdownMenu/README.md
tristanjasper: 😱

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,133 @@
...
+  input[type="checkbox"] {
+    ${stylers.visuallyHidden};
+
+    &:focus {
+      & + label {
+        box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
+        border-color: ${tokens.highlight.active.noBorder.borderColor};
+        outline: none;
+      }
+    }
```
mikrotron: Suggestion:
```js
&:focus + label::before {
    box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
    border-color: ${tokens.highlight.active.noBorder.borderColor};
}
```

packages/Checkbox/src/Checkbox.js
```diff
@@ -0,0 +1,75 @@
...
+        id={checkboxId}
+        checked={checkedState === checkboxStates.CHECKED}
+        disabled={isDisabled}
+        ref={inputRef}
+        {...inputProps}
+      />
+      <label htmlFor={checkboxId}>
+        {children}
+        <CheckIcon className="checkbox-icon" aria-hidden data-pka-anchor="checkbox.icon.check" />
+        <DashIcon aria-hidden className="checkbox-icon" data-pka-anchor="checkbox.icon.indeterminate" />
```
mikrotron: Oh, probably the don't need the `classNames`.

tristanjasper: I am using the classnames for some styling of the icons...

mikrotron: Let's use the `pka-anchors` instead (in the next PR for the `<Radio>`)

tristanjasper: At least in the paprika components i think we should keep the testing attributes separate from styling purposes. (Consuming applications may use the testing attributes for styling though)

Discussed with @mikrotron, agreed that when I building out the radio input component, we should refactor the styles applied on these icon components to be variables coming from the `checkbox.styles.js`.

johnpeele:
A couple of small UI tweaks:
1. Font size for `small` size should be `14px`
2. Font size for `medium` size should be `16px` 

I am including a screenshot of the latest design mocks for referencing the checkmark and dash icon sizes
Checkboxes

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,130 @@
...
+const topOffset = size => (getSpacingInt(size) - integize(stylers.fontSizeValue()) * stylers.lineHeightValue(-1)) / 2;
+const isCheckboxBigger = size => topOffset(size) > 0;
+const labelPadding = ({ hasChildren, size }) => {
+  const top = isCheckboxBigger(size) ? `${topOffset(size)}px` : 0;
+  const left = hasChildren ? `${getSpacingInt(size) + integize(tokens.space)}px` : getSpacing(size);
+  return `${top} 0 ${top} ${left}`;
+};
+
+const checkerTop = size => (isCheckboxBigger ? "-1px" : `${Math.abs(topOffset(size) / 2)}px`);
+const checkBoxIconLeft = ({ size }) => `${getSpacingInt(size) / 2}px`;
```
jamiek-acl: This is a bit of a rat's nest of complexity. Could it be simplified so guys like me can understand it? (map it, dont calculate it)

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,130 @@
...
+      border-radius: ${tokens.border.radius};
+      content: "";
+      height: ${size};
+      left: 0;
+      width: ${size};
+      z-index: 1;
+    }
+
+    & + label:hover {
+      &:before {
```
jamiek-acl: could just do this on one line instead of two
`& + label:hover:before`
Also it should be `::before`

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    &:checked,
+    &:indeterminate {
+      & + label::before {
+        background: ${tokens.color.black};
+        border: none;
+      }
+      & + label:hover {
+        &:before {
```
jamiek-acl: could just do this on one line instead of two. Also it should be `::before`

packages/Checkbox/stories/Checkbox.stories.js
```diff
@@ -0,0 +1,27 @@
...
+import CheckboxAlign from "./examples/CheckboxAlign";
+import CheckboxGrouping from "./examples/CheckboxGrouping";
+import ShowcaseStory from "./examples/Showcase";
+
+storiesOf("Checkbox", module)
+  .addDecorator(withKnobs)
+  .add("Showcase", ShowcaseStory)
+
+  .add("Checkbox group", () => (
+    <L10n locale="zh">
```
jamiek-acl: Need these `L10n`s?

packages/Checkbox/stories/Checkbox.stories.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+import { Story } from "storybook/assets/styles/common.styles";
+
+// Common Checkbox story styles
+
+export const CheckboxStory = styled(Story)`
```
jamiek-acl: Just my opinion that this file adds unnecessary noise

packages/Checkbox/stories/examples/CheckboxAlign.js
```diff
@@ -0,0 +1,40 @@
...
+  const [checkedState, setCheckedState] = React.useState(props.value || UNCHECKED);
+
+  const handleChange = () => setCheckedState(checkedState === CHECKED ? UNCHECKED : CHECKED);
+
+  return (
+    <div>
+      <Heading level={1} displayLevel={2} isLight>
+        Checkbox Alignment Right Example
+      </Heading>
+      <Rule />
```
jamiek-acl: Maybe _this_ should be a component

packages/Checkbox/stories/examples/CheckboxGrouping.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
```
jamiek-acl: What is this story demonstrating? (dont say "grouping")

packages/Checkbox/stories/examples/Showcase.js
```diff
@@ -0,0 +1,42 @@
...
+      <Heading level={1} displayLevel={2} isLight>
+        Showcase
+      </Heading>
+      <Tagline>Use the knobs to tinker with the props.</Tagline>
+      <Rule />
+      <Checkbox {...props} onChange={handleChange} checkedState={checkedState}>
+        Authentic VHS beard.
+      </Checkbox>
+      <Rule />
+      <Button onClick={setIndeterminate}>Set Indeterminate</Button>
```
jamiek-acl: Shouldn't this be a knob instead?

packages/Icon/src/Add.js
```diff
@@ -14,10 +14,10 @@ const SvgAdd = ({ title, ...props }) => (
     focusable={false}
     {...props}
   >
-    <title>{title}</title>
+    {title ? <title>{title}</title> : null}
```
jamiek-acl: Why'd all these Icons get updated in this PR?

tristanjasper: I'm no sure. I ran `npx lerna bootstrap` and it generated these changes in the js files. Even, so the changes are not for the worse as far as i can tell.

packages/Checkbox/src/Checkbox.styles.js
alexzherdev: @mikrotron didn't you add a `toInt` function somewhere?

mikrotron: @alexzherdev no, just the `spacer()`, which eliminated _most_ of the cases where I needed to use `parseInt()`.  `toInt()` is probably a better name though.

packages/Checkbox/stories/examples/CheckboxGrouping.js
mikrotron: I think this is a relic of some a rough story I had committed.  I wanted to see the line-height and vertical spacing between `<Checkboxes>` at different sizes.  It's not quite doing that anymore, but 🤷‍♂️.  I regard it as a "dev" story that we can file away, delete, or glean something good from later for a curated "Variations" story (or something).

packages/Icon/src/Add.js
mikrotron: I bet there was an update to `svgr`.

packages/Checkbox/stories/examples/Showcase.js
mikrotron: This distracted me. I got it to work (mostly) but found an issue in the component code as a result – when the `checkedState` changes _from_ indeterminate, and it is NOT the result of a user click, it should imperatively unset the `indeterminate` property of the `<input>`.

packages/Checkbox/stories/Checkbox.stories.styles.js
tristanjasper: We can refactor this out when we create some common storybook components and approaches

packages/Checkbox/stories/examples/CheckboxAlign.js
tristanjasper: Same as above - https://github.com/acl-services/paprika/pull/199/files#r340298417

packages/Checkbox/src/Checkbox.js
```diff
@@ -0,0 +1,81 @@
...
+  const inputRef = React.useRef(null);
+
+  React.useEffect(() => {
+    if (!inputRef.current) return;
+
+    if (checkedState === INDETERMINATE) {
+      inputRef.current.indeterminate = INDETERMINATE;
+    } else {
+      inputRef.current.indeterminate = false;
+      inputRef.current.checked = checkedState === CHECKED;
```
tristanjasper: @mikrotron Why do you think setting the `checked` value here is necessary? Won't it handled through https://github.com/acl-services/paprika/pull/199/files#diff-70cab65379b60a6be2550b5a72d611dfR61. I think only the `indeterminate` is required to be set manually

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,138 @@
...
+
+    & + label::before {
+      background: ${tokens.color.white};
+      border: 2px solid ${tokens.border.color};
+      border-radius: ${tokens.border.radius};
+      content: "";
+      height: ${checkBoxSize};
+      left: 0;
+      width: ${checkBoxSize};
+      z-index: 1;
```
jamiek-acl: do we still have the `z()` function?

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,138 @@
...
+    & + label > .checkbox-icon {
+      color: ${tokens.color.white};
+      font-size: ${iconSizeValue};
+      height: ${checkBoxSize};
+      left: ${checkBoxHalfSize};
+      opacity: 0;
+      pointer-events: none;
+      transform: translateX(-50%);
+      transition: opacity 0.15s ease-out;
+      z-index: 2;
```
jamiek-acl: `z()`

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,138 @@
...
+
+const labelSizeValue = ({ size }) => `${fontSizeValue(labelSizeMap[size])}px`;
+const iconSizeValue = ({ size }) => `${fontSizeValue(iconSizeMap[size])}px`;
+
+const labelPadding = ({ hasLabel, size }) => {
+  const top = `${labelTopOffsetMap[size]}px`;
+  const left = hasLabel ? `${toInt(getSpacing(size)) + toInt(tokens.space)}px` : getSpacing(size);
+  return `${top} 0 0 ${left}`;
+};
+
```
jamiek-acl: If everyone else is fine with it, then so be it, but I still dont like all this. Can't it just be something like:

```
switch(size) {
  case ShirtSizes.SMALL:
    $fontSize = font-scale();
    $paddingTop = '4';
    $height = '10';
    break;
  case ShirtSizes.MEDIUM:
    $fontSize = font-scale(1);
    $paddingTop = '0';
    $height = '14';
    break;
  case ShirtSizes.LARGE:
    $fontSize = font-scale(2);
    $paddingTop = '0';
    $height = '15';
    break;
}
```
Then use `fontSize, paddingTop, height` in the css below?

packages/Stylers/src/helpers.js
```diff
@@ -1,12 +1,19 @@
...
+
 //
 // Typography
 //
 
 const fontScaleRatio = 1.125; // 8:9 - Major second
 
-const fontSizeValue = (scale = 0) => Math.round(Number.parseInt(tokens.fontSize.default, 10) * fontScaleRatio ** scale);
+export const fontSizeValue = (scale = 0) =>
+  Math.round(Number.parseInt(tokens.fontSize.default, 10) * fontScaleRatio ** scale);
```
jamiek-acl: I also dont like this and wish it were just a map. Not relevant to this PR, though.

https://www.amazon.ca/Dont-Make-Think-Revisited-Usability/dp/0321965515/ref=pd_sbs_14_img_0/144-3293126-0814043?_encoding=UTF8&pd_rd_i=0321965515&pd_rd_r=005d21f3-1cf3-41ce-bef8-f44eb4b18598&pd_rd_w=oxoLl&pd_rd_wg=5HCng&pf_rd_p=9926bb69-42b9-46e4-b788-f665992e326d&pf_rd_r=C4WXAX75XRKMZV2PY9T3&psc=1&refRID=C4WXAX75XRKMZV2PY9T3

packages/Checkbox/src/Checkbox.styles.js
tristanjasper: I think for now we can go with the approach we have, perhaps refactor later if we think is really too complex. I'm not sure how your `switch would work` if we are needing those styles in on different elements either..

packages/Stylers/src/helpers.js
tristanjasper: Agreed, we can revisit in future but i'm a fan of simplification as well!

packages/Checkbox/src/Checkbox.styles.js
nahumzs: any reason to postpone the refactor @tristanjasper ? I think we could achieve the same either with a switch or an object and help with clarity. 

```js
const styles = {
     [ShirtSizes.SMALL]: {},
     [ShirtSizes.MEDIUM]: {},
     [ShirtSizes.LARGE]: {},
}
```

I will let @mikrotron and you to decide if worth to refactor it now or later.

tristanjasper: @nahumzs @jamiek-acl @mikrotron I've refactored. Let me know if any further changes required.

johnpeele:
@tristanjasper Can we add the `cursor: not-allowed;` to the disabled states? This is probably something that we should be doing for all components, if we are not already ;)

Also, did the `indeterminant` story/knob get lost somewhere?

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,176 @@
...
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import tokens from "@paprika/tokens";
+
+const getLabelLeftPadding = (checkboxSize, hasLabel) => {
+  return hasLabel ? `${toInt(checkboxSize) + toInt(tokens.space)}px` : checkboxSize;
+};
+
+const smallCheckboxSize = "16px";
+const mediumCheckboxSize = "20px";
+const largeCheckboxSize = "24px";
```
mikrotron: I'd rather see tokens used here (except for the `20px` value).

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,176 @@
...
+
+const styles = {
+  [ShirtSizes.SMALL]: {
+    baseFontSize: {
+      fontSize: "14px",
```
mikrotron: I'd rather see `fontSizeValue()` here (and below).

tristanjasper:
> @tristanjasper Can we add the `cursor: not-allowed;` to the disabled states? This is probably something that we should be doing for all components, if we are not already ;)
> 
> Also, did the `indeterminant` story/knob get lost somewhere?

@johnpeele Indeterminant state can be set under the `checkedState` knob in storybook
image

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -0,0 +1,176 @@
...
+
+    & + label:hover::before {
+      border: 2px solid ${tokens.color.black};
+    }
+
+    & + label > .checkbox-icon {
+      color: ${tokens.color.white};
+      ${({ size }) => {
+        return styles[size].checkBoxIconStyles;
+      }}
```
mikrotron: This would be a little more succinct as:
```js
${({ size }) => styles[size].checkBoxIconStyles;
```

packages/Checkbox/src/Checkbox.styles.js
tristanjasper: @mikrotron why no token for the 20px?

packages/Checkbox/src/Checkbox.styles.js
jamiek-acl: I think that's more clear

mikrotron:
🦙 🐐
LGTM
🦖 🐒

PR #203, opened 2019-09-23
ux-586-datepicker-prevent acl-ui3 style overrides

tristanjasper:
### 🛠 Purpose

Use the Paprika dropdown in a module that also uses acl-ui 3 buttons. Eg in results questionnaires.

Some styles are being overridden by acl-ui 3 styles.

Fixes include
Ensure Paprika buttons are non-semantic
Ensure component has a background colour
Ensure label is displaying correct styles for flex layout

## Screenshots / Gifs / Codepens
Before:
image

image

---
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/ux-586-datepicker-prevent-aclui-style-overrides' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/ux-586-datepicker-prevent-aclui-style-overrides</a>

### DIFF
--- a/packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.js
+++ b/packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.js
@@ -119,11 +119,11 @@ function ShortcutPanel(props) {
         </div>
         <div css={yearListStyles}>
           <div css={columnHeaderStyles} isYear>
-            <Button.Icon onClick={handleClickPrev} kind="minor" size="small">
+            <Button.Icon isSemantic={false} onClick={handleClickPrev} kind="minor" size="small">
               <ArrowLeft color={tokens.textColor.icon} />
             </Button.Icon>
             {I18n.t("datePicker.year")}
-            <Button.Icon onClick={handleClickNext} kind="minor" size="small">
+            <Button.Icon isSemantic={false} onClick={handleClickNext} kind="minor" size="small">
               <ArrowRight color={tokens.textColor.icon} />
             </Button.Icon>
           </div>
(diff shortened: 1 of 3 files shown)


packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.js
```diff
@@ -119,22 +119,34 @@ function ShortcutPanel(props) {
...
           {I18n.t("actions.apply")}
         </Button>
-        <Button kind="minor" onClick={onCancel} size="small" data-pka-anchor="datepicker.calendar.cancel">
+        <Button
+          isSemantic={false}
+          kind="minor"
+          onClick={onCancel}
+          size="small"
+          data-pka-anchor="datepicker.calendar.cancel"
+        >
```
tristanjasper: Just changed all buttons to use `isSemantic={false}` in this file

packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.styles.js
```diff
@@ -77,6 +77,12 @@ export const containerStyles = css`
     margin: 0;
     overflow: hidden;
     padding: 5px ${tokens.spaceLg} ${tokens.spaceSm} ${tokens.spaceLg};
+
+    /* prevent acl-ui3 label style overrides */
+    input[type="radio"] + label {
+      display: flex;
+      margin: auto;
+    }
```
tristanjasper: I kind of feel we should include a global style in all of our paprika components (or perhaps only the ones that use inputs with labels) to deal with this instead of adding this style rule in specific locations throughout our paprika components. Saying that though inputs/labels won't always need to be presented in a flex layout.... 🤔

packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.js
tristanjasper: I know we've discussed setting the `isSemantic={false}` in a global hook for our styled components  (so that it is a default) however until that is setup i'm setting it on the individual buttons

packages/DatePicker/src/components/ShortcutPanel/ShortcutPanel.styles.js
allison-c: I don't know what's the best solution. I think we can go with your fix. We can also use `styled.label` here to avoid overrides.
Something like:
```js
const StyledLabel = styled.label`
  ...
  // styles are not going to be override
`;

export const containerStyles = css`
  ...
  StyledLabel {
    ...
    // Styles are going to be override
  }
`;
```

tristanjasper: ok, yeah i think for now i'll leave it with the css attribute as this is the common approach we are taking. Perhaps we will change again later on

PR #206, opened 2019-09-25
XD-5802-Adding onError prop to DatePicker

tristanjasper:
### 🛠 Purpose

Allow the consuming application to know an invalid date string has been parsed in the component so that it can react accordingly displaying error messages if required. (Needed in results)

@nahumzs minor version bump required

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/XD-5802-add-date-picker-parse-error' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/XD-5802-add-date-picker-parse-error</a>

### DIFF
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -170,7 +170,7 @@ function DatePicker(props) {
     } else {
       setConfirmationResult("");
       setHasParsingError(true);
-      onDateParseError();
+      onDateParseError(inputtedString);
     }
   }
 
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -209,7 +209,7 @@ function DatePicker(props) {
     handleChange(selectedDate);
   }
 
-  const hasError = extendedInputProps.hasError || hasParsingError;
+  const { hasError, ...moreExtendedInputProps } = extendedInputProps;
 
   return (
     <Popover
@@ -221,6 +221,7 @@ function DatePicker(props) {
       shouldKeepFocus
     >
       <Input
+        hasError={hasError || hasParsingError}
         icon={<CalendarIcon />}
         id={id}
         isDisabled={isDisabled}
@@ -232,8 +233,7 @@ function DatePicker(props) {
         onKeyUp={handleKeyUp}
         inputRef={inputRef}
         value={confirmationResult || inputtedString}
-        {...extendedInputProps}
-        hasError={hasError}
+        {...moreExtendedInputProps}
       />
 
       <Popover.Content>
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -209,7 +209,7 @@ function DatePicker(props) {
     handleChange(selectedDate);
   }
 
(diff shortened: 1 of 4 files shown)


packages/DatePicker/src/DatePicker.js
```diff
@@ -165,6 +170,7 @@ function DatePicker(props) {
     } else {
       setConfirmationResult("");
       setHasParsingError(true);
+      onDateParseError();
```
allison-c: Do you need any information for example `inputtedString`?

tristanjasper: Hmm, not in my case but perhaps someone else _could_? I'm don't think they would but i can pass it as well.

tristanjasper: I've changed just in case someone may want to display it somehow

nahumzs:
can you elaborate on why we need it in one of our features? having context either on the comments or via the PR comments helps future devs to find possible answers to their questions. thanks @tristanjasper

packages/DatePicker/src/DatePicker.js
```diff
@@ -44,6 +44,9 @@ const propTypes = {
 
   /** Callback when date is selected or input. */
   onChange: PropTypes.func.isRequired,
+
+  /** Callback when input string is not correctly parsed as a date . */
+  onDateParseError: PropTypes.func,
```
nahumzs: what about having `onError()` 
```
<DatePicker onError={({ input, errorMessage }) => {
   // do whatever with the error.
}}
```

I think onError is more reusable pattern across our components and we could communicate what kind of error is by passing an object describing the error.

tristanjasper: Sure but we already have a `hasError` prop (which is quite similar) in this component so I didn't want to change that too much at this point. I agree that `onError` would be preferable. Perhaps we can rename `hasError` prop to `hasConsumerError` or `hasExternalError`? @allison-c @nahumzs wdyt?

nahumzs: `hasError` is a prop that our inputs (any kind) use, so I wouldn't suggest renaming that one. I guess you would want to manipulate that one when an event onError occur. So I think is safe to use onError. isn't?

nahumzs: `hasError` will only describe if the input has a red border or red font as an example right?

tristanjasper: Ok yeah I just think it is quite easy for someone to confuse the two `onError` and `hasError` some wanted to be more declarative. I think the 'hasError` prop should really be on the `DatePicker.Input` component only and not on the `DatePicker` component? At least that would be less confusing?

packages/DatePicker/src/DatePicker.js
```diff
@@ -231,6 +233,7 @@ function DatePicker(props) {
         inputRef={inputRef}
         value={confirmationResult || inputtedString}
         {...extendedInputProps}
+        hasError={hasError}
```
tristanjasper: @allison-c I'm wondering if it would be better to render the paprika input component in the DatePicker.Input component instead of it just being a shell component? Perhaps something to consider in the future

PR #207, opened 2019-09-26
XD-5802-Prevent DatePicker calendar flicker on close

tristanjasper:
### 🛠 Purpose

Prevent flicker of Datepicker calendar on close.

### ✏️ Notes to Reviewer

Giphy captures below not showing flicker very well, better to run locally in storybook to see change

---

### 🖥 Screenshots
Context action (closing calendar)
![flicker-before](https://user-images.githubusercontent.com/4040102/65713573-5040b680-e04e-11e9-9da8-95aee4ddcf1e.
 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/xd-5802-datepicker-prevent-flicker' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/xd-5802-datepicker-prevent-flicker</a>

PR #210, opened 2019-09-27
Xd 5802 date picker input focus on calendar hidden

tristanjasper:
### 🛠 Purpose

Previous approach of preventing date picker calendar flicker was causing the input to lose focus.
This Pr takes a different approach to the problem using an `opacity:0` style to prevent the calendar flash.

Additionally fixes a background colour style not being applied correctly on shortcut panel

@nahumzs Minor bump only

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots
Before
flicker-input-focus-before

After
flicker-input-focus-fix

 
 ### 📙 Storybook 
 <a href='http://storybooks.highbond-s3.com/paprika/XD-5802-DatePicker-Input-focus-on-calendar-hidden' target="_blank" rel="noopener">http://storybooks.highbond-s3.com/paprika/XD-5802-DatePicker-Input-focus-on-calendar-hidden</a>


packages/DatePicker/src/DatePicker.js
```diff
@@ -244,19 +238,22 @@ function DatePicker(props) {
...
-            />
-          </div>
-        </Popover.Content>
-      ) : null}
+      <Popover.Content>
+        <div
+          css={calendarPopoverStyles}
+          data-pka-anchor="datepicker.calendar"
+          ref={calendarRef}
+          shouldShowCalendar={shouldShowCalendar}
```
tristanjasper: This file is now reverted back to its original logic pre https://github.com/acl-services/paprika/pull/207.
Only change now is that  I am now passing `shouldShowCalendar` prop instead in order to set a style

packages/DatePicker/src/DatePicker.styles.js
```diff
@@ -2,6 +2,12 @@ import { css } from "styled-components";
 
 import tokens from "@paprika/tokens";
 
+const opacityNoneStyles = css`
+  opacity: 0;
```
tristanjasper: Needs to be opacity rather than `visibility` in order for the click to propagate correct and the input getting focus

PR #213, opened 2019-09-30
Ux 588 list box browser

nahumzs:
## Storybook link

http://storybooks.highbond-s3.com/paprika/UX-588-ListBoxBrowser

## About the API decision

Some information about the API decision.
Regularly APIs that reassemble html are favour in Paprika over javascript objects, but for this component, which is recursive, I'm favouring the use of a javascript object.

The main reason for my decision is because javascript objects are easier to read when comparing it with nested html elements.

As well for this specific case is simpler to shape a javascript array that matches the required data structure than map over nested elements to create the data structure. 

Finally is also easier to parse a simple javascript array than nested JSX children.

```js
const data = [{
    label: "one",  
    options: [{ 
         label: "two", 
	       options: [{ label: "three" }] 
     }] 
}]
// vs
/* hypothetical API */
<ListBoxBrowser>
   <ListBoxBrowser.Option label="one">
       <ListBoxBrowser.Option label="two">
	       <ListBoxBrowser.Option label="three" />
       </ListBoxBrowser.Option>
   </ListBoxBroswer.Option>
</ListBoxBrowser>
```

---

# ListBoxBrowser

It is a React component that allows users to select one or multiple options using two columns, similar to the experience of browsing a file manager in any giving Operative System.

## Installation

```sh
$ npm install @paprika/listbox-browser
```

or

```sh
$ yarn add @paprika/listbox-browser
```

## Props

```
data
isMulti
height
onChange
isParentSelectable
rootTitle
browserTitle
children
hasBreadcrumb
defaultSelectedOptions
defaultSelectedView
```

## Usage

The `<ListBoxBrowser />` is primarily an uncontrolled component which received a `data` prop allowing the consumer to set the initial state.

Also, has two functions to set defaults for selected options and a default selected view for the browser column.

The `onChange` prop receives a parameter with an array of the selected options.

### Data prop

The most important prop for the `<ListBoxBrowser />` is the `data` prop, which initializes the state for the component.

The `data` prop has a shape of an array of objects where each object requires to have at least a `label` attribute on it, and at least one of those items should include an `options` property.

### Data prop shape

```js
{
	label: "require attribute",
	options: [{}{}{}], // more object with the same shape
}
```

The **object** can have an options attribute, which is an array of the same kind of objects.

A data prop with **options** looks like:

```js
const data = [
   {
        label: "NBA",
        options: [{ label: "Toronto Raptors"}, {...}]
   },
   {
        label: "NHL",
        options: [{ label: "Vancouver Canucks"}, {...}]
   },
   {
        label: "MLS",
        options: [{ label: "Vancouver Whitecaps"}, {...}]
   },
]
```

Recursively the component runs over all the options on the array and assembles the appropriated UI for the user.

**NOTE** You can add extra properties to the objects; those are and pass to you back on the `onChange` function or when using `defaultSelectedOptions`, `defaultSelectedView` functions.

---

#### Basic example

```js
import ListBoxBrowser from "@paprika/listbox-browser";

   const data = [{...}...{...}]; // your data;
   return <ListBoxBrowser data={data} />
}
```

---

#### ListBoxBrowser.SelectedOptions

By default, the `<ListBoxBrowser />` does not display a list of the selected options.  
But you can activate this feature to display multiple options using the `<ListBoxBrowser.SelectedOptions />` subcomponent.

```js
<ListboxBrowser data={data}>
   <ListBoxBrowser.SelectedOptions />
</ListBoxBrowser>
```

#### defaultSelectedOptions

The `defaultSelectedOptions` received a function to be executed to determine if an option should start as selected or not. This function is called as many times as options are pass down on the data prop.

Ex.

```js
const data = [
	{ key: 1, label: "one", options: [{...}]},
	{ key: 2, label: "two", options: [{...}]},
	...
]

<ListBoxBrowser
	data={data}
	defaultSelectedOptions={(option) =>  (option.key === 1)}  />

/**
  The previous result on option with `key 1` to be selected */
```

#### defaultSelectedView

This prop worked exactly like the previous one, receiving a function which you can use to compare and decide what option should be the one as initial view.

**Note**: Be sure to selected an `option` property with `options`; otherwise, this function doesn't work and fall back to the default behaviour, which is to select the first option with options.

### Lazy loading

This component provides a way to load async options. To achieve this, you should declare the `options` array property empty `[]`, this indicates to the component to fire the `onFetch={option => ()}` prop when the user interacts with it.

#### Lazy loading example

To help you to sync your data with the newest options, the ListBoxBrowser provides you with some tools to make it easier:

A Subcomponent `<ListBoxBrowser.Browser isLoading />` which lets you set the right column into a pending state.
And a `ListBoxBrowser.findOption(data, fn)` which helps you in finding an option in your data so you can modify it.

```js
export default function App(props) {
  const [isBrowserLoading, setIsBrowserLoading] = React.useState(false);
  const [data, setData] = React.useState(props.defaultData);

  async function handleFetch(option) {
    setIsBrowserLoading(() => true);
    const key = option.key;
    const response = await fakeFetch({ key });

    setData(data => {
      const cloneData = data.splice(0);
      const o = ListBoxBrowser.findOption(cloneData, option => option.key === key);
      o.options = response;

      return cloneData;
    });
    setIsBrowserLoading(() => false);
  }

  return (
    <ListBoxBrowser onFetch={handleFetch} data={data} rootTitle="Universes" browserTitle="Heroes">
      <ListBoxBrowser.Browser isLoading={isBrowserLoading} />
      <ListBoxBrowser.OptionsSelected />
    </ListBoxBrowser>
  );
}
```

## Images

Screen Shot 2019-10-17 at 9 37 20 AM
Screen Shot 2019-10-17 at 9 37 28 AM
Screen Shot 2019-10-17 at 9 38 20 AM
Screen Shot 2019-10-17 at 9 38 34 AM

### No address on this PR

- Will be a need for a search and we have a design for it but there is not implementation currently
- We might need in the future a renderOption prop for custom render each option, currently there is not need
- IsDisabled status, we don't have a current need for that state but we already have a design for it
- There is a flickering happening when it is a single select, this is because of the inherent behaviour from the ListBox but can be solved in other PR.
- It is not left and right key implementation, though you can navigate correctly the component with tab and shift, we need to expand the ListBox to support this behaviour so this component can be implemented.


tristanjasper:
nit: image

Perhaps ensure the div with focus z index is higher?

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
+# ListBoxBrowser
+
+It is a React component that allows users to select one or multiple options using two columns, similar to the experience of browsing a file manager in any giving Operative System.
```
tristanjasper: "any given operating system"

packages/ListBoxBrowser/src/helpers.js
```diff
@@ -0,0 +1,259 @@
...
+        };
+      } else {
+        refDefaultSelectedOptions.current = { ...refDefaultSelectedOptions.current, [option.parent]: [option] };
+      }
+      /* eslint-enable no-param-reassing */
+    }
+
+    // checking if its initial view
+    if (_hasOptions && defaultSelectedView && defaultSelectedView(option.attributes)) {
+      refDefaultSelectedView.current = newPath;
```
tristanjasper: what happens if there are more than one option returned, it will just set the default view to the latest iterated option?, perhaps if this is set more than once based on the data there should be an error... 🤷‍♂

packages/ListBox/src/components/Trigger/Trigger.styles.js
```diff
@@ -27,6 +27,12 @@ const triggerStyles = `
 `;
 
 const triggerStylesProps = () => props => {
+  if (props.isHidden) {
+    return css`
+      border: 1px solid #d7d7d7;
```
tristanjasper: Perhaps add a token for this hex value, seems to be currently be used in 4 paprika components

packages/ListBox/src/hooks/hooks.js
```diff
@@ -77,7 +77,9 @@ export function useAdjustWidth() {
     }
 
     const handleResize = () => {
-      dispatch({ type: useListBox.types.setTriggerWidth, payload: $triggerContainer.offsetWidth });
+      if ($triggerContainer) {
+        dispatch({ type: useListBox.types.setTriggerWidth, payload: $triggerContainer.offsetWidth });
+      }
```
tristanjasper: move lines 80-82 to a function? The same thing being called on line 75

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+  */
+  defaultSelectedView: PropTypes.func,
+};
+
+const defaultProps = {
+  browserTitle: "",
+  children: null,
+  hasBreadcrumb: true,
+  hasError: false,
+  height: 220,
```
tristanjasper: track the height in a variable?

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+        return;
+      }
+      handleClickBrowser({ $$key: option.parent, hasOptions: false, isClickFromButton: true })();
+    },
+    [handleClickBrowser, handleClickRoot]
+  );
+
+  const onRemove = React.useCallback(function handleRemove(option) {
+    setSelectedOptions(selectedOptions => {
+      console.log(option);
```
tristanjasper: rm console

packages/ListBoxBrowser/src/components/Breadcrumb/Breadcrumb.styles.js
```diff
@@ -0,0 +1,8 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens/lib/tokens";
+
+export const button = css`
+  color: ${tokens.color.blue};
+  font-size: 14px;
```
tristanjasper: `${stylers.fontSize(-1)};`

packages/ListBoxBrowser/src/components/CustomListBox/CustomListBox.styles.js
```diff
@@ -0,0 +1,100 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens/lib/tokens";
+import stylers from "@paprika/stylers";
+
+export const arrowRigthContainer = css`
```
tristanjasper: arrowRightContainer

packages/ListBoxBrowser/src/components/CustomListBox/CustomListBox.styles.js
```diff
@@ -0,0 +1,100 @@
...
+`;
+
+export const loading = css`
+  align-items: center;
+  border: 0;
+  border-left: 1px solid ${tokens.border.color};
+  display: flex;
+  flex-basis: 50%;
+  justify-content: center;
+  min-height: 220px;
```
tristanjasper: pass in a height prop for this value?

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.js
```diff
@@ -0,0 +1,96 @@
...
+  remove,
+  counter,
+  optionLabel,
+  optionBreadcrum,
+  noOptionsSelectedContainer,
+  noOptionsSelected,
+} from "./OptionsSelected.styles";
+
+function getLength(options) {
+  let i = 0;
```
tristanjasper: perhaps use lodash size function instead of this?

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
```diff
@@ -0,0 +1,94 @@
...
+  flex-basis: 100%;
+  flex-grow: 1;
+  padding: ${tokens.spaceSm} ${tokens.space};
+`;
+
+export const counter = css`
+  background: ${tokens.color.white};
+  border: 1px solid ${tokens.border.color};
+  border-radius: ${tokens.border.radius};
+  font-size: 13px;
```
tristanjasper: use stylers helper

packages/ListBox/src/components/Content/Content.js
```diff
@@ -53,6 +53,7 @@ export default function Content(props) {
         {...getDOMAttributesForListBoxContainer({ isInline: true })}
         onKeyDown={handleKeyboardKeys(state, dispatch)}
         ref={refListBoxContainer}
+        data-ppk-anchor="listbox-content-inline"
```
oscarkwan:
```suggestion
        data-pka-anchor="listbox-content-inline"
```

packages/ListBoxBrowser/README.md
oscarkwan: `any given operating system`

oscarkwan: Don't think we need to start off the sentence with "It is a React component".

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+height
+isMulti
+isParentSelectable
+onChange
+rootTitle
+```
+
+## Usage
+
+The `<ListBoxBrowser />` is primarily an uncontrolled component which received a `data` prop allowing the consumer to set the initial state.
```
oscarkwan:
```suggestion
The `<ListBoxBrowser />` is primarily an uncontrolled component which receives a `data` prop which allows the consumer to set the initial state.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+Also, has two functions to set defaults for selected options and a default selected view for the browser column.
```
oscarkwan:
```suggestion
It has two functions to set defaults. One for selected options and another for a default selected view for the browser column.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+The `onChange` prop receives a parameter with an array of the selected options.
+
+### Data prop
+
+The most important prop for the `<ListBoxBrowser />` is the `data` prop, which initializes the state for the component.
+
+The `data` prop has a shape of an array of objects where each object requires to have at least a `label` attribute on it, and at least one of those items should include an `options` property.
```
oscarkwan:
```suggestion
The `data` prop is an array of objects where each object requires to have at least a `label` attribute on it and at least one of those items should include an `options` property.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+        options: [{ label: "Vancouver Canucks"}, {...}]
+   },
+   {
+        label: "MLS",
+        options: [{ label: "Vancouver Whitecaps"}, {...}]
+   },
+]
+```
+
+Recursively the component runs over all the options on the array and assembles the appropriated UI for the user.
```
oscarkwan:
```suggestion
Recursively the component runs over all the options on the array and assembles the UI for the user.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+**NOTE** You can add extra properties to the objects; those are and pass to you back on the `onChange` function or when using `defaultSelectedOptions`, `defaultSelectedView` functions.
```
oscarkwan: This one i'm not really sure about the grammar on this part
"those are and pass to you back on the `onChange` function or when using `defaultSelectedOptions`, `defaultSelectedView` functions.".

Maybe pro @alexzherdev can chime in with his english skills :)

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+```js
+<ListboxBrowser data={data}>
+   <ListBoxBrowser.SelectedOptions />
+</ListBoxBrowser>
+```
+
+#### defaultSelectedOptions
+
+The `defaultSelectedOptions` received a function to be executed to determine if an option should start as selected or not. This function is called as many times as options are pass down on the data prop.
```
oscarkwan:
```suggestion
The `defaultSelectedOptions` receives a function to be executed to determine if an option should start as selected or not. This function is called as options are passed down on the data prop.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+	data={data}
+	defaultSelectedView={(option) =>  (option.key === 1)}  />
+
+/**
+  The previous result on option with `key 1` to be selected */
+```
+
+#### defaultSelectedView
+
+This prop worked exactly like the previous one, receiving a function which you can use to compare and decide what option should be the one as initial view.
```
oscarkwan:
```suggestion
This prop works exactly like the `defaultSelectedOptions`. It receives a function which you can use to compare and decide what option should be the one as initial view.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+**Note**: Be sure to selected an `option` property with `options`; otherwise, this function doesn't work and fall back to the default behaviour, which is to select the first option with options.
```
oscarkwan:
```suggestion
**Note**: Be sure to select an `option` property with `options`. Otherwise, this function doesn't work and falls back to the default behaviour, which selects the first option with options.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Lazy loading
+
+This component provides a way to load async options. To achieve this, you should declare the `options` array property empty `[]`, this indicates to the component to fire the `onFetch={option => ()}` prop when the user interacts with it.
```
oscarkwan:
```suggestion
This component provides a way to load async options. To achieve this, you can declare the `options` array property empty `[]`. This indicates to the component to fire the `onFetch={option => ()}` prop when the user interacts with it.
```

packages/ListBoxBrowser/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### Lazy loading example
+
+To help you to sync your data with the newest options, the ListBoxBrowser provides you with some tools to make it easier:
+
+A Subcomponent `<ListBoxBrowser.Browser isLoading />` which lets you set the right column into a pending state.
+And a `ListBoxBrowser.findOption(data, fn)` which helps you in finding an option in your data so you can modify it.
```
oscarkwan:
```suggestion
A `ListBoxBrowser.findOption(data, fn)` which helps you in finding an option in your data so you can modify it.
```

packages/ListBoxBrowser/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@paprika/listbox-browser",
+  "version": "0.1.0",
+  "description": "ListBox Browser component let you select options in a multi level env",
```
oscarkwan:
```suggestion
  "description": "ListBox Browser component lets you select options in a multi level environment",
```

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+const propTypes = {
+  /**
+    An array of javascript objects holding the data structure for the ListBoxBrowser
+    The object shape **must have** at least a string **label** property and an optional array **options** property.
+    Also can hold any other kind of data for your own use.
+    ex.
+
+    [{ label: "One" }, { label: "Two", options: [{ label: "Three" }] }]
+  */
+  data: PropTypes.any.isRequired, // eslint-disable-line
```
oscarkwan: Should it be `any`? Is there a way we can use prop-types to enforce (must have a label).

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+import OptionsSelected from "./components/OptionsSelected";
+import CustomListBox from "./components/CustomListBox";
+import Root from "./components/Root";
+import Browser from "./components/Browser";
+import Title from "./components/Title";
+
+const propTypes = {
+  /**
+    An array of javascript objects holding the data structure for the ListBoxBrowser
+    The object shape **must have** at least a string **label** property and an optional array **options** property.
```
oscarkwan:
```suggestion
    The object shape **must have** at least a string **label** property and an array **options** property in one of the objects.
```

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+    Also can hold any other kind of data for your own use.
+    ex.
+
+    [{ label: "One" }, { label: "Two", options: [{ label: "Three" }] }]
+  */
+  data: PropTypes.any.isRequired, // eslint-disable-line
+
+  /**
+    Indicates if the user can select one or multiple options
```
oscarkwan:
```suggestion
    Indicates if the user can select multiple options
```

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+  /**
+    Indicates if the right column should display a breadcrumb
+  */
+  hasBreadcrumb: PropTypes.bool,
+  /**
+    Set a border red color around the component indicating that has an error
+  */
+  hasError: PropTypes.bool,
+  /**
+    When declaring the array options empty, this will be execute to retrive the
```
oscarkwan:
```suggestion
    When declaring the array options empty, this will be executed to retrieve the
```

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+    data, useful if you want to do a lazy load.
+
+    ex.
+    <ListBoxBrowser data={[{ label: "lazy", options: [] }]} onFetch={(option) => {** logic**}}>
+      <ListBoxBrowser.Browser isLoading={isBrowserLoading} />
+    </ListBoxBrowser>
+  */
+  onFetch: PropTypes.func,
+  /**
+    A function that set an option selected returning true or false
```
oscarkwan:
```suggestion
    A function that sets an option selected returning true or false
```

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -0,0 +1,355 @@
...
+    const data = [{ key: 1, label: "one", options: [...] }, { key: 2, label: "two" }, { key: 3, label: "three" }]
+
+    ex.
+    <ListBoxBrowser data={data} defaultSelectedOptions={(option) => {
+      return option.key === 2 or option.key === 3
+    }} />
+  */
+  defaultSelectedOptions: PropTypes.func,
+  /**
+    A function that set the intial view for the right columns (Browser) of the ListBoxBrowser
```
oscarkwan:
```suggestion
    A function that sets the initial view for the right columns (Browser) of the ListBoxBrowser
```

packages/ListBoxBrowser/src/components/CustomListBox/CustomListBox.styles.js
```diff
@@ -0,0 +1,99 @@
+import { css } from "styled-components";
+import tokens from "@paprika/tokens/lib/tokens";
+import stylers from "@paprika/stylers";
+
+export const arrowRigthContainer = css`
+  display: flex;
+  height: ${stylers.spacer(4)};
+  justify-content: flex-end;
+  position: absolute;
+  right: -8px;
```
oscarkwan:
```suggestion
  right: -${tokens.space};
```

packages/ListBoxBrowser/src/components/CustomListBox/CustomListBox.styles.js
```diff
@@ -0,0 +1,99 @@
...
+  justify-content: center;
+  width: 100%;
+
+  &:hover:after {
+    background: ${tokens.color.black};
+    border-radius: 50%;
+    content: "";
+    height: ${stylers.spacer(3)};
+    left: 50%;
+    margin-left: -12px;
```
oscarkwan:
```suggestion
    margin-left: -${tokens.spaceLg};
```

packages/ListBoxBrowser/src/components/CustomListBox/CustomListBox.styles.js
```diff
@@ -0,0 +1,99 @@
...
+    margin-top: -12px;
```
oscarkwan:
```suggestion
    margin-top: -${tokens.spaceLg};
```

packages/ListBoxBrowser/src/components/CustomListBox/CustomListBox.styles.js
```diff
@@ -0,0 +1,99 @@
...
+export const checkbox = css`
+  position: relative;
+  top: -2px;
+`;
+
+export const backButton = css`
+  align-items: center;
+  cursor: pointer;
+  display: inline-flex;
+  font-size: 14px;
```
oscarkwan:
```suggestion
  ${stylers.fontSize(-1)}
```

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.js
```diff
@@ -0,0 +1,96 @@
...
+        {Object.keys(selectedOptions).map(key => {
+          return selectedOptions[key].map(option => {
+            const breadcrumb = getBreadcrumb({ data: localData, option });
+            return (
+              <div key={option.$$key} css={optionStyles} data-ppk-anchor="listbox-browser-selected-options">
+                <RawButton css={button} onClick={handleClick(option)}>
+                  <div css={optionLabel} data-ppk-anchor="listbox-browser-selected-options-label">
+                    {option.attributes.label}
+                  </div>
+                  <div css={optionBreadcrum}>
```
oscarkwan: Spelling (optionBreadcrumb), and all other places that uses this

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
```diff
@@ -0,0 +1,96 @@
...
+  align-items: center;
+  background: ${tokens.backgroundColor.white};
+  border: 1px solid ${tokens.border.color};
+  border-radius: ${tokens.border.radius};
+  box-sizing: border-box;
+  color: ${tokens.color.black};
+  display: flex;
+  flex-basis: 100%;
+  flex-grow: 1;
+  font-size: 14px;
```
oscarkwan:
```suggestion
  ${stylers.fontSize(-1)}
```

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
```diff
@@ -0,0 +1,96 @@
...
+  justify-content: flex-start;
+  margin: ${tokens.spaceSm};
+`;
+
+export const title = css`
+  background: ${tokens.backgroundColor.level0};
+  border-bottom: 1px solid ${tokens.border.color};
+  border-top: 1px solid ${tokens.border.color};
+  box-sizing: border-box;
+  font-size: 14px;
```
oscarkwan:
```suggestion
  ${stylers.fontSize(-1)}
```

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
```diff
@@ -0,0 +1,96 @@
...
+  font-weight: bold;
+  padding: ${spacer(0.5)} ${spacer(1)} ${spacer(0.5)} ${spacer(1.5)};
```
oscarkwan:
```suggestion
  padding: ${tokens.spaceSm} ${tokens.space} ${tokens.spaceSm} ${tokens.spaceLg};
```

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
oscarkwan: yeesss

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
```diff
@@ -0,0 +1,96 @@
...
+`;
+
+export const optionBreadcrum = css`
+  color: ${tokens.color.blackLighten20};
+  font-size: 13px;
+`;
+
+export const noOptionsSelectedContainer = css`
+  background: ${tokens.backgroundColor.level0};
+  padding: ${spacer(1)};
```
oscarkwan:
```suggestion
  padding: ${tokens.space};
```

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
```diff
@@ -0,0 +1,96 @@
...
+`;
+
+export const noOptionsSelected = css`
+  background: ${tokens.color.white};
+  border: 1px dashed ${tokens.border.color};
+  color: ${tokens.textColor.subtle};
+  font-style: italic;
+  padding: ${spacer(1)};
```
oscarkwan:
```suggestion
  padding: ${tokens.space};
```

packages/ListBoxBrowser/src/components/Title/Title.js
```diff
@@ -0,0 +1,47 @@
...
+import { ListBoxBrowserContext } from "../../ListBoxBrowser";
+
+const propTypes = {
+  rootTitle: PropTypes.node.isRequired,
+  browserTitle: PropTypes.node.isRequired,
+  data: PropTypes.arrayOf(PropTypes.object), // eslint-disable-line
+  browserKey: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+  onClickBreadcrumb: PropTypes.func.isRequired,
+};
+const defaultProps = {};
```
oscarkwan: ✂️

packages/ListBoxBrowser/src/components/Title/Title.styles.js
```diff
@@ -0,0 +1,26 @@
...
+import { spacer } from "@paprika/stylers/lib/helpers";
+
+export const title = css`
+  background: ${tokens.backgroundColor.level0};
+  border-bottom: 1px solid ${tokens.border.color};
+  border-top-left-radius: ${tokens.border.radius};
+  border-top-right-radius: ${tokens.border.radius};
+  flex-basis: 50%;
+  flex-grow: 1;
+  font-size: 14px;
```
oscarkwan:
```suggestion
  ${stylers.fontSize(-1)}
```

packages/ListBoxBrowser/src/components/Title/Title.styles.js
```diff
@@ -0,0 +1,26 @@
...
+  font-weight: bold;
+  min-height: 20px;
+  padding: ${spacer(0.5)} ${spacer(1)} ${spacer(0.5)} ${spacer(1.5)};
```
oscarkwan:
```suggestion
  padding: ${tokens.spaceSm} ${tokens.space} ${tokens.spaceSm} ${tokens.spaceLg};
```

packages/ListBoxBrowser/src/helpers.js
```diff
@@ -0,0 +1,259 @@
...
+  data,
+  path = "root",
+  refDefaultSelectedOptions,
+  defaultSelectedOptions,
+  defaultSelectedView,
+  refDefaultSelectedView,
+}) {
+  return data.map(({ label = null, options, ...moreAttributes }, index) => {
+    if (!label) {
+      throw new Error("A Label attribute is required for each option object, can't process data.");
```
oscarkwan: Oh i see.. you validate here, you can ignore my prop-types comment.

packages/ListBoxBrowser/src/helpers.js
```diff
@@ -0,0 +1,259 @@
...
+  });
+}
+
+export function getOptionByKey(data, path) {
+  let node = null;
+  function runner(data, path) {
+    return data.map(option => {
+      if (!option.$$key.includes("/") && (typeof data[path] !== "undefined" && option.$$key === path)) {
+        node = option;
+        throw new Error("option found");
```
oscarkwan: Something more descriptive for the error would be nice :)

packages/ListBoxBrowser/src/helpers.js
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  setSelectedOptions(selectedOptions => {
+    return { ...selectedOptions, [browserKey]: getOptions(indexes, list) };
+  });
+}
+
+export function focusListBoxBrowser() {
+  window.requestAnimationFrame(() => {
+    document.querySelectorAll('[data-ppk-anchor="listbox-content-inline"]')[1].focus();
```
oscarkwan: I wonder what if in a future scenario, we have two listbox browsers on the same page. Will there be issues then with this?

packages/ListBoxBrowser/stories/ListBoxBrowser.multiple.stories.js
```diff
@@ -0,0 +1,124 @@
...
+      <ListBoxBrowser
+        data={issues}
+        rootTitle="Phase"
+        onChange={selectedOptions => {
+          console.log("selected options:", selectedOptions);
+        }}
+      />
+    </Story>
+  ))
+  .add("Allowing select parent", () => (
```
oscarkwan:
```suggestion
  .add("Allows selecting parent", () => (
```

packages/ListBoxBrowser/stories/ListBoxBrowser.multiple.stories.js
```diff
@@ -0,0 +1,124 @@
...
+        browserTitle="Heroes"
+        onChange={selectedOptions => {
+          console.log("selected options:", selectedOptions);
+        }}
+      >
+        <ListBoxBrowser.OptionsSelected />
+      </ListBoxBrowser>
+    </Story>
+  ))
+  .add("Has not breadcrumb", () => (
```
oscarkwan:
```suggestion
  .add("Has no breadcrumb", () => (
```

packages/ListBoxBrowser/stories/examples/Lazy.js
```diff
@@ -0,0 +1,64 @@
+import React from "react";
+import { Story } from "storybook/assets/styles/common.styles";
+import ListBoxBrowser from "../../src";
+
+const propTypes = {};
+const defaultProps = {};
```
oscarkwan: ✂️

packages/ListBoxBrowser/src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,33 @@
...
+import PropTypes from "prop-types";
+import RawButton from "@paprika/raw-button";
+import { button } from "./Breadcrumb.styles";
+import { isRoot } from "../../helpers";
+
+const propTypes = {
+  onClick: PropTypes.func.isRequired,
+  breadcrumb: PropTypes.arrayOf(PropTypes.object),
+};
+const defaultProps = {};
```
oscarkwan: `breadcrumb` prop is not required, do we need a default prop here? does eslint not complain?

packages/ListBoxBrowser/src/components/Browser/Browser.js
```diff
@@ -0,0 +1,16 @@
...
+  isLoading: false,
+};
+
+export default function Browser() {
+  return null;
+}
+
+Browser.propTypes = propTypes;
+Browser.defaultProps = defaultProps;
+Browser.displayName = "ListBoxBrowser.Browser";
```
oscarkwan: What's the purpose of this component, is there a better naming for this? `Browser.Browser` sounds kind of weird

packages/ListBoxBrowser/test/ListBoxBrowser.spec.js
```diff
@@ -0,0 +1,69 @@
...
+import data from "./fixtures/multiple";
+import dataMultipleFirstOptionNoOptions from "./fixtures/multiple.firstOptionNoOptions";
+import dataNoOptions from "./fixtures/multiple.withNoOptions";
+import { getData, getOptionByKey } from "../src/helpers";
+import ListBoxBrowser from "../src";
+
+const selectedOptions = { current: {} };
+const dataParameters = { data, selectedOptions, defaultSelectedOptions: () => false };
+
+describe("ListBoxBrowser", () => {
```
oscarkwan: Will you be adding more interactive tests? Most of these tests are just snapshot / see if it renders properly.

nahumzs: yeap I going to add cypress, before merging it.

packages/ListBoxBrowser/src/components/Browser/Browser.js
nahumzs: is the right side of the component, honestly, I don't know if there is a better name for it. right now let you set a `isLoading` prop .

This component exists to split props on the subcomponent without overwhelms the main `<ListBoxBrowser />` with tons of props.

packages/ListBoxBrowser/src/components/Breadcrumb/Breadcrumb.js
nahumzs: ah!!!! you right why is the linter not catching it. I will check this.

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -23,6 +23,7 @@ const propTypes = {
   onClickClear: PropTypes.func,
   onClickFooterAccept: PropTypes.func,
   placeholder: PropTypes.string,
+  isHidden: PropTypes.bool,
```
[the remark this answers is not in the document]
nahumzs: currently, we don't, but one future scenario could be to have a multiple-listbox in one side of the screen and in a different container and side of the screen display the selected item. Pretty much this lets you detach the trigger from the listbox.

packages/ListBoxBrowser/README.md
nahumzs: feel free to modify it :)

packages/ListBoxBrowser/src/components/Browser/Browser.js
oscarkwan: `RightColumn` ? :D? But i'm not going to die on this hill, if can't think of something better than we can leave it

packages/ListBoxBrowser/README.md
oscarkwan:
```suggestion
A two column layout that allows users to select one or multiple options. It is similar to the experience of browsing a file manager in any operating system.
```

packages/ListBoxBrowser/src/components/Browser/Browser.js
alexzherdev: yes @oscarkwan please don't we need you

packages/ListBoxBrowser/src/ListBoxBrowser.js
nahumzs: yeap, we have to use a custom prop, don't know if it's that relevant to have it, I'm throwing an error already if I don't find the label. what do you think?

packages/ListBox/src/components/Trigger/Trigger.styles.js
nahumzs: this is part of the Listbox, right now this is not part of this PR. maybe it would be good to add an issue with this concern?

packages/ListBoxBrowser/src/helpers.js
nahumzs: will get the latest options selected, not a big deal this is handled by the consumer so it's up to them to return the correct option.

packages/ListBoxBrowser/src/helpers.js
nahumzs: so map doesn't support the break statement, what I'm doing is a technique where you throw an error to stop the loop, so what is actually happening I'm just describing the option has been found stop looping and during the handling exception, I return the value.

```js
  try {
    runner(data, path);
    return [];
  } catch (e) {
    return node;
  }
```

I took the inspiration from this gist I saw time ago.

https://gist.github.com/sebmarkbage/2c7acb6210266045050632ea611aebee#file-infrastructure-js-L23

alexzherdev: A+ for inventiveness 😮 👏 but I'm absolutely bummed to say that this is overengineered 😭 everything here is achievable and arguably made clearer with a plain loop. We don't need `.map` here because its return value is never used. Could replace it with `.forEach` but then it only overcomplicates by not being able to stop the loop.

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.js
nahumzs: I would prefer to have a small function and avoid the dependency :)

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
nahumzs: what is the stylers/token for 13px?

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.js
alexzherdev: This seems more complicated than size though because it counts the lengths of the arrays in the values?
The implementation looks a bit off to me: I don't think the inner `forEach` and the return are necessary (`forEach` itself returns undefined anyway).

```js
Object.values(options).reduce((acc, option) => acc + option.length, 0);
```

Maybe also rename the function to something like `getTotalOptionsLength`?

packages/ListBox/src/hooks/hooks.js
nahumzs: won't do, duplication sometimes is not necessarily bad in my opinion.

packages/ListBoxBrowser/src/helpers.js
nahumzs: I agree map, is no necessary, I removed it and replaced for foreach and keep it the same way, throwing an exception on a loop make it easier to implement the recursive function. :)

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
oscarkwan: `stylers.fontSize(-2)` i believe.

nahumzs: mmm I think at that level is better to have it as 13px is clearer

Issue #96, opened 2019-06-21
Replace componentType for displayName


nahumzs:
currently, some components use `componentType` as an escape to filter the type of the component while iterating over props.children.

After some reading and suggestions seems that the "correct" way to do this is using 
 "Component.displayName" [see more](https://github.com/facebook/react/issues/4915#issuecomment-335803765)

Help is asked to go and replace any `componentType` in the repo and replace it for `diplayName`

alexzherdev:
I don't think that particular comment is the actual solution for the question though? [This](https://github.com/facebook/react/issues/4915#issuecomment-162730765) looks like one (and the code sample below it).

nahumzs:
Would you prefer to import the instance and compare the type directly? I open to both to me is simpler to just declare displayName than import a component just to be compared. But I will leave this decision to you, I'm good with both.

PR #214, opened 2019-10-01
UX-565 Takeover component migration

### LINKED ISSUE #208: Takeover Component
# Takeover

### Description
**Takeover** component can toggle a full-screen view to help the user focus on complex UI tasks.
More information at the design system site: https://design.wegalvanize.com/p/components/takeover

### Context
The Takeover is part of the Panel family which includes the Modal and [SidePanel](https://github.com/acl-services/paprika/tree/master/packages/SidePanel), currently, we have a Takeover component working in our previous library which we want to migrate to Paprika.

The main tasks for doing so are:
- [ ] Make it work as a *controlled* component.
- [ ] Replace `scss` with styled-components, preferable using the `css` prop. In case you need it during this task you can take advantage of `@paprika/tokens` and `@paprika/stylers`
- [ ] Remove react-motion from the component, let's keep it for now with a basic fade-in/fade-out animation when opening the Takeover, later we can discuss with the designer how to improve it.
- [ ] Implementing trap focus via `focus-trap-react` [focus trap on sidepanel](https://github.com/acl-services/paprika/blob/master/packages/SidePanel/src/SidePanel.js#L11)
- [ ] Implementing a similar API to the SidePanel [readme](https://github.com/acl-services/paprika/blob/master/packages/SidePanel/readme.md)
- [ ] Add proper a11y [requirements](https://design.wegalvanize.com/p/components/takeover#accessibility) 

## Visual representation of the component

Raw Visual Representation

Takeover with Wizard

## API

API intended, please feel free to modified or improve it. 

```js
// Content taking full width without padding
<Takeover>
   <Takeover.Header></Takeover.Header>
   [... your conent ...]
</Takeover>

// Content with proper padding according to design [${tokens.space * 2}](https://design.wegalvanize.com/tokens)
<Takeover>
   <Takeover.Header></Takeover.Header>
   <Takeover.Content>[... your conent ...]</Takeover.Content> {/* with padding */}
</Takeover>

// Ability to override the focus trap:
<Takeover>
   <Takeover.FocusTrap {...anyPropThatFocusTrapAccepst} />
   <Takeover.Header></Takeover.Header>
   <Takeover.Content>[... your conent ...]</Takeover.Content> {/* with padding */}
</Takeover>
```

**Note:** The Takeover should be able to either render in a React.Portal or as a regular Inline Element taking over an entire page. [ref1](https://design.wegalvanize.com/p/components/takeover#disable-body-scrolling), [ref2](https://github.com/acl-services/paprika/blob/master/packages/SidePanel/src/SidePanel.js#L45)

## A11y considerations
*Add any specific consideration related to a11y, keyboard shortcuts, screen reader behaviour.*

## Notes
Please take a moment and read some tips about migrating component to paprika in our [wiki page](https://github.com/acl-services/acl-ui/wiki/Migrating-to-Paprika)

DanilAgafonov:
### 🛠 Purpose

**Takeover** component can toggle a full-screen view to help the user focus on complex UI tasks.
More information at the design system site: design.wegalvanize.com/p/components/takeover

Resolves #208 

---

### ✏️ Notes to Reviewer

—

---

### 🖥 Screenshots

Basic:
image

With nested SidePanel:
image

With full-width content:
image

---

[http://storybooks.highbond-s3.com/paprika/UX-565-migrate-takeover](http://storybooks.highbond-s3.com/paprika/UX-565-migrate-takeover)


nahumzs:
Could you add a `README.md` file explaining the use of the component, the API seems good to me 👍 follows a similar approach with the sidepanel, and how the modal will work.

packages/SidePanel/src/SidePanel.js
```diff
@@ -43,6 +43,9 @@ const propTypes = {
 
   /** Render the sidepanel inline */
   isInline: PropTypes.bool,
+
+  /** Sidepanel will capture and will not propagate event after pressing ESC. Helps in case we have many overlapped components. */
+  shouldStopEscapePropagation: PropTypes.bool,
```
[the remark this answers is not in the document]
DanilAgafonov: Is it OK to introduce breaking change here? If we do this, than default behavior will change. Maybe someone who use this component relies on current behavior? But maybe it is ok to introduce this breaking changes? WDYT?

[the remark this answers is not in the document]
nahumzs: Second thought maybe is better to not do it automatically, and just in cases when we need it.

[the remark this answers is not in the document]
alexzherdev: recently read this, seems logical to always stop propagation https://sarahmhigley.com/writing/escaping-101/

packages/SidePanel/src/components/Overlay/Overlay.styles.js
```diff
@@ -2,11 +2,22 @@ import tokens from "@paprika/tokens/lib/tokens";
...
   height: 100%;
   position: fixed;
   top: 0;
   width: 100%;
+  
+  &:hover {
+    background: ${tokens.modal.backdrop.backgroundColor} none;
+  }
+  
+  &:active {
```
nahumzs: this is extra styling is no longer required, @oscarkwan already changed the <Button isSemantic={false} /> for <RawButton />, the first one introduced weird styles to the overlay, just rebase and remove this style and should be ok.

packages/SidePanel/src/hooks/useEscapeKey.js
```diff
@@ -1,19 +1,23 @@
 import React from "react";
```
nahumzs: maybe we can unified this component on the helpers and reuse it in modal, sidepanel and takeover component?

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+
+const defaultProps = {
+  isInline: false,
+  shouldStopEscapePropagation: false,
+  onAfterClose: () => {},
+  onClose: () => {},
+  onAfterOpen: () => {},
+};
+
+const Portal = ({ children, active }) =>
```
nahumzs: another candidate to be consolidate into a helper package.

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+  active ? ReactDOM.createPortal(children, document.body) : <React.Fragment>{children}</React.Fragment>;
```
nahumzs: It might be a good idea to replace `<React.Fragment>` for a `<div data-ppk-anchor="takeover" />`  this comes handy if we want to target it either for testing or for custom styling.

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+
+const StyledWrapper = styled(Wrapper)`
```
nahumzs: I think in name of consistency at this point, I will have to ask you to move these styles into a new file, Takeover.styles.js and use the `css`

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+const StyledHeader = styled(Header)`
+  flex: none;
+`;
+
+const ContentWrapper = styled.div`
+  overflow-y: auto;
+  flex: 1 1 auto;
+`;
+
+const Takeover = ({ isOpen, onClose, isInline, shouldStopEscapePropagation, onAfterClose, onAfterOpen, ...props }) => {
```
nahumzs: please use a `function Takeover(props)`  instead of declaring a variable, and use destructuring inside of the function, again this is just for the sake of consistency and how other components are doing it on our library.

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+  const refWrapper = React.useRef(null);
+
+  function handleTransitionEnter(node) {
+    // https://github.com/reactjs/react-transition-group/blob/6dbadb594c7c2a2f15bc47afc6b4374cfd73c7c0/src/CSSTransition.js#L44
+    node.scrollTop;
+  }
+
+  const {
```
nahumzs: this is another candidate to move into the helper package.

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+    fallbackFocus: () => document.createElement("div"),
+    ...extendedFocusTrapOptions,
+  };
+
+  return (
+    <>
+      {isOpen && <EscListener on={onClose} shouldStopEscapePropagation={shouldStopEscapePropagation} />}
+      {isOpen && <LockBodyScroll />}
+      <Portal active={!isInline}>
+        <Transition
```
nahumzs: Can we extract maybe this component `<Transition />` as well so we can reuse it in modal, sidepanel and takeover 🤔?

packages/Takeover/src/components/Content.js
```diff
@@ -0,0 +1,16 @@
...
+const Content = styled.div`
+  margin: ${spacer(2)};
+  padding: ${spacer(2)};
+  background-color: ${tokens.color.white};
+  border: 1px solid ${tokens.border.color};
+  border-radius: 3px;
+  box-shadow: ${tokens.shadow};
+`;
+
+Content.componentType = "Takeover.Content";
```
nahumzs: while there are some components using `componentType`, the latest agreement is to use from now and forward`displayName` instead to detect component.

packages/Takeover/src/components/EscListener.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const propTypes = {
+  /** Callback triggered on ESC key press */
+  on: PropTypes.func.isRequired,
+};
+
+const EscListener = ({ on, shouldStopEscapePropagation }) => {
```
nahumzs: To me, this seems more a hook than a component, I think @alexzherdev commented about this in a previous PR.

packages/Takeover/src/components/FocusTrap.js
```diff
@@ -0,0 +1,36 @@
+import PropTypes from "prop-types";
```
nahumzs: Let's extract this component as well into the helpers seems a one that we want to reuse in more than one place.

packages/Takeover/src/components/Header.js
```diff
@@ -0,0 +1,83 @@
...
+  kind: "default",
+  onClose: () => {},
+};
+
+const kind = {
+  default: `background: ${tokens.color.white}; color: ${tokens.color.black};`,
+  primary: `background: ${tokens.color.purple}; color: ${tokens.color.white};`,
+};
+
+const Wrapper = styled.div`
```
nahumzs: same here in the name of consistency please move this into their own file and use css prop.

packages/Takeover/src/components/Header.js
```diff
@@ -0,0 +1,83 @@
...
+
+const StyledCloseButton = styled(Button.Close)`
+  .button__icon {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+`;
+
+const Header = React.forwardRef(({ children, hasCloseButton, kind, onClose, ...moreProps }, ref) => (
```
nahumzs: same here can you use the paramenter as prop and then destructure? 

The only place where destructuring from the parameters is encouraged or at least is consistency in our library is when using a function inside of styledComponent or in a regular function that are not component.

The use of the prop word helps developers to spot faster which are component and which are not.

packages/Takeover/src/components/Header.js
```diff
@@ -0,0 +1,83 @@
...
+          size="small"
+        />
+      )}
+    </div>
+  </Wrapper>
+));
+
+Header.propTypes = propTypes;
+Header.defaultProps = defaultProps;
+Header.componentType = "Takeover.Header";
```
nahumzs: displayName

packages/Takeover/src/components/LockBodyScroll.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
```
nahumzs: Can go into the helpers package for sure.

packages/Takeover/src/helpers/extractChildren.js
```diff
@@ -0,0 +1,27 @@
+import React from "react";
```
nahumzs: candidate for the helpers package as well.

packages/Takeover/src/tokens.js
```diff
@@ -0,0 +1 @@
+export const animationDuration = 700;
```
nahumzs: I like this, for package local const variable. cool!

packages/SidePanel/src/SidePanel.js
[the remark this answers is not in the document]
mikrotron: > stopPropagation: Just Do It™ 
💯🙂

packages/Takeover/src/Takeover.js
mikrotron: `pka-anchor`

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+  flex-direction: column;
+`;
+
+const StyledHeader = styled(Header)`
+  flex: none;
+`;
+
+const ContentWrapper = styled.div`
+  overflow-y: auto;
+  flex: 1 1 auto;
```
mikrotron: This is just me, but I much prefer the long-form `flex` rules... as a courtesy to anyone reading who has to think about what each one means.  In this case, I think you only need to specify `flex-grow: 1`.

packages/Takeover/src/Takeover.js
mikrotron: Since it should wrap the `children` in both cases, maybe it should actually go down in `render()`.

packages/Takeover/src/Takeover.js
```diff
@@ -0,0 +1,116 @@
...
+          mountOnEnter
+          unmountOnExit
+          in={isOpen}
+          timeout={animationDuration}
+          onEnter={handleTransitionEnter}
+          onEntered={onAfterOpen}
+          onExited={onAfterClose}
+        >
+          {state => (
+            <FocusTrapLibrary active={!isInline} focusTrapOptions={focusTrapOptions}>
```
mikrotron: Even an `inline` Takeover should trap the focus, shouldn't it?

packages/Takeover/src/components/Content.js
```diff
@@ -0,0 +1,16 @@
+import styled from "styled-components";
+import { spacer } from "@paprika/stylers/lib/helpers";
+import tokens from "@paprika/tokens/lib/tokens";
+
+const Content = styled.div`
+  margin: ${spacer(2)};
+  padding: ${spacer(2)};
+  background-color: ${tokens.color.white};
+  border: 1px solid ${tokens.border.color};
+  border-radius: 3px;
```
mikrotron: `tokens.border.radius`

packages/Takeover/src/components/EscListener.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const propTypes = {
+  /** Callback triggered on ESC key press */
+  on: PropTypes.func.isRequired,
```
mikrotron: `onPress`?

packages/Takeover/src/components/FocusTrap.js
```diff
@@ -0,0 +1,36 @@
...
+  initialFocus: PropTypes.oneOfType([PropTypes.node, PropTypes.string, PropTypes.func]),
+
+  /**
+    By default, an error will be thrown if the focus trap contains no elements in its tab order. With this option you can specify a fallback element to programmatically receive focus if no other tabbable elements are found. For example, you may want a popover's <div> to receive focus if the popover's content includes no tabbable elements. Make sure the fallback element has a negative tabindex so it can be programmatically focused. The option value can be a DOM node, a selector string (which will be passed to document.querySelector() to find the DOM node), or a function that returns a DOM node.
+    escapeDeactivates {boolean}: Default: true. If false, the Escape key will not trigger deactivation of the focus trap. This can be useful if you want to force the user to make a decision instead of allowing an easy way out.
```
mikrotron: Can we just remove this line?

packages/Takeover/src/components/FocusTrap.js
```diff
@@ -0,0 +1,36 @@
...
+  fallbackFocus: PropTypes.oneOfType([PropTypes.node, PropTypes.string, PropTypes.func]), //
+
+  /** Default: false. If true, a click outside the focus trap will deactivate the focus trap and allow the click event to do its thing. */
+  clickOutsideDeactivates: PropTypes.bool,
+
+  /** Default: true. If false, when the trap is deactivated, focus will not return to the element that had focus before activation. */
+  returnFocusOnDeactivate: PropTypes.bool,
+};
+
+const defaultProps = FocusTrapLibrary.defaultProps.focusTrapOptions;
```
mikrotron: Is there any drawback to copying these here, for improved readability / documentation?

packages/Takeover/src/components/FocusTrap.js
```diff
@@ -0,0 +1,36 @@
...
+    escapeDeactivates {boolean}: Default: true. If false, the Escape key will not trigger deactivation of the focus trap. This can be useful if you want to force the user to make a decision instead of allowing an easy way out.
+  */
+  fallbackFocus: PropTypes.oneOfType([PropTypes.node, PropTypes.string, PropTypes.func]), //
+
+  /** Default: false. If true, a click outside the focus trap will deactivate the focus trap and allow the click event to do its thing. */
+  clickOutsideDeactivates: PropTypes.bool,
+
+  /** Default: true. If false, when the trap is deactivated, focus will not return to the element that had focus before activation. */
+  returnFocusOnDeactivate: PropTypes.bool,
```
mikrotron: Can we leave this out? I think we will _always_ want to do this, and if we really want to focus something else (like a newly created item) it could still be done anyway either in `onAfterClose()` or by a parental component.

packages/Takeover/src/components/FocusTrap.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+const defaultProps = FocusTrapLibrary.defaultProps.focusTrapOptions;
+
+const FocusTrap = () => null;
+
+FocusTrap.propTypes = propTypes;
+FocusTrap.defaultProps = defaultProps;
+FocusTrap.componentType = "Takeover.FocusTrap";
```
mikrotron: `displayName`

packages/Takeover/src/components/Header.js
mikrotron: I'm not 100% convinced, but definitely swayed by Danil's argument _against_ using the css prop.  But I definitely like to separate the styling into a `styles.js` file.

packages/Takeover/src/tokens.js
mikrotron: Me too.  Would it be more appropriate in `/helpers` though?

packages/Takeover/src/components/Header.js
```diff
@@ -0,0 +1,83 @@
...
+    ${fontSize()};
+    font-weight: 400;
+    margin: 0;
+  }
+
+  ${props => kind[props.kind]}
+`;
+
+const StyledCloseButton = styled(Button.Close)`
+  .button__icon {
```
mikrotron: I would actually like to get rid of this class in the `<Button>` (and avoid classes as much as possible in Paprika components).  Instead I suggest using the `[data-pka-anchor='button.icon']` selector.

packages/Takeover/src/components/FocusTrap.js
alexzherdev: I thought we agreed against `displayName`? I don't remember exactly

mikrotron: Are you gaslighting me? :trollface: 
I'm like 87% sure we decided on `displayName`.

https://github.com/acl-services/paprika/pull/214#discussion_r336592869

mikrotron: Make that 97%
https://github.com/acl-services/paprika/issues/96

alexzherdev: > Are you gaslighting me? :trollface:

I'm not 😄I got this confused with https://github.com/acl-services/paprika/issues/101
I think that argument still stands here. There was also this https://github.com/acl-services/paprika/pull/123#discussion_r304175906

packages/Takeover/src/Takeover.js
DanilAgafonov: IDK if we should support IE10, but if we should...

> Internet Explorer 10 uses 0 instead of 1 as the initial value for the flex-shrink property. (https://github.com/philipwalton/flexbugs#flexbug-6)

packages/Takeover/src/Takeover.js
DanilAgafonov: I just copied this behavior from SidePanel. Should SidePanel trap the focus in inline mode too?

packages/Takeover/src/components/Header.js
DanilAgafonov: In my opinion, the best option is to convert [`ButtonIcon` component](https://github.com/acl-services/paprika/blob/eb1ee6765fce360df1d296e2ffa1c9d0e417a917/packages/Button/src/Button.js#L96) to styled component:

```jsx
export const ButtonIcon = styled(props => props.children ? <span {...props} /> : null)`
  ${iconStyles};
`;
```


And export `ButtonIcon` const from `Button.js`:

```jsx
...

export const ButtonIcon = ...;

const Button = ...;

export default Button;
```

And then style icon in place in following way:
```jsx
import { ButtonIcon } from '@paprika/button';

const StyledCloseButton = styled(Button.Close)`
  ${ButtonIcon} {
    ...
  }
`;
```

packages/Takeover/src/components/FocusTrap.js
DanilAgafonov: Probably, you missed the purpose of `FocusTrap` component.

As you can see it always renders `null`. The only one purpose of this component is to declare propTypes. When someone uses `<Takeover>` component, they have ability to override focus trap options by passing this component as `<Takeover>` child.

All props passed to `<Takeover.FocusTrap>` will be [extracted](https://github.com/acl-services/paprika/blob/fbd96f0b0397e3af5e69d3bdae6c8d01508b8dfe/packages/Takeover/src/Takeover.js#L63-L75) and [applied to actual focus trap component](https://github.com/acl-services/paprika/blob/fbd96f0b0397e3af5e69d3bdae6c8d01508b8dfe/packages/Takeover/src/Takeover.js#L99).

And I guess we should keep an ability to override/add any `focus-trap-react` option.

packages/Takeover/src/components/LockBodyScroll.js
DanilAgafonov: Should it be in `helpers` dir or directly in `packages`? For example, we have `Guard` component in `packages`. And I think `LockBodyScroll` works on the same level of responsibility as `Guard` component.

packages/Takeover/src/Takeover.js
DanilAgafonov: Same question as here: https://github.com/acl-services/paprika/pull/214#discussion_r342030869

packages/Takeover/src/Takeover.js
DanilAgafonov: IMHO, modal, takeover and sidepanel is a one component with different props:
Modal: `<Dialog>...</Dialog>`
Takeover: `<Dialog align="stretch">...</Dialog>`
Sidepanel: `<Dialog align="right">...</Dialog>`

So I suggest to not extract `<Transition />`, but at some point create `Dialog` component that combines modal, takeover and sidepanel functionality. (And deprecate Sidepanel and Takeover after that)

I have task about Modal component in backlog. Probably, it would be better to implement common Dialog, than separate Modal.

packages/Takeover/src/tokens.js
nahumzs: I think is clearer having their own tokens files. Helpers tend to be more functions/methods specifics for the logic, and this is for styling and constant variables.

packages/Takeover/src/Takeover.js
nahumzs: We only support IE11 and up. we deprecated IE10 a couple years ago.

packages/Takeover/src/Takeover.js
nahumzs: mmm, I think when the sidepanel is inline is a little more complex because we have an array of sidepanels, so maybe we will need to add extra code to trap the focus across all the sidepanels, so far that is not implemented and didn't think about that until now. 

Maybe the takeover shouldn't care if it's inline or not? 
cc @mikrotron

packages/Takeover/src/components/Header.js
nahumzs: what about then if you want to use style components directly which I'm not against, start implementing `import * as styled from "..."`

and do `<styled.Wrapper>`  which to me feels better than Wrapper or WrapperStyled

what do you think? (danil, mike)?

cc @mikrotron

packages/Takeover/src/components/LockBodyScroll.js
nahumzs: what about creating a new folder named `components`, https://github.com/acl-services/paprika/tree/master/packages/helpers/src ?

packages/Takeover/src/components/FocusTrap.js
nahumzs: I'm fine with displayName for now, is easier to understand and use. If you feel strongly about this @alexzherdev, we can refactor our displayName and componentName in a ux-kaisen day.

packages/SidePanel/src/SidePanel.js
```diff
@@ -118,6 +117,14 @@ function SidePanel(props) {
     ...extendedFocusTrapOptions,
   };
 
+  function handleEscKey(event) {
+    if (event.key === "Escape") {
```
nahumzs: thank you :)

packages/Takeover/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/Takeover"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/button": "^0.1.13",
```
nahumzs: we recently upgrade paprika with the breaking change on the L10n component issue.

you need to bump button to 0.2.0 at least and I'm not sure the other ones but you can check their own package json to see the latest version.

packages/Takeover/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@paprika/stylers": "^0.1.5",
+    "@paprika/tokens": "^0.1.4",
+    "focus-trap-react": "^6.0.0",
+    "prop-types": "^15.7.2",
+    "react-transition-group": "^4.3.0"
+  },
+  "peerDependencies": {
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
+    "styled-components": "^4.2.0"
```
nahumzs: are we using L10n in this component if yes we should add it as a peerDependency if not ignore my comment and mark it as `resolve`.

packages/Takeover/src/tokens.js
mikrotron: I mean `/src/helpers/tokens.js`, just so the `/src` folder stays a bit tidier.

packages/Takeover/src/Takeover.js
mikrotron: Hm. I'm not sure if we need to have an inline option or not for Takeover... _maybe_ there's a reason to keep it's code inline in the DOM, rather than a portal appended to the end...but I can't really think of why you would _need_ that.  

For the SidePanel though , you may want an "inline" implementation that does _NOT_ trap focus, since the user can still see and interact with other page content.  The Takeover on the other hand blocks everything else out (it is "modal").

packages/Takeover/src/components/FocusTrap.js
mikrotron: True, but having the `escapeDeactivates` prop here, in the comments, seems strange.  I think a link to the online documentation for the `focus-trap-react` library, maybe directly to the options, would be more helpful?

packages/Takeover/src/components/Header.js
mikrotron: Agreed. I had a similar conversation recently with Tristan, and we came to the same conclusion.

packages/Takeover/src/Takeover.js
DanilAgafonov: @mikrotron @nahumzs I don't see any reason to have inline mode for Takeover. WDYT if I completely remove `inline` prop?

nahumzs: Yes, let's remove it @DanilAgafonov

packages/Takeover/src/Takeover.js
DanilAgafonov: Ok, let's start from implementing Modal as separated component. I will try to extract `<Transition />` later when start working on Modal component.

packages/Takeover/src/components/FocusTrap.js
DanilAgafonov: I will do this as part of work on Modal component. Does it work for you?

packages/Takeover/src/components/Header.js
DanilAgafonov: It feels much better for me, than using `<div css={...}>...</div>`, but anyway I think it is not how styled-components should be used.

The main idea of SC is remove mapping between styles and components. Every styled component is just normal React component. It is an what author of SC trying to enforce across SC community (he was saying about this many times on twitter and conferences). For example, you can watch Max's (author of SC) talk: https://www.youtube.com/watch?v=jaqDA7Btm3c

Some examples of using styled components from real production grade applications:
https://github.com/outline/outline/blob/master/app/components/DropdownMenu/DropdownMenu.js
https://github.com/withspectrum/spectrum/blob/alpha/src/components/logo/index.js
https://github.com/caspg/space-exp/blob/master/src/components/Header/Header.jsx
https://github.com/prisma-labs/graphql-playground/blob/master/packages/graphql-playground-react/src/components/Settings.tsx

packages/Takeover/src/components/FocusTrap.js
DanilAgafonov: It doesn't matter if the props declared in this `propTypes` object or nor. It is just for documentation.

If we remove some prop from this list, it can be passed by user anyway and `focus-trap` will change internal behavior.

So removing `returnFocusOnDeactivate` just will make this feature undocumented, but will not disable an ability to use it.

packages/Takeover/src/components/FocusTrap.js
nahumzs: ok make sense 👍

packages/Takeover/src/components/Header.js
nahumzs: When we started the paprika repository, we decided to go with the css approach since we didn't want to enforce all our developer to jump into styled-component was just too disruptive at that time, but now I think its safe to start using `styled` instead `css` whenever make sense or the majority of the time. 

The only thing that I don't understand from your comment is:

> but anyway I think it is not how styled-components should be used.

We are not mapping anything we are just describing what kind of component it's, it's the same that declaring it on the file, but in our case, we want to split it into a different file, so we can distinguish which is a styled component from a regular component.  

```js
// Some.styles.js
export const MyStuff = styled.div`` 
// or
export const MyStuff = styled(Some)``

// Some.js
import * as styled from "./Some.styles.js"

return <styled.MyStuff>content</styled.MyStuff>
```

you still could import MyStuff at any file and will keep the styling and make it a Styled component without **`css`**. 
From how I see it the only thing we are not doing is declaring it in the same file. or am I missing something?

Ps. https://github.com/prisma-labs/graphql-playground/blob/master/packages/graphql-playground-react/src/components/Settings.tsx#L14

^ that is the kind of declaration I want to avoid, `Wrapper`, `IconWrapper`, IMO read better and fits better using `<styled.Wrapper />` and `<styled.Icon />`

/// 
https://github.com/caspg/space-exp/blob/master/src/components/Header/Header.jsx#L59
^ this one as well is not describing that is a styled component. 

CC @mikrotron

packages/Takeover/src/components/FocusTrap.js
nahumzs: yes, let's keep the documentation as it's.

packages/Takeover/src/components/Header.js
DanilAgafonov: Thanks, @nahumzs! I got your point. I will refactor this PR to export styled components from `*.styled.js` files to move forward, but suggest to continue this discussion.

----

From my point of view, sometimes it is hard to say whether component should be placed in `*.styled.js` file or should be placed like regular React component.

Examples:
1. https://github.com/acl-services/paprika/blob/f9443c6f4126fcf4f110154386269664f5f6efab/packages/Takeover/src/components/Content.js
2. ```jsx
   const Checkbox = styled.input.attrs({
     type: 'checkbox',
   })`
     margin: 0 0.7em 0 0;
   `;
   ```
3. ```jsx
   const SignInLink = styled(ButtonLink).attrs({
     to: '/signin',
     children: 'Sign In',
     neutral: true,
     inline: true,
   })`
     margin-top: 1em;
   `;
   ```
4. ```jsx
   const RequiredMark = styled.span.attrs({
     children: '*',
   })`
     color: #ff0000;
   `;
   ```

> we want to split it into a different file, so we can distinguish which is a styled component from a regular component.

Is there any reason to have an ability to distinguish which is a styled component from a regular component?
It looks like "Separation of technologies", which doesn't look like best practice in React ecosystem. It would be better to follow "Separation of concerns".

Moreover, I think SC is implementation details. When we use component, we shouldn't know whether it implemented with SC or not. `*.styled.js` suffix in file name exposes implementation details.

For example:
1. Say, we have `RequiredMark`. This component is a part of `SignUpForm`.
2. Say, we put all styled components used for `SignUpForm` to `SignUpForm.styled.js`.
3. And now we use it as `<styled.RequiredMark>`.
4. In some day we decided to add tooltip for `RequiredMark`.
5. And now `RequiredMark` looks like:
   ```jsx
   const Asterisk = styled.span.attrs({
     children: '*',
   })`
     color: #ff0000;
   `;

   const RequiredMark = (props) => (
     <Tooltip text="This field is required">
		<Asterisk {...props} />
     </Tooltip>
   );
   ```

Should `RequiredMark` be extracted from `SignUpForm.styled.js`? If yes, than we should change `<styled.RequiredMark>` to `<RequiredMark>`. This is what I want to avoid: we don't want to change all usages on changing implementation of used component.

DanilAgafonov: By the way, how should I name files with SC? `*.styles.js` or `*.styled.js`?

DanilAgafonov:
For some reason Jest tests are stucked on:
```
packages/ListBoxBrowser/test/specs/ListBoxBrowser.spec.js
packages/ListBox/test/specs/single/ListBox.Option.spec.js
packages/ListBox/test/specs/multi/ListBox.multi.spec.js
packages/ListBox/test/specs/single/ListBox.single.spec.js
```

It occurred only in this branch. No problems on master. I will fix it on Friday (Nov 15).

packages/Takeover/src/components/Header.js
nahumzs: About your question `*.styles.js`. 

Lots of good comments Danil.

Definitely we are not trying to do a separation of concerns or technologies, I think the intention was/is to help other developers to find faster where are the styles declare and where is the logic more than separating the technologies. That's all.

I understand your concerns and they are valid. Also, those might be specific cases. The 80/90% of the time, we wouldn't have to worry about those, and when those happen I don't see anything wrong about implementing it as you suggested.

I think most of the things we follow on paprika are guidelines, we are not trying to impose the decision. And most of our decision are trying to solve 80%/90% of the common cases.

cc @mikrotron

PR #215, opened 2019-10-01
Fix failing spec

tristanjasper:
### 🛠 Purpose

Fix failing spec tests for date picker.

@nahumzs minor bump required

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

image

---

### DIFF
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -131,7 +131,11 @@ function DatePicker(props) {
   }
 
   function handleClosePopover() {
-    if (!isElementContainsFocus(calendarRef.current) && !isElementContainsFocus(inputRef.current)) {
+    if (
+      calendarRef.current &&
+      !isElementContainsFocus(calendarRef.current) &&
+      (inputRef.current && !isElementContainsFocus(inputRef.current))
+    ) {
       if (!hasParsingError) {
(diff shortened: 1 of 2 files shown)


packages/DatePicker/src/DatePicker.js
```diff
@@ -131,7 +131,11 @@ function DatePicker(props) {
...
   function handleClosePopover() {
-    if (!isElementContainsFocus(calendarRef.current) && !isElementContainsFocus(inputRef.current)) {
+    if (
+      calendarRef.current &&
+      !isElementContainsFocus(calendarRef.current) &&
+      (inputRef.current && !isElementContainsFocus(inputRef.current))
+    ) {
       if (!hasParsingError) {
         setConfirmationResult(formatDateProp(humanFormat));
         setInputtedString(formatDateProp(dataFormat));
```
AndreyChernykh: That is a good example of `Don't Sync State. Derive It!` https://kentcdodds.com/blog/dont-sync-state-derive-it

```js
const confirmationResult = formatDateProp(humanFormat);
const confirmationResult = formatDateProp(dataFormat);
```
should be enough. 

if you are worried about recalculating stuff - you could use `Reac.useMemo`

AndreyChernykh: @nahumzs fyi

PR #216, opened 2019-10-01
XD-5802-set datePicker date with Utc

tristanjasper:
### 🛠 Purpose

Fix Date picker failing spec by setting specific time

@nahumzs minor bump

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "storybook": "start-storybook -p 9009 -c .storybook",
     "test:a11y": "echo 'Axe will run on this command'",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
-    "test": "jest --passWithNoTests",
+    "test": "TZ=UTC jest --passWithNoTests",
     "transpile": "lerna exec --parallel -- babel --root-mode upward src --out-dir lib --source-maps --copy-files"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "storybook": "start-storybook -p 9009 -c .storybook",
     "test:a11y": "echo 'Axe will run on this command'",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
-    "test": "TZ=UTC jest --passWithNoTests",
+    "test": "jest --passWithNoTests",
     "transpile": "lerna exec --parallel -- babel --root-mode upward src --out-dir lib --source-maps --copy-files"
   },
   "engines": {
--- a/packages/DatePicker/test/DatePicker.spec.js
+++ b/packages/DatePicker/test/DatePicker.spec.js
@@ -58,4 +58,14 @@ describe("DatePicker", () => {
 
     expect(document.getElementsByClassName("form-input--has-error").length).toEqual(1);
   });
+
+  it("should render date correctly with different timezones", () => {
+    const utcOffSetPlusOneHour = moment("2019-04-02").utcOffset(60);
+    const { getByTestId, rerender } = render({ date: utcOffSetPlusOneHour });
+    expect(getByTestId("datepicker.input").value).toEqual("April 02, 2019");
+
+    const utcOffSetMinusOneHour = moment("2019-04-02").utcOffset(-60);
+    rerender({ date: utcOffSetMinusOneHour });
+    expect(getByTestId("datepicker.input").value).toEqual("April 02, 2019");
+  });
 });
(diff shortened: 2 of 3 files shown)


nahumzs:
what about the comment from @DanilAgafonov about that this has a legit but when working with a different UTC? did we already take into consideration his comment?

allison-c:
> what about the comment from @DanilAgafonov about that this has a legit but when working with a different UTC? did we already take into consideration his comment?

The error came from the spec(`moment('mock_date_string')`). In the component, we use the UTC timezone(`moment.utc('user_inputted_string')`) to convert user inputted date, so it will be fine.

braska:
I guess, we need to check what passed to `date` prop in end applications (Projects, etc.). DatePicker converts passed date to UTC. If we pass something like `moment('2019-01-01')` (as in tests), then it becomes `31 December, 2018` for users with UTC+ (e.g. CET) timezone. Because 01.01.2019 00:00 CET = 31.12.2018 23:00 UTC. My point: probably, we should fix real component, not a test.

If we don't need to fix component and we want to fix only tests, then we need to introduce changes to Storybook story for DatePicker.

tristanjasper:
> Why date is converted to UTC in component?
> 
> I guess, we need to check what passed to `date` prop in end applications (Projects, etc.). Probably, some apps can pass date with non-UTC timezone. If we pass something like `moment('2019-01-01')` (as in tests), then it becomes `31 December, 2018` for users with UTC+ (e.g. CET) timezone. Because 01.01.2019 00:00 CET = 31.12.2018 23:00 UTC. My point: probably, we should fix DatePicker component (get rid of "to UTC" conversion), not a test.
> 
> If we don't need to fix component and we want to fix only tests, then we need to introduce changes to Storybook story for DatePicker.

@braska We are not converting the date to UTC in the component, we are just passing the component a date converted to UTC in the spec test. I don't quite understand if you feel we need to make changes in the component?

allison-c:
> > Why date is converted to UTC in component?
> > I guess, we need to check what passed to `date` prop in end applications (Projects, etc.). Probably, some apps can pass date with non-UTC timezone. If we pass something like `moment('2019-01-01')` (as in tests), then it becomes `31 December, 2018` for users with UTC+ (e.g. CET) timezone. Because 01.01.2019 00:00 CET = 31.12.2018 23:00 UTC. My point: probably, we should fix DatePicker component (get rid of "to UTC" conversion), not a test.
> > If we don't need to fix component and we want to fix only tests, then we need to introduce changes to Storybook story for DatePicker.
> 
> @braska We are not converting the date to UTC in the component, we are just passing the component a date converted to UTC in the spec test. I don't quite understand if you feel we need to make changes in the component?

I think what Danil mean is we shouldn't convert the prop `date` by UTC https://github.com/acl-services/paprika/blob/master/packages/DatePicker/src/DatePicker.js#L80 
That's a good question, I'd like to try if we can remove it or not.

tristanjasper:
> > > Why date is converted to UTC in component?
> > > I guess, we need to check what passed to `date` prop in end applications (Projects, etc.). Probably, some apps can pass date with non-UTC timezone. If we pass something like `moment('2019-01-01')` (as in tests), then it becomes `31 December, 2018` for users with UTC+ (e.g. CET) timezone. Because 01.01.2019 00:00 CET = 31.12.2018 23:00 UTC. My point: probably, we should fix DatePicker component (get rid of "to UTC" conversion), not a test.
> > > If we don't need to fix component and we want to fix only tests, then we need to introduce changes to Storybook story for DatePicker.
> > 
> > 
> > @braska We are not converting the date to UTC in the component, we are just passing the component a date converted to UTC in the spec test. I don't quite understand if you feel we need to make changes in the component?
> 
> I think what Danil mean is we shouldn't convert the prop `date` by UTC https://github.com/acl-services/paprika/blob/master/packages/DatePicker/src/DatePicker.js#L80
> That's a good question, I'd like to try if we can remove it or not.

👌 Got it. I didn't realise it was doing a conversion in the component as well. I think is fine without it.

package.json
```diff
@@ -20,7 +20,7 @@
     "storybook": "start-storybook -p 9009 -c .storybook",
     "test:a11y": "echo 'Axe will run on this command'",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
-    "test": "jest --passWithNoTests",
+    "test": "TZ=UTC jest --passWithNoTests",
```
DanilAgafonov: Can we remove `TZ=UTC`? We have changed DatePicker component and there is no need to specify timezone here.

packages/DatePicker/test/DatePicker.spec.js
```diff
@@ -40,21 +40,21 @@ describe("DatePicker", () => {
   });
 
   it("should display initial date", () => {
-    const { getByTestId } = render({ date: moment("2019-01-02") });
+    const { getByTestId } = render({ date: moment.utc("2019-01-02") });
```
DanilAgafonov: Can we revert all changes in this file?

PR #218, opened 2019-10-04
UX-480 Popover tooltip fixes 🐐

mikrotron:
### 🛠 Purpose
Don't let the `<Tip>` fire point events (`onMouseOver`, `onMouseOut`, `focus`) which will close the "tooltip" (`isEager`) version.

### ✏️ Notes
- Also added some `data-pka-anchor` attributes throughout `<Popover>`.
- **💥 [BREAKING CHANGE]:** Adjusted `data-pka-anchor` values to use a `.` delimiter instead of a `-`, for consistency (ex `popover-trigger` → `popover.trigger`)

### DIFF
--- a/packages/Popover/src/components/Content/Content.js
+++ b/packages/Popover/src/components/Content/Content.js
@@ -82,10 +82,9 @@ const Content = React.forwardRef((props, ref) => {
   /* eslint-disable jsx-a11y/mouse-events-have-key-events */
   return ReactDOM.createPortal(
     <ContentStyled
-      {...moreProps}
       aria-hidden={!isOpen}
       data-component-name="PopoverContent"
-      data-pka-anchor="popover-content"
+      data-pka-anchor="popover.content"
       ref={handleRef}
       isOpen={isOpen}
       onBlur={handleBlur}
@@ -94,6 +93,7 @@ const Content = React.forwardRef((props, ref) => {
       style={contentStyles}
       tabIndex={isOpen ? 0 : -1}
       zIndex={content.zIndex}
+      {...moreProps}
     >
       {children}
     </ContentStyled>,
--- a/packages/DropdownMenu/test/dropdownMenu.cypress.js
+++ b/packages/DropdownMenu/test/dropdownMenu.cypress.js
@@ -5,10 +5,10 @@ describe("<DropdownMenu />", () => {
 
   it("should show hide dropdown content when when clicking trigger", () => {
     cy.getByTestId("dropdown-menu__trigger").click();
-    cy.getByTestId("popover-content").should("be.visible");
+    cy.getByTestId("popover.content").should("be.visible");
     cy.wait(100).then(() => {
       cy.getByTestId("dropdown-menu__trigger").click();
-      cy.getByTestId("popover-content").should("not.be.visible");
+      cy.getByTestId("popover.content").should("not.be.visible");
     });
   });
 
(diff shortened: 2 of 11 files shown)


packages/Popover/src/components/Content/Content.js
```diff
@@ -82,10 +82,9 @@ const Content = React.forwardRef((props, ref) => {
   /* eslint-disable jsx-a11y/mouse-events-have-key-events */
   return ReactDOM.createPortal(
     <ContentStyled
-      {...moreProps}
       aria-hidden={!isOpen}
       data-component-name="PopoverContent"
-      data-pka-anchor="popover-content"
+      data-pka-anchor="popover.content"
```
mikrotron: Would this be considered a "breaking change" for semver?

mikrotron: cc @allison-c

alexzherdev: I’d say so, given that anchors are explicitly not a private implementation detail, but something for others to depend on.

allison-c: For myself, I always try to avoid using dependency's test id since they might be removed or changed anytime...
But I saw we use this test id in the Listbox and the Dropdown component, so maybe other developers will do the same thing.

mikrotron: If we bump the version from `v0.1.12` → `v0.2.0`, that will indicate a breaking change, and so it won't break anyone's tests until they upgrade the npm package in their project.  

I'd like to keep our naming aligned to the convention (that we haven't really documented anywhere 😬).  

Also, it's a pretty straightforward fix, so I'm going to suggest we go ahead anyway.  Unless anyone has strong reservations about that? Should we bundle other breaking changes together?

cc @nahumzs

oscarkwan:
Are we sticking with this new syntax? Do we need a ticket to make all the other `data-pka-anchor` use the same structure?

packages/Popover/src/components/Content/Content.js
nahumzs: Oh I thought we were using css `style` names for those `data-anchors` using dash `-`.

I don't have a problem with this new approach if we keep doing it consistently, I think right now there are cases where we have been using `-` dashes.

nahumzs: Do you think introducing the `.` is that relevant that justify to break our current approach so far? or is it just a small upgrade from our current approach? 

I'm asking because if we introduce the `.` we should then rename the anchors for other components, which will take time and for some time we will hold both syntaxis on our repo.

mikrotron: The first component we used the `data-pka-anchor` for was the `<Sortable>` and I used the `.` to delimit the top level component (`sortable`) from the sub components (`sortable.item`).  So I'm not introducing something new here, I'm sticking to the original plan (I should have caught this particular one in code review when Allison added it before, that was my bad).  But the `-` is still used to separate words... like `progress-accordion.item`. 

https://github.com/acl-services/paprika/blob/master/packages/ProgressAccordion/src/ProgressAccordion.js#L78

So, essentially, the `data-pka-anchor` will typically be the kebab-case version of the component's `displayName`.

oscarkwan: Oh i see, makes more sense now

nahumzs: do you mean this? https://github.com/acl-services/paprika/blob/master/packages/ListBox/test/helpers/selectors.js#L2

nahumzs: I don't want to stop this PR unnecessarily,  I'm ok with whatever route we take if it's consistent :)

PR #219, opened 2019-10-08
PD-13496 migrate pill component

oscarkwan:
### 🛠 Purpose

Migrate over the pill component to be used in an upcoming epic (requirements interlinking)

---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

Screen Shot 2019-10-09 at 5 12 11 PM


packages/Pill/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/raw-button": "0.1.5",
+    "@paprika/stylers": "^0.1.5",
+    "@paprika/tokens": "^0.1.4",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "react": "^16.4.0",
+    "react-dom": "^16.4.2",
```
AndreyChernykh: Please, bump it to ^16.8.0 (both, react and react-dom)

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
+import { css } from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+
+const PillColorStyles = {
+  "#d7d7d7": css`
```
alexzherdev: What about other colours like blue, green, etc.?

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
+import { css } from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+
+const PillColorStyles = {
```
alexzherdev: nit: `pill` lowercase, and also are we using color or colour @mikrotron 😄

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+  `,
+
+  highRisk: css`
+    background: #cd3c44;
+  `,
+};
+
+const PillSizeStyles = {
+  small: css`
+    border-radius: 8px / 50%;
```
alexzherdev: Is there no token for these 8px?

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+    line-height: ${stylers.spacer(2)};
+    padding: 0 ${tokens.space};
+    ${stylers.fontSize(-3)};
+
+    .aclui-icon::before {
+      ${stylers.fontSize(-3)};
+    }
+  `,
+  medium: css`
+    border-radius: 12px / 50%;
```
alexzherdev: and this?

packages/Pill/src/Pill.styles.js
oscarkwan: Those colours are okay with a white text, in this specific colour (blackLighten60) we need it to be black to pass contrast.

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+  `,
+
+  highRisk: css`
+    background: #cd3c44;
+  `,
+};
+
+const PillSizeStyles = {
+  small: css`
+    border-radius: ${tokens.space} / 50%;
```
alexzherdev: Sorry, I actually wasn't thinking of using `space` here 😄 I was just recalling we had some tokens for border radius sizes, and wondering whether these cases deserve tokens or not. Up to you anyway 👍

oscarkwan: Oooh, yeah, i would need to use "button border radius" for 4px. I think this is fine :)

packages/Pill/src/Pill.styles.js
oscarkwan: We use `color` everywhere :D, css uses `color` looool

mikrotron: Yeah, `color` is pretty standard in our code (and pretty much everyone's).  That doesn't mean I like it, or that it's right, but it is consistent.

packages/Pill/src/Pill.js
```diff
@@ -0,0 +1,74 @@
...
+
+export const tokenPillColors = [
+  tokens.color.black,
+  tokens.color.blackLighten60,
+  tokens.color.blue,
+  tokens.color.greenDarken10,
+  tokens.color.orangeDarken10,
+];
+export const severityPillColors = ["noRisk", "lowRisk", "mediumRisk", "highRisk"];
+export const pillSizes = ["small", "medium"];
```
mikrotron: I think that this will be a common enough use case to define another "set", like:
https://github.com/acl-services/paprika/blob/master/packages/helpers/src/customPropTypes.js#L10
But maybe `ShirtSizes.LIMITED`?

packages/Pill/src/Pill.js
```diff
@@ -0,0 +1,74 @@
...
+import tokens from "@paprika/tokens";
+import RawButton from "@paprika/raw-button";
+import pillStyles from "./Pill.styles";
+
+export const tokenPillColors = [
+  tokens.color.black,
+  tokens.color.blackLighten60,
+  tokens.color.blue,
+  tokens.color.greenDarken10,
+  tokens.color.orangeDarken10,
```
mikrotron: I really never liked this part of the Pill API.  I like the `severityPillColors` options, they are nicely semantic.  I'd support generic colour options (like `black`, `grey`, `blue`, `green`, `red`) but only supporting select _token_ colours seems like the wrong level of abstraction.

packages/Pill/src/Pill.js
```diff
@@ -0,0 +1,74 @@
...
+export const pillSizes = ["small", "medium"];
+
+const propTypes = {
+  a11yText: PropTypes.string,
+  children: PropTypes.node.isRequired,
+  isDisabled: PropTypes.bool,
+  onClick: PropTypes.func,
+  pillColor: PropTypes.oneOf([...tokenPillColors, ...severityPillColors]),
+  size: PropTypes.oneOf(pillSizes),
+  textColor: PropTypes.string,
```
mikrotron: I feel like we should just impose this, based on the `pillColor`.  Either programmatically, or with a hashmap, since we have a limited set of options we accept for `pillColor`.

packages/Pill/src/Pill.js
```diff
@@ -0,0 +1,74 @@
...
+  onClick: null,
+  pillColor: tokens.color.blackLighten60,
+  size: "medium",
+  textColor: tokens.color.white,
+};
+
+function Pill(props) {
+  const { a11yText, pillColor, textColor, isDisabled, size, onClick, ...moreProps } = props;
+
+  const tabIndex = onClick === null ? -1 : 0;
```
mikrotron: if `onClick` is `null`, can we just omit `tabIndex`?  If the user _really_ needs a `tabIndex` of `-1`, they could always provide it via `moreProps`, right? `<RawButton>` will apply a `tabIndex='0'` automatically if it gets an `onClick` prop (unless it also gets `isDisabled`).

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
+import { css } from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+
+const pillColorStyles = {
+  "#d7d7d7": css`
+    color: ${tokens.color.black};
+
+    svg {
+      fill: ${tokens.color.black};
```
mikrotron: If we assume the `svg` is a Paprika `<Icon>`, then it will inherit the font color (above), so I don't think this line is needed.

packages/Pill/src/Pill.styles.js
mikrotron: Tokens _are_ basically just variables 😄 
So I do think it is a good idea to add a new `pill--radius` token.

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+  highRisk: css`
+    background: #cd3c44;
+  `,
+};
+
+const pillSizeStyles = {
+  small: css`
+    border-radius: ${tokens.space} / 50%;
+    font-weight: bold;
+    line-height: ${stylers.spacer(2)};
```
mikrotron: Would a pill ever need to break its text into two lines?

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+    padding: 0 ${tokens.space};
+    ${stylers.fontSize(-3)};
+
+    .aclui-icon::before {
```
mikrotron: ☹️ Do we really need this? Can we just expect any `acl-ui` icons that are used to have their own css applied by the app?

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+      ${stylers.fontSize(-3)};
+    }
+  `,
+  medium: css`
+    border-radius: ${tokens.spaceLg} / 50%;
```
mikrotron: Oh, I see, we will need two `pill--radius` tokens! I'm still okay with that.

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+    line-height: ${stylers.spacer(3)};
+    padding: 0 ${stylers.spacer(4)};
+    ${stylers.fontSize(-1)};
+  `,
+};
+
+const pillStyles = css`
+  align-items: center;
```
mikrotron: Oh, since you have this, maybe the `line-height` above can be omitted?

packages/Pill/src/Pill.styles.js
```diff
@@ -0,0 +1,76 @@
...
+  background: ${props => props.pillColor};
+  color: ${props => props.textColor};
+  display: inline-flex;
+  white-space: nowrap;
```
mikrotron: Why? If it's for truncation, it's missing a couple rules (in fact here's a `styler` for that).

packages/Pill/src/Pill.styles.js
oscarkwan: This makes the height of the pill

mikrotron: Maybe just `height` would make more sense then?

packages/Tokens/src/tokens.yaml
```diff
@@ -148,6 +148,10 @@ popover:
   border-radius: $card--border-radius
   shadow: 0 2px 4px rgba(0, 0, 0, 0.1)
 #
+pill:
+  small-radius: 8px / 50%
+  medium-radius: 12px / 50%
```
mikrotron: I think this would be clearer as `4px` and `6px` 😆

oscarkwan: We don't want that as, border radius is a bit special. It doesn't look right with one value, with this above syntax we get elliptical radius which is what we want

mikrotron: TIL that the CSS `border-radius` property is more advanced than I thought!

alexzherdev: Was it not a joke? 😄 
Also 8px / 50% is 16px 🤓 :trollface:

mikrotron:
When the pill gets squished into a small space, would it be better to truncate? Like this:
Screen Shot 2019-10-10 at 9 58 25 AM

Instead of like this:
Screen Shot 2019-10-10 at 9 56 32 AM

PR #227, opened 2019-10-10
UX-595 button stories for starling

oscarkwan:
### 🛠 Purpose

Curate the Button in Storybook so that it’s a nice, clean view in Storybook. Create some stories designed for the Starling site integration. The idea is to have all of the copy in Sanity so it’s easy for other designers to update. The stories for Starling are not intended to be exhaustive—that can be reserved for a story in Storybook

### 🖥 Screenshots

Screen Shot 2019-10-17 at 2 39 14 PM
Screen Shot 2019-10-17 at 2 39 25 PM
Screen Shot 2019-10-17 at 2 39 40 PM

Screen Shot 2019-10-10 at 7 05 44 PM


nahumzs:
how there is not marge/space between these buttons? 

http://storybooks.highbond-s3.com/paprika/UX-595-Starling-update-button/?path=/story/button-starling--button-states

johnpeele:
For the Button States story:

1. Flat, Secondary, Destructive, Dropdown, Icon only, Icon + Text, and Full Width are not **states**, they are variations or button types.
2. Hover, active, focused, pending, and disabled are **states**

So we might need separate stories for variations and states? Or better explanation of the 2 if they are all combined into 1 story.

packages/Button/stories/examples/ButtonStates.js
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <Story>
+      <Button kind="flat">Flat</Button>
+      <Button kind="secondary">Secondary</Button>
+      <Button kind="destructive">Destructive</Button>
+      <Button isActive>Active</Button>
+      <Button isDisabled>Disabled</Button>
+      <Button isPending>Pending</Button>
+      <Button isDropdown>Dropdown</Button>
+      <Button icon={<PlusIcon />} />
```
mikrotron: You should use `<Button.Icon>` for this example.  The "Basic" story has examples of how it's used, but essentially it's: `<Button.Icon><Icon /></Button.Icon>`

packages/Button/stories/examples/Showcase.js
```diff
@@ -21,7 +21,7 @@ const iconSelections = {
 };
 
 const buttonProps = () => ({
-  children: text("label", "Take Action"),
+  children: text("label", "Buy now"),
```
mikrotron: Gratuitous consumerism :trollface:

pxsmith:
> how there is not marge/space between these buttons?
> 
> http://storybooks.highbond-s3.com/paprika/UX-595-Starling-update-button/?path=/story/button-starling--button-states

Good point @nahumzs. Is there a reason for not including default spacing built into the button? That’s one less thing for consuming developer to have to think about, and one more win for consistency.

nahumzs:
@pxsmith yes there is a reason, none component should affect the space outside of their own container, which means that our component shouldn't have a margin between them. 

The main reason for that is when the component starts composing with other components, if you set a margin can affect the layout of your new compose component and force you to override the style, instead all component should have defined boundaries and play nice with other components, like real legos. 

So space on the layout is the responsibility of the developer, so in this case, I asked why there is not space because didn't know if that was intentional or not. I would think that someone needs to add space between those buttons so they can have some space.

pxsmith:
Awesome thanks for the answer @nahumzs! 👍 The developer will add margin. 😄

packages/Button/stories/examples/ButtonStates.js
pxsmith: @mikrotron is `<Button icon={<PlusIcon />}>Icon</Button>` okay though? I see that no text shows up for `<Button.Icon><PlusIcon />Icon</Button.Icon>`.

packages/Button/stories/examples/Showcase.js
pxsmith: @mikrotron better now? heheheh

PR #230, opened 2019-10-11
PD-13507 SidePanel: fix bodyScroll

AndreyChernykh:
### 🛠 Purpose

After closing the SidePanel, <body> still has `overflow: hidden;` 🐞  

---

### ✏️ Notes to Reviewer

1. `useBodyOverflow` -> `< LockBodyScroll />`
2.  `useEscapeKey` -> `< EscListener  />`

---

### 🖥 Screenshots

---


packages/SidePanel/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/sidepanel",
-  "version": "0.1.14",
+  "version": "0.2.0",
```
nahumzs: you shouldn't bump this one, we will do it when we release via lerna.

packages/SidePanel/src/components/EscListener.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+const propTypes = {
+  /** Callback triggered on ESC key down */
+  onEscKeydown: PropTypes.func.isRequired,
+};
+
+function EscListener({ onEscKeydown }) {
```
alexzherdev: This looks exactly like a custom hook, not a component

packages/SidePanel/src/components/EscListener.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
nahumzs: `fyi` I think this change might conflict with @DanilAgafonov takeover migration PR, but might not be that bad.

packages/SidePanel/src/components/LockBodyScroll.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import ReactDOM from "react-dom";
+
+function LockBodyScroll() {
```
alexzherdev: Not a fan of repeating this file in every component. Perhaps something we could add into `helpers`?

nahumzs: let's merge this one, and maybe @DanilAgafonov can consolidate these two into the helpers in his PR.

Issue #225, opened 2019-10-10
Ensure all react peer dependencies are at least set to `^16.8.0` 


nahumzs:
# Bug Report

Ensure all react peer dependencies are at least set to `^16.8.0`, there are several package.json which are set below the hooks release. 

Solution: 

```json
    "react": "^16.8.0",
    "react-dom": "^16.8.0",
```
--- 

Some packages with peerdendecy less than 16.8.0 :

https://github.com/acl-services/paprika/blob/1e13e22889724b7e629fac2fed7d7f534aa93f66/packages/Icon/package.json

https://github.com/acl-services/paprika/blob/1e13e22889724b7e629fac2fed7d7f534aa93f66/packages/Icon/package.json

https://github.com/acl-services/paprika/blob/cd908fc16fe7b06b6a6f5a2ad0e39a6e50a2ab8a/packages/Heading/package.json

https://github.com/acl-services/paprika/blob/320525ad89d162a636041b7a54071b6ee18b99a5/packages/RawButton/package.json

https://github.com/acl-services/paprika/blob/bc155f4690be5e6d4bc319ff6e236c430f8b0065/packages/Spinner/package.json

https://github.com/acl-services/paprika/blob/bc155f4690be5e6d4bc319ff6e236c430f8b0065/packages/Input/package.json

https://github.com/acl-services/paprika/blob/bb7c55038f274bce52f9fa9615775c34f9f631e6/packages/Popover/package.json

and more ...

sowmiyamuthuraman:
can I work on this?

nahumzs:
We should also check that all our components have ReactDom as a peer dependency as well.

AndreyChernykh:
https://github.com/acl-services/paprika/pull/234

nahumzs:
resolve by #234

PR #234, opened 2019-10-15
Fix peer deps

AndreyChernykh:
### 🛠 Purpose
Ensure all react peer dependencies are at least set to `^16.8.0`

See https://github.com/acl-services/paprika/issues/225
---

### ✏️ Notes to Reviewer
1. Fixed some prop-types warning 
2. Fixed one unit-test warning
---

### 🖥 Screenshots

Warnings: 
image

image

### DIFF
--- a/packages/Button/src/Button.js
+++ b/packages/Button/src/Button.js
@@ -43,7 +43,7 @@ const propTypes = {
   isSubmit: PropTypes.bool,
 
   /** The visual style of the button. */
-  kind: PropTypes.oneOf(Kinds),
+  kind: PropTypes.oneOf(Kinds.ALL),
 
   /** Callback to be executed when the button is clicked or activated by keyboard. Typically required. */
   onClick: PropTypes.func,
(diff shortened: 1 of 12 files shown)


packages/Toast/tests/Toast.spec.js
```diff
@@ -58,7 +58,9 @@ describe("Toast", () => {
     const { getByText, queryByText } = renderComponent({ canAutoClose: true, onClose: handleClose });
 
     expect(getByText(/content/i)).toBeInTheDocument();
-    jest.runAllTimers();
+    act(() => {
+      jest.runAllTimers();
```
nahumzs: thank you!!!!!

packages/Button/src/Button.js
```diff
@@ -43,7 +43,7 @@ const propTypes = {
   isSubmit: PropTypes.bool,
 
   /** The visual style of the button. */
-  kind: PropTypes.oneOf(Kinds),
+  kind: PropTypes.oneOf(Kinds.ALL),
```
nahumzs: I think we are going to change this approach later right @mikrotron ?

mikrotron: I like the idea of having a `types` package that can be consumed by our components, but also by consuming apps to make it easier to use typed props.  

We should probably open a "proposal" issue and get a little feedback before we assume we _will_ change this, but it's probably a safe bet.

PR #235, opened 2019-10-16
PD-13509 fix group and overlay

oscarkwan:
### 🛠 Purpose
- Read comments
---

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

### DIFF
--- a/packages/SidePanel/src/components/Group/Group.styles.js
+++ b/packages/SidePanel/src/components/Group/Group.styles.js
@@ -7,6 +7,9 @@ export const groupCSS = css`
   justify-content: flex-end;
   position: fixed;
   right: 0;
+  width: 100%;
+  z-index: ${props => props.zIndex};
+
   ${props => {
     return `
       top: ${props.offsetY}px;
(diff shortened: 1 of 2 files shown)


packages/SidePanel/src/components/Group/Group.styles.js
```diff
@@ -7,6 +7,9 @@ export const groupCSS = css`
   justify-content: flex-end;
   position: fixed;
   right: 0;
+  width: 100%;
+  z-index: ${props => props.zIndex};
```
oscarkwan: Need to set this to 100% or else, using % width on the side panels inside the group will not work correctly.

Z-index is added, was missing before.

packages/SidePanel/src/components/Overlay/Overlay.js
```diff
@@ -38,9 +38,9 @@ export default function Overlay(props) {
   const vh = visuallyHidden;
 
   return (
-    <Button isSemantic={false} {...moreProps} onClick={handleClick} css={overlayCSS}>
+    <RawButton {...moreProps} onClick={handleClick} css={overlayCSS}>
```
oscarkwan: <Button> caused weird styling on overlay which caused it to not look correct

nahumzs:
🙇‍♀ 🌮 👉 👨

PR #239, opened 2019-10-18
PD-13509 fix side panel overlay with percentage width

oscarkwan:
### 🛠 Purpose

---

### ✏️ Notes to Reviewer
---
- Make it so that the Group and Overlay component has the same z-index
- Make it so that the Overlay is clickable even if the Group has a width of 100%

### 🖥 Screenshots

---
Screen Shot 2019-10-18 at 2 19 15 PM

### DIFF
--- a/packages/SidePanel/src/components/Group/Group.styles.js
+++ b/packages/SidePanel/src/components/Group/Group.styles.js
@@ -17,6 +17,7 @@ export const groupCSS = css`
     `;
   }}
 
+  /* Read more here: https://github.com/acl-services/paprika/pull/239#discussion_r336679762 */
   > div[role="dialog"] {
     visibility: visible;
   }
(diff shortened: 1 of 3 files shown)


packages/SidePanel/src/components/Group/Group.styles.js
```diff
@@ -7,6 +7,7 @@ export const groupCSS = css`
   justify-content: flex-end;
   position: fixed;
   right: 0;
+  visibility: hidden;
```
oscarkwan: This is so that the <group> div is not visible, but the side panels inside it are. Then the overlay can be clickable even if the group has a `width: 100%` to support percentage widths for side panels.

nahumzs: can you add it to the comments on the styles? or just referred with a link to this comment?

oscarkwan: Sure

Issue #141, opened 2019-07-24
Create a variation for <Button /> to be treated like a link


oscarkwan:
# Feature request
## Is your feature request related to a problem?
- The problem is that we want to render a link but with button styles. We don't offer this right now in our button component.

## Describe the solution you'd like
`<Button.Link href="blah" />"
- This will actually render a <a> and be able to pass in `target` and other `<a>` attributes
or it might be best to just create a prop for "openNewTab"

## Describe alternatives you've considered
*A clear and concise description of any alternative solutions or features you've considered.*

## Additional context
*Add any other context or screenshots about the feature request here.*

PR #243, opened 2019-10-23
link button

slee1130:
### 🛠 Purpose

---link button is done but needs to be checked 

### ✏️ Notes to Reviewer
@oscarkwan 
---

### 🖥 Screenshots

---

### DIFF
--- a/packages/Button/src/LinkButton.js
+++ b/packages/Button/src/LinkButton.js
@@ -1,0 +1,49 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+// import iconButtonStyles from "./IconButton.styles";
+import Button from "./Button";
+
+const LinkPropTypes = {
+  children: PropTypes.node.isRequired,
+  kind: PropTypes.oneOf(Button.Kinds.ALL),
+  size: PropTypes.oneOf(ShirtSizes.MEDIUM),
+  url: PropTypes.string,
+};
+
+const LinkDefaultProps = {
+  kind: Button.Kinds.DEFAULT,
+  size: ShirtSizes.MEDIUM,
+  url: null,
+};
+function redirect_to(url) {
+  window.location = url;
+}
+const LinkButton = React.forwardRef((props, ref) => {
+  const buttonProps = {
+    children: null,
+    icon: props.children,
+    isFullWidth: false,
+  };
+  // debugger/
+  // buttonProps.children = <a href={props.url} />
+  return (
+    <Button
+      {...props}
+      {...buttonProps}
+      ref={ref}
+      onClick={() => {
+        redirect_to(props.url);
+      }}
+    >
+      {buttonProps.children}
+    </Button>
+  );
+});
+
+LinkButton.Kinds = Button.Kinds;
+LinkButton.displayName = "LinkButton";
+LinkButton.propTypes = LinkPropTypes;
+LinkButton.defaultProps = LinkDefaultProps;
+
+export default LinkButton;
--- a/packages/Button/stories/examples/Basic.js
+++ b/packages/Button/stories/examples/Basic.js
@@ -7,6 +7,7 @@ import PlusIcon from "@paprika/icon/lib/Add";
 import InfoIcon from "@paprika/icon/lib/InfoCircle";
 import { ButtonStory } from "../Button.stories.styles";
 import Button from "../../src";
+// import LinkButton from "LinkButton";
 
 const DarkBackground = styled.span`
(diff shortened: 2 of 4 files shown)


nahumzs:
Welcome @slee1130  👋 thank you for contributing 🎉

nahumzs:
closes https://github.com/acl-services/paprika/issues/141

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+// import iconButtonStyles from "./IconButton.styles";
```
oscarkwan: Hey @slee1130,

Remember when creating a pull request, it's good to clean up formatting and delete commented out code in your branch that is not needed. This makes it a more clean view for the reviewers to look at.

So just keep in mind to do some "polishing" before you push a branch up for review! Thanks

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,49 @@
...
+const LinkButton = React.forwardRef((props, ref) => {
+  const buttonProps = {
+    children: null,
+    icon: props.children,
+    isFullWidth: false,
+  };
+  // debugger/
+  // buttonProps.children = <a href={props.url} />
+  return (
+    <Button
```
oscarkwan: We actually want to render a `<a>` tag. This is because we don't want to use `window.location`. It isn't good for accessibility.

https://webaim.org/techniques/hypertext/

I think the challenge here is being able to render a <a> but still make it look like a <Button>.

oscarkwan:
In the future for `git` purposes and better readability.

Please name your branches
`OS-<Task you are working on>`

OS standing for open source
cc: @mikrotron @nahumzs

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from "prop-types";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+// import iconButtonStyles from "./IconButton.styles";
+import Button from "./Button";
+
+const LinkPropTypes = {
+  children: PropTypes.node.isRequired,
+  kind: PropTypes.oneOf(Button.Kinds.ALL),
+  size: PropTypes.oneOf(ShirtSizes.MEDIUM),
+  url: PropTypes.string,
```
oscarkwan: Besides a url prop, we want to also support if the link wants to be opened in a new tab.

If the link is opened in a new tab we need to support a security vulnerability.
Read more about it here: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

oscarkwan: After your props are completed, please add a comment above each prop for documentation purposes.

Example: https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.js#L12

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,49 @@
...
+  size: ShirtSizes.MEDIUM,
+  url: null,
+};
+function redirect_to(url) {
+  window.location = url;
+}
+const LinkButton = React.forwardRef((props, ref) => {
+  const buttonProps = {
+    children: null,
+    icon: props.children,
```
oscarkwan: Try not to copy paste too much. For learning purposes its good to understand what you are copy pasting. For example this line wouldn't make sense here.

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,49 @@
...
+    isFullWidth: false,
```
oscarkwan: Same here, i don't think we need to force fullWidth to be false.

packages/Button/stories/examples/Basic.js
```diff
@@ -76,6 +78,9 @@ const ExampleStory = () => (
       <Button kind="link" onClick={clickHandler}>
         link
       </Button>
+      <Button.Link onClick={clickHandler} kind="primary" url="/">
```
oscarkwan: After you fix up the API, have a separate "section" for the Button.Link component for one that opens normally, and another that opens in a new tab.

packages/Button/stories/examples/ButtonVariations.js
```diff
@@ -15,6 +15,7 @@ const ExampleStory = () => {
       <Button isDropdown>Dropdown</Button>
       <Button.Icon>
         <PlusIcon />
+        <Button.Link />
```
oscarkwan: This is not needed here.

PR #244, opened 2019-10-23
add useMountedRef hook

AndreyChernykh:
### 🛠 Purpose

- Add `usePrevious` hook.
---

### ✏️ Notes to Reviewer

- Add `@testing-library/react-hooks` 
- Add `packages/helpers/CHANGELOG.md`
---

### 🖥 Screenshots

---


packages/helpers/test/useMountedRef.spec.js
```diff
@@ -0,0 +1,29 @@
...
+
+import useMountedRef from "../src/hooks/useMountedRef";
+
+describe("useMountedRef", () => {
+  test("should increment counter", () => {
+    const {
+      result: { current: mountedRef },
+      rerender,
+      unmount,
+    } = renderHook(() => {
```
nahumzs: ohhh, so this is like you were coding your component and using the hook inside of it?

nahumzs:
I'm not seeing the `CHANGELOG.md` file, did you omit on purpose?

packages/helpers/test/useMountedRef.spec.js
AndreyChernykh: Sort of

packages/helpers/test/useMountedRef.spec.js
```diff
@@ -0,0 +1,29 @@
+import { renderHook } from "@testing-library/react-hooks";
+
+import useMountedRef from "../src/hooks/useMountedRef";
+
+describe("useMountedRef", () => {
+  test("should increment counter", () => {
```
mikrotron: I'm confused by the description "should increment counter". Was it copied from somewhere else?

AndreyChernykh: Oh no! I'll fix it

packages/helpers/src/hooks/useMountedRef.js
```diff
@@ -0,0 +1,14 @@
...
+  const mountedRef = React.useRef(false);
+
+  React.useEffect(() => {
+    mountedRef.current = true;
+    return function cleanup() {
+      mountedRef.current = false;
+    };
+  }, []);
+
+  return mountedRef;
```
alexzherdev: Late to the game, but why not return `mountedRef.current` here? The end user probably never cares about the ref itself, they're interested in the boolean value of whether we're mounted or not (and so `ref` in the hook name looks irrelevant to me)

AndreyChernykh: @mikrotron @alexzherdev 

Ref is safer to use with effects. Let's say you want to make a request `on first mount`: 

```js
useEffect(() => {
  fetch().then(() => {
    if (ref.current) { // do something }
  })
}, [ref]);
```

Since `ref` is immutable, the change will not trigger an effect to reapply. 

But if we do `const isMounted = useMounted()`

then we have to pass it as parameter. And the effect will be reapplied when the component dismounts.  

```js
useEffect(() => {
  fetch().then(() => {
    if (isMounted) { // do something }
  })
}, [isMounted]);
```

alexzherdev: Hmm I see. In the former case, does eslint warn if you don't specify `ref` as dependency? I think refs are guaranteed to keep referential integrity.
I'm assuming in the latter case, if you don't specify `isMounted`, it should definitely warn.

AndreyChernykh: @alexzherdev Yes, it'll warn you. 
But with `ref` it is safe to add it as a dependency because of `referential integrity`. 

We might think of having a warning instead of an error (as we have now) for that. But I'm worried that people might ignore warnings 🤔

PR #245, opened 2019-10-23
AN-8117 Move l10 component into peer dependency

allison-c:
### 🛠 Purpose

**Breaking change**

Phoenix reported an issue that a paprika label didn't get translated correctly(https://aclgrc.atlassian.net/browse/AN-8117). The label component(from FormElement) was trying to create another <L10n> context instead of using the existing one created by the app. This issue can be solved by moving `@paprika/l10n` into peer dependencies.

### ✏️ Notes to Reviewer

Consumers need to add `@paprika/l10n` dependency. However, they should already have it since paprika components need to be wrapped by it.

### 🖥 Screenshots

---


nahumzs:
@allison-c this would force all consumers to install the L10n component together with styled-components, if this is approved you will also need to update the readme file. 

Why we can't solve this without adding this component as peer dependency?

allison-c:
> @allison-c this would force all consumers to install the L10n component together with styled-components, if this is approved you will also need to update the readme file.
> 
> Why we can't solve this without adding this component as peer dependency?

Currently I only have this solution in my mind. I'm open to any suggestions!

nahumzs:
> was trying to create another context instead of using the existing one created by the app

Do you know why it is trying to create a new context? it should consume the provider that has been declared at the top of your APP isn't?

did you check if your app is by mistaking declaring different L10n providers?

allison-c:
> > was trying to create another context instead of using the existing one created by the app
> 
> Do you know why it is trying to create a new context? it should consume the provider that has been declared at the top of your APP isn't?
> 
> did you check if your app is by mistaking declaring different L10n providers?

I checked the app. I was because some components might not get updated in the app. For example, FormElement in AN is 0.1.4, and the L10n version in `@paprika/form-element 0.1.4` is 0.2.0. So <FormElement> is using the `useI18n` and L10n context from its dependency. But the rest of other parts in the app is using `@paprika/l10  0.2.4`. There're two sets of L10n code, and top L10n context provider is from `@paprika/l10  0.2.4`, but the FormElement try to find the context from `@paprika/l10  0.2.0`

This is the dependency tree in AN
image

Upgrade FormElement in AN will also solve the issue, but that means we need to force developers to upgrade all paprika component version if they bump L10n version

nahumzs:
what about if we upgrade all the L10n version in paprika, and then we upgrade your APP and from this point forward we know that we have to keep L10n in sync in paprika, and if we modify that component is a break change? cc @mikrotron @AndreyChernykh

allison-c:
> what about if we upgrade all the L10n version in paprika, and then we upgrade your APP and from this point forward we know that we have to keep L10n in sync in paprika, and if we modify that component is a break change? cc @mikrotron @AndreyChernykh

L10n version in paprika is correct. I don't think we need to upgrade or fix it. I'm worry about my APP is using an old version of the FormElement component which has a dependency of an older version of L10n. Then I'll have '2' L10n in my code. What if I don't want to upgrade FormElement?

nahumzs:
I discussed offline with @allison-c about this issue, I think for now upgrading her own app is the way to go, let's wait a little bit to see how this problem evolves and see if we need to have the L10n as peer component. 
Also when @alexzherdev comeback can provide another perspective.

alexzherdev:
If we want to have a single `<L10n>` at the top that serves as the provider for both the app **and** paprika, then there needs to be a single L10n package in the tree, thus it must be a peerDependency. Otherwise what happens is the app's copy of L10n creates a context `C1`, renders a provider `C1.Provider`, but then paprika components have their own L10n copy with its own context `C2` and look for it here https://github.com/acl-services/paprika/blob/6fdb61322c16db1d15cdcc5396009683a39b928b/packages/L10n/src/useI18n.js#L31
It's just a coincidence we haven't run into this sooner.

nahumzs:
Seems like the best approach, for now, is to have L10n component as a peer dependency. Let's do that. Please remember to update the readme file specifying this.

Issue #240, opened 2019-10-21
Popover – Self-aware Repositioning


mikrotron:
#### Problem
Sometimes the content of a Popover will change due to an asynchronous event (like lazy-loading data) while the Popover is open, and the size of the Popover.Content will be updated, but the position will not.

#### Suggested Solution
Use a `useLayoutEffect` hook, with a dependency on `children` that will reposition the Popover.

PR #248, opened 2019-10-25
Fix for date picker calendar positioning respecting viewport edges

tristanjasper:
### 🛠 Purpose

Date Picker wasn't positioning correctly and respecting viewport edges.

This solution sets a fixed height and width for the calendar component. 
The height is set to be at the months that have the max height (may and august show more days)

It may be that we can removed these width and height style values if we are able to fix https://github.com/acl-services/paprika/issues/240 at which point if the content changes.

The negative of this solution is there is a small gap when the calendar is shown above the input.

Minor bump required

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

Before
datepicker-positioning-before

After
calendarpos-v2


packages/DatePicker/src/components/Calendar/Calendar.js
```diff
@@ -178,7 +178,7 @@ function Calendar(props) {
   }
 
   const CalendarKey = `${currentMonth && currentMonth.format("YYYY-MM")}/${possibleDate &&
-    possibleDate.format("YYYY-MM")}/${date && date.format("YYYY-MM")}/${isVisible}`;
+    possibleDate.format("YYYY-MM")}/${date && date.format("YYYY-MM")}`;
```
tristanjasper: @allison-c  Was there a particular reason you wanted the calendar to re-render every time someone opened the popup?

packages/DatePicker/stories/DatePicker.stories.js
```diff
@@ -23,9 +23,11 @@ storiesOf("DatePicker", module)
       hasError: boolean("hasError", false),
     });
     return (
-      <Example locale="en" {...datePickerProps()}>
-        <DatePicker.Input {...inputProps()} />
-      </Example>
+      <div style={{ marginTop: "270px" }}>
```
mikrotron: Just a suggestion for the future, I like the `css` prop syntax better:
`<div css={`margin-top: 270px;`}>`

packages/DatePicker/src/components/Calendar/Calendar.js
allison-c: I was trying to force `<SDPController />` re-render to run `getInitialVisibleMonth`.  However I tested your storybook, it works fine now..

allison-c: Oh I think I found the bug

1. Go to http://storybooks.highbond-s3.com/paprika/ux-242-date-picker-positioning/?path=/story/datepicker--datepicker
2. Use Calendar select Oct 30
3. Click Input again and keep clicking next month button
4. Click outside
4. Click Input again and see the Calendar

You can see the Calendar shows future but it should be October cause we didn't select any date in the second time

packages/DatePicker/src/components/Calendar/Calendar.styles.js
```diff
@@ -34,6 +37,9 @@ const iconButtonStyles = css`
 `;
 
 const calendarStyles = css`
+  height: ${calendarHeight}px;
```
mikrotron: This seems to introduce a gap for most months:
Screen Shot 2019-10-25 at 3 21 45 PM

tristanjasper: Yes, as I have mentioned in the description. That is the negative of this solution

PR #249, opened 2019-10-25
Update popover content position based on child change

tristanjasper:
### 🛠 Purpose

Resolves https://github.com/acl-services/paprika/issues/240

Update popover content position on children change. This to fix issues where the content of the popover isn't positioning itself in the right position sometimes or with right size.

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---

### DIFF
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -215,7 +215,6 @@ function DatePicker(props) {
 
   return (
     <Popover
-      style={{ marginTop: "150px" }}
       offset={8}
       {...extendedPopoverProps}
       isOpen={shouldShowCalendar}
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -215,6 +215,7 @@ function DatePicker(props) {
 
   return (
     <Popover
+      style={{ marginTop: "150px" }}
       offset={8}
       {...extendedPopoverProps}
       isOpen={shouldShowCalendar}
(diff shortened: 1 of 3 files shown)


packages/DatePicker/src/DatePicker.js
```diff
@@ -215,6 +215,7 @@ function DatePicker(props) {
 
   return (
     <Popover
+      style={{ marginTop: "150px" }}
```
mikrotron: Don't forget to remove this... I'd suggest adding something like it to `DatePicker.stories.js`

packages/Popover/src/Popover.js
```diff
@@ -438,14 +445,22 @@ class Popover extends React.Component {
...
       </ThemeContext.Provider>
     );
   }
 }
 
+// Todo Refactor this when we convert popover component to use hooks
+function PopoverChildren(props) {
+  React.useLayoutEffect(() => {
+    if (props.children) props.onChildChange();
+  }, [props.children]);
```
alexzherdev: So I think I discussed this with @mikrotron?.. but because `children` are usually react elements, which are JS objects, this will basically trigger any time the popover's parent re-renders. We can probably go with this for now and see whether this causes issues in practice.

tristanjasper: Ok yeah not sure of best solution to prevent that at this point, but se can see what issues arise.

mikrotron: Yeah, let's see. Popovers are closed most of the time, and when they're open, the user is likely focussed on them and not causing re-renders, so it seems like this won't come up too often anyway?

PR #253, opened 2019-10-28
OS-Link button

slee1130:
### 🛠 Purpose

---

### ✏️ Notes to Reviewer
Changed the branch name, prefix with OS and made sure committed the right change I've made.
The rest will be the same as this morning: 
Clean up the codes
Rendering <a> tag to open a new tab 
Added horizontal rule to make the buttons separated
Added styling on Button.stories.styles.js

---

### 🖥 Screenshots

---


packages/Button/src/Button.styles.js
```diff
@@ -20,7 +20,7 @@ const commonStyles = `
   font-weight: bold;
   text-align: center;
   vertical-align: middle;
-
+  text-decoration: none;
```
oscarkwan: add an empty line after

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,58 @@
...
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import Button from "./Button";
+import buttonStyles from "./Button.styles";
+
+const LinkPropTypes = {
+  children: PropTypes.node.isRequired,
+  kind: PropTypes.oneOf(Button.Kinds.ALL),
+  size: PropTypes.oneOf(ShirtSizes.MEDIUM),
+  href: PropTypes.string,
+  text: PropTypes.string,
```
oscarkwan: We should not need a `text` prop but just rely on `children` instead. Let me know if you have any questions

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,58 @@
...
+  href: null,
+  text: "Link",
+  isOpenNewTab: true,
+};
+
+const LinkButton = React.forwardRef((props, ref) => {
+  const buttonProps = {
+    children: null,
+  };
+  console.log(props);
```
oscarkwan: Remove console.logs

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,58 @@
...
+        css={buttonStyles}
+        {...props}
+        {...buttonProps}
+        ref={ref}
+        href={props.href}
+        target="_blank"
+        rel="noopener noreferrer"
+      >
+        {props.text}
+        {buttonProps.children}
```
oscarkwan: can we use the button prop children for the text instead?

packages/Button/src/LinkButton.js
```diff
@@ -0,0 +1,58 @@
...
+  text: "Link",
+  isOpenNewTab: true,
+};
+
+const LinkButton = React.forwardRef((props, ref) => {
+  const buttonProps = {
+    children: null,
+  };
+  console.log(props);
+  if (props.isOpenNewTab) {
```
oscarkwan: I'd create a separate variable that can hold

```
target="_blank"
rel="noopener noreferrer"
```

based on if there is a `prop.isOpenNewTab` or not. Then we don't need to render two different `<a>`. We just render one.

For example
`<a {...isOpenNewTabProps} {...buttonProps} etc....>`

packages/Button/stories/Button.stories.styles.js
```diff
@@ -3,6 +3,7 @@ import { Story } from "storybook/assets/styles/common.styles";
 import tokens from "@paprika/tokens";
 
 // Common Button story styles
+// Added link button story styles
```
oscarkwan: We don't need this comment

packages/Button/stories/Button.stories.styles.js
```diff
@@ -12,4 +13,11 @@ export const ButtonStory = styled(Story)`
       margin-left: ${tokens.space};
     }
   }
+  a,
```
oscarkwan: You just need to add `a, button, ... { margin-left: ${tokens.space}; }`. You don't need an additional block that has the same css rule.

packages/Button/stories/examples/Basic.js
```diff
@@ -78,6 +80,11 @@ const ExampleStory = () => (
       </Button>
     </p>
     <Rule />
+    <p>
+      <Button.Link onClick={clickHandler} isOpenNewTab={false} kind="primary" href="/about" text="Link" />
+      <Button.Link onClick={clickHandler} kind="secondary" href="/about" text="Link that opens in new tab" />
```
oscarkwan: `/about` is not a valid link. Just use google

packages/Button/src/LinkButton.js
slee1130: done

packages/Button/stories/Button.stories.styles.js
slee1130: done

packages/Button/stories/Button.stories.styles.js
slee1130: done

packages/Button/stories/examples/Basic.js
slee1130: done

packages/Button/src/LinkButton.js
slee1130: done

packages/Button/src/LinkButton.js
slee1130: done

packages/Button/src/Button.styles.js
slee1130: done

packages/Button/src/LinkButton.js
slee1130: done

Issue #241, opened 2019-10-21
Popover – Follow-up Issues


mikrotron:
## Feature requests

1) `<Popover.Content>` should not require `props.children` (should allow `null`)
  https://github.com/acl-services/paprika/blob/master/packages/Popover/src/components/Content/Content.js#L10

2) `width` and `max-width` should not be applied to inline styles. It would also be nice to have a `minWidth` prop to accompany the `maxWidth` for dynamic content.
  https://github.com/acl-services/paprika/blob/master/packages/Popover/src/components/Content/Content.js#L79

3) `moreProps` should be propagated on `<Popover.Trigger>` and `<Popover.Card>` root nodes (this will allow classes and testid attributes to be used).
  https://github.com/acl-services/paprika/blob/master/packages/Popover/src/components/Trigger/Trigger.js#L48

PR #256, opened 2019-10-31
UX-627 Allow Popover.Content to have no children 🐐

mikrotron:
### 🛠 Purpose
Primarily to allow `<Popover.Content>` `children` prop to be `null` to unblock a bug in Storyboards.

Resolves https://github.com/acl-services/paprika/issues/241

### ✏️ Notes
- **[PATCH]** update to `<Popover>`.
- Added a "Dynamic Content" story that renders `<Popover isOpen>` with no `children`, with a `<Spinner>`, then with actual content.
- **BONUS:** Propagate `moreProps` for `<Popover.Trigger>` and `<Popover.Card>`.
- **BONUS:** Add a `minWidth` prop for `<Popover>`.

### 🖥 Screenshots
popover

📙 Storybook: http://storybooks.highbond-s3.com/paprika/UX-627--popover-content-null/?path=/story/popover--with-dynamic-content


packages/Popover/src/Popover.js
```diff
@@ -455,10 +461,13 @@ class Popover extends React.Component {
...
 // Todo Refactor this when we convert popover component to use hooks
 function PopoverChildren(props) {
+  const { children, onChildChange } = props;
+
   React.useLayoutEffect(() => {
-    if (props.children) props.onChildChange();
-  }, [props.children]);
-  return props.children;
+    if (children) onChildChange();
+  }, [children, onChildChange]);
```
tristanjasper: @mikrotron Can you explain why you needed to add the `onChildChange` prop here? Is that so it detects if that prop is dynamically added?

mikrotron: In _theory_ that prop could change, and this should react on that.  Really though, it was just to make eslint happier (it was warning about it).

packages/Popover/src/components/Content/Content.js
```diff
@@ -64,6 +65,10 @@ const Content = React.forwardRef((props, ref) => {
     }, parseInt(PopoverConstants.transition, 10));
   }
 
+  function cssValue(value) {
```
tristanjasper: `convertToPx` instead of `cssValue`? A bit more descriptive

packages/Popover/src/components/Trigger/Trigger.js
```diff
@@ -45,10 +45,14 @@ function Trigger(props) {
...
   }
 
-  const { children, a11yText } = props;
+  const { children, a11yText, ...moreProps } = props;
 
   if (typeof children === "function") {
-    return <React.Fragment>{children(handleTriggerEvent, { "aria-describedby": content.ariaId })}</React.Fragment>;
+    return (
+      <React.Fragment>
+        {React.cloneElement(children(handleTriggerEvent, { "aria-describedby": content.ariaId }), { ...moreProps })}
```
tristanjasper: so if the child of the trigger component is a `node` it won't receive the `...moreProps`. Is that intentional?

tristanjasper: oh i see you add below

packages/Popover/src/components/Content/Content.js
tristanjasper: Or i guess you are handling other values such as `100%` in which case i guess you won't want the conversion...

mikrotron: Yeah, it's only adding `px` to numbers, if it's something like `100%` or `auto`, then I leave it as is.

PR #257, opened 2019-10-31
AN-8115 Adding space before form element optional label for edge cases

allison-c:
### 🛠 Purpose

Sometimes the label is an extremely long string like this:
image
We can avoid the label cut-off by adding a space before `optional` `required` label.
Note: Japanese and Chinese don't need this cause they'll be broke up naturally.

### ✏️ Notes to Reviewer

**Patch update**

### 🖥 Screenshots

Before:
image

After:
image

### DIFF
--- a/packages/L10n/src/locales-js/de.js
+++ b/packages/L10n/src/locales-js/de.js
@@ -56,8 +56,8 @@ const locales = {
         "next": "Nächste Seite"
       },
       "formElement": {
-        "optional": "(optional)",
-        "required": "(erforderlich)"
+        "optional": " (optional)",
+        "required": " (erforderlich)"
       },
       "prevNextArrowButton": {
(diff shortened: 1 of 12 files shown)


mikrotron:
If you put an additional space (`{' '}`) at the beginning of this line:
https://github.com/acl-services/paprika/blob/master/packages/FormElement/src/components/Label/Label.js#L31

Would the _(optional)_ part align with the label above it (and not appear a wee bit indented?)

allison-c:
> If you put an additional space (`{' '}`) at the beginning of this line:
> https://github.com/acl-services/paprika/blob/master/packages/FormElement/src/components/Label/Label.js#L31
> 
> Would the _(optional)_ part align with the label above it (and not appear a wee bit indented?)

That's a good idea, I'll close this PR and open another one

PR #258, opened 2019-10-31
L10N-2019-10-31-30578 - translations batch

alexandermccate:


alexandermccate:
@allison-c I applied the fixes for the ES PT date format.

PR #260, opened 2019-11-01
no divider yet

slee1130:
### 🛠 Purpose

---

### ✏️ Notes to Reviewer

--- I am in still WIP but please give me some feedbacks 

### 🖥 Screenshots

---

### DIFF
--- a/packages/Heading/src/Heading.js
+++ b/packages/Heading/src/Heading.js
@@ -1,6 +1,5 @@
 import React from "react";
 import PropTypes from "prop-types";
-import classNames from "classnames";
 import headingStyles from "./Heading.styles";
 
 const propTypes = {
@@ -9,9 +8,6 @@ const propTypes = {
 
   /** Heading content is required */
   children: PropTypes.node.isRequired,
-
-  /** Optional custom class name */
-  className: PropTypes.string,
 
   /** Optional display level(1-6) affects styles only */
   displayLevel: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
@@ -41,7 +37,6 @@ const propTypes = {
 
 const defaultProps = {
   a11yText: null,
-  className: null,
   displayLevel: null,
   domRef: () => {},
   hasDivider: false,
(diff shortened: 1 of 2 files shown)


packages/Heading/src/Heading.js
```diff
@@ -53,55 +48,46 @@ const defaultProps = {
...
-    const elementProps = this.getElementProps(safeDisplayLevel, safeLevel);
-
-    return (
-      <div css={headingStyles} {...elementProps} as={isSemantic ? `h${safeLevel}` : "div"}>
-        {this.renderHeadingContent()}
-        {hasDivider ? divider : null}
-      </div>
-    );
-  }
+function getElementProps(safeDisplayLevel, safeLevel, props) {
```
oscarkwan: safeDisplayLevel is no being used in this function anymore. I think we still ned to use this. The logic before was `safeDisplayLevel || safeLevel`

oscarkwan: I believe this is why the display level is not working properly

packages/Heading/src/Heading.js
```diff
@@ -53,55 +48,46 @@ const defaultProps = {
...
+
+function renderHeadingContent(a11yText, children) {
+  return a11yText ? <span aria-hidden>{children}</span> : children;
+}
+
+function Heading(props) {
+  const { a11yText, children, hasDivider, isSemantic, level, displayLevel } = props;
+  const safeLevel = safeValue(level);
+  const safeDisplayLevel = displayLevel ? safeValue(displayLevel) : null;
+  const divider = <span className="heading__divider" />;
```
oscarkwan: You could export a style just for this <span>

For ex
```
<span css={dividerStyles} />
```

Then you export dividerStyles from the styles file. You can see some examples. We do this in other components

packages/Heading/src/Heading.styles.js
```diff
@@ -14,64 +16,75 @@ const headingStyles = () => `
...
-  }
+  }`,
+};
 
-  &.heading--has-underline {
+// underline
+const underline = `
     border-bottom: 1px solid ${tokens.border.color};
-    &.heading--level-1 {
+    h1 {
```
oscarkwan: We can't use these because we have a prop that is `isSemantic`, which when false does not use . h1 to h6. It uses <div> to avoid styles override by global css instead.

packages/Heading/src/Heading.styles.js
```diff
@@ -14,64 +16,75 @@ const headingStyles = () => `
...
-  }
+// light
+
+const light = `
+font-weight: 400;
+`;
+
+const headingStyles = props => `
+${defaultHeadingStyles}
+${heading[`HEADING${props.level}`]}
```
oscarkwan: `HEADING_LEVEL_${props.level}` perhaps?

packages/Heading/src/Heading.styles.js
```diff
@@ -14,64 +16,75 @@ const headingStyles = () => `
...
+${props.isHidden ? hidden : ""}
+${props.hasUnderline ? underline : ""}
+${props.isLight ? light : ""}
```
oscarkwan: using null instead of empty string may be better.

packages/Heading/src/Heading.styles.js
```diff
@@ -1,7 +1,9 @@
 import tokens from "@paprika/tokens";
 import stylers from "@paprika/stylers";
 
-const headingStyles = () => `
+// default heading styles
```
oscarkwan: We don't need comments if it is self-explanatory

slee1130: done

packages/Heading/src/Heading.styles.js
slee1130: done

packages/Heading/src/Heading.styles.js
slee1130: done

packages/Heading/src/Heading.styles.js
slee1130: done

packages/Heading/src/Heading.js
slee1130: done

nahumzs:
Can we improve the PR name as well the description, is kind of difficult to know what is the intention for this PR :)

PR #261, opened 2019-11-01
Ux 603 New Radio component

tristanjasper:
### 🛠 Purpose

Add the radio component to Paprika.
http://storybooks.highbond-s3.com/paprika/ux-603-radio-component

Tests to be added in future PR

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---
image


johnpeele:
@tristanjasper You might already be aware, but the `checked` state for the Large size is a bit off.

Screen Shot 2019-11-04 at 12 21 24 PM

tristanjasper:
> @tristanjasper You might already be aware, but the `checked` state for the Large size is a bit off.
> 
> Screen Shot 2019-11-04 at 12 21 24 PM

@johnpeele What browser were you seeing this out of interest? I'll look into it.

packages/Radio/stories/examples/Grouping.js
```diff
@@ -0,0 +1,47 @@
...
+
+const radioProps = () => ({
+  size: select("size", ShirtSizes.DEFAULT, "medium"),
+  isDisabled: boolean("isDisabled", false),
+  isChecked: boolean("isChecked", false),
+  canDeselect: boolean("canDeselect", false),
+  a11yText: text("a11yText", ""),
+});
+
+const ExampleStory = props => {
```
nahumzs: To me sound that this Group radio button should be a uncontrolled component, my concern is that is too much boiler plate when what a consumer pretty much wants it o know which component is checked.

I would approach this API like the following:
```jsx
<Radio.Group 
   defaultCheck={(check) => check.value.id === "something"} 
   onChange={(value) => { console.log(value) }}
>
   <Radio value={whatever}>First</Radio>
   <Radio value={whatever}>second</Radio>
   <Radio value={whatever}>third</Radio>
<Radio.Group>
```

This achieve two things one removes the need of any state from the consumer and give the ability to set the initial value via a function that can be used to check which checkbox should be initially selected. 

what do you think? @mikrotron @tristanjasper

nahumzs: Also this way we are more align the way that most input component works like having a onChange value. like the input, select and pretty much any react component.

tristanjasper: Sure that is another option, I guess there isn't really a case for a single radio component to be needed, ie without the surrounding radioGroup container so yeah would be cleaner. I'll move forward with this suggestion

tristanjasper: > Also this way we are more align the way that most input component works like having a onChange value. like the input, select and pretty much any react component.

Agreed

johnpeele:
@tristanjasper I am using Brave, which is a version of Chrome.

packages/Radio/README.md
```diff
@@ -0,0 +1,31 @@
...
+`> npm install --save @paprika/radio`
+or
+`> yarn add @paprika/radio`
+
+### Usage
+
+```js
+import Radio from "@paprika/radio";
+
+<Radio onChange={handleChange} checkState={checkedStateValue>
```
nahumzs: can you add an example of radio group please, I think that one would be the most common use case.

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,68 @@
...
+const defaultProps = {
+  a11yLabelledByText: "",
+  canDeselect: false,
+  children: null,
+  defaultCheck: () => {},
+  isDisabled: false,
+  size: ShirtSizes.MEDIUM,
+};
+
+const Group = ({ a11yLabelledByText, canDeselect, children, defaultCheck, ...moreGroupProps }) => {
```
nahumzs: please switch this to a form a function instead of arrow function and destructured the properties on the body.

function Group(props) {
  const {a11yLabelledByText, canDeselect, children, defaultCheck, ...moreGroupProps} = props;
}

packages/Radio/stories/examples/Showcase.js
```diff
@@ -0,0 +1,41 @@
...
+const ExampleStory = props => {
+  return (
+    <RadioStory>
+      <Heading level={1} displayLevel={2} isLight>
+        Showcase
+      </Heading>
+      <Tagline>Use the knobs to tinker with the props.</Tagline>
+      <Rule />
+      <Radio.Group
+        defaultCheck={check => check.props.value.id === "value 1"}
```
nahumzs: 💯 thank you 😻

nahumzs:
@johnpeele We officially support chrome, firefox, Edge, Safari, and we try to do our best to make it render as similar as possible on IE11. I think is quite expensive to try to adjust each browser. cc @mikrotron

packages/Radio/src/components/Group/Group.js
tristanjasper: ok out of interest what is the reason for the preference?

packages/Radio/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@paprika/radio",
+  "version": "0.0.1",
```
mikrotron: `v0.1.0`, please

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
+import React from "react";
+import PropTypes from "prop-types";
+import uuid from "uuid/v4";
```
mikrotron: Wanna try `nanoid` instead here? (https://github.com/ai/nanoid)
Danil suggested it as a lighter weight alternative.

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+
+const propTypes = {
+  /** aria-labelledby prop on the containing group element */
+  a11yLabelledByText: PropTypes.string,
```
mikrotron: `a11yText` please (this is a way to insulate app developers from the implementation details).

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,69 @@
...
+
+  /** Can deselect any radio */
+  canDeselect: PropTypes.bool,
+
+  /** Function used to evaluate which radio is selected by default. */
+  defaultCheck: PropTypes.func,
+
+  /** The individual radio itemss. */
```
mikrotron: nit: ss 🐍

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,69 @@
...
+
+const propTypes = {
+  /** aria-labelledby prop on the containing group element */
+  a11yLabelledByText: PropTypes.string,
+
+  /** Can deselect any radio */
+  canDeselect: PropTypes.bool,
+
+  /** Function used to evaluate which radio is selected by default. */
+  defaultCheck: PropTypes.func,
```
mikrotron: 🤔 ...

packages/Radio/src/Radio.styles.js
```diff
@@ -0,0 +1,196 @@
...
+const smallRadioSize = tokens.radio.sizeSm;
+const mediumRadioSize = tokens.radio.sizeMd;
+const largeRadioSize = tokens.radio.sizeLg;
+
+const getHalfSizeCss = sizeCss => `${toInt(sizeCss) / 2}px`;
+const smallRadioHalfSize = getHalfSizeCss(smallRadioSize);
+const mediumRadioHalfSize = getHalfSizeCss(mediumRadioSize);
+const largeRadioHalfSize = getHalfSizeCss(largeRadioSize);
+
+const styles = {
```
mikrotron: A lot of this file is copied from the `<Checkbox>`.  I don't mind the copy pasta as phase one, but do you have any ideas yet for how to DRY this up in a later PR perhaps?

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,69 @@
...
+  const { a11yLabelledByText, canDeselect, children, defaultCheck, ...moreGroupProps } = props;
+  const defaultCheckedId = React.Children.toArray(children).find(defaultCheck).props.value.id;
+  const [checkedId, setCheckedId] = React.useState(defaultCheckedId || null);
+  const deselectableId = id => (checkedId === id ? null : id);
+  const handleRadioClick = id => {
+    setCheckedId(canDeselect ? deselectableId(id) : id);
+  };
+
+  return (
+    <div role="radiogroup" aria-labelledby={a11yLabelledByText} data-pka-anchor="radio.group">
```
mikrotron: `aria-labelledby` accepts DOM `ids` as its value.  What you need here is `aria-label`.

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
...
+        onChange={() => {}}
+        onClick={onClick}
+        checked={isChecked}
+        disabled={isDisabled}
+        id={radioId}
+        onKeyUp={handleKeyUp}
+        onKeyDown={handleKeyDown}
+        ref={inputRef}
+        type="radio"
+        {...inputProps}
```
mikrotron: This seems a bit odd here, maybe all the props should actually go up above on line 56?

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
...
+      onClick(event);
+    }
+  };
+
+  const inputProps = {};
+  if (a11yText) inputProps["aria-label"] = a11yText;
+  return (
+    <div data-pka-anchor="radio" css={radioStyles} {...styleProps} {...moreProps}>
+      <input
+        onChange={() => {}}
```
mikrotron: Is this needed?

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
...
+  isDisabled: false,
+  size: ShirtSizes.MEDIUM,
+};
+
+function Radio(props) {
+  const { a11yText, children, isChecked, canDeselect, isDisabled, size, onClick, ...moreProps } = props;
+  const radioId = React.useRef(uuid()).current;
+  const inputRef = React.useRef(null);
+
+  const styleProps = {
```
mikrotron: This would probably be more readable if it were placed down by `inputProps`.

packages/Radio/stories/examples/Showcase.js
```diff
@@ -0,0 +1,41 @@
...
+        onChange={value => {
+          console.log(value);
```
mikrotron: Recommended: use Storybook `action` here, like this:

https://github.com/acl-services/paprika/blob/master/packages/Button/stories/examples/Basic.js#L19

packages/Radio/src/components/Group/Group.js
mikrotron: This might be the most significant aspect I would like to challenge... This seems a bit awkward, and I like the way the API for the `<ListBox>` handles this better, though I'm not so familiar with the implementation details.  

This prop could be omitted in favour of adding an `isSelected` (or `defaultSelected`) prop directly on the `<Radio>` item that is selected. Consumers will likely map through an array in their application data to generate those `<Radios>` and so adding `defaultSelected` should be pretty straightforward.  But then the `<Radio>` component is going to have to set that in the state for the `RadioGroup` somehow... 🤔 

Does this approach make sense to you, @nahumzs?

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  const inputProps = {};
+  if (a11yText) inputProps["aria-label"] = a11yText;
+  return (
+    <div data-pka-anchor="radio" css={radioStyles} {...styleProps} {...moreProps}>
+      <input
+        onChange={() => {}}
+        onClick={onClick}
+        checked={isChecked}
```
mikrotron: I'm a little unsure _why_, but with this `checked` attribute, I don't see it being added to the DOM.  But if I use `defaultChecked` instead, it works as expected.  Maybe it's a React thing, like `className` or `htmlFor`... 🤷‍♂️

packages/Tokens/README.md
```diff
@@ -7,7 +7,7 @@ or
 # How to use it
 
 Run
-`> yarn start`
+`> yarn pretranspile`
```
mikrotron: 🙇

packages/Radio/src/Radio.styles.js
```diff
@@ -0,0 +1,196 @@
...
+  line-height: ${({ hasLabel }) => (hasLabel ? lineHeightValue(-1) : "0")};
+  margin: ${tokens.space} 0;
+  position: relative;
+
+  input[type="radio"] {
+    ${visuallyHidden};
+
+    &:focus + label::before {
+      box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
+      border-color: ${tokens.highlight.active.noBorder.borderColor};
```
mikrotron: This doesn't seem to be taking affect, the border is still black when focussed:

Screen Shot 2019-11-25 at 1 26 50 PM

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
...
+        onKeyDown={handleKeyDown}
+        ref={inputRef}
+        type="radio"
+        {...inputProps}
+      />
+      <label onKeyUp={handleKeyUp} className={canDeselect ? "deselectable" : ""} htmlFor={radioId}>
+        {children}
+
+        {canDeselect ? (
+          <CheckIcon className="radio-icon" aria-hidden data-pka-anchor="radio.icon.check" />
```
mikrotron: Can we export the classes from `Radio.styles.js` and apply directly instead of using `classNames`? 

Same for the `<div>` check below.

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,69 @@
...
+      {React.Children.map(children, (child, index) => {
+        if (child && child.type && child.type.displayName === "Radio") {
+          const childKey = { key: `Radio${index}` };
+          return React.cloneElement(child, {
```
alexzherdev: I suggest reviewing this to see if we can migrate to a context-based structure instead, see https://github.com/acl-services/paprika/pull/134/commits/f5063084fb5c90ab0477b913fc0ac8bd7b759b04 for an example. `cloneElement` is the least composition-friendly approach.

packages/Radio/src/Radio.js
tristanjasper: image
Is showing for me?

packages/Radio/src/Radio.js
tristanjasper: without it i do get a warning 
image
however i'll change it to have the `readOnly` attribute instead

packages/Radio/src/Radio.js
mikrotron: Ah yes, on initial load, the first `<input>` does have `checked`.  But when you change it, the DOM doesn't change.  I think it may be related to the other warning you were seeing:
https://github.com/acl-services/paprika/pull/261#discussion_r351460309

packages/Radio/src/Radio.js
tristanjasper: Tried taking this a step further and making some changes as suggested but quickly find it getting a little messy and not necessarily easier to understand. I think perhaps that could do with some refactoring in a secondary Pr but take a look at the in progress branch i have - https://github.com/acl-services/paprika/commit/297f96081768df1aa28a89ec408da2af3f27902f as I was having some issues with the disabled state

packages/Radio/src/Radio.js
alexzherdev: It's the difference between attributes and properties described in https://github.com/facebook/react/issues/13525#issuecomment-417818906 under "DOM Properties Are Not Just "Attributes in JS". Looks like in this case React manipulates the property, not the attribute, so you see no change in the inspector.
radio

tristanjasper: Yeah strange, I see what you mean now. Bit of a puzzler as to why the dom element isn't updating correctly and showing the attribute changes....

tristanjasper: Ok thanks @alexzherdev

tristanjasper: @mikrotron So i'm not sure its a problem that the dom element attributes are not updated, testing with voiceover i can still hear it announce that the radio button is selected, so I think this is fine?

packages/Radio/src/components/Group/Group.js
tristanjasper: Ok I'll look into changing to use this approach in a follow up PR, but I agree would be better.

packages/Radio/src/Radio.styles.js
tristanjasper: Not yet but I'll give it some thought in the next refactor where I remove any classes being applied on elements....

packages/Radio/src/components/Group/Group.js
tristanjasper: I see in the list box option component we have both `defaultIsSelected` and `isSelected` props available, which i find a bit confusing. @nahumzs What was the reason again for including both? Do you think that would be beneficial to introduce both here?

packages/Radio/src/Radio.js
mikrotron: I think you're on the right track there.  Maybe once you merge this one I'll take that branch and try finesse it into a viable PR.

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
...
+    size,
+  };
+
+  const inputProps = {
+    readOnly: true,
+    onClick,
+    checked: isChecked,
+    disabled: isDisabled,
+    id: radioId,
+    name: children,
```
mikrotron: This isn't quite right.  All the Radios in this group need to have the same name.  It should be a prop on `<Radio.Group>`, but if it's not provided it can be a `nanoid` generated by the Group component and provided to each Radio.

packages/Radio/src/Radio.js
```diff
@@ -0,0 +1,90 @@
...
+
+const propTypes = {
+  a11yText: PropTypes.string,
+  children: PropTypes.node,
+  isChecked: PropTypes.bool,
+  isDisabled: PropTypes.bool,
+  canDeselect: PropTypes.bool,
+  onClick: PropTypes.func.isRequired,
+  size: PropTypes.oneOf(ShirtSizes.DEFAULT),
+};
```
mikrotron: Should `value` be a required prop?

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,71 @@
...
+    const id = child.props.value.id;
+    onChange(id);
+    setCheckedId(canDeselect ? deselectableId(id) : id);
+  };
+
+  return (
+    <div role="radiogroup" aria-labelledby={a11yText} data-pka-anchor="radio.group">
+      {children.map((child, index) => {
+        if (child && child.type && child.type.displayName === "Radio") {
+          const childKey = { key: `Radio${index}` };
```
mikrotron: I know you're not supposed to use `index` as the `key`, but not sure what else you could use here 🤔 
Maybe if `value` was a required prop of `Radio`, then the `key` could  be a jsonified version of `child.value`?

packages/Radio/src/components/Group/Group.js
nahumzs: For the Listbox the reason is that sometimes you don't want to handle the isSelected prop for the component you want it to be uncontrolled and only listening for the onChange event (defaultIsSelected); but also there are cases where there is a need for the consumer to fully control the isSelected prop (control component).

`Btw` Most of react components are controlled and uncontrolled components like the `<input />` where you can use defaultValue or value. 

Handling both cases on the listbox required to evaluate each item and useEffect to listening for changings to synchronize the internal state. (because listbox has an useReducer)

In contrast in the listBoxBrowser I'm using a function instead of adding a prop to each element, the idea of doing that was as inspiration after talking about this problem with Andrei. 

Extracting which element is selected and which no to a function make the component more flexible, now we don't have to care about how to store the values and the consumer can use any kind of data structure to store the selected values.

It is more flexible in my experience and removes complexity from the author and the consumer. 

So for this specific component, might be easier for the consumer to use `isSelected`, `defaultSelected` to each individual checkbox. For that I'm sorry because I suggested using the function approach.  

But in a more complex scenario is easier to go with the function pattern. 

Let me know if you need any assistance to make this change and I can help you. Sorry again to suggested the function approach at the beginning.

packages/Radio/src/components/Group/Group.js
alexzherdev: This would be obsoleted if https://github.com/acl-services/paprika/pull/261#discussion_r351053143 is implemented

packages/Radio/src/Radio.js
```diff
@@ -12,8 +12,9 @@ const propTypes = {
   isChecked: PropTypes.bool,
   isDisabled: PropTypes.bool,
   canDeselect: PropTypes.bool,
-  onClick: PropTypes.func.isRequired,
+  onClick: PropTypes.func,
   size: PropTypes.oneOf(ShirtSizes.DEFAULT),
+  value: PropTypes.shape({ id: PropTypes.string }).isRequired,
```
mikrotron: Hm, how about:

```js
PropTypes.oneOfType([
  PropTypes.bool,
  PropTypes.string,
  PropTypes.number,
  PropTypes.shape({ id: PropTypes.string }),
]).isRequired
```

mikrotron: This would require a bit of logic when you use the `value` as the `key`, but gives the consumer more flexibility.

packages/Radio/src/components/Group/Group.js
```diff
@@ -30,21 +27,33 @@ const defaultProps = {
...
 function Group(props) {
-  const { a11yText, canDeselect, children, defaultCheck, onChange, ...moreGroupProps } = props;
-  const defaultCheckedId = React.Children.toArray(children).find(defaultCheck).props.value.id;
-  const [checkedId, setCheckedId] = React.useState(defaultCheckedId || null);
-  const deselectableId = id => (checkedId === id ? null : id);
+  const { a11yText, canDeselect, children, onChange, ...moreGroupProps } = props;
+  const defaultCheckedIndex = React.Children.toArray(children).findIndex(child => child.props.defaultIsSelected);
+  const selectedIndex = React.Children.toArray(children).findIndex(child => child.props.isSelected);
+
+  const defaultIndex = () => {
```
mikrotron: Why is this a function?

packages/Radio/src/components/Group/Group.js
tristanjasper: @alexzherdev I'm unsure how that approach would work if we are using the index to evaluate white child is selected...

packages/Radio/src/components/Group/Group.js
tristanjasper: Previously I was evaluating other props as well, changed to just use the variable now

packages/Radio/src/Radio.js
tristanjasper: Yep, no longer using the value however as consumer can work with the `activeIndex` to get the data values

packages/Radio/src/Radio.js
tristanjasper: No longer passing as a prop

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,77 @@
...
+  a11yText: "",
+  canDeselect: false,
+  children: null,
+  isDisabled: false,
+  size: ShirtSizes.MEDIUM,
+};
+
+function Group(props) {
+  const { a11yText, canDeselect, children, isDisabled, onChange, ...moreGroupProps } = props;
+  const defaultCheckedIndex = React.Children.toArray(children).findIndex(child => child.props.defaultIsChecked);
```
alexzherdev: This still assumes `Radio` is a direct child of `Group` which defeats the purpose of using context. (Recommend to create a story with code like
```js
<Radio.Group>
  <>
    <Radio />
    <Radio />
  <>
</Radio.Group>
```
to see that it doesn't work)
If it's acceptable to put the checked status on the group itself (i.e. `checkedIndex` prop), then I think context can still be used. Otherwise we should revert that change since it's not achieving what we want.

tristanjasper: I would have thought it is ok to expect the usage of this component is to always use Radio components as the siblings? I understand that there will be issues if they use a different child but is that a scenario that we need to cover? We do provide storybook examples and a readme to show correct usage...

tristanjasper: We could throw an error if the wrong child type was passed as well..

mikrotron: What if you need side-by-side radios though? You'll need some flexbox divs in there for layout.

tristanjasper: Perhaps we will need to revisit the required api in a future PR. I know in the case of formElement we pass an `isInline` prop to dictate the layout rather than style the children directly,  - https://github.com/acl-services/paprika/blob/master/packages/FormElement/src/FormElement.js#L34 But perhaps we are not too keen on that approach? Will need discussion. For now I have changed it back to use `cloneElement` rather than context. Discussing with @alexzherdev we have established that if we wanted to retain the index to dictate which radio is selected (we are also need to track which radio is deselectable) we would perhaps have to go down the route of https://gist.github.com/ryanflorence/10e9387f633f9d2e6f444a9bddaabf6e#option-4-the-unholy-option.

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,76 @@
...
+  const deSelectableIndex = index => (checkedIndex === index ? null : index);
+  const name = nanoid();
+  const handleRadioClick = index => {
+    onChange(index);
+    setCheckedIndex(canDeselect ? deSelectableIndex(index) : index);
+  };
+
+  return (
+    <div role="radiogroup" aria-labelledby={a11yText} data-pka-anchor="radio.group">
+      {children.map((child, index) => {
```
mikrotron: I think we should still use `React.Children.map()` here, to be a little more defensive.

packages/Radio/src/components/Group/Group.js
```diff
@@ -0,0 +1,76 @@
...
+  const { a11yText, canDeselect, children, isDisabled, onChange, ...moreGroupProps } = props;
+  const defaultCheckedIndex = React.Children.toArray(children).findIndex(child => child.props.defaultIsChecked);
+  const selectedIndex = React.Children.toArray(children).findIndex(child => child.props.isChecked);
+
+  const [checkedIndex, setCheckedIndex] = React.useState(defaultCheckedIndex);
+  if (selectedIndex !== -1 && selectedIndex !== checkedIndex) {
+    setCheckedIndex(selectedIndex);
+  }
+
+  const deSelectableIndex = index => (checkedIndex === index ? null : index);
```
mikrotron: nit: `deSelectableIndex` → `deselectableIndex`

tristanjasper: Ok, i've changed to `getDeselectableIndex` as it is a function so I think that is more appropriate

PR #263, opened 2019-11-04
Os heading style components refactored

slee1130:
### 🛠 Purpose

---

### ✏️ Notes to Reviewer

** Important: this heading styled component is a breaking change! **

--- 1. heading paddings are fixed
--- 2. safeDisplayLevel and safeLevel is added in the function, inversed heading order is fixed
--- 3. CSS {dividerStyles} is done
--- 4. &.heading--level-1 … section is changed to avoid styles override by global CSS instead
--- 5. changed to HEADING_LEVEL_${props.level} 
--- 6. stick to "" instead of null 
--- 7. variables are declared in alphabetical order
--- 8. no console.log/ or commented out coding is left
--- 9. no self-explanatory comments

  

### 🖥 Screenshots

--- 
Screen Shot 2019-11-04 at 12 58 19 PM
Screen Shot 2019-11-04 at 12 57 54 PM


packages/Heading/src/Heading.styles.js
```diff
@@ -14,64 +14,64 @@ const headingStyles = () => `
...
+  } else {
+    result += `padding-bottom: 2px;`;
   }
+  return result;
+};
 
-  &.heading--is-hidden {
-    ${stylers.visuallyHidden}
-  }
+const hidden = `
```
oscarkwan: Still some indentation issues in this file, let me know if you are confused about it

oscarkwan:
@nahumzs Can i get a quick look from you 🙇

packages/Heading/README.md
```diff
@@ -0,0 +1,68 @@
+## Heading
+
+The `<Heading>` component provides heading styles without className, which is already implemented in heading.js so can be used with the attributes to control the styling without importing from Heading.styles.js.If displayLevel exists, the heading will display <h> tag with displayLevel, otherwise it will display <h> tag with level.
```
oscarkwan: We want this to be human readable friendly and very simple. We don't need to discuss technical implementations here. Just What the component is.

packages/Heading/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+It is rendered as a `<h> tag`by default, but can also be rendered as a generic `<div>`, if global CSS conflicts are an issue (when `isSemantic` prop is set to `false`).
```
oscarkwan:
```suggestion
It is rendered as a `<h> tag` by default, but can also be rendered as a generic `<div>`. Set `isSemantic` to false if there are global CSS conflicts
```

oscarkwan:
@mikrotron Hm, i wonder if this is a breaking change. do we rely on those class names in consuming apps? For overwriting purposes?

packages/Heading/src/Heading.styles.js
```diff
@@ -14,64 +14,64 @@ const headingStyles = () => `
...
+  border-bottom: 2px solid ${tokens.color.blackLighten70};
+  flex: 1;
+  margin-left: ${tokens.space};
 `;
 
-export default headingStyles;
+export const headingStyles = props => `
+  ${defaultHeadingStyles}
+  ${heading[`HEADING_LEVEL_${props.safeDisplayLevel || props.safeLevel}`]}
+  ${props.isHidden ? hidden : ""}
```
mikrotron: I'd suggest just putting `stylers.visuallyHidden` inline here, instead of declaring `hidden` above.

packages/Heading/src/Heading.styles.js
```diff
@@ -14,64 +14,64 @@ const headingStyles = () => `
...
+  ${props.hasUnderline ? underline(props.safeDisplayLevel || props.safeLevel) : ""}
+  ${props.isLight ? light : ""}
```
mikrotron: Same here, since it's one line.

packages/Heading/src/Heading.styles.js
slee1130: Ok!

packages/Heading/src/Heading.styles.js
```diff
@@ -1,7 +1,7 @@
 import tokens from "@paprika/tokens";
 import stylers from "@paprika/stylers";
 
-const headingStyles = () => `
+const defaultHeadingStyles = `
```
mikrotron: While you're at it, can you replace:
`line-height: 1.2` with `${stylers.lineHeight(-1)}`, please?

(Line 10 below)

packages/Heading/src/Heading.js
```diff
@@ -53,55 +48,50 @@ const defaultProps = {
...
+}
+
+function Heading(props) {
+  const { a11yText, children, displayLevel, hasDivider, isSemantic, level } = props;
+  const safeLevel = safeValue(level);
+  const safeDisplayLevel = displayLevel ? safeValue(displayLevel) : null;
+  const divider = <span css={dividerStyles} />;
+  const elementProps = getElementProps(safeDisplayLevel, safeLevel, props);
+  return (
+    <div css={headingStyles} {...elementProps} as={isSemantic ? `h${safeLevel}` : "div"}>
```
mikrotron: Can we please also add `data-pka-anchor='heading'`?

mikrotron:
@oscarkwan, short answer to your question is "yes".  This will be a breaking change.

nahumzs:
we should make a note that we don't support classes as targeting mechanism and do it is under the consumer risk, what do you think? So renaming classes don't end in a breaking change.

mikrotron:
@nahumzs 

For authors:

> Since styling is implemented with the `styled-components` library, classes are typically not needed and best avoided.

https://github.com/acl-services/paprika/wiki/Coding-Conventions#anchors

But I agree, we should make this more clear for consumers. Essentially we should make it clear to consumers that they should use the anchors, and not internal implementation details.  That will establish our "API" contract more clearly, so these kind of changes can be made confidently without considering them "breaking changes".

packages/Heading/README.md
```diff
@@ -0,0 +1,68 @@
...
+</Heading>;
+
+<Heading level={1} isLight>
+  Heading One With isLight
+</Heading>;
+
+<Heading level={1} isHidden>
+  Heading One With isHidden
+</Heading>;
+```
```
mikrotron: This is more detailed than our other README files.  I'm not saying we should remove it, but the README doesn't need to be comprehensive. This is a very simple component, but if we tried to match this level of detail for a more complex component, the README could be quite huge. There are also prop descriptions and Storybook to pick up where this leaves off, and ideally we wouldn't have too much redundancy, since it means more maintenance to keep everything up to date as we make changes.

packages/Heading/src/Heading.styles.js
slee1130: done

packages/Heading/src/Heading.styles.js
slee1130: done

packages/Heading/src/Heading.styles.js
slee1130: done

packages/Heading/README.md
slee1130: done

packages/Heading/README.md
slee1130: done

packages/Heading/src/Heading.js
slee1130: done

packages/Heading/README.md
slee1130: done

PR #264, opened 2019-11-04
UX-632 Add new tokens for hilighting text

jamiek-acl:
### 🛠 Purpose

Patch update to Tokens.

https://aclgrc.atlassian.net/browse/UX-632

Create tokens for hilighting text. This will initially be used by Workpapers' Citation Mode and History Log (diff view).

1. green highlight
2. red highlight
3. yellow highlight

Colours seen here: https://acl.invisionapp.com/share/TEFKPZTXF5H#/screens/275081116


packages/Tokens/src/tokens.yaml
```diff
@@ -105,6 +105,16 @@ highlight:
...
       inset-box-shadow: inset 0 0 0 1px $color--blue, inset 0 0 0 3px $color--blue-lighten-30
+  text:
+    added:
+      background: $color--green-lighten-50
+      font: $color--green-lighten-20
+    removed:
+      background: $color--orange-lighten-40
+      font: $color--orange-darken-20
+    selected:
+      background: '#f7f3ba'
```
mikrotron: Neither of these are close enough?
https://github.com/acl-services/acl-ui/blob/master/src/components/helpers/tokens.json#L62-L63

jamiek-acl: Nope :)  I spoke with Jayde about the colours and she wanted a custom one for the yellow (but standard for the green and red).

nahumzs:
something is wrong, semaphore is complaining about `EEXIST: file already exists, symlink '../../../../../../Icon' -> '/home/semaphore/paprika/packages/Toast/node_modules/@paprika/button/node_modules/@paprika/icon' `

jamiek-acl:
Thanks allison for the test fix. @mikrotron or @nahumzs please reivew allison's changes.

packages/Tokens/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/tokens",
-  "version": "0.1.7",
+  "version": "0.1.8",
```
mikrotron: This will get bumped automatically when we do the `lerna publish`, so you can leave it as `0.1.7` for now and just include a comment in the PR description that this is a "patch update" to `tokens`.

PR #267, opened 2019-11-07
Os adding test select.js

slee1130:
### 🛠 Purpose

---

### ✏️ Notes to Reviewer

--- first trial testing:
isDisabled
isReadonly
hasError
--- needs some feedback @oscarkwan 

### 🖥 Screenshots

---


packages/Heading/README.md
```diff
@@ -0,0 +1,68 @@
+## Heading
```
oscarkwan: Reminder to check out from `master` branch so your heading changes are not here

packages/Select/tests/Select.spec.js
```diff
@@ -22,4 +22,34 @@ describe("Select", () => {
...
   });
+
+  it("should not display the options", () => {
+    const { queryByText } = render(
+      <Select placeholder="placeholder value">
+        <option value="Select value">Select value</option>
+        <option value="Another option">Another option</option>
+      </Select>
+    );
+    expect(queryByText(/place holder value/i)).not.toBeInTheDocument();
```
oscarkwan: I think it's better to test if "Select value" and "Another option" are not in the DOM instead of place holder value for this scenario.

packages/Select/tests/Select.spec.js
```diff
@@ -22,4 +22,34 @@ describe("Select", () => {
...
+  });
+
+  it("should be an error if clicks the wrong option", () => {
+    const { container } = render(
+      <Select placeholder="placeholder value" hasError>
+        <option value="Select value">Select value</option>
+        <option value="Another option">Another option</option>
+      </Select>
+    );
+    expect(container.querySelector(".form-select--has-error")).toBeInTheDocument();
```
oscarkwan: Can we capture the text in the error instead of the class?

PR #269, opened 2019-11-07
PD-13298 Support more props in CollapsibleCheckLists.Item

jamiek-acl:
### 🛠 Purpose

Make CollapsibleChecklists.Item accept more props, to pass to the input (i.e. a `ref`).


jamiek-acl:
Screen Shot 2019-11-07 at 10 56 17 AM

PR #270, opened 2019-11-07
Pd 13695 fix token colour

oscarkwan:
### 🛠 Purpose

---

- The token rename, is because added / selected / removed doesn't make sense in citation mode. We should make it a bit more generic and most simple way is just to use the colour.

- Modified one token from lighten -> darken

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---


packages/Tokens/src/tokens.yaml
```diff
@@ -106,13 +106,13 @@ highlight:
       box-shadow: 0 0 0 1px $color--blue, 0 0 0 3px $color--blue-lighten-30
       inset-box-shadow: inset 0 0 0 1px $color--blue, inset 0 0 0 3px $color--blue-lighten-30
   text:
-    added:
+    green:
       background: $color--green-lighten-50
-      font: $color--green-lighten-20
-    removed:
+      font: $color--green-darken-20
```
jamiek-acl: thanks for catching that!

PR #272, opened 2019-11-08
Ux 606 form element flexible layout

tristanjasper:
### 🛠 Purpose

Context:
We could use the formElement for questions in questionnaires (in results). Some questions responses have multiple check boxes so we need to support this, in addition to allowing the question text to be shown above the response field.

At this time we do not want to change the layout of questions. Discussed with @krismckinnonacl on this point.

Form element component changes
Allow component to handle multiple children.
Add an additional extra panel component for displaying text field above the children, allowing us to use form element in results questionnaires.

Side fixes
Fix storybook knobs not working in form element showcase story
Add addition knobs in showcase for showing errors or description
Add alternate layouts stories

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

image

---


packages/ProgressAccordion/stories/examples/Showcase.js
```diff
@@ -49,4 +49,4 @@ const ExampleStory = () => {
   );
 };
 
-export default () => <ExampleStory />;
+export default () => <ExampleStory {...propKnobs()} />;
```
tristanjasper: fixes the knobs in the story

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,74 @@ function FormElement(props) {
...
-                </React.Fragment>
-              )}
-            </React.Fragment>
-          );
-        }
-
-        return child;
-      })}
+      <div css={isInline ? inlineContainerStyles : null}>
+        {extractedChildren["FormElement.ExtraPanel"] ? extractedChildren["FormElement.ExtraPanel"] : null}
```
tristanjasper: adding the new extra panel here

packages/FormElement/src/components/ExtraPanel/ExtraPanel.js
```diff
@@ -0,0 +1,30 @@
...
+
+const defaultProps = {
+  ariaExtraPanelId: null,
+};
+
+function ExtraPanel(props) {
+  const { children, ariaExtraPanelId, ...moreProps } = props;
+
+  return (
+    <div id={ariaExtraPanelId} css={extraPanelStyles} data-pka-anchor="formElement.extrapanel" {...moreProps}>
```
allison-c: `ariaExtraPanelId ` is not necessary, it can be passed in through moreProps

packages/FormElement/src/FormElement.js
mikrotron: Can you just do `{extractedChildren["FormElement.ExtraPanel"]}` ?

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,73 @@ function FormElement(props) {
...
+                  readOnly: isReadOnly,
+                }
+              : {
+                  ...childAttributes,
+                  hasError,
+                  isDisabled,
+                  isReadOnly,
+                  size,
+                };
+            return React.cloneElement(child, extendedProps);
```
mikrotron: I think you need to retain the `key` prop.

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,73 @@ function FormElement(props) {
...
         ariaDescriptionId,
       });
     }
 
     return null;
   }
 
+  const childAttributes = {
+    className: "form-element--child",
+    "aria-describedby": ariaDescriptionId,
```
mikrotron: The new component will also need an id that can be included here.  The form control needs to have a list of element ids that describe it, including the description, the error, AND the new component.

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,73 @@ function FormElement(props) {
...
-        }
-
-        return child;
-      })}
+      <div css={isInline ? inlineContainerStyles : null}>
+        {extractedChildren["FormElement.ExtraPanel"] ? extractedChildren["FormElement.ExtraPanel"] : null}
+
+        {extractedChildren.children.map(child => {
+          if (React.isValidElement(child)) {
+            const extendedProps = isString(child.type)
```
mikrotron: It's not obvious what condition you really care about with `isString(child.type)`.  Can we do something like:
`const isNativeFormControl = isString(child.type);` to improve readability?
cc @allison-c

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,73 @@ function FormElement(props) {
...
+      return React.cloneElement(extractedChildren["FormElement.Description"], {
         ariaDescriptionId,
       });
     }
 
     return null;
   }
 
+  const childAttributes = {
+    className: "form-element--child",
```
mikrotron: Lets avoid classNames and use a `data-pka-anchor='form-element.child'` instead.

packages/FormElement/src/FormElement.js
mikrotron: Maybe we should wrap a `<div>` or `<span>` around this, to apply the margin via `css` prop (instead of using the `.form-element--child` class).

packages/FormElement/src/components/ExtraPanel/ExtraPanel.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import extraPanelStyles from "./ExtraPanel.styles";
+
+const propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+function ExtraPanel(props) {
```
mikrotron: `ExtraPanel` isn't sitting right with me.  I'm not sure what name would best describe this, especially in contrast to the `Description` which is basically the same thing, but in a different layout position... 🤔 

How about `Instructions`? Would that be suggestive if its actual purpose? A user would want to read instructions before proving input, right?

cc @krismckinnonacl

packages/FormElement/stories/examples/Showcase.js
```diff
@@ -0,0 +1,54 @@
...
+  isDisabled: boolean("isDisabled", false),
+  isInline: boolean("isInline", false),
+  isLabelVisuallyHidden: boolean("isLabelVisuallyHidden", false),
+  isReadOnly: boolean("isReadOnly", false),
+  label: text("label", "Form element"),
+  size: select("size", ShirtSizes.DEFAULT, "medium"),
+});
+
+const errorPropKnobs = () => ({
+  showError: boolean("show error", true),
```
mikrotron: I see now what you're trying to do with this... but I think it might be better leave this out, and allow the viewer to discover how the component works. This suggests there's a prop called `showError`, which is not the case.

packages/FormElement/stories/examples/Showcase.js
```diff
@@ -0,0 +1,54 @@
...
+  errorText: text("error text", "This field cannot be blank."),
+});
+
```
mikrotron: Can we add knobs for `Description` and the new component too? Maybe even for `Help`?

packages/FormElement/stories/examples/Showcase.js
```diff
@@ -0,0 +1,54 @@
...
+import Input from "@paprika/input";
+import { Rule, Tagline } from "storybook/assets/styles/common.styles";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import { FormElementStory } from "../FormElement.stories.styles";
+import FormElement from "../../src";
+
+const formElementProps = () => ({
+  hasOptionalLabel: boolean("hasOptionalLabel", false),
+  hasRequiredLabel: boolean("hasRequiredLabel", false),
+  id: text("id", "input-id"),
```
mikrotron: I don't think this is really necessary for a knob.  It's part of the API, but not a very visible one that can be shown off in the Showcase.

packages/FormElement/stories/examples/Showcase.js
mikrotron: Maybe for the showcase it makes more sense for the `errorText` to be blank by default.

sunminsarahlee:
Btw I think we should add red border for date picker error state. 
Screen Shot 2019-11-12 at 2 20 22 PM

packages/FormElement/src/components/ExtraPanel/ExtraPanel.js
tristanjasper: I'm figured we would want something agnostic rather than specific? I'm fine with instructions anyway.

packages/FormElement/src/FormElement.js
tristanjasper: 👌

alexzherdev: `cloneElement` preserves keys automatically https://reactjs.org/docs/react-api.html#cloneelement

packages/FormElement/src/components/ExtraPanel/ExtraPanel.js
tristanjasper: Going with Instructions for now

packages/Checkbox/src/Checkbox.js
```diff
@@ -62,6 +62,7 @@ const Checkbox = props => {
         id={checkboxId}
         ref={inputRef}
         type="checkbox"
+        onChange={onChange}
```
tristanjasper: fixing missing prop error

tristanjasper:
> Btw I think we should add red border for date picker error state.
> Screen Shot 2019-11-12 at 2 20 22 PM

@sunminsarahlee Isn't it already doing that? Or you mean something else
image

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +71,91 @@ function FormElement(props) {
...
     return null;
   }
 
+  function renderInstructions() {
+    if (extractedChildren["FormElement.Instructions"]) {
+      return React.cloneElement(extractedChildren["FormElement.Instructions"], {
+        ariaInstructionsId,
+      });
+    }
+  }
```
mikrotron: This should always return something (`null` if there is no Instructions).

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +71,91 @@ function FormElement(props) {
...
+
+  const nativeChildTypes = ["input", "textarea", "select"];
+
+  const renderFormElementChild = (child, index) => (
+    <div key={`child-key_${index}`} css={formElementChildStyle}>
```
mikrotron: Ideally we wouldn't use `index` as the `key`.  The original code was referencing `child.key`, which is a good idea, but if it doesn't exist, maybe we can fallback on the `index`?  Or else maybe `uuidv4()`?

cc @alexzherdev

mikrotron: Also, I still think it would be nice to include the `data-pka-anchor='form-element.child'` attribute here to give the consumer to ability to apply styling overrides cleanly if necessary.

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +71,91 @@ function FormElement(props) {
...
+                  "aria-describedby": ariaDescribedByIds,
+                  id: inputId,
+                }
+              : {
+                  hasError,
+                  isDisabled,
+                  isReadOnly,
+                  size,
+                  "aria-describedby": ariaDescribedByIds,
+                  id: child.type.displayName === "Input" ? inputId : null,
```
mikrotron: I think we should pass the `id` along to any React component, not just an `<Input>`.   I believe the `<DatePicker>` depends on this to associate the label to it, for instance.

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +71,91 @@ function FormElement(props) {
...
-            </React.Fragment>
-          );
-        }
-
-        return child;
-      })}
+      <div css={isInline ? inlineContainerStyles : null}>
+        {renderInstructions()}
+        {extractedChildren.children.map((child, index) => {
+          if (React.isValidElement(child)) {
```
mikrotron: I think the logic here is backwards? 
If `isValidElement` then we should pass `isDisabled`, not `disabled`, for example.

packages/FormElement/stories/examples/Showcase.js
```diff
@@ -0,0 +1,69 @@
...
+      <FormElement {...props}>
+        <FormElement.Instructions>{instructionsPropKnobs().instructionsText}</FormElement.Instructions>
+        <Input onChange={handleChange} value={value} placeholder="Form placeholder" />
+        <FormElement.Error>{errorPropKnobs().errorText}</FormElement.Error>
+        <FormElement.Description>
+          <span>{descriptionPropKnobs().descriptionText}</span>
+        </FormElement.Description>
+        <FormElement.Help>
+          {helpPropKnobs().helpText}
+          <a href="wegalvanize.com">Learn more</a>.
```
mikrotron: I'd suggest cutting this... so we can see that the Icon disappears if we clear the `help text` knob.

packages/ProgressAccordion/stories/examples/Showcase.js
```diff
@@ -7,7 +7,7 @@ import generatedItems from "../fixtures/generateItems";
 import ProgressAccordion from "../../src";
 
 const propKnobs = () => ({
-  activeIndex: select("activeIndex", [0, 1, 2, 3, 4], 2),
+  activeIndex: select("activeIndex", [0, 1, 2, 3, 4], 3),
```
mikrotron: 🤨why?

packages/FormElement/src/FormElement.js
tristanjasper: I tried using `uuidv4()`but i was losing input focus on every key press. Also child.key doesn't always exist as you say. I'll try and add some fallback logic.

mikrotron:
@sunminsarahlee @tristanjasper 

> > Btw I think we should add red border for date picker error state.
> > Screen Shot 2019-11-12 at 2 20 22 PM

Sarah's right, but it's a DatePicker issue.  It doesn't have a `hasError` prop that it should pass down to the actual `<Input>`:
https://github.com/acl-services/paprika/blob/master/packages/DatePicker/src/DatePicker.js#L23-L49

packages/FormElement/src/FormElement.js
mikrotron: Good point about uuid.  I think we might have to require a `key` be included on children of `<FormElement>`.  We should add that to the README as part of the API. Can we possibly throw a warning if it's not included?

packages/FormElement/src/FormElement.js
allison-c: Maybe we can add a set of helper functions like this? https://stackoverflow.com/questions/33199959/how-to-detect-a-react-component-vs-a-react-element sometimes `React.isValidElement` isn't enough, that's why I have to add `isString(child.type)` checking before

tristanjasper: Oh As the code was previously I figure the `isString` method was evaluating if it was a react component, and that `React.isValidElement` was checking that it was just an element of some kind, native or otherwise. I'll adjust it 👍

packages/FormElement/src/FormElement.js
allison-c: `child.key` is auto-generated by React, should it be always there?

mikrotron: In that case, yeah, let's just use `child.key`. Thanks, @allison-c!

tristanjasper: Oh i see the `extractChildren` function is adding the key

allison-c:
> @sunminsarahlee @tristanjasper
> 
> > > Btw I think we should add red border for date picker error state.
> > > Screen Shot 2019-11-12 at 2 20 22 PM
> 
> Sarah's right, but it's a DatePicker issue. It doesn't have a `hasError` prop that it should pass down to the actual `<Input>`:
> https://github.com/acl-services/paprika/blob/master/packages/DatePicker/src/DatePicker.js#L23-L49

Sorry about that 😢 we probably have to bring some of these back https://github.com/acl-services/paprika/commit/7d6ed78e55d7ce0f2d3cfa89d5ffdf05c76669d9

packages/ProgressAccordion/stories/examples/Showcase.js
tristanjasper: 😄 i was testing if the knobs were even working in progress accordion they are after the other change on line 52 
`export default () => <ExampleStory {...propKnobs()} />;`

packages/FormElement/src/FormElement.js
tristanjasper: hmm, ok yeah i was trying to restrict it, ie in the case of checkbox component, but we can deal with that in a future pr

packages/FormElement/src/FormElement.js
tristanjasper: I'm not a big fan of using the `data-pka-anchor` attributes for styling but thats for another discussion.

mikrotron: I agree with you, _internally_ (within Paprika).  But once someone consumes it, they will need some kind of selector to apply styling to, otherwise you'll end up with gross stuff like `div > div:nth-child(3) > span`.

packages/FormElement/src/FormElement.js
tristanjasper: Actually i believe all of the children will be react components at this point as they have been extracted via the extractChildren function, so I am removing the check for React.isValidElement as seems redundant. I am checking the type however to determine if it is a native element or not.

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,92 @@ function FormElement(props) {
...
+        ariaInstructionsId,
+      });
+    }
+    return null;
+  }
+
+  const nativeChildTypes = ["input", "textarea", "select"];
+
+  const renderFormElementChild = child => (
+    <div data-pka-anchor="form-element.child" key={`${child.key}`} css={formElementChildStyle}>
```
mikrotron: nit: just `key={child.key}`

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,92 @@ function FormElement(props) {
...
+      {child}
+    </div>
+  );
+
+  const ariaDescribedByIds = `${ariaErrorId} ${ariaInstructionsId} ${ariaDescriptionId}`;
```
mikrotron: I'd probably prefer to put this inline – I think there's already enough abstraction/indirection with `childExtendedProps`.

packages/FormElement/src/FormElement.js
```diff
@@ -71,85 +72,92 @@ function FormElement(props) {
...
-            </React.Fragment>
-          );
-        }
-
-        return child;
-      })}
+      <div css={isInline ? inlineContainerStyles : null}>
+        {renderInstructions()}
+        {extractedChildren.children.map(child => {
+          if (isString(child.type) && nativeChildTypes.includes(child.type)) {
```
mikrotron: I think it would read better to extract the `isString(child.type)` into a function that also does the `nativeChildTypes.includes(child.type)`.

packages/FormElement/stories/examples/Showcase.js
```diff
@@ -0,0 +1,66 @@
...
+      <Tagline>Form Element.</Tagline>
+      <Rule />
+      <FormElement {...props}>
+        <FormElement.Instructions>{instructionsPropKnobs().instructionsText}</FormElement.Instructions>
+        <Input onChange={handleChange} value={value} placeholder="Form placeholder" />
+        <FormElement.Error>{errorPropKnobs().errorText}</FormElement.Error>
+        <FormElement.Description>
+          <span>{descriptionPropKnobs().descriptionText}</span>
+        </FormElement.Description>
+        <FormElement.Help>{helpPropKnobs().helpText}</FormElement.Help>
```
mikrotron: Hmpf. Still renders the icon even if it's `""`.  Oh well 🤷‍♂️

tristanjasper: 🤷‍♂ can revisit later as any content can be passed

PR #275, opened 2019-11-12
PD-13719 make overflow hidden important

oscarkwan:
### 🛠 Purpose
- Resolves https://github.com/acl-services/paprika/issues/277
- make lock body scroll overflow hidden !important to fight with inline css conflicts
---

### ✏️ Notes to Reviewer
---
- there is some inline css overflow auto in projects... that conflicts with this, so need to make it !important

### 🖥 Screenshots

---


nahumzs:
My dear `projects` and they lovely UI bugs :).

PR #278, opened 2019-11-13
Ux 271 fix date picker errors

tristanjasper:
### 🛠 Purpose

Date Pickers usage of the Popover component has changed now that we have updated the way that the popover handles its child components. Side effect of https://github.com/acl-services/paprika/pull/249

This Pr aims to primarily fix the errors reported in failing jest test on Date Picker component

Additionally I have cleaned up the way we are rendering calendar, which now unmounts the component when it should not be visible.

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

image

### DIFF
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -235,12 +235,7 @@ function DatePicker(props) {
       />
 
       <Popover.Content>
-        <div
-          css={calendarPopoverStyles}
-          data-pka-anchor="datepicker.calendar"
-          ref={calendarRef}
-          shouldShowCalendar={shouldShowCalendar}
-        >
+        <div css={calendarPopoverStyles} data-pka-anchor="datepicker.calendar" ref={calendarRef}>
           {shouldShowCalendar ? (
             <Calendar
               date={date}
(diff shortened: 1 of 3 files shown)


packages/DatePicker/src/DatePicker.js
```diff
@@ -241,13 +241,14 @@ function DatePicker(props) {
...
           shouldShowCalendar={shouldShowCalendar}
         >
-          <Calendar
-            date={date}
-            isVisible={shouldShowCalendar}
-            onSelect={handleSelect}
-            possibleDate={debouncedPossibleDate}
-            resetPossibleDate={handleResetPossibleDate}
-          />
+          {shouldShowCalendar ? (
```
tristanjasper: Unless i'm missing something it seems we no longer need to keep the calendar rendered (but invisible) in the dom after the recent popup changes - https://github.com/acl-services/paprika/pull/249

@allison-c Can you check Is there some functionality that this is meant to be doing now i have removed this complexity?

nahumzs:
omg, this make me soooo happy that error was driving me crazy, thank yooo!!

packages/DatePicker/src/DatePicker.js
```diff
@@ -241,13 +241,14 @@ function DatePicker(props) {
           ref={calendarRef}
           shouldShowCalendar={shouldShowCalendar}
```
allison-c: This line can be removed too because only used in `calendarPopoverStyles `

PR #281, opened 2019-11-14
Form Element - add extra accessibility features, fix described by ids

tristanjasper:
### 🛠 Purpose

This PR makes the following changes on the formElement component

- Ensure Ids are being added correctly to the Description, Error and Instructions components so that the `aria-describedby` accessibility features are working correctly.
- Sets additional aria attributes on child components where applicable
- Introduces `aria-live` attribute that will announce errors that are happening asynchronously via screen readers.
- Adds a story for Accessibility features, allowing us to show that `aria-live` accessibility feature is working correctly

### ✏️ Notes to Reviewer

---

### 🖥 Screenshots

---


packages/FormElement/src/FormElement.js
```diff
@@ -87,14 +87,18 @@ function FormElement(props) {
 
   function renderFooter() {
     if (hasError) {
-      return React.cloneElement(extractedChildren["FormElement.Error"], {
-        ariaErrorId,
-      });
+      return (
+        <div aria-live="assertive">
```
tristanjasper: @sunminsarahlee I am using the aria-live attribute of "assertive" here has this is an error being announced, and as the visual representation is an exclamation mark, is red and is designed to draw attention, i felt the screen reader should also reflect that importance. Once the error has been read out, the screen reader (on mac) defaults to letting the listener know they are on a button, so the listener won't lose context.

Let me know what you think.

tristanjasper: Interestingly here we are rendering the `div` with the `aria-live` in the same render cycle as its content and this works fine on chrome. In our accessibility course we were advised that the container div would have to be rendered pre-contents, so will be good to test this . Will also be a good idea to test this approach on other accessibility platforms (jaws, navx?).

oscarkwan:
@mikrotron did u test with nvda / jaws already?

Issue #97, opened 2019-06-22
SidePanel.Footer Flickering while scrolling


nahumzs:
## Expected behavior

While scrolling the SidePanel the footer should smoothly move and stay stuck at the bottom of the SidePanel

## Current behavior
Every time the user scrolls there is a flickering occurring while repositioning the Footer.

## Screenshots / Gifs / Codepens
*Include media to help illustrate the bug. Gifs help a lot!*

## Additional context
*Please provide any more additional details to illustrate the bug*

mikrotron:
This should be achievable by using a flexbox layout instead of using the `useFooterOffset()` hook.

PR #282, opened 2019-11-15
OS-Side panel without flickering

slee1130:
### 🛠 Purpose

---

### ✏️ Notes to Reviewer

--- Used “position: fixed” since the “sticky” is not supported IE. And the width of the footer is passed by the props so the styling is the same and the footer is still moving smoothly and staying stuck at the bottom of the side panel as it is asked in issue 97.  Resolves https://github.com/acl-services/paprika/issues/97

@oscarkwan 

### 🖥 Screenshots

---


packages/SidePanel/src/hooks/useFooterOffset.js
```diff
@@ -2,8 +2,8 @@ import React from "react";
 import debounce from "lodash.debounce";
 
 const useFooterOffset = (height, ref = null, isSticky) => {
```
oscarkwan: can you check if `useFooterOffset` this is used? If not we can just delete the whole thing

packages/SidePanel/src/components/Footer/Footer.styles.js
```diff
@@ -13,7 +13,8 @@ export const footerCSS = css`
   right: 0;
   transition: opacity 0.3s ease-in;
   width: 100%;
+
   ${props => {
-    return props.isSticky ? "position: absolute;" : "position: relative;";
+    return props.isSticky ? `position: fixed;width:${props.width};` : "position: relative; width: 100%";
```
oscarkwan: let's create a variable for these styles. 

for example

```
const fixedFooterStyles = css`
  position: fixed;
  width: ${props => props.width}
`;

... etc
```

packages/SidePanel/src/components/Footer/Footer.js
```diff
@@ -1,32 +1,32 @@
...
   children: PropTypes.node.isRequired,
   height: PropTypes.number,
+  width: PropTypes.number,
   isSticky: PropTypes.bool,
 };
 
 const defaultProps = {
   height: 48,
   isSticky: false,
+  width: 100,
```
oscarkwan: Can you use the same default prop as the `SidePanel`. This width prop gets passed down from the <SidePanel> component. so it will make sense to use the same PropTypes, and default prop

PR #284, opened 2019-11-15
Ux 633 data table

nahumzs:
# DataTable

### 🔗 Storybook 

http://storybooks.highbond-s3.com/paprika/UX-633-DataTable/iframe.html?id=datatable--showcase

---

### 🛠 Purpose
Users need to be able to find the view of their data in the table in a format that makes the most sense to them. The easiest way to do this is by filtering the contents of the table down to something more meaningful and easily digestible.

An individual user could filter all issues down to those that are assigned just to them. A manager could filter a table of all controls down to those that their team owns and will be testing in Q2. Many different use cases benefit from a quality filtering experience.

Our goal is to build a filtering capability that is configured centrally (filters for all columns are configured in the same area), but can be invoked either from a central place (a "Filter" button, for example) or directly from a popover menu inside each column header.

Note: the idea is we'll follow the same pattern for sorting and column management

Key Details

- [ ] Filters are applied to individual columns
- [ ] Filter options depend on the field type
- [ ] There should be some kind of indication when a filter has been applied (likely within the central "Filter" button)

Reference
https://whimsical.com/CBQ3d5Sricjst6mfuP8xWa

---

## ✏️ API

The API for this component has its inspiration from other table's API on the community but primarily by [Griddle](http://griddlegriddle.github.io/Griddle/docs/) their approach of defining Columns is elegant and straightforward to understand.
 
We also took a look at other established components like [react-table](https://github.com/tannerlinsley/react-table), which cover more cases than the ones we need in our platform and their API don't align too well with the way we are writing components.

Another solution we took a look was the [react-virtualize](https://github.com/bvaughn/react-virtualized/blob/master/docs/Table.md), the real benefit of using it this component is the virtualization part which, after researching a little was not as hard to implement in this current PR.
 
Other libraries we paid some attention are:
[ag-grid](https://www.ag-grid.com/react-getting-started/?utm_source=ag-grid-readme&utm_medium=repository&utm_campaign=github)
[reactables(https://github.com/GigaTables/reactable)
[mui-datatables](https://github.com/gregnb/mui-datatables)
[react-data-grid](https://adazzle.github.io/react-data-grid/docs/examples/simple-grid)
[react-pivot](https://github.com/davidguttman/react-pivot)
and reacttabular, rsuit-table, sematable, shin, telerik table

After our research, we end with the following `API` which use `wai-aria role="grid"` attribute to support virtualization.

### data structure
The data table data structure is an array of objects, ex:
```js
const data = [
{ name: "Mariane", age: 43, hobby: "watch netflix"},
{ name: "John", age: 15, hobby: "play sports"},
...
{ name: "Kathy", age: 34, hobby: "hiking"},
];
```
### Column Definition Subcomponent
The most important subcomponent for the *DataTable* is the `ColumnDefinition`, which dictates how the data should be display on the component.

```jsx
<DataTable data={[...]}>
  <DataTable.ColumnDefinition />
</DataTable>
```

Using the cell `prop` over this subcomponent, allowed the consumer to indicate if they want to map the column with a value in their object, or they want to use a render function to have more control over how to display the data.

```js
const data = [
  { name: "Mariane", age: 43, hobby: "watch netflix"},
  ...,
  ...
];

// if you assigned `age` to the `cell` prop, the component will map 
// the object attribute *age* to the displayed value on the cell.

<DataTable data={data}>
  <DataTable.ColumnDefinition header="name" cell="name" />
  <DataTable.ColumnDefinition header="age" cell="age" />
</DataTable>
```

Output:
| name      | age  |
|--------   |----- | 
| Mariane  |  43   |

Using a render function inside the cell to have more granular
 control about how to render the cell:

```jsx
<DataTable.ColumnDefinition 
  header="name / age" 
  cell={(row) =>  <span>{row.name} / {row.age} </span>} 
/>
```

Output:
| name / age    |
|--------------| 
| Mariane / 43 |

*Other props are:*
*header* // a node
*id* // unique identifier for the columns, is required
*width* will fix the width of the column in px and add ellipsis to the string content

#### Real basic example
```js
    <DataTable data={data} height={window.innerHeight}>
      <DataTable.ColumnDefinition 
          id="country" 
          width="190" 
          header="Country" 
          cell={cell => getFlag(cell)} 
      />
      <DataTable.ColumnDefinition id="name" header="Name" cell="name" />
      <DataTable.ColumnDefinition id="goals" header="Goals" cell="goals" />
      <DataTable.ColumnDefinition id="status" header="Status" cell="status" />
    </DataTable>
```

#### Navigation component

---

### 🖥 Screenshots

---


alexzherdev:
> Another solution we took a look was the react-virtualize, the real benefit of using it this component is the virtualization part which, after researching a little was not as hard to implement in this current PR.

Honestly, don't think this is the right bet. Personally virtualization is very high on my "avoid rolling your own" list, probably right after implementing an authentication system or an encryption algorithm from scratch. Current implementation looks janky, and that's not to say we can't work to improve it, but the library author has had years to improve react-virtualized and then react-window AND he's on react core at least for now (btw I think it'd be recommended to use react-window's FixedSizeGrid if it fits our purpose).

> We also took a look at other established components like react-table, which cover more cases than the ones we need in our platform and their API don't align too well with the way we are writing components.

A bit less critical for me, but it's a shame that we have to reimplement all the table functionality, given how many libraries already exist to do the same things. The ratio of reinventing the wheel per line of code looks very high to me. Curious what others think.

PR #286, opened 2019-11-15
Pd 13298 popover focus

jamiek-acl:
Bug fix in the Popover component.

### 🛠 Purpose

Before this PR, when you open the popover, focus goes into it.  Then when you tab through its content, focus goes to the address bar (since the Popover is added to the bottom of the DOM).

After this PR, when you open the popover, focus goes into it. Then when you tab through its content, focus goes to the next item on screen and the popover closes.

This is maybe premature and it needs to be tested with `isEager` and some other props.... just want to get some feedback.

### 🖥 Screenshots

Notice the `tabIndex` before (3), how it is removed (-1), then how it is restored (3).
Notice I can tab right through the popover, in both directions.

focus2

### DIFF
--- a/packages/Popover/src/Popover.js
+++ b/packages/Popover/src/Popover.js
@@ -102,7 +102,6 @@ class Popover extends React.Component {
     this.$portal = document.body.appendChild(portalNode);
 
     this.focusableElements = [];
-    this.originalTabIndexes = [];
     this.triggerFocusIndex = null;
 
     this.state = {
@@ -341,21 +340,10 @@ class Popover extends React.Component {
     return false;
   };
 
-  restoreTabIndexes = () => {
-    this.focusableElements.forEach(focusableElement => {
-      if (focusableElement !== this.$trigger && !this.elementIsDescendentOfPopover(focusableElement)) {
-        focusableElement.tabIndex = this.originalTabIndexes.shift(); // eslint-disable-line no-param-reassign
-      }
-    });
-  };
-
-  removeTabIndexes = () => {
+  getPopoverTriggerFocusIndex = () => {
     this.focusableElements.forEach((focusableElement, index) => {
       if (focusableElement === this.$trigger) {
         this.triggerFocusIndex = index;
-      } else if (!this.elementIsDescendentOfPopover(focusableElement)) {
-        this.originalTabIndexes.push(focusableElement.tabIndex);
-        focusableElement.tabIndex = -1; // eslint-disable-line no-param-reassign
       }
     });
   };
@@ -365,7 +353,7 @@ class Popover extends React.Component {
     //       find the first focusable element like button, input, etc?
     //       can focus automatically
     if (!this.props.shouldKeepFocus && !this.props.isEager && this.isOpen() && event.propertyName === "visibility") {
-      this.removeTabIndexes();
+      this.getPopoverTriggerFocusIndex();
       event.target.focus();
     }
   };
@@ -415,13 +403,11 @@ class Popover extends React.Component {
   close() {
     // NOTE: Even if uncontrolled, the app may want to be notified when closed via the onClose callback
     if (this.props.onClose !== null) {
-      this.restoreTabIndexes();
       this.props.onClose();
     }
 
     if (this.props.isOpen === null) {
       this.setState({ isOpen: false });
-      this.restoreTabIndexes();
       this.removeListeners();
     }
   }
--- a/packages/Popover/src/Popover.js
+++ b/packages/Popover/src/Popover.js
@@ -7,9 +7,7 @@ import tokens from "@paprika/tokens";
 import { AlignTypes } from "@paprika/helpers/lib/customPropTypes";
 import isInsideBoundaries from "./helpers/isInsideBoundaries";
 import { getContentCoordinates, getTipCoordinates } from "./helpers/getPosition";
-// import { isActiveElementPopover } from "./helpers/isActiveElementPopover";
 import getBoundingClientRect from "./helpers/getBoundingClientRect";
-
 import PopoverContext, { ThemeContext } from "./PopoverContext";
 import Content from "./components/Content/Content";
 import Card from "./components/Card/Card";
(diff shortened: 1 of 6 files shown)


packages/Popover/src/Popover.js
```diff
@@ -26,6 +26,10 @@ import PopoverStyled from "./Popover.styles";
 const openDelay = 350;
 const closeDelay = 150;
 const throttleDelay = 20;
+const focusableElementSelector = 'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
```
oscarkwan: Will our Buttons that are (spans) work?

jamiek-acl: They'd have to have `tabIndex` so ya

nahumzs: do we need to add `[role="button"]`?

jamiek-acl: Makes sense to me

packages/Popover/src/Popover.js
```diff
@@ -175,6 +180,8 @@ class Popover extends React.Component {
     // about setState for Popovers and Tooltips in ComponentDidMount
     // https://reactjs.org/docs/react-component.html#componentdidmount
 
+    focusableElements = document.querySelectorAll(focusableElementSelector);
```
mikrotron: I think this should be set when the Popover opens.  The elements on the page could vary while the Popover is mounted.

packages/Popover/src/Popover.js
```diff
@@ -26,6 +26,11 @@ import PopoverStyled from "./Popover.styles";
 const openDelay = 350;
 const closeDelay = 150;
 const throttleDelay = 20;
+const focusableElementSelector =
+  'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"]), [role="button"]';
+let focusableElements;
+const originalTabIndexes = [];
+let triggerFocusIndex = -1;
```
mikrotron: Why `-1`? Can it just be left undefined?

jamiek-acl:
Aside: I just noticed that if `isEager` or `shouldKeepFocus` is true, you can't get into the content of the Popover with the tab keys (e.g. if it had a link in it). This is true for the existing Popover, and also for this revised one. Is that a problem?

Also, what's the difference between `isEager` and `shouldKeepFocus`...? They seem to have the same purpose, and are only checked together [in the code].

mikrotron:
@jamiek-acl I think `shouldKeepFocus` still requires a click to open, it just doesn't transfer the focus to the `<Content>`.  This is important for the `<DatePicker>` I think, but @allison-c knows better than me.

I think it's fine that `isEager`, tooltip-style `<Popovers>` don't allow focus to their contents.  The content is linked to the trigger element by `aria-describedby`, and our design patterns do not support having interactive elements in a hover-activated tooltip (I think). One of the designers would be able to verify that though.

packages/Popover/src/Popover.js
```diff
@@ -18,14 +16,11 @@ import Tip from "./components/Tip/Tip";
...
-// ACCESSIBILITY
-// NOTE: When closing the popover seems to be better to focus the trigger button
-//       only when the close method is the ESC key, when clicking outside doesn't feel a fluid action
-//       assigning the focus to the button again
-
 const openDelay = 350;
 const closeDelay = 150;
 const throttleDelay = 20;
+const focusableElementSelector =
+  'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"]), [role="button"]';
```
mikrotron: Just `'button, [href], input, select, textarea, [tabindex]'` should work, and is "cleaner", but 🤷‍♂️

jamiek-acl: Yeah I saw that one. It is a bit easier to read, but no point in touching elements that already have the value we want

packages/Popover/src/Popover.js
```diff
@@ -181,8 +180,15 @@ class Popover extends React.Component {
     }
   }
 
+  componentWillReceiveProps(nextProps) {
```
alexzherdev: This lifecycle is deprecated. Perhaps in the future we can migrate this to `getDerivedStateFromProps`.

packages/Popover/src/Popover.js
```diff
@@ -296,11 +302,66 @@ class Popover extends React.Component {
...
+      }
+      elementsParent = elementsParent.parentNode;
+    }
+    return false;
+  };
+
+  restoreTabIndexes = () => {
+    this.focusableElements.forEach(focusableElement => {
+      if (focusableElement !== this.$trigger && !this.elementIsDescendentOfPopover(focusableElement)) {
+        focusableElement.tabIndex = this.originalTabIndexes.reverse().pop(); // eslint-disable-line no-param-reassign
```
alexzherdev: `reverse()` is in-place, and this is done in a loop, so probably there's an issue with reversing multiple times. `this.originalTabIndexes.shift()` should work?

nahumzs:
`shouldKeepFocus` works when you want to have as example a search, where the input will open the popover but will keep the focus on it. 
DatePicker, ListBoxWithTags, ComboBox are some of the possible components that can and will benefit from having that prop. @jamiek-acl

packages/Popover/src/Popover.js
```diff
@@ -296,11 +308,64 @@ class Popover extends React.Component {
...
+    });
+  };
+
+  removeTabIndexes = () => {
+    this.focusableElements.forEach((focusableElement, index) => {
+      if (focusableElement === this.$trigger) {
+        this.triggerFocusIndex = index;
+      } else if (!this.elementIsDescendentOfPopover(focusableElement)) {
+        this.originalTabIndexes.push(focusableElement.tabIndex);
+        focusableElement.tabIndex = -1; // eslint-disable-line no-param-reassign
```
mikrotron: Did you try out removing the `tabindex` manipulation? This is the part that makes me the most nervous (this individual component alters a lot of external DOM elements, which isn't very component-like).

jamiek-acl: Latest push does not touch the other items' tabIndex

PR #289, opened 2019-11-19
PD-13299 Fix Axe error on SortableItem

jamiek-acl:
MINOR (backward compatible) change to SortableItem

### 🛠 Purpose
Using the paprika/sortable i was getting this axe error:
`Ensure ARIA attributes are allowed for an element's role.`

Digging into it, it seems to be because the `<li>` elements have an `aria-roledescription` (added both by us, and by the react-beautiful-dnd library) and I need to remove those.

I updated to the latest version of the react-beautiful-dnd library, as they no longer add that attribute. I also removed where we add the attribute. Now the axe error is gone, yay.

But when using screen reader in Chrome, it no longer says “Draggable Item. Press space to lift”.
No announcement is made in Chrome (though one is made in Safari) even though the sortable item has `aria-labelledby="rbd-lift-instruction-0"` and the page includes:
```<div id="rbd-lift-instruction-0" style="display: none;">Draggable item. Ensure your screen reader is not in browse mode and then press space bar to lift.</div>```

I read this page: https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/ which basically said aria-labelledby won't work consistently on `li` items.  So i added the `role="treeitem"` and now it does announce it in Chrome.

### DIFF
--- a/packages/Sortable/src/Sortable.js
+++ b/packages/Sortable/src/Sortable.js
@@ -75,6 +75,7 @@ const Sortable = ({ children, onChange, hasNumbers, hasZebraStripes, onRemove, .
             data-is-dragging-over={snapshot.isDraggingOver ? true : undefined}
             isDraggingOver={snapshot.isDraggingOver}
             ref={provided.innerRef}
+            role="tree"
           >
             {validChildren.length > 0 &&
               validChildren.map((child, index) => (
--- a/packages/Sortable/src/Sortable.js
+++ b/packages/Sortable/src/Sortable.js
@@ -80,7 +80,6 @@ const Sortable = ({ children, onChange, hasNumbers, hasZebraStripes, onRemove, .
             data-is-dragging-over={snapshot.isDraggingOver ? true : undefined}
             isDraggingOver={snapshot.isDraggingOver}
             ref={provided.innerRef}
-            role="tree"
           >
             {validChildren.length > 0 &&
               validChildren.map((child, index) => (
--- a/packages/Sortable/src/Sortable.js
+++ b/packages/Sortable/src/Sortable.js
@@ -63,7 +63,12 @@ const Sortable = ({ children, onChange, hasNumbers, hasZebraStripes, onRemove, .
   };
 
   return (
-    <DragDropContext onDragEnd={handleDragEnd} onDragStart={handleDragStart} onDragUpdate={handleDragUpdate}>
+    <DragDropContext
+      onDragEnd={handleDragEnd}
+      onDragStart={handleDragStart}
+      onDragUpdate={handleDragUpdate}
+      liftInstruction={I18n.t("sortable.aria_description")}
+    >
       <Droppable droppableId={dropId.current}>
         {(provided, snapshot) => (
           <ul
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by nahumzs** [comment]
@oscarkwan  do you think that maybe we can see if something similar is happening to the listbox? It might have the wrong aria attributes.

File: packages/Sortable/src/Sortable.js [hunk 1]
```diff
@@ -63,7 +63,12 @@ const Sortable = ({ children, onChange, hasNumbers, hasZebraStripes, onRemove, .
   };
 
   return (
-    <DragDropContext onDragEnd={handleDragEnd} onDragStart={handleDragStart} onDragUpdate={handleDragUpdate}>
+    <DragDropContext
+      onDragEnd={handleDragEnd}
+      onDragStart={handleDragStart}
+      onDragUpdate={handleDragUpdate}
+      liftInstruction={I18n.t("sortable.aria_description")}
```
mikrotron: This is the new prop for `react-beautiful-dnd`?

packages/Sortable/tests/Sortable.cypress.js
```diff
@@ -25,7 +25,7 @@ describe("<Sortable />", () => {
       .trigger("mousemove", { button: 0, ...dragOffset })
       .wait(animationDelay);
     cy.get(selector.root)
-      .trigger("mousemove", { button: 0, ...dragOffset })
+      .trigger("mousemove", { button: 0, ...dragOffset, clientY: 100 })
```
mikrotron: If this is really needed, maybe it makes more sense on line 13?

packages/Sortable/src/Sortable.js
jamiek-acl: Yes, without this prop it does not announce `Press space to lift` when you tab over the items.

packages/Sortable/tests/Sortable.cypress.js
jamiek-acl: But doing that would affect the other tests.  After the upgrade, all the other tests still worked... this one just started failing until I added this.

packages/Sortable/src/Sortable.js
```diff
@@ -75,6 +80,7 @@ const Sortable = ({ children, onChange, hasNumbers, hasZebraStripes, onRemove, .
             data-is-dragging-over={snapshot.isDraggingOver ? true : undefined}
             isDraggingOver={snapshot.isDraggingOver}
             ref={provided.innerRef}
+            role="tree"
```
mikrotron: You're onto something here... but... it's _not_ a tree.  Lemme dig a little into this.

jamiek-acl: Ok, it's my understanding from the URL in the description that the `<li>` has to be one of these roles, or the screen reader wont work predictably: https://www.w3.org/TR/wai-aria-1.1/#widget_roles

packages/Sortable/tests/Sortable.cypress.js
jamiek-acl: Just tried, doesnt work.  Has to be on this line only 🤷‍♂

PR #291, opened 2019-11-20
PD-13719 change portal and add anchors

oscarkwan:
### 🛠 Purpose
---

### ✏️ Notes to Reviewer
---
- Fix portal so that we don't rely on default props for the container (document.body)
- Add side panel anchors for testing

### 🖥 Screenshots

---

### DIFF
--- a/packages/SidePanel/package.json
+++ b/packages/SidePanel/package.json
@@ -16,7 +16,7 @@
     "@babel/runtime-corejs2": "^7.3.1",
     "@paprika/button": "^0.2.3",
     "@paprika/heading": "^0.3.1",
-    "@paprika/helpers": "^0.2.5",
+    "@paprika/helpers": "^0.2.4",
     "@paprika/icon": "^0.2.5",
     "@paprika/raw-button": "^0.2.2",
     "@paprika/stylers": "^0.2.2",
--- a/packages/SidePanel/package.json
+++ b/packages/SidePanel/package.json
@@ -16,7 +16,7 @@
     "@babel/runtime-corejs2": "^7.3.1",
     "@paprika/button": "^0.2.3",
     "@paprika/heading": "^0.3.1",
-    "@paprika/helpers": "^0.2.4",
+    "@paprika/helpers": "^0.2.5",
     "@paprika/icon": "^0.2.5",
     "@paprika/raw-button": "^0.2.2",
     "@paprika/stylers": "^0.2.2",
--- a/packages/SidePanel/src/SidePanel.js
+++ b/packages/SidePanel/src/SidePanel.js
@@ -134,19 +134,20 @@ function SidePanel(props) {
   if (isVisible) {
     const dialog = (
       <Dialog
-        onAnimationEnd={handleAnimationEnd}
-        header={headerExtracted}
+        data-pka-anchor="sidepanel"
         footer={footerExtracted}
-        onClose={onClose}
-        refSidePanelContent={refSidePanelContent}
-        width={width}
-        kind={kind}
+        header={headerExtracted}
         isCompact={isCompact}
         isInline={isInline}
-        refHeader={refHeader}
-        offsetY={offsetScroll}
         isOpen={isOpen}
+        kind={kind}
+        offsetY={offsetScroll}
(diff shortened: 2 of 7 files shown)


packages/SidePanel/src/SidePanel.js
```diff
@@ -134,19 +134,20 @@ function SidePanel(props) {
   if (isVisible) {
     const dialog = (
       <Dialog
-        onAnimationEnd={handleAnimationEnd}
-        header={headerExtracted}
+        data-pka-anchor="sidepanel"
```
oscarkwan: Only change here, rest is just me alphabetizing the props

packages/helpers/src/components/Portal.js
```diff
@@ -11,11 +11,16 @@ const propTypes = {
...
   active: true,
-  container: document.body,
+  container: null,
 };
 
-const Portal = ({ children, active, container }) =>
-  active ? ReactDOM.createPortal(children, container) : <React.Fragment>{children}</React.Fragment>;
+function Portal({ children, active, container }) {
+  return active ? (
+    ReactDOM.createPortal(children, container || document.body)
```
oscarkwan: @DanilAgafonov, this change is needed as we are having an issue in projects where the container is null which then throws a JS error in the app.

cc: @AndreyChernykh

packages/helpers/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/helpers",
-  "version": "0.2.4",
+  "version": "0.2.5",
```
oscarkwan: do i need to bump this? @nahumzs

nahumzs: no only the dependency inside of the side panel. pretty much bump the version only on the sidepanel package.json

oscarkwan: okok my bad

packages/SidePanel/package.json
```diff
@@ -16,7 +16,7 @@
     "@babel/runtime-corejs2": "^7.3.1",
     "@paprika/button": "^0.2.3",
     "@paprika/heading": "^0.3.1",
-    "@paprika/helpers": "^0.2.4",
+    "@paprika/helpers": "^0.2.5",
```
nahumzs: 👍

packages/helpers/package.json
nahumzs: lerna will do this automatically after I describe it, the sidepanel is a breaking change. as well this one.

packages/SidePanel/package.json
oscarkwan: But then now side panel points to a "future version" and npx lerna bootstrap fails
`error Couldn't find any versions for "@paprika/helpers" that matches "^0.2.5"`

packages/helpers/src/components/Portal.js
```diff
@@ -11,11 +11,16 @@ const propTypes = {
 const defaultProps = {
   children: null,
   active: true,
-  container: document.body,
+  container: null,
 };
 
-const Portal = ({ children, active, container }) =>
-  active ? ReactDOM.createPortal(children, container) : <React.Fragment>{children}</React.Fragment>;
+function Portal({ children, active, container }) {
```
nahumzs: please accept prop parameter and destructure the values inside of the function body, lets be literal that we are accepting those values from the props.

packages/SidePanel/package.json
oscarkwan: I reverted for now, since its breaking. Let me know if thats okay.

packages/helpers/src/components/Portal.js
alexzherdev: Good 🎣 @oscarkwan

AndreyChernykh: @DanilAgafonov yes, it looks like in some cases when `defaultProps` are defined, `document.body` does not exist yet.

alexzherdev: @AndreyChernykh since `defaultProps` are defined at the top level in the module, they will be evaluated when the bundle script is loaded. Since in projects the bundle script element is in the `head`, `document.body` will not exist yet.

PR #294, opened 2019-11-21
PD-13298 Tab focus logic adjustments (suggested) 🐐

mikrotron:
### 🛠 Purpose

Handle all the cases.

### ✏️ Notes to Reviewer

- Pick and choose this parts you like.

### DIFF
--- a/packages/Popover/src/Popover.js
+++ b/packages/Popover/src/Popover.js
@@ -304,29 +304,25 @@ class Popover extends React.Component {
 
   handleKeyDown = event => {
     if (event.key === "Tab" && this.isOpen() && this.$trigger) {
-      if (event.shiftKey && this.focusIsOnFirstFocusableElementInPopover()) {
+      const isFocusOnFirst = this.focusIsOnCertainElementInPopover("first") || document.activeElement === this.$content;
+      const isFocusOnLast = this.focusIsOnCertainElementInPopover("last");
+
+      if (event.shiftKey && isFocusOnFirst) {
         event.preventDefault();
-        this.focusableElements[this.triggerFocusIndex].focus(); // this closes the popover, which restores tabIndexes
-      } else if (!event.shiftKey && this.focusIsOnLastFocusableElementInPopover()) {
+        this.restoreTabIndexes();
+        focusableElements[triggerFocusIndex].focus();
+      } else if (!event.shiftKey && isFocusOnLast) {
         event.preventDefault();
-        this.focusableElements[this.triggerFocusIndex + 1].focus(); // this closes the popover, which restores tabIndexes
+        this.restoreTabIndexes();
+        focusableElements[triggerFocusIndex + 1].focus();
       }
     }
   };
 
-  focusIsOnFirstFocusableElementInPopover = () => {
-    return this.focusIsOnCertainElementInPopover("first");
-  };
-
-  focusIsOnLastFocusableElementInPopover = () => {
-    return this.focusIsOnCertainElementInPopover("last");
-  };
-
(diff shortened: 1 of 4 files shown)


packages/Popover/src/Popover.js
```diff
@@ -179,15 +173,18 @@ class Popover extends React.Component {
...
-
     if (this.isOpen()) {
       this.addListeners();
       this.setVisibilityAndPosition();
     }
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (!this.props.isOpen && nextProps.isOpen) {
+      this.open();
```
mikrotron: This results in `$trigger` being set, even when it's a controlled component.

packages/Popover/src/Popover.js
```diff
@@ -305,29 +302,25 @@ class Popover extends React.Component {
 
   handleKeyDown = event => {
     if (event.key === "Tab" && this.isOpen() && this.$trigger) {
-      if (event.shiftKey && this.focusIsOnFirstFocusableElementInPopover()) {
+      const isFocusOnFirst = this.focusIsOnCertainElementInPopover("first") || document.activeElement === this.$content;
```
mikrotron: Need to handle the case where popover is opened, then `shift+tab` is pressed immediately.

packages/Popover/src/Popover.js
```diff
@@ -305,29 +302,25 @@ class Popover extends React.Component {
...
         event.preventDefault();
-        focusableElements[triggerFocusIndex].focus(); // this closes the popover, which restores tabIndexes
-      } else if (!event.shiftKey && this.focusIsOnLastFocusableElementInPopover()) {
+        this.restoreTabIndexes();
+        this.focusableElements[this.triggerFocusIndex].focus(); // this closes the popover, which restores tabIndexes
+      } else if (!event.shiftKey && isFocusOnLast) {
         event.preventDefault();
-        focusableElements[triggerFocusIndex + 1].focus(); // this closes the popover, which restores tabIndexes
+        this.restoreTabIndexes();
+        this.focusableElements[this.triggerFocusIndex + 1].focus(); // this closes the popover, which restores tabIndexes
```
mikrotron: Oops, forgot to remove the comment... but in the case with controlled Popover with no `<Popover.Trigger>`, it wasn't being restored, but it works when it's done explicitly above.  Not sure why.

packages/Popover/src/Popover.js
```diff
@@ -305,29 +302,25 @@ class Popover extends React.Component {
...
-
-  focusIsOnLastFocusableElementInPopover = () => {
-    return this.focusIsOnCertainElementInPopover("last");
-  };
-
   focusIsOnCertainElementInPopover = which => {
-    const popoverContent = document.querySelector("[data-pka-anchor='popover.content']"); // TODO: use react
-    const focusableElementsInPopover = popoverContent.querySelectorAll(focusableElementSelector);
-
+    const focusableElementsInPopover = this.$content.querySelectorAll(focusableElementSelector);
```
mikrotron: This was an important fix to support multiple `<Popovers>` on the page.

packages/Popover/src/components/Content/Content.js
```diff
@@ -100,7 +100,7 @@ const Content = React.forwardRef((props, ref) => {
       onMouseOut={handleMouseEvent}
       onMouseOver={handleMouseEvent}
       style={contentStyles}
-      tabIndex={isOpen ? 0 : -1}
+      tabIndex="-1"
```
mikrotron: Not important to toggle this. It never needs to be in the natural tab order and can always be focussed programmatically when it's `-1`.

PR #300, opened 2019-11-22
refactor(@DataTable): reducers api and renaming

allison-c:
### Purpose 🚀
- `plugins` -> `reducers`
- Moving `<Sort />` into `<Navigation>`

### Notes ✏️

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/DataTable/src/components/Navigation/Navigation.js
```diff
@@ -4,12 +4,34 @@ import * as styled from "./Navigation.styles";
...
+  ColumnsDefinition: [],
 };
 
 export default function Navigation(props) {
-  const { children } = props;
-  return <styled.Navigation>{children}</styled.Navigation>;
+  const { children, ColumnsDefinition } = props;
+  return (
+    <styled.Navigation>
+      {React.Children.map(children, child => {
```
nahumzs: I think would be more useful to pass all the props to all component no only to the Sort.
```
<styled.Navigation>
  {React.Children.map(Children, child => React.CloneElement(child, {...props,  ...child.props)) 
</styled.Navigation>
```

packages/DataTable/src/components/Navigation/components/Sort/Sort.js
```diff
@@ -0,0 +1,84 @@
...
+        </DropdownMenu.Trigger>
+      )}
+    >
+      {ColumnsDefinition.map(({ props: columnProp }) => {
+        const { id: columnId, header, sortDirections } = columnProp;
+
+        if (!sortDirections || sortDirections.length === 0) return null;
+
+        return (
+          <DropdownMenu.Item key={columnId} onClick={noop}>
```
nahumzs: why do we need to have our own button? and our own onClick, didn't that's why we have `DropDown.Item` ?

packages/DataTable/src/components/Navigation/Navigation.js
allison-c: Ok! I was unsure about what kind of child components in the future, but it makes sense to keep it simple and clear

packages/DataTable/src/components/RowHeight/RowHeight.js
```diff
@@ -61,6 +61,7 @@ export default function RowHeight(props) {
 
 RowHeight.propTypes = propTypes;
 RowHeight.defaultProps = defaultProps;
+RowHeight.displayName = "DataTable.Navigation.RowHeight";
```
nahumzs: Why you changed the name? This is rendered on the Navigation but the logic doesn't belong to the Navigation, also if you change this here you should change it on the `DataTable.js` where use it to extract the children with extract children.

packages/DataTable/src/components/Navigation/components/Sort/Sort.js
allison-c: From my understanding, `DropDown.Item` is each item in the dropdown list which is the blue box here:
image
But the `Item` itself is not clickable, the small pink button on the right is clickable

packages/DataTable/src/components/RowHeight/RowHeight.js
allison-c: Maybe `DataTableControls.RowHeight` ? or do you prefer `DataTable.RowHeight` ? I understand `RowHeight` means the row height of the table, but this component is for changing the row height, the table itself doesn't care the value of row height. Actually if i remove this part https://github.com/acl-services/paprika/pull/300/files#diff-ccd22f487246440a452bd36678a0f647R23 I don't really need to know the display name of the component

packages/DataTable/src/components/Navigation/components/Sort/Sort.js
nahumzs: then I guess we don't need a menu, we can refactor it later when we are more sure about what UI we are going to have at the end. :)

nahumzs:
lgtm :D

Issue #131, opened 2019-07-18
Dropdown Component - Followup issues


tristanjasper:
# Bug Report
## Prerequisites
**This is just a checklist, please delete this section after**
Please answer the following questions for yourself before submitting a bug report.
- [x] I checked to make sure that this bug has not already been filed
- [ ] I reproduced the bug in the latest version of paprika
- [ ] Please label as, low / medium / high priority

*Please fill in as much detail as possible below*

## Expected behavior
Suggested issues

1. `<Dropdown.Item>` currently has an `isDestructive` prop however we would like to consider refactoring to use a `<Dropdown.DestructiveItem>` component to make it more declarative in the html approach of the consuming application

~~2. Flash of menu when closing confirmation panel, this seems to happen also in the datepicker component so is related to how we are implementing popover component
![flash](https://user-images.githubusercontent.com/4040102/61750660-c842e580-ad5a-11e9-8c93-dfb1701b78ce.gif)~~

~~3.  Need to add margin between buttons on the confirmation panel
![image](https://user-images.githubusercontent.com/4040102/61909305-9f4f5b80-aee6-11e9-8ff2-6923b3ee72e9.png)~~

~~4. Clean up Api of dropdown component to take same form as other components, removing the need for a renderTrigger attribute on the root element to render the Trigger~~

~~5. Add keyboard up/down arrow support for the Dropdown content as well as the first dropdown content item should be focused when opening menu~~

6. Change approach of focusing first dropdown list item (on menu open) to happen as a result of a popover event rather than using timeout

7. Cleanup over usage of React.cloneElement and use Context Sharing instead

## Current behavior
*What is the current behavior?*

## Screenshots / Gifs / Codepens
-Flash of menu when closing confirmation panel
![flash](https://user-images.githubusercontent.com/4040102/61750660-c842e580-ad5a-11e9-8c93-dfb1701b78ce.gif)

## Additional context
*Please provide any more additional details to illustrate the bug*

tristanjasper:
Issues 2 and 3 of this list are now resolved as part of https://github.com/acl-services/paprika/pull/160

PR #303, opened 2019-11-22
Ux 131 dropdown api change

tristanjasper:
### Purpose 🚀
Simplify the trigger Api for the dropdownComponent

### Notes ✏️

Secondary fixes include

- Added the edge property on dropdown component for alignment to the trigger
- Cleanup of the story knobs 
- Changed the Dropdown.Trigger component api to use `buttonType` prop allowing us to pass any content and set the button appearance.

### Updates 📦
- [x ] MAJOR (breaking) change to _these packages_

### Storybook 📕

### Screenshots 📸

### References 🔗
https://github.com/acl-services/paprika/issues/131_

### DIFF
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -79,6 +79,8 @@ const DropdownMenu = props => {
       const confirmationComponent = renderConfirmation(handleCloseMenu);
       return React.cloneElement(confirmationComponent, {
         align,
+        edge,
+        defaultIsOpen: true,
         getPositioningElement: () => document.getElementById(triggerId.current),
         offset: popoverOffset,
         onClose: handleCloseMenu,
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -32,7 +32,6 @@ const DropdownMenu = props => {
   const [isOpen, setIsOpen] = React.useState(false);
   const [isConfirming, setIsConfirming] = React.useState(false);
   const [renderConfirmation, setRenderConfirmation] = React.useState(null);
-
   const triggerRef = React.useRef(null);
   const menuId = React.useRef(uuid());
   const triggerId = React.useRef(uuid());
@@ -41,10 +40,8 @@ const DropdownMenu = props => {
     setIsOpen(false);
 
     if (isConfirming) {
-      setTimeout(() => {
-        setIsConfirming(false);
-        setRenderConfirmation(null);
-      }, 0);
+      setIsConfirming(false);
+      setRenderConfirmation(null);
     }
 
     if (triggerRef.current) triggerRef.current.focus();
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -3,6 +3,7 @@ import PropTypes from "prop-types";
 import uuid from "uuid/v4";
 import { AlignTypes } from "@paprika/helpers/lib/customPropTypes";
 import Popover from "@paprika/popover";
+import extractChildren from "@paprika/helpers/lib/extractChildren";
 import Divider from "./components/Divider";
 import Trigger from "./components/Trigger";
 import LinkItem from "./components/LinkItem";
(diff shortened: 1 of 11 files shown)


packages/DropdownMenu/src/components/Trigger/Trigger.js
```diff
@@ -1,38 +1,55 @@
...
+  simple: Button,
+};
+
+ButtonTypes.ALL = Object.values(ButtonTypes);
+
 const propTypes = {
   children: PropTypes.node,
-  icon: PropTypes.node,
-  isOpen: PropTypes.bool.isRequired,
+  buttonType: PropTypes.oneOf(ButtonTypes.ALL),
```
tristanjasper: changing the api of the trigger component to take a `buttonType` prop. This makes it easy for us to pass any content we like, and just change a prop to say it should show the content in a button as default or as a raw button for example which doesnt' show the button styles

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -80,6 +76,8 @@ const DropdownMenu = props => {
       const confirmationComponent = renderConfirmation(handleCloseMenu);
       return React.cloneElement(confirmationComponent, {
         align,
+        edge,
```
tristanjasper: ensuring the confirmation component uses the same edge value as the rest of the items in the dropdown

packages/DropdownMenu/stories/examples/Showcase.js
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import { Rule, Tagline } from "storybook/assets/styles/common.styles";
-import { select } from "@storybook/addon-knobs";
+import { select, text } from "@storybook/addon-knobs";
+import { AlignTypes } from "@paprika/helpers/lib/customPropTypes";
```
nahumzs: `out of the scope for this PR`

I think we should start creating a types file on helper package and have something like:

```
// types.js
export default {
  left: "LEFT",
 right: "RIGHT",
 xsmall: "XSMALL",
 large: "LARGE",
 ...etc
}
```

```js
import types from "@paprika/helpers/lib/types";

const defaultProps: {
  size: types.xsmall
}

function Something(props) {
   const {size} = props;
   if (size === types.xsmall) {
        render ""xsmall"
   }

   return "something else"
}
```

cc @mikrotron  what do you think?

packages/DropdownMenu/stories/examples/DropdownMenuDividersExample.js
```diff
@@ -3,14 +3,8 @@ import DropdownMenu from "../../src";
...
-      align="bottom"
-      renderTrigger={({ isOpen, handleOpenMenu }) => (
-        <DropdownMenu.Trigger isOpen={isOpen} onOpenMenu={handleOpenMenu}>
-          Trigger
-        </DropdownMenu.Trigger>
-      )}
-    >
+    <DropdownMenu align="bottom">
+      <DropdownMenu.Trigger>Trigger</DropdownMenu.Trigger>
       <DropdownMenu.Item onClick={() => {}}>Item 1</DropdownMenu.Item>
```
nahumzs: In an unrelated but taking advantage of this PR.

I don't think onClick should be required on DropDown.Item, 
what bout if you just want to display a `<hr />` ? do we have a `<DropDown.Divider />`?

nahumzs:
Thank you tristan :)

packages/DropdownMenu/stories/examples/Showcase.js
mikrotron: Yeah, I think having a `types` packages is a good idea.  I don't think we should export the t-shirt sizes as the default export, but that's a small implementation detail.  

Another possibility is to have things like `types`, `helpers`, maybe even `tokens` and `stylers` all bundled together in a `core` package...

nahumzs: shouldn't all of them be part of the helper package?

packages/DropdownMenu/stories/examples/DropdownMenuDividersExample.js
tristanjasper: Yes we have a Dropdown.Divider. Currently the `onClick` isn't required either. We have `Dropdown.Item` that have an `onClick` prop, but also `Dropdown.LinkItem` that doens't have an `onClick` prop as it is rendered as an `a` tag

PR #304, opened 2019-11-22
Ux 666 side panel tweaks

### LINKED ISSUE #298: Side Panel Design Tweaks
### Design tweaks

**Default Header**
1. Use the `<Heading>` component for the headings, but keep the font-weight `700` for both the Default and Compact variations
2. Make the close button `size={ShirtSizes.MEDIUM}`
3. Close button icon should be white

**Default Footer**
1. Use `size="large"` for buttons 

**Compact Header**
1. Reduce height from `56px` to `48px`
2. You could make the `padding: 12px 16px` instead of `padding: 0 16px` but it should work either way
3. Make the close button `size={ShirtSizes.SMALL}`

**Compact Footer**
1. Use `size="large"` for buttons

🗒️**Actually, all footers should be `72px` tall and use `size="large"` buttons**

**Child Panel offsets**
1. Default size = `32px` top and bottom
2. Compact size = `24px` top and bottom

oscarkwan:
### Purpose 🚀
Address side panel tweaks based off of John's design tweaks #298 

### Notes ✏️
- See Github Issue #298  for details.

### Updates 📦
-  ✅ MAJOR (breaking) change to _these packages_

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-666-side-panel-tweaks

### Screenshots 📸
Screen Shot 2019-11-22 at 2 13 23 PM

### References 🔗
Resolves #298

### DIFF
--- a/packages/SidePanel/src/components/Header/Header.js
+++ b/packages/SidePanel/src/components/Header/Header.js
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import Button from "@paprika/button";
+import Heading from "@paprika/heading";
 import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
 
 import { headerCSS } from "./Header.styles";
@@ -10,6 +11,7 @@ const propTypes = {
   hasCloseButton: PropTypes.bool,
   isCompact: PropTypes.bool,
   kind: PropTypes.oneOf([Button.Kinds.DEFAULT, Button.Kinds.PRIMARY]),
+  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]).isRequired,
   onClose: PropTypes.func,
 };
 
@@ -31,8 +33,9 @@ function darkBackgroundProps(kind) {
 const Header = React.forwardRef((props, ref) => {
   const {
     hasCloseButton,
-    kind,
     isCompact,
+    kind,
+    level,
     // injected by Dialog.js
     onClose,
     ...moreProps
@@ -40,7 +43,9 @@ const Header = React.forwardRef((props, ref) => {
 
   return (
     <div data-pka-anchor="sidepanel.header" ref={ref} css={headerCSS} kind={kind} isCompact={isCompact} {...moreProps}>
-      <div tabIndex="-1">{props.children}</div>
+      <Heading tabIndex="-1" level={level || 2}>
+        {props.children}
+      </Heading>
       <div>
         {hasCloseButton ? (
           <Button.Close
@@ -48,7 +53,7 @@ const Header = React.forwardRef((props, ref) => {
             isSemantic={false}
             onClick={onClose}
             {...darkBackgroundProps(kind)}
-            size={ShirtSizes.SMALL}
+            size={isCompact ? ShirtSizes.SMALL : ShirtSizes.MEDIUM}
           />
         ) : null}
       </div>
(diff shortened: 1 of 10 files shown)


packages/SidePanel/src/components/Header/Header.js
```diff
@@ -31,24 +33,27 @@ function darkBackgroundProps(kind) {
...
     onClose,
     ...moreProps
   } = props;
 
   return (
     <div data-pka-anchor="sidepanel.header" ref={ref} css={headerCSS} kind={kind} isCompact={isCompact} {...moreProps}>
-      <div tabIndex="-1">{props.children}</div>
+      <Heading tabIndex="-1" level={level || 2}>
+        {props.children}
+      </Heading>
```
oscarkwan: This is the breaking change, we enforce Heading now, consumer does not need to pass in <Heading> to the <SidePanel.Header> anymore.

oscarkwan: @mikrotron Do we have cases where we don't want to pass in Heading into the side panel title?

packages/SidePanel/src/components/Header/Header.js
```diff
@@ -31,24 +33,27 @@ function darkBackgroundProps(kind) {
...
+    level,
     // injected by Dialog.js
     onClose,
     ...moreProps
   } = props;
 
   return (
     <div data-pka-anchor="sidepanel.header" ref={ref} css={headerCSS} kind={kind} isCompact={isCompact} {...moreProps}>
-      <div tabIndex="-1">{props.children}</div>
+      <Heading tabIndex="-1" level={level || 2}>
```
alexzherdev: Why `level || 2` if `level` is required?

oscarkwan: True, i should not make it required above then, and just make it defaulted to `2`

packages/SidePanel/src/components/Header/Header.js
mikrotron: Probably a better question for @johnpeele.  But _maybe_ we'd want to have the close button, but no actual heading...?

johnpeele:
We want to enforce heading font sizes each in the Default and Compact variations of the panels. So I’ll leave that up to you guys on how best to implement.
________________________________

nahumzs:
what else do we need to do to merge this one :), or is ready to go?  @oscarkwan

johnpeele:
@oscarkwan For the close button in the Compact variation is there any way to easily make the font-size `14px` instead of `13px`. This will make the `X` icon perfectly centered within the button. But I would guess that the font-size is hard-coded into the Button component?

packages/SidePanel/src/components/Header/Header.js
```diff
@@ -31,24 +34,27 @@ function darkBackgroundProps(kind) {
...
+    level,
     // injected by Dialog.js
     onClose,
     ...moreProps
   } = props;
 
   return (
     <div data-pka-anchor="sidepanel.header" ref={ref} css={headerCSS} kind={kind} isCompact={isCompact} {...moreProps}>
-      <div tabIndex="-1">{props.children}</div>
+      <Heading tabIndex="-1" level={level}>
```
mikrotron: As per JP's comment, we should include the appropriate `displayLevel` prop here.

Reference: https://github.com/acl-services/paprika/pull/304#issuecomment-558300711

oscarkwan:
> @oscarkwan For the close button in the Compact variation is there any way to easily make the font-size `14px` instead of `13px`. This will make the `X` icon perfectly centered within the button. But I would guess that the font-size is hard-coded into the Button component?

Yeah it is hard coded in the button as 13px for Small Size, it will be weird if we have to make a custom size just for this case. It doesn't really look too off on my screen 🤔

PR #311, opened 2019-11-27
UX-646 Show and hide columns

allison-c:
### Purpose 🚀
Show/hide columns

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-646-show-hide-columns

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/DataTable/src/components/ColumnDefinition/ColumnDefinition.js
+++ b/packages/DataTable/src/components/ColumnDefinition/ColumnDefinition.js
@@ -10,12 +10,16 @@ ColumnDefinition.propTypes = {
   momentParsingFormat: PropTypes.string,
   sortDirections: PropTypes.arrayOf(PropTypes.oneOf([sortDirections.ASCEND, sortDirections.DESCEND])),
   type: PropTypes.oneOf([columnTypes.TEXT, columnTypes.NUMBER, columnTypes.DATE]),
+  canHide: PropTypes.bool,
+  isHidden: PropTypes.bool,
 };
 
 ColumnDefinition.defaultProps = {
   momentParsingFormat: null,
   sortDirections: [],
   type: null,
+  canHide: true,
+  isHidden: false,
 };
 
 ColumnDefinition.displayName = "DataTable.ColumnDefinition";
--- a/packages/DataTable/src/components/Navigation/components/ColumnManaging/ColumnManaging.js
+++ b/packages/DataTable/src/components/Navigation/components/ColumnManaging/ColumnManaging.js
@@ -1,0 +1,62 @@
+import React from "react";
+import PropTypes from "prop-types";
+import DropdownMenu from "@paprika/dropdown-menu";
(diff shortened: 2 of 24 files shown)


packages/DataTable/src/components/ColumnDefinition/ColumnDefinition.js
```diff
@@ -10,12 +10,16 @@ ColumnDefinition.propTypes = {
...
+  canHide: PropTypes.bool,
+  isHidden: PropTypes.bool,
 };
 
 ColumnDefinition.defaultProps = {
   momentParsingFormat: null,
   sortDirections: [],
   type: null,
+  canHide: true,
+  isHidden: false,
```
nahumzs: shouldn't this be `defaultIsHidden` because you are going to own the state once the consumer passes it? or are you planning to add an effect and watch it? and update the context state?

packages/DataTable/src/components/Navigation/components/ColumnManaging/ColumnManaging.js
```diff
@@ -0,0 +1,62 @@
...
+  if (action.type === "TOGGLE_COLUMN") {
+    return {
+      ...action.changes,
+      columns: {
+        ...action.changes.columns,
+        [action.payload]: {
+          ...action.changes.columns[action.payload],
+          isHidden: !action.changes.columns[action.payload].isHidden,
+        },
+      },
```
nahumzs: so much spread, quite hard to understand what is happening, maybe assigned those to values with a name so it is easier to read what are we merging on the state? What do you think?

packages/DataTable/stories/ShowHideColumns.stories.js
```diff
@@ -0,0 +1,51 @@
...
+import ColumnManaging from "../src/components/Navigation/components/ColumnManaging";
+
+const mockData = fixtures(1);
+function App() {
+  return (
+    <DataTable keygen="id" data={mockData} height={viewPortHeight()}>
+      <DataTable.Navigation>
+        <Sort />
+        <RowHeight />
+        <ColumnManaging />
```
nahumzs: Maybe later we can asked for better name than RowHeight and ColumnManaging, but for now is cool :)

packages/DataTable/src/components/ColumnDefinition/ColumnDefinition.js
allison-c: I was planning to add an effect and watch it, so this can be controlled but also uncontrolled

packages/DataTable/src/components/Navigation/components/ColumnManaging/ColumnManaging.js
allison-c: Yes, I agree. There's another way if we can handle this logic in the main reducer `packages/DataTable/src/reducers/table.js` we can create a sub-reducer for it.

PR #314, opened 2019-11-27
Add a testing story for the Popover in the Takeover

jamiek-acl:
### Purpose 🚀
This PR isn't a solution, it demonstrates a problem and I am looking for suggestions how to remedy this.

Since the Popover gets added at the end of the DOM, and the Takeover hijacks behaviour outside of its scope, a Popover within a Takeover doesn't work very well.

### Notes ✏️

1. The Takeover uses `focus-trap-react`, which uses `focus-trap`.
2. The Popover gets added at the end of the DOM -- outside of the Takeover's HTML element.

When you open a popover, focus goes to its content: https://github.com/acl-services/paprika/blob/master/packages/Popover/src/Popover.js#L357

Except that the Takeover says `I dont think so!` and takes back the focus:  https://github.com/davidtheclark/focus-trap/blob/master/index.js#L253
This causes the takeover to open quickly, then close, and focus moves to a weird spot.

### Interacting with a Popover with the `isEager` or `shouldKeepFocus` prop:

The Popover opens, but the Takeover prevents you from interacting with its contents with the mouse or keyboard.
Nov-27-2019 15-30-45

### Interacting with a Popover with no props:

The Popover opens then closes right away.
Nov-27-2019 15-31-28

### DIFF
--- a/packages/Popover/stories/Popover.stories.js
+++ b/packages/Popover/stories/Popover.stories.js
@@ -8,6 +8,7 @@ import PositioningElement from "./examples/PositioningElement";
 import ScrollContainer from "./examples/ScrollContainer";
 import Transformed from "./examples/Transformed";
 import FocusTest from "./examples/FocusTest";
+import FocusTakeoverTest from "./examples/FocusTakeoverTest";
 import A11y from "./examples/A11y";
 import Cypress, { propHandles } from "./examples/Cypress";
 import Screener from "./examples/Screener";
@@ -24,7 +25,8 @@ storiesOf("Popover", module)
 
 storiesOf("Popover/Dev", module)
   .add("Has container with a CSS transform", () => <Transformed />)
-  .add("Testing Focus Management", () => <FocusTest />);
+  .add("Testing Focus Management", () => <FocusTest />)
+  .add("Testing Focus Management in Takeover", () => <FocusTakeoverTest />);
 
 storiesOf("Popover/Automation Tests", module)
   .add("Accessibility", () => <A11y />)
--- a/packages/Popover/stories/examples/FocusTakeoverTest.js
+++ b/packages/Popover/stories/examples/FocusTakeoverTest.js
@@ -5,101 +5,70 @@ import InfoIcon from "@paprika/icon/lib/InfoCircle";
 import Popover from "../../src";
 
 export default function ExampleStory() {
-  const [isOpen, setIsOpen] = React.useState(true);
+  const viewInIframeLink = (
+    <a href="/iframe.html?id=popover-dev--testing-focus-management-in-takeover&viewMode=story" target="_blank">
+      View in iframe
+    </a>
+  );
 
-  function handleOpen() {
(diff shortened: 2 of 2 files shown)


jamiek-acl:
We could put the Popover back in-line (not at the end of the DOM) -- or have it as an option.

jamiek-acl:
We could use a different focus trapper, or fork theirs.

allison-c:
I think it's good to have a `parent` prop in `<Popover.Content />`, by default the `parent` is `document.body`. In some cases like above, we can say `<Popover.Content parent={myTakeoverContent} />` so that the popover panel can be rendered inside the takeover, or sidepanel

oscarkwan:
I think having the Popover as a Portal should be an option -- it certain cases it makes sense to make in line?

oscarkwan:
I'm actually making Toast as an option to render as a Portal

tristanjasper:
Perhaps we could use `focus-trap-react` in the popover content component as well to prevent it from propagating to the parent?

jamiek-acl:
> Perhaps we could use `focus-trap-react` in the popover content component as well to prevent it from propagating to the parent?

I dont think we want to trap the focus in the popover (...although, if the user pressed `enter` to open it, shouldnt they have to press `esc` to get out of it...?)

mikrotron:
I like Allison's suggestion, I think that could work.

Alternatively maybe we can add an attribute, like `data-pka-portal` to the `<Popover>` and then utilize the `onDeactivate` prop of the `focus-trap` to check and see if an element with that attribute has received the click, and then let the focus stay there?
https://github.com/davidtheclark/focus-trap#usage

@DanilAgafonov, I think you are more familiar with the `focus-trap` library, do you think that approach would work?

If we can avoid forking the `focus-trap`, I think we should.

nahumzs:
One possible problem with that suggestion is that the popover can't float free outside of the modal so potentially could end be wrapped or cut by an overflow position.

The main reason for the popover to be rendered on the portal is to be independent of any container al always be visible. 

We have that implementation in `acl-ui-3` and we didn't follow it because of all the side effects of having the popover render inside of any modal. 

Forking the repo / open a PR for adding a custom data attribute can be a possible solution, is that we will need to implement on our popovers a way to find the next focusable element after the tab is clicked.

DanilAgafonov:
Maybe we can migrate from stale `focus-trap` to https://github.com/theKashey/react-focus-lock? It supports React Portals.

allison-c:
> Maybe we can migrate from stale `focus-trap` to https://github.com/theKashey/react-focus-lock? It supports React Portals.

I like this idea too! Also I noticed they support `whiteList` we can check if the `node` is our popover component by `pka-id` inside the function

nahumzs:
let's do that @DanilAgafonov

jamiek-acl:
Resolved by changing the focus trap library in this PR: https://github.com/acl-services/paprika/pull/363

PR #318, opened 2019-11-28
Ux 131 dropdown keyboard accessible

tristanjasper:
### Purpose 🚀

Adding keyboard shortcuts to dropdown component item navigation.
Side notes:
- On opening the menu the first element will get focus
- Disabled items should also be focusable as per https://www.w3.org/TR/wai-aria-practices/#keyboard-interaction-12
- Modified `RawButton` component to prioritize `tabIndex` prop over `isDisabled` prop to dictate if it can receive focus.

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-131-dropdown-keyboard-accessible

### Screenshots 📸
dropdown-keyboard

### References 🔗

### DIFF
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -86,6 +86,8 @@ function DropdownMenu(props) {
       focusAndSetIndex(0);
     } else if (event.key === "End") {
       focusAndSetIndex(dropdownLastItemIndex);
+    } else if (event.key === "Enter" || event.key === " ") {
+      // do nothing
     } else {
       handleCloseMenu();
     }
@@ -154,12 +156,6 @@ function DropdownMenu(props) {
       align={align}
       offset={popoverOffset}
       isOpen={isOpen}
-      onOpen={() => {
-        console.log("calling on open dropdown");
-      }}
-      onDelayedOpen={() => {
-        console.log("calling on delayed open dropdown");
-      }}
       onClose={() => {
         if (!isConfirming) handleCloseMenu();
       }}
@@ -167,16 +163,7 @@ function DropdownMenu(props) {
       {...moreProps}
     >
       <Popover.Trigger>{renderTrigger()}</Popover.Trigger>
-      <Popover.Content
-        onOpen={() => {
-          console.log("calling on open content dropdown");
-        }}
-        onDelayedOpen={() => {
-          console.log("calling on delayed open content dropdown");
-        }}
-        id={menuId.current}
-        role={!isConfirming ? "menu" : null}
-      >
+      <Popover.Content id={menuId.current} role={!isConfirming ? "menu" : null}>
         {isOpen && <Popover.Card>{renderContent()}</Popover.Card>}
       </Popover.Content>
     </Popover>
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -79,6 +79,8 @@ const DropdownMenu = props => {
       const confirmationComponent = renderConfirmation(handleCloseMenu);
       return React.cloneElement(confirmationComponent, {
         align,
+        edge,
+        defaultIsOpen: true,
         getPositioningElement: () => document.getElementById(triggerId.current),
         offset: popoverOffset,
         onClose: handleCloseMenu,
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -32,7 +32,6 @@ const DropdownMenu = props => {
   const [isOpen, setIsOpen] = React.useState(false);
   const [isConfirming, setIsConfirming] = React.useState(false);
   const [renderConfirmation, setRenderConfirmation] = React.useState(null);
-
   const triggerRef = React.useRef(null);
   const menuId = React.useRef(uuid());
   const triggerId = React.useRef(uuid());
@@ -41,10 +40,8 @@ const DropdownMenu = props => {
     setIsOpen(false);
 
     if (isConfirming) {
-      setTimeout(() => {
-        setIsConfirming(false);
-        setRenderConfirmation(null);
-      }, 0);
+      setIsConfirming(false);
+      setRenderConfirmation(null);
     }
 
     if (triggerRef.current) triggerRef.current.focus();
--- a/packages/DropdownMenu/src/DropdownMenu.js
+++ b/packages/DropdownMenu/src/DropdownMenu.js
@@ -57,8 +57,8 @@ function DropdownMenu(props) {
 
   const handleOpenMenu = () => {
     setIsOpen(true);
-    // https://github.com/acl-services/paprika/issues/131 #6
-    // Todo should be called via an onOpen callback via popover mounting content rather than timeout.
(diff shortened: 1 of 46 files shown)


packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -49,10 +58,42 @@ const DropdownMenu = props => {
...
   };
 
   const extractedChildren = extractChildren(children, ["DropdownMenu.Trigger"]);
 
+  const dropdownLastItemIndex =
+    React.Children.toArray(
+      extractedChildren.children.filter(
+        child =>
+          child.type &&
+          (child.type.displayName === "DropdownMenu.Item" || child.type.displayName === "DropdownMenu.LinkItem")
```
tristanjasper: Being specific as we do not want to include the Dropdown.Divider in this count

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -49,10 +58,42 @@ const DropdownMenu = props => {
...
+    ).length - 1;
+
+  const onKeyDown = (event, currentIndex) => {
+    if (event.key === "ArrowDown") {
+      const indexToFocus = currentIndex === dropdownLastItemIndex ? 0 : currentIndex + 1;
+      focusAndSetIndex(indexToFocus);
+    } else if (event.key === "ArrowUp") {
+      const indexToFocus = currentIndex === 0 ? dropdownLastItemIndex : currentIndex - 1;
+      focusAndSetIndex(indexToFocus);
+    } else if (event.key === "Home") {
```
tristanjasper: As discussed with @mikrotron `ArrowRight` and `ArrowLeft` keys should only be relevant if the dropdown Item has nest submenus, otherwise those keys should just close the menu

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -49,10 +58,42 @@ const DropdownMenu = props => {
...
+      focusAndSetIndex(0);
+    } else if (event.key === "End") {
+      focusAndSetIndex(dropdownLastItemIndex);
+    } else if (event.key === "Enter" || event.key === " ") {
+      // do nothing
```
tristanjasper: If i do not intercept these keys then this cause the popover content to flash and reopen the menu. This I believe is due to the default browser behaviour of these events happening.

packages/DropdownMenu/src/components/Item/Item.js
```diff
@@ -28,27 +31,39 @@ const propTypes = {
...
+    onClose,
+    onShowConfirmation,
+    renderConfirmation,
+    onKeyDown,
+    ...moreProps
+  } = props;
 
   const handleClickItem = () => {
-    onClick();
+    onClick(children);
```
tristanjasper: returning the children here to the consuming application click handler, although perhaps it isn't required as they can get the value themselves as the onClick is on each item.. 🤷‍♂ I decided it doesn't hurt to pass it though

packages/DropdownMenu/src/components/Item/Item.js
```diff
@@ -28,27 +31,39 @@ const propTypes = {
...
     onClose();
   };
 
   const itemProps = {
     onClick: renderConfirmation !== null ? onShowConfirmation : handleClickItem,
     role: "menuitem",
+    className: "dropdown-item",
```
tristanjasper: Adding the classname so that the on onFocus event can get the right element - https://github.com/acl-services/paprika/pull/318/files#diff-edd13737a415dcf236bf45816da474cbR42

packages/DropdownMenu/src/components/Item/Item.js
```diff
@@ -28,27 +31,39 @@ const propTypes = {
...
+    onKeyDown,
     isDestructive,
   };
 
   return (
-    <RawButton hasInsetFocusStyle css={ItemStyles} {...itemProps} {...moreProps}>
+    <RawButton tabIndex={0} hasInsetFocusStyle css={ItemStyles} {...itemProps} {...moreProps}>
```
tristanjasper: I am setting the tabIndex specifically as 0 so that even if this item is disabled it can take focus. This has required a small change in how the `RawButton` component evaluates tab index. https://github.com/acl-services/paprika/pull/318/files#diff-57b7f13a30b6174f5a9c7477773ad940R99

packages/DropdownMenu/src/components/LinkItem/LinkItem.js
```diff
@@ -2,27 +2,32 @@ import React from "react";
...
+  onKeyDown: () => {},
 };
 
 const LinkItem = props => {
-  const { children, isExternal, link, ...moreProps } = props;
+  const { children, onKeyDown, isExternal, link, ...moreProps } = props;
 
   const linkItemProps = {
     role: "menuitem",
+    className: "dropdown-item",
```
tristanjasper: This is pretty much the same as what is being applied to the `Item.js` component. I think we should perhaps consolidate much of the code between `Item.js` and `LinkItem.js` so  we keep it DRY.

packages/RawButton/src/RawButton.js
```diff
@@ -86,7 +96,7 @@ const RawButton = React.forwardRef((props, ref) => {
       onKeyDown={handleKeyDown}
       onKeyUp={handleKeyUp}
       ref={rawButtonRef}
-      tabIndex={isDisabled ? null : tabIndex}
+      tabIndex={getTabIndex()}
```
tristanjasper: Changing so that if there is a tabIndex provided on the component it uses it, rather than the `disabled` prop overriding it. In the case of a menubar we want even disabled items to be focusable.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -27,15 +27,24 @@ const defaultProps = {
...
+  const [currentFocusIndex, setFocusIndex] = React.useState(0);
   const triggerRef = React.useRef(null);
   const menuId = React.useRef(uuid());
   const triggerId = React.useRef(uuid());
 
+  const dropdownListRef = React.useRef(null);
+
+  function focusAndSetIndex(index) {
+    if (dropdownListRef && dropdownListRef.current && index !== undefined)
+      dropdownListRef.current.querySelectorAll(".dropdown-item")[index].focus();
```
oscarkwan: We could use pka-anchors instead

packages/DropdownMenu/src/components/LinkItem/LinkItem.js
```diff
@@ -2,27 +2,32 @@ import React from "react";
...
 };
 
 const defaultProps = {
   isExternal: false,
+  onKeyDown: () => {},
 };
 
 const LinkItem = props => {
-  const { children, isExternal, link, ...moreProps } = props;
+  const { children, onKeyDown, isExternal, link, ...moreProps } = props;
```
oscarkwan: add some description for aria label if its isExternal

packages/DropdownMenu/src/DropdownMenu.js
tristanjasper: 👌

packages/DropdownMenu/src/components/LinkItem/LinkItem.js
tristanjasper: 🆗

PR #305, opened 2019-11-22
Os creating new button package

slee1130:
### Purpose 🚀
- DialogAction component
- Added a common StoryHeader component
- Changed some Readme

### Notes ✏️
_details of code change / secondary purposes of this PR_
-DialogAction.js, DialogAction.stories.js, DialogAction.spec.js, and README are added. 
-There is no styling for DialogAction so gave button space for &nbsp for now but if other tips are provided, it would be appreciated. 
@oscarkwan 

### Updates 📦
Adding new component `DialogActions`
- [x] PATCH (big fix) change to _Toast, L10n_

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/OS-creating_new_button_package

### Screenshots 📸
_optional but highly recommended_
Screen Shot 2019-11-22 at 2 37 09 PM

### References 🔗
_relevant Jira ticket / GitHub issues_

PR #320, opened 2019-11-29
UX-457 Modal component

DanilAgafonov:
### Purpose 🚀
1. Introduce Modal component. See description in Design System: https://design.wegalvanize.com/p/components/modal
2. Extract commons between `@paprika/modal` and `@paprika/takeover`

### Notes ✏️

### Updates 📦
- [x] First release for `@paprika/overlay` and `@paprika/modal`
- [x] MINOR (backward compatible) change to `@paprika/tokens`
- [x] MAJOR change to `@paprika/takeover` (see commits with `!`)

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-457-modal

### Screenshots 📸
localhost_9009_iframe html_id=modal--basic

### References 🔗
https://aclgrc.atlassian.net/browse/UX-457


packages/Overlay/src/Overlay.js
```diff
@@ -0,0 +1,145 @@
...
+import tokens from "@paprika/tokens";
+import Wrapper from "./Wrapper";
+import Backdrop from "./Backdrop";
+
+const propTypes = {
+  className: PropTypes.string,
+  backdropClassName: PropTypes.string,
+  isOpen: PropTypes.bool.isRequired,
+  children: PropTypes.func,
+  showBackdrop: PropTypes.bool,
```
nahumzs: this should be `hasBackDrop` cc @mikrotron

packages/Modal/README.md
```diff
@@ -0,0 +1,39 @@
+# Modal
+
+https://design.wegalvanize.com/p/components/modal
+
+## Components
+
+You can use any of the following components to compose the Modal:
+
+```jsx
+<Modal.Header />
```
nahumzs: Shouldn't we have also `<Modal.Overlay />` like we do on the sidepanel? then:

<Modal>
  <Modal.Overlay {...customProps} />
</Modal>

packages/Overlay/src/Overlay.js
```diff
@@ -0,0 +1,145 @@
...
+import { Transition } from "react-transition-group";
+import FocusTrap from "focus-trap-react";
+import LockBodyScroll from "@paprika/helpers/lib/components/LockBodyScroll";
+import Portal from "@paprika/helpers/lib/components/Portal";
+import tokens from "@paprika/tokens";
+import Wrapper from "./Wrapper";
+import Backdrop from "./Backdrop";
+
+const propTypes = {
+  className: PropTypes.string,
```
nahumzs: shouldn't this be part of `{...moreProps}` part of the root node?

nahumzs: we shouldn't have it as a prop because we can be assigned on `...moreProps`

packages/Overlay/src/Overlay.js
```diff
@@ -0,0 +1,145 @@
+import React from "react";
```
nahumzs: 💯 cool to have an overlay component now :)

packages/Takeover/src/Takeover.js
```diff
@@ -49,54 +39,27 @@ const Takeover = props => {
...
-                {headerExtracted && <styled.Header {...headerExtracted.props} onClose={onClose} />}
-                {contentExtracted && (
-                  <styled.ContentWrapper role="region" tabIndex="0">
-                    {contentExtracted}
-                  </styled.ContentWrapper>
-                )}
-                {children}
-              </styled.Wrapper>
-            </FocusTrapLibrary>
+    <Overlay
```
nahumzs: cool! cleaner!!

packages/Modal/src/index.js
```diff
@@ -0,0 +1,12 @@
+import Modal from "./Modal";
+import FocusTrap from "./components/FocusTrap";
+import Header from "./components/Header";
+import Content from "./components/Content";
+import Footer from "./components/Footer";
+
+Modal.FocusTrap = FocusTrap;
```
nahumzs: Same I think we should use `<Modal.Overlay />` and keep the same pattern as the sidepanel.

packages/Modal/src/Modal.js
```diff
@@ -0,0 +1,74 @@
...
+      onClose={onClose}
+      onAfterOpen={onAfterOpen}
+      onAfterClose={onAfterClose}
+      focusTrapOptions={extendedFocusTrapOptions}
+    >
+      {state => (
+        <styled.Wrapper size={size}>
+          <styled.Dialog state={state} role="dialog" data-pka-anchor="modal">
+            {headerExtracted && <styled.Header {...headerExtracted.props} onClose={onClose} />}
+            <styled.ContentWrapper role="region" tabIndex="0">
```
mikrotron: I don't think this should have a tabIndex of `0` – it is not interactive and doesn't need to be in the keyboard tab order.  

Does it need a `-1` so that it can be the fallback focussable element on open?

packages/Modal/src/Modal.js
```diff
@@ -0,0 +1,74 @@
...
+    <styled.Overlay
+      isOpen={isOpen}
+      onClose={onClose}
+      onAfterOpen={onAfterOpen}
+      onAfterClose={onAfterClose}
+      focusTrapOptions={extendedFocusTrapOptions}
+    >
+      {state => (
+        <styled.Wrapper size={size}>
+          <styled.Dialog state={state} role="dialog" data-pka-anchor="modal">
```
mikrotron: The element with `role=dialog` should also have `aria-modal=true` and either `aria-label` or `aria-labelledby`.  While the preferred approach is probably to use `aria-labelledby` with the `id` for the `<Heading>` element, that's a bit awkward with this composition, and I think it's probably easier for consumer to simply put the same content into the `aria-label` (but lets call the prop `a11yText`, to maintain our convention, and to insulate typical consumer from the implementation details).

Reference:
https://www.w3.org/TR/wai-aria-practices/#dialog_roles_states_props

packages/Modal/src/Modal.styles.js
```diff
@@ -0,0 +1,88 @@
+import styled, { css } from "styled-components";
+import tokens from "@paprika/tokens";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import OriginalOverlay from "@paprika/overlay";
+import OriginalHeader from "./components/Header";
+import OriginalFooter from "./components/Footer";
+
+/*
+Some comments according modal positioning:
```
mikrotron:
```suggestion
Some comments concerning modal positioning:
```

packages/Modal/src/Modal.styles.js
```diff
@@ -0,0 +1,88 @@
...
+  ${({ state }) => states[state]};
+`;
+
+export const Wrapper = styled.div`
+  flex: 0 1 auto;
+  margin: ${tokens.modal.margin};
+  max-height: calc(100% - ${tokens.modal.margin} - ${tokens.modal.margin});
+  max-width: calc(100% - ${tokens.modal.margin} - ${tokens.modal.margin});
+  width: ${({ size }) => mapShirtSizesToValues[size] || mapShirtSizesToValues[ShirtSizes.MEDIUM]};
+  z-index: 1;
```
mikrotron:
```suggestion
  ${stylers.z(1)};
```

packages/Modal/src/Modal.styles.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+`;
+
+const mapShirtSizesToValues = {
+  [ShirtSizes.SMALL]: tokens.modal.sizes.sm,
+  [ShirtSizes.MEDIUM]: tokens.modal.sizes.md,
+  [ShirtSizes.LARGE]: tokens.modal.sizes.lg,
+};
+
+export const Dialog = styled.div`
```
mikrotron: I'd recommend including `stylers.boxSizingStyles`, for more resilience.

https://github.com/acl-services/paprika/blob/master/packages/Stylers/src/includes.js#L8-L15

packages/Modal/src/components/Content.js
```diff
@@ -0,0 +1,10 @@
+import styled from "styled-components";
+import { spacer } from "@paprika/stylers/lib/helpers";
+
+const Content = styled.div`
+  padding: 0 ${spacer(2)} ${spacer(2)};
```
mikrotron: Pretty minor detail, but I'm kinda 😕 on there being bottom padding, but no top padding.  It makes sense this way if the first element is a `<p>` or another `<h>`, since they have margin built in, but I feel like the `Content` should be agnostic.

packages/Modal/src/components/FocusTrap.js
```diff
@@ -0,0 +1,13 @@
+import Overlay from "@paprika/overlay";
+
+const propTypes = Overlay.propTypes.focusTrapOptions;
+
+const defaultProps = Overlay.defaultProps.focusTrapOptions;
```
mikrotron: I know that this is consistent with our standard, but for such a small component, it seems like unnecessary indirection.  WDYT?

packages/Modal/src/components/Footer.js
```diff
@@ -0,0 +1,20 @@
...
+import { spacer } from "@paprika/stylers/lib/helpers";
+import tokens from "@paprika/tokens";
+
+const Footer = styled.div`
+  align-items: center;
+  background-color: ${tokens.modal.footer.backgroundColor};
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-end;
+  padding: ${spacer(1.5)};
```
mikrotron:
```suggestion
  padding: ${tokens.spaceLg};
```

packages/Modal/src/components/Footer.js
```diff
@@ -0,0 +1,20 @@
...
+
+  & > * {
+    margin: ${spacer(0.5)};
```
mikrotron: padding: ${tokens.spaceSm};

mikrotron: On second though, this seems a bit intrusive 🤔 
Again I think the parent container should be agnostic of its contents.

packages/Modal/src/components/Header.js
```diff
@@ -0,0 +1,37 @@
...
+  hasCloseButton: true,
+  level: 3,
+  onClose: () => {},
+};
+
+const Header = React.forwardRef((props, ref) => {
+  const { children, level, hasCloseButton, onClose, ...moreProps } = props;
+
+  return (
+    <styled.Wrapper ref={ref} {...moreProps}>
```
mikrotron: Can we call it `<styled.Header>` instead?

packages/Modal/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "access": "public"
+  },
+  "directories": {
+    "lib": "lib",
+    "test": "tests"
+  },
+  "peerDependencies": {
+    "react": "^16.12.0",
+    "react-dom": "^16.12.0",
+    "styled-components": "^4.4.1"
```
mikrotron: Does it matter that this component has a higher version dependency than the rest? Should we be keeping them all in sync?

packages/Modal/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "main": "lib/index.js",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika/tree/master/packages/Modal"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "directories": {
```
alexzherdev: Since this doesn't seem to be doing anything useful yet (https://docs.npmjs.com/files/package.json#directories) I'd omit for now? Otherwise there may be confusion from people seeing this and going "hmm what's this do? this is probably doing something useful... why is it only in this package but not others though? is this one different somehow?"

packages/Modal/stories/Modal.stories.js
```diff
@@ -0,0 +1,129 @@
...
+    </LongBlock>
+  );
+};
+
+storiesOf("Modal", module)
+  .addDecorator(withKnobs)
+  .add("Basic", () => (
+    <ModalStory>
+      <Modal.Content>
+        {Array(25)
```
mikrotron: Maybe this helper could be extended to include some custom content?

https://github.com/acl-services/paprika/blob/master/.storybook/assets/styles/common.styles.js#L57

packages/Modal/stories/Modal.stories.js
```diff
@@ -0,0 +1,129 @@
...
+
+storiesOf("Modal / screener", module)
+  .add("when little content", () => (
+    <ModalStory>
+      <Modal.Content>
+        <p>Some content here</p>
+      </Modal.Content>
+    </ModalStory>
+  ))
+  .add("when height bigger than (100% - (124px * 2)) on 1280x1024", () => (
```
mikrotron: Is there a way to name this that explains what you're _really_ trying to test here? These details could maybe be included in the actual story content.

packages/Overlay/src/Overlay.js
```diff
@@ -0,0 +1,145 @@
...
+          unmountOnExit
+          in={isOpen}
+          timeout={+tokens.overlay.animationDuration}
+          onEnter={handleTransitionEnter}
+          onEntered={onAfterOpen}
+          onExited={onAfterClose}
+        >
+          {state => (
+            <Wrapper {...moreProps} onKeyDown={handleEscKey}>
+              {showBackdrop && <Backdrop className={backdropClassName} state={state} onClick={onClose} />}
```
mikrotron: If we add a `data-pka-anchor='overlay.backdrop'` on the `<Backdrop>` then a custom `className` prop on `<Wrapper>` should allow consumers to target both, and we can eliminate the `backdropClassName` prop.

packages/Overlay/src/Backdrop.js
```diff
@@ -0,0 +1,32 @@
...
+`;
+
+const states = {
+  entering: openedCss,
+  entered: openedCss,
+  exiting: closedCss,
+  exited: closedCss,
+};
+
+const Backdrop = styled.div`
```
mikrotron: Can we consolidate `Backdrop.js` with `Wrapper.js` into `Overlay.styles.js`, and then use `styled.Overlay` and `styled.Backdrop` in `Overlay.js`?

packages/Overlay/src/index.js
```diff
@@ -0,0 +1,3 @@
+export { default } from "./Overlay";
+export { default as Backdrop } from "./Backdrop";
+export { default as Wrapper } from "./Wrapper";
```
mikrotron: Who needs to consume these styled-components besides `<Overlay>`?

packages/Takeover/src/Takeover.js
```diff
@@ -1,12 +1,8 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { Transition } from "react-transition-group";
-import FocusTrapLibrary from "focus-trap-react";
-import LockBodyScroll from "@paprika/helpers/lib/components/LockBodyScroll";
-import Portal from "@paprika/helpers/lib/components/Portal";
+import Overlay from "@paprika/overlay";
```
mikrotron: 🙌 🎉

packages/Modal/README.md
mikrotron: Also, `<Modal.FocusTrap>`?

packages/Tokens/src/tokens.yaml
```diff
@@ -155,6 +150,24 @@ card:
...
+#
+modal:
+  background-color: $color--white
+  border-radius: $card--border-radius
+  shadow: 0 8px 12px -4px rgba(63, 61, 60, 0.2), 0 1px 3px 0 rgba(63, 61, 60, 0.1)
+  sizes:
+    sm: 450px
+    md: 675px
+    lg: 900px
+  margin: 24px
```
mikrotron:
```suggestion
  margin: ($space * 3)
```

mikrotron:
Not sure if it's an issue with the `<Modal>` or the `<SidePanel>`, but in the "with nested SidePanel" story, if I click on the `<SidePanel.Overlay>`, then try click the close button, it does not work.

packages/Modal/README.md
DanilAgafonov: Added `<Modal.Overlay />`, but removed `<Modal.FocusTrap>`. Options for FocusTrap can be passed as:
```jsx
<Modal>
  ...
  <Modal.Overlay
    focusTrapOptions={{
      initialFocus: ...,
    }}
  />
  ...
</Modal>
```

packages/Overlay/src/Overlay.js
DanilAgafonov: Ah, I see. Just forgot to drop className from propTypes when introduced `...moreProps` (previously `className` was directly passed to `Wrapper`).

packages/Modal/src/index.js
DanilAgafonov: > keep the same pattern as the sidepanel

What do you mean by this?

packages/Modal/src/Modal.js
DanilAgafonov: `tabIndex=0` adds ability to focus this area. It is required to allow scroll this area by keyboard. Without this user don't have an ability scroll content area by keyboard.

packages/Modal/src/components/Content.js
DanilAgafonov: > `Content` should be agnostic of what it contains.

Yes, definitely. But it is not a reason why there is no top padding. There is no top padding, because `Header` already have bottom padding. There will be too much space if we have paddings for both `Header` and `Content`.

But I am thinking about case when there will be only `Content` and no `Header`:

```jsx
<Modal>
  <Modal.Content>
    ...
  </Modal.Content>
</Modal>
```

I will try to implement some solution for this case.

packages/Modal/package.json
DanilAgafonov: This lines automatically generated by `npx lerna create ...`. I don't think that it is good idea to remove them every time we start new package.

packages/Modal/package.json
DanilAgafonov: Say user install `@paprika/takeover` and `@paprika/modal`.

Takeover require `styled-components@^4.2.0`
Modal require `styled-components@^4.4.1`

Peer dependencies will be resolved on user side.

So yarn/npm will install only one version of SC (compatible with all packages). So `styled-components@4.2.1` will be installed.

So I think it is fine to have different versions of SC in packages.

packages/Modal/src/components/Header.js
DanilAgafonov: Say we use BEM methodology. Than we will have following structure: `.header > .header__wrapper` (not `.header > .header__header`).

When I think about semantic, it feels a bit strange for me having `Header` inside `Header`.

packages/Overlay/src/Overlay.js
DanilAgafonov: Yes, I will add `data-pka-anchor` attr for backdrop, but don't think that we should remove `backdropClassName`.

I think we should avoid having following CSS on consumer side:
```scss
.my-custom-class-for-overlay {
  // some styles for overlay wrapper  

  [data-pka-anchor='overlay.backdrop'] {
    // some styles for overlay backdrop
  }
}
```

```jsx
import styles from './styles.scss';

export default () => (
  <Modal>
    ...
    <Modal.Overlay className={styles.myCustomClassForOverlay} />
    ...
  </Modal>
);
```

When we use `[data-pka-anchor='overlay.backdrop']` we rely on the fact that element with this selector located somewhere inside in `.my-custom-class-for-overlay`. And this sounds like we rely on internal details of `Overlay` implementation.

I think this way is much cleaner:
```scss
.my-custom-class-for-overlay {
  // some styles for overlay wrapper
}

.my-custom-class-for-overlay-backdrop {
  // some styles for overlay backdrop
}
```

```jsx
import styles from './styles.scss';

export default () => (
  <Modal>
    ...
    <Modal.Overlay className={styles.myCustomClassForOverlay} backdropClassName={styles.myCustomClassForOverlayBackdrop} />
    ...
  </Modal>
);
```

packages/Overlay/src/Backdrop.js
DanilAgafonov: I was thinking about this. But decided to create `Backdrop` and `Wrapper` as separated components to have an ability to expose them in `index.js`. I saw your question about why I expose this components. I will answer you about this below your comment.

packages/Overlay/src/index.js
DanilAgafonov: I decided to expose this components to have ability to style them when we consume them with `styled-components`.

Example:
```jsx
import Overlay, { Backdrop } from '@paprika/overlay';

const StyledOverlay = styled(Overlay)`
  // some styles for overlay

  ${Backdrop} {
    // some styles for backdrop
  }
`;
```

If we remove `Backdrop` from exports, than only way to style backdrop will be:
```jsx
import Overlay from '@paprika/overlay';

let StyledOverlay = styled(Overlay)`
  // some styles for overlay
`;

StyledOverlay = styled(({ className, ...rest }) => <StyledOverlay backdropClassName={className} {...rest} />)`
  // some styles for backdrop
`;
```

I agree that probably exporting `Wrapper` is a bit unnecessary. I can remove it, but think it does not affect anything.

DanilAgafonov:
@mikrotron 
> Not sure if it's an issue with the <Modal> or the <SidePanel>, but in the "with nested SidePanel" story, if I click on the <SidePanel.Overlay>, then try click the close button, it does not work.

The issue with Storybook. It binds click listener that prevent listening click on overlay. If I remove this listener in Chrome DevTools, than everything works fine.

packages/Modal/src/components/Header.js
nahumzs: I think this is minor, is ok going with wrapper mike :)

packages/Modal/package.json
nahumzs: only if those versions are compatible between them?, I wouldn't like to have a package with sc 4 and another with 3 and eventually other with 5, which would make it harder for us to help our consumers in an episode of not having the right version.

DanilAgafonov: > only if those versions are compatible between them?

Yes.

>  I wouldn't like to have a package with sc 4 and another with 3 and eventually other with 5, which would make it harder for us to help our consumers in an episode of not having the right version.

I definitely wouldn't like to have multiple different major SC versions too. But I think it is ok to have different SC versions which are compatible.

packages/Modal/src/components/Content.js
DanilAgafonov: @mikrotron Could you please take a look? What do you think about this solution? https://github.com/acl-services/paprika/pull/320/commits/ec0630cb9e835e51863ffd4957d93ca8ce33c4d8

packages/Modal/src/components/Footer.js
DanilAgafonov: @nahumzs @mikrotron What do we think if we wrap each child to `div` an assign styles to this `div`? So we can avoid problems if children have its own margins. They will never overridden.

https://github.com/acl-services/paprika/pull/320/commits/feae3ecbe58dce863f1b203b1b4e9afeffae82af
https://github.com/acl-services/paprika/pull/320/commits/6989fbc9bfb30fde54b4c576a117d4ab31bcb245

mikrotron: I think we can simplify the `<Modal.Footer>` and render more as a pure container.  In some cases we might want to include a "do not show again" checkbox or something similar, not sure if this will handle that?  It does allow for a convenient API, however...

Soon we will have a new component, `<DialogActions>` that will control the layout of the usual footer (two or three buttons).  It is in progress, but you may have some suggestions for it:
https://github.com/acl-services/paprika/pull/305

What do you think?

packages/Modal/src/components/Header.js
mikrotron: 👌

packages/Modal/src/components/Content.js
mikrotron: I see what you're going for there, but it feels pretty complex.  What about simply using `margin` instead of `padding` for the `Content`, so that the margins will collapse and even if the first child does have some of its own top margin, it won't increase the spacing?

packages/Modal/src/Modal.js
mikrotron: This is a good point, and a bit of a blind spot in the accessibility material I've been studying.  Let's leave it like this for now, but I've asked @sunminsarahlee  to mock up an alternative focus style for this kind of situation where the focussed element is _NOT_ actionable, which we can apply in a follow up task.

nahumzs:
@DanilAgafonov @mikrotron do you think guys think this component is good enough to merge? or there is any major change that we need to do before merging?

packages/Modal/package.json
alexzherdev: I guess that's fair (although IMO we should feel free to change lerna's output if it doesn't fit our needs), but my point still stands. Not blocking though

This also brings up the point of having a blessed way to create new packages, e.g. I think @nahumzs had a script for that in the beginning.

packages/Modal/src/components/Content.js
DanilAgafonov: Good idea, but:
1. Margins doesn't collapse inside `flex` container.
2. Even if container will not be `flex`, `Content` wrapped with `ContentWrapper`, so margins will not be collapsed anyway.

packages/Modal/src/components/Footer.js
DanilAgafonov: IMO, `<DialogActions>` does nothing. My suggestion:
1. Don't have hardcoded set of buttons. Buttons should be defined on consumer side.
2. `<Modal.Footer>` should be responsible for positioning of buttons.
3. `In some cases we might want to include a "do not show again" checkbox or something similar` For this case I suggest to add `disablePositioning` to `<Modal.Footer>`. When this props passed, `<Modal.Footer>` disable all style rules except `background-color` and `padding`.

In other words, I suggest to move all responsibilities of `<DialogActions>` to `<Modal.Footer>`, but do not hardcode set of buttons to it.

@mikrotron WDYT?

DanilAgafonov:
@nahumzs there are some unfinished discussions, but I am ok with current implementation. The only think that we should decide before merging is https://github.com/acl-services/paprika/pull/320#discussion_r356682473

packages/Modal/src/components/Footer.js
mikrotron: The original `<DialogActions>` was created as a way to share the `<Modal.Footer>` in acl-ui with `<SidePanel>` and other "dialogs".  If we continue with this new implementation of the `<Modal.Footer>`, we will likely find ourselves in the same situation, where we want to extract it to share it.  Neither implementations do much, besides control the layout of buttons (a deliberate design choice that has been debated and committed to by the UX designers – but also subject to change) as well as a bit of styling for the spacing.  

I think I prefer for the new `<Modal.Footer>` to be a fairly basic container (with a bit of styling for background color, for instance) and the `<DialogActions>` to contain the styling / layout for the buttons we will use 95% of the time. 

That said, since the `<DialogActions>` isn't ready yet, we can merge this for now and refactor a little later (including the `<SidePanel>`).

@nahumzs ?

packages/Modal/package.json
alexzherdev: @DanilAgafonov this will be removed from Lerna
2B1F6FEA-1E84-4CED-9A8B-AFF58E240580

packages/Modal/src/Modal.js
```diff
@@ -0,0 +1,79 @@
...
+    <styled.Overlay
+      isOpen={isOpen}
+      onClose={onClose}
+      onAfterOpen={onAfterOpen}
+      onAfterClose={onAfterClose}
+      {...overlayProps}
+    >
+      {state => (
+        <styled.Wrapper size={size}>
+          <styled.Dialog state={state} role="dialog" aria-modal="true" aria-label={ariaLabel} data-pka-anchor="modal">
```
oscarkwan: Can we also support role="alertdialog"?

https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/alertdialog.html

The alertdialog role is a special-case dialog role designed specifically for dialogs that divert users' attention to a brief, important message. Its usage is described in the alert dialog design pattern.

DanilAgafonov: @oscarkwan You mean add a new property, that controls `role`?

packages/Modal/package.json
DanilAgafonov: Oh, thanks. I removed `"directories"` in this commit: https://github.com/acl-services/paprika/pull/320/commits/e728bbe1781d14142130b9267660c5d712a604d2

packages/Modal/src/Modal.js
mikrotron: Do we need this for an upcoming feature, @oscarkwan?
I think it's a good idea, but easily added in a later PR.  I might want to give a bit of thought to the API.

packages/Modal/src/components/Footer.js
nahumzs: Sorry, little late response, but we already agree that all the footer header and body/content should be agnostic of whatever is render on them, pretty much the only thing they do is providing padding/margins/background as necessary if they are use it. 

for now, let's merge it, but let's agree that any area that will display content in our modals shouldn't affect or mess with it's content unless the consumer uses Modal.Footer or Modal.Header, Sidepanel.Header, Takeover.Footer etc. 

If you don't use neither of Header, Footer or Content, the consumer "content" should expand as much as possible under our modals. 

cc @mikrotron

packages/Modal/package.json
nahumzs: Yes, we had a script at the beginning we later remove it, we can bring it back if we think will provide value.

packages/Modal/src/Modal.js
oscarkwan: No, not needed for an upcoming feature, it could be added later. Just thought we should also support other dialog roles.

DanilAgafonov: I created issue: https://github.com/acl-services/paprika/issues/355

mikrotron: Thank you, @DanilAgafonov

PR #326, opened 2019-12-03
UX-670 Render isFixed Toasts as portals

oscarkwan:
### Purpose 🚀
Main purpose is to render `isFixed` `<Toasts>` as React Portals.
Secondary purpose is to improve accessibility.

### Notes ✏️
- Render `isFixed` Toasts, as a `Portal`.
- Refactor stories into, Showcase, Variations, Examples ...etc
- Remove prop `ariaAlert` and replace with `isPolite` (default is `false`).
- Rely on aria `role=status` and `role=alert` as they have implicit `aria-live` values.
- Since most screen readers do not announce content in a `role=status` live region, the container must be mounted to the DOM first, then given a short delay before dynamically updating its content. This was tested with several combinations:
  - Success: 
    - NVDA + Chrome / Firefox
    - Jaws + Chrome / Firefox / IE11
    - Narrator + Edge / IE11
  - Fail: 
    - NVDA + Edge / IE11
    - Jaws + Edge
    - Narrator + Chrome / Firefox
- Toasts that are `kind='visually-hidden'` (for screen readers only) will `autoClose` (so they will be announced, but then removed from the DOM so they can't be navigated to out of context).
- `autoClose` default is now 5 seconds (was 1.5).

### Updates 📦
- [x] MAJOR (breaking) change to `Toast`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-670-render-toast-portal-fxied/?path=/story/toast--showcase

### References 🔗
https://aclgrc.atlassian.net/browse/UX-670


.storybook/config.js
```diff
@@ -19,6 +19,12 @@ const stories = [
   // Welcome
   require.context("./", false, /welcome.story.js/),
 
+  // Toast
```
alexzherdev: Why can't this just say `require.context("../packages", true, /\.stories\.(js|mdx)$/),`?

oscarkwan: This is our way to sort it on the side nav :S cc @mikrotron

mikrotron: Definitely still experimental + unsatisfactory, but I'm very interested in being able to to define the ordering for stories.  For now, alphabetic would be better than having these _two_ out of place, but I have other plans for later.

packages/Toast/src/Toast.styles.js
```diff
@@ -37,7 +37,8 @@ const fixedStyles = css`
   position: fixed;
   top: ${stylers.spacer(2)};
   transform: translateX(-50%);
-  width: calc(100% - #{${stylers.spacer(4)}});
+  width: calc(100% - ${stylers.spacer(4)});
+  z-index: 10;
```
mikrotron: This should use `props.zIndex`, shouldn't it?

mikrotron: Actually, it looks like it'll just be overridden by line 97...

packages/Toast/stories/Toast.examples.stories.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const controlledToast = () => <ControlledToastExample />;
+controlledToast.story = {
+  parameters: {
+    options: {
+      isToolshown: true,
+      showPanel: true,
+    },
+  },
+};
```
mikrotron: Since these options are all the same, there's probably an opportunity to DRY this up a bit.

packages/Toast/src/Toast.js
```diff
@@ -73,77 +71,125 @@ const icons = {
...
+  const memoizedStartAutoCloseTimer = React.useCallback(() => {
     function handleDelayedClose() {
-      clearTimeout(timerRef.current);
+      clearTimeout(autoCloseTimer);
       if (isOpen === undefined) setIsToastOpen(false);
       onClose();
     }
 
-    timerRef.current = setTimeout(handleDelayedClose, Math.max(autoCloseDelay, minimumCloseTimeout));
+    autoCloseTimer = setTimeout(handleDelayedClose, Math.max(autoCloseDelay, minimumCloseTimeout));
```
alexzherdev: why this change? This does not work. It's just assigning to a local variable, which will be discarded after the current render. You do need to assign to `.current`.

mikrotron: 🤦‍♂

PR #340, opened 2019-12-10
add jest and cypress in Checkbox

slee1130:
### Purpose 🚀

### Notes ✏️
- add jest and cypress testing in checkbox component
- feedback needed for cypress 
@oscarkwan 

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/Checkbox/stories/Checkbox.stories.js
+++ b/packages/Checkbox/stories/Checkbox.stories.js
@@ -20,3 +20,5 @@ storiesOf("Checkbox", module)
       <CheckboxAlign />
     </Story>
   ));
+
+storiesOf("Checkbox/Automation Tests", module).add("Cypress", () => <ShowcaseStory />);
--- a/packages/Checkbox/stories/Checkbox.stories.js
+++ b/packages/Checkbox/stories/Checkbox.stories.js
@@ -20,5 +20,3 @@ storiesOf("Checkbox", module)
       <CheckboxAlign />
     </Story>
   ));
-
-storiesOf("Checkbox/Automation Tests", module).add("Cypress", () => <ShowcaseStory />);
--- a/packages/Checkbox/tests/Checkbox.spec.js
+++ b/packages/Checkbox/tests/Checkbox.spec.js
@@ -25,7 +25,7 @@ describe("Checkbox", () => {
     expect(getByRole("checkbox")).not.toHaveAttribute("checked");
   });
 
-  it("Renders checkedState is INDETERMINATE", () => {
+  it("Renders checkedState is indeterminate", () => {
     const { getByRole } = renderComponent({ checkedState: checkboxStates.INDETERMINATE });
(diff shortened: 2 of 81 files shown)


packages/Checkbox/tests/Checkbox.spec.js
```diff
@@ -0,0 +1,37 @@
...
+    expect(getByRole("checkbox")).toBeInTheDocument();
+    expect(getByRole("checkbox")).toHaveAttribute("checked");
+  });
+
+  it("Renders checkedState is unchecked", () => {
+    const { getByRole } = renderComponent({ checkedState: checkboxStates.UNCHECKED });
+    expect(getByRole("checkbox")).not.toHaveAttribute("checked");
+  });
+
+  it("Renders checkedState is INDETERMINATE", () => {
```
oscarkwan: no need to all caps the word

packages/Checkbox/tests/cypress/Checkbox.cypress.js
```diff
@@ -0,0 +1,9 @@
+describe("<CheckBox />", () => {
+  before(() => {});
+
+  it("should be checked when you click the checkbox", () => {
+    cy.visitStorybook("checkbox-automation-tests--cypress");
+    cy.get("label").click();
+    cy.get("label").click();
```
oscarkwan: This text is actually not expecting any results. If you think we don't need a cypress test, and jest is sufficient then we don't need to add it here.  It seems like this is testing the same thing as jest

packages/Checkbox/tests/Checkbox.spec.js
```diff
@@ -0,0 +1,37 @@
...
+    a11yText: null,
+    checkedState: checkboxStates.CHECKED,
+    children: null,
+    isDisabled: false,
+    onChange: () => {},
+  };
+  return render(<Checkbox {...defaultProps} {...props} />);
+}
+
+describe("Checkbox", () => {
```
oscarkwan: Can we add a couple more tests?

One for checking a box, and other checkboxes are also selected.
Another for disabled

packages/Checkbox/tests/cypress/Checkbox.cypress.js
slee1130: ok

packages/Checkbox/tests/Checkbox.spec.js
slee1130: done

packages/Checkbox/stories/Checkbox.stories.js
```diff
@@ -20,3 +20,5 @@ storiesOf("Checkbox", module)
       <CheckboxAlign />
     </Story>
   ));
+
+storiesOf("Checkbox/Automation Tests", module).add("Cypress", () => <ShowcaseStory />);
```
oscarkwan: You also don't need this anymore

Issue #342, opened 2019-12-10
Async action in confirmation Component not showing pending state correctly


tristanjasper:
# Bug Report

Added story to demonstrate isPending state
Ensure onClose callback is closed after the confirmation has animated closed.
Ensured that the isPending button state doesn't cause wrapping issues in the Confirmation content area.

## Prerequisites
**This is just a checklist, please delete this section after**
Please answer the following questions for yourself before submitting a bug report.
- [ ] I checked to make sure that this bug has not already been filed
- [ ] I reproduced the bug in the latest version of paprika
- [ ] Please label as, low / medium / high priority

*Please fill in as much detail as possible below*

## Expected behavior
*Please describe the behavior you are expecting to happen.*

## Current behavior
*What is the current behavior?*

## Screenshots / Gifs / Codepens
*Include media to help illustrate the bug. Gifs help a lot!*

## Additional context
*Please provide any more additional details to illustrate the bug*

PR #343, opened 2019-12-11
UX-ISSUE-342 Confirmation component isPending and animated close support

tristanjasper:
### Purpose 🚀

Update confirmation component to support pending state correctly and add storybook example

https://github.com/acl-services/paprika/issues/342

This PR :
- Added story to demonstrate isPending state
- Ensures onClose callback is closed after the confirmation has animated closed.
- Ensured that the isPending button state doesn't cause wrapping issues in the Confirmation content area by removing the cancel button.

-Update dropdowncomponent examples to use confirmation component `onClose` callback 

### Notes ✏️

### Updates 📦

### Storybook 📕

### Screenshots 📸
Before:
Dropdown Component
dropdownClosingBefore

After
Confirmation Component
confirmationPending
Dropdown Component
dropdownClosing

### References 🔗


packages/Confirmation/src/Confirmation.js
```diff
@@ -77,7 +77,7 @@ const Confirmation = props => {
     if (isConfirmOpen) {
       setIsConfirmOpen(false);
       if (triggerRef.current) triggerRef.current.focus();
-      setTimeout(onClose, 0);
+      setTimeout(onClose, 250);
```
tristanjasper: Added a delay to accommodate the animation of the confirmation closing. In the case of the dropdown Component (which uses confirmation component) we can then use this callback to close the menu.

packages/Confirmation/src/Confirmation.js
```diff
@@ -98,32 +98,28 @@ const Confirmation = props => {
...
-                isSemantic={false}
-                ref={confirmButtonRef}
-                kind={confirmButtonType}
-                size={buttonSize}
-                onClick={handleOnConfirm}
-              >
-                {confirmLabel}
-              </Button>
+  const popoverContent = (
+    <Popover.Content id={confirmId.current}>
```
tristanjasper: Moved Content into a variable for easier readability

packages/Confirmation/src/Confirmation.js
```diff
@@ -98,32 +98,28 @@ const Confirmation = props => {
...
+              isPending={isPending}
+              isSemantic={false}
+              ref={confirmButtonRef}
+              kind={confirmButtonType}
+              size={buttonSize}
+              onClick={handleOnConfirm}
+            >
+              {confirmLabel}
+            </Button>
+            {!isPending ? (
```
tristanjasper: I am hiding the cancel button if the state of the confirm button is now Pending. If we do not do this then we get wrapping issues 
image.
Perhaps another approach would be to to re-render then entire popover content tree (which would ensure the popover renders content width correctly) or to call an imperative api callback on the popover which re-calculates the content width, but i had issues with both approaches. Open to suggestions for either if we are adamant we don't want the cancel button to disappear, Personally I think this approach is fine.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -133,22 +133,24 @@ function DropdownMenu(props) {
     }
 
     return (
-      <div css={contentStyles} ref={dropdownListRef}>
-        {extractedChildren.children.map((child, index) => {
-          const childKey = { key: `DropdownMenuItem${index}` };
-          if (child && child.type && child.type.displayName === "DropdownMenu.Item") {
-            if (child.props.renderConfirmation) {
+      <Popover.Card>
```
tristanjasper: Only wrapping the content with the popover card when it is the dropdown list of items, and not when it is just showing the confirmation panel. This avoids the rendering of the card (with no contents) when we are closing the confirmation panel.

packages/DropdownMenu/stories/examples/DropdownMenuExample.js
```diff
@@ -2,8 +2,11 @@ import React from "react";
 import Confirmation from "@paprika/confirmation";
 import DropdownMenu from "../../src";
 
-const handleConfirm = onCloseMenu => onCloseConfirm => {
-  onCloseConfirm();
+const handleConfirm = onConfirm => {
```
tristanjasper: Updated the examples to show correct implementation of the confirmation component, where we close the menu after the confirmation panel has been animated closed.

packages/Confirmation/src/Confirmation.js
jamiek-acl: I dont think removing the cancel button is the ideal way to go (1: for a11y, 2: if some day we support actually cancelling the async request)

packages/Confirmation/stories/examples/ConfirmationExampleWithAsyncAction.js
```diff
@@ -0,0 +1,34 @@
...
+const ConfirmationExampleWithAsyncAction = () => {
+  const [isPending, setIsPending] = React.useState();
+  const handleConfirm = handleCloseConfirm => {
+    setIsPending(true);
+    setTimeout(() => {
+      handleCloseConfirm();
+    }, 2000);
+  };
+  return (
+    <div>
```
jamiek-acl: Dont need these divs

packages/DropdownMenu/stories/examples/DropdownMenuExample.js
```diff
@@ -34,7 +37,8 @@ const DropdownMenuExample = () => {
               body="Lorem ipsum dolor amet vexillologist tacos selvage narwhal butcher twee ethical hot chicken."
               confirmLabel="Delete filter"
               heading="Delete filter?"
-              onConfirm={handleConfirm(onCloseMenu)}
+              onConfirm={handleConfirm}
+              onClose={handleCloseConfirm(onCloseMenu)}
```
jamiek-acl: I think adding an `isPending` example to the DropdownMenu would be helpful

packages/Confirmation/src/Confirmation.js
tristanjasper: Cancelling the async could still be done, just change the `isPending` prop back to true and the cancel button would show. I'll look into re-rendering of the popup content as I agree it's not ideal..

packages/DropdownMenu/stories/examples/DropdownMenuExample.js
tristanjasper: I'll add to the showcase

packages/Confirmation/src/Confirmation.js
allison-c: Maybe adding the explanation in the code in case we forget it later on?

tristanjasper: Yeah ok i shall, although i'm not keen on leaving comments in the code in general

packages/Confirmation/src/Confirmation.js
```diff
@@ -51,7 +51,8 @@ const Confirmation = props => {
   } = props;
   const confirmButtonRef = React.useRef(null);
   const triggerRef = React.useRef(null);
-  const confirmId = React.useRef(uuid());
+  const confirmId = React.useRef(uuid()).current;
+  let popoverKey = uuid();
```
allison-c: I think this one can be ref as well. In the useEffect, we can do `popoverKeyRef.current = uuid()`

packages/Confirmation/src/Confirmation.js
```diff
@@ -68,6 +69,10 @@ const Confirmation = props => {
     }
   }, [confirmButtonRef, defaultIsOpen]);
 
+  React.useEffect(() => {
+    popoverKey = uuid();
+  }, [isPending]);
```
allison-c: Why do we need to force the popover rerendering?

tristanjasper: Yeah it was needed because when the `isPending` prop is true on the button compoennt it shows spinning arrows and causes a width change. If the popover isn't rerendered (and width evaluated correctly) it causes layout flow issues - https://github.com/acl-services/paprika/pull/343#discussion_r356345840

PR #344, opened 2019-12-12
Ux 338 list box browser accessibility issues

tristanjasper:
### Purpose 🚀
Update listbox browser and listbox focus outline so that it does not distract too much as per Item 1 on https://aclgrc.atlassian.net/wiki/spaces/UX/pages/662410769/Listbox+Browser+Component+Review+11+22+2019

Additionally fixed a z-index issue where the focus outline was being but partly hidden.

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-338-list-box-browser-issues

### Screenshots 📸
Before:
accissues-before

After:
accissues

### References 🔗


packages/ListBox/src/components/Content/Content.js
```diff
@@ -46,20 +47,18 @@ export default function Content(props) {
   const { refListBoxContainer } = state;
 
   /* NOTE no idea what ROLE should be this div when the ListBox is INLINE */
-  /* eslint-disable jsx-a11y/no-static-element-interactions */
   if (state.isInline) {
     return (
-      <div
+      <ContentStyled
```
nahumzs: this is only wrapping the inline style of the listbox, do this style only applies when its inline? @tristanjasper

jayde-acl:
Looks great but can you bump that dotted outline up a bit to make it a bit darker? `#8A8A8A` Sarah says if we are going to have it then it should pass 3:1 contrast ratio. Thanks!

packages/ListBox/src/components/Content/Content.js
tristanjasper: Ok Yeah it makes sense to be applied inline and not inline so will make sure it is

tristanjasper:
> Looks great but can you bump that dotted outline up a bit to make it a bit darker? `#8A8A8A` Sarah says if we are going to have it then it should pass 3:1 contrast ratio. Thanks!

👌

packages/ListBox/src/components/Content/Content.js
tristanjasper: Discussed with @jayde-acl  and @nahumzs for now this dotted style is only being applied to the listBox when it is inline

tristanjasper:
> > Looks great but can you bump that dotted outline up a bit to make it a bit darker? `#8A8A8A` Sarah says if we are going to have it then it should pass 3:1 contrast ratio. Thanks!
> 
> 👌

Ok I have updated to that value

tristanjasper:
Closed PR and created new PR with correct branch number - https://github.com/acl-services/paprika/pull/375

PR #345, opened 2019-12-12
UX-650 saving navigation state

allison-c:
### Purpose 🚀
- Being able to save/read navigation status with localStorage

Receiving initial navigation status will be another PR

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-650-Saving-navigation-state

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/DataTable/src/components/Navigation/components/Filters/FilterItem.js
+++ b/packages/DataTable/src/components/Navigation/components/Filters/FilterItem.js
@@ -57,6 +57,14 @@ export default function FilterItem(prop) {
     });
   }
 
+  function handleChangeDatePicker(momentDate) {
+    if (momentDate === null) {
+      handleChangeValue("");
+    } else {
+      handleChangeValue(momentDate.format(columns[selectedColumnId].momentParsingFormat));
+    }
+  }
+
   function renderTextInput() {
     if (
       selectedRule === rules.IS_BLANK ||
@@ -89,7 +97,8 @@ export default function FilterItem(prop) {
       </select>
       {selectedColumnType === columnTypes.DATE ? (
         <DatePicker
-          onChange={momentDate => handleChangeValue(momentDate.format(columns[selectedColumnId].momentParsingFormat))}
+          initialDate={value}
+          onChange={handleChangeDatePicker}
           parsingFormat={columns[selectedColumnId].momentParsingFormat}
         />
       ) : (
--- a/packages/DataTable/src/context.js
+++ b/packages/DataTable/src/context.js
@@ -122,6 +122,7 @@ function TableProvider(props) {
   }, [columns]);
 
   function updateLocalStorage(changes) {
+    if (!localStorageId) return;
     const originalStatus = JSON.parse(localStorage.getItem(storageKey));
     localStorage.setItem(storageKey, JSON.stringify({ ...originalStatus, ...changes }));
   }
--- a/packages/DataTable/src/context.js
+++ b/packages/DataTable/src/context.js
@@ -10,8 +10,9 @@ const TableDispatchContext = React.createContext();
 const TableLocalStorageContext = React.createContext();
 
 function TableProvider(props) {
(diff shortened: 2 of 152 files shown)


packages/DataTable/src/context.js
```diff
@@ -4,28 +4,60 @@ import isMatchWith from "lodash.ismatchwith";
...
+      keygen,
+      sortColumn: null,
+      sortDirection: null,
+      columnsOrder: columns.map(column => column.id),
+      columns: columns.reduce((columnsObject, column) => ({ ...columnsObject, [column.id]: column }), {}),
+      filters: [],
+      logicalFilterOperator: logicalFilterOperators.AND,
+    };
+    let currentTableState = {};
+    if (localStorageId) {
```
allison-c: We'll save the status when there's localStorageId

packages/DataTable/src/components/Navigation/components/Filters/FilterItem.js
```diff
@@ -57,6 +57,14 @@ export default function FilterItem(prop) {
     });
   }
 
+  function handleChangeDatePicker(momentDate) {
+    if (momentDate === null) {
+      handleChangeValue("");
+    } else {
+      handleChangeValue(momentDate.format(columns[selectedColumnId].momentParsingFormat));
```
nahumzs: 🤔 what if they pass an incorrect momentParsingFormat? shouldn't we check if the date produced by moment its valid?

packages/DataTable/src/hooks/useIsUpdated.js
```diff
@@ -0,0 +1,7 @@
+import usePrevious from "@paprika/helpers/lib/hooks/usePrevious";
+
+export default function useIsUpdated(prop) {
```
nahumzs: `nip`

This is no exactly a custom hook, there is no state to manage here,  but this could totally be a helper function

export default function(prop, prevProp) {
  return prevProp !== undefined && prop !== prevProp;
}

but of course you will have to import usePrevious in each file.

packages/DataTable/stories/MemorizedNavigation.stories.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import DataTable, { Filters, Sort, ColumnManaging } from "../src";
+import fixtures from "./fixtures";
+import { viewPortHeight } from "./helpers";
+
+const mockData = fixtures(1);
+function App() {
+  return (
+    <DataTable localStorageId="my-table-id" keygen="id" data={mockData} height={viewPortHeight()}>
```
nahumzs: I don't think we should force the user to create localStorageId, this is the same con ide on the ColumnDefinition, honestly I think those should be handle by our selves, is quite simple to create a `nanoid` in DataTable, and let our consumer don't worry about this.
```js
  import nanoid from "nanoid"
  const id = React.useState(nanoid)
```

nahumzs: mmm wait, but then wouldn't be possible to know which table to retrive froom the localstorage because can be multiple tables, ok. maybe let's change th ename to `<DataTable id="" />`

And let's eventually remove the requirement from adding a `id` on Column definition.

packages/DataTable/src/components/Navigation/components/Filters/FilterItem.js
allison-c: I think date picker already handled the case when users type invalid date, I'll double-check

packages/DataTable/src/hooks/useIsUpdated.js
allison-c: But there's state management in `usePrevious`
Isn't it just like this? https://github.com/streamich/react-use/blob/master/src/usePreviousDistinct.ts

PR #350, opened 2019-12-17
Xd 7194 simplify confirmation component api

tristanjasper:
### Purpose 🚀
Confirmation panel could do with a simpler api for consuming applications and to make it consistant with the DropdownMenu component api recent changes
https://aclgrc.atlassian.net/browse/XD-7194

### Notes ✏️

Secondary changes
- Adjusted visual gap to trigger button to be consistant with DropdownMenu component
- Added some more unit tests for component
- Ensure onClose callback isn't being called more than once.
- Updated Confirmation components `triggerButton` component to have same api as dropdownMenu components `trigger` component api (changes to icon prop and support for raw html as child)
- Minor cleanup of `isSquare` prop in Button component

### Updates 📦
- [x] MAJOR (breaking) change to confirmation package
- [x] PATCH (bug fix) change to Button package

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/xd-7194-simplify-confirmation-component-api

### Screenshots 📸

### References 🔗


packages/Confirmation/src/Confirmation.js
```diff
@@ -73,15 +73,11 @@ const Confirmation = props => {
...
-    isConfirmOpen,
-    handleOpenConfirm,
-  };
-
   const handleCloseConfirm = () => {
     if (isConfirmOpen) {
       setIsConfirmOpen(false);
-      if (triggerRef.current) triggerRef.current.focus();
+      // being called repeatedly by popoverclose and cancel button?
+      console.log("calling close confirm");
```
tristanjasper: @nahumzs Any idea why this would be getting called repeatedly when clicking the cancel button in confirmation popup?
image

packages/Confirmation/src/Confirmation.js
```diff
@@ -12,7 +12,7 @@ import { confirmStyles, confirmBodyStyles, confirmFooterStyles } from "./Confirm
 const propTypes = {
   buttonSize: PropTypes.oneOf(ShirtSizes.DEFAULT),
   /** Children should be a render prop in the form of a function to display trigger */
-  children: PropTypes.func,
+  children: PropTypes.node,
```
tristanjasper: No longer passing a function for the trigger component

packages/Confirmation/src/Confirmation.js
```diff
@@ -54,6 +54,8 @@ const Confirmation = props => {
   const confirmId = React.useRef(uuid()).current;
   let popoverKey = uuid();
 
+  const popoverOffset = 4;
```
tristanjasper: Adding an offset that is the same as the `dropdownMenu` component

packages/Confirmation/src/Confirmation.js
```diff
@@ -73,29 +75,26 @@ const Confirmation = props => {
...
-      setIsConfirmOpen(false);
-      if (triggerRef.current) triggerRef.current.focus();
-      setTimeout(onClose, 250);
-    }
+    setIsConfirmOpen(prevIsConfirmOpen => {
+      if (prevIsConfirmOpen) {
+        setTimeout(onClose, 250);
+      }
+      return false;
+    });
```
tristanjasper: Due to confirm and cancel button clicks calling this function as well as popover `onClose` event this is being called twice, and both time `isConfirmOpen` value in state is true, even though it is being set as `false` in the `handleCloseConfirm` function. By checking the previous state of `isConfirmOpen` instead (which is available from the `setIsConfirmOpen` function we are able to get the correct value and contents of the `if` are evaluated only once.

packages/Confirmation/test/confirmation.spec.js
```diff
@@ -18,28 +18,31 @@ function renderComponent() {
...
+  //   const onCloseFunc = jest.fn();
+  //   const { getByText } = renderComponent({
+  //     onClose: onCloseFunc,
+  //   });
+  //   fireEvent.click(getByText(/trigger/i));
+  //   fireEvent.click(getByText(/Confirm Delete/i));
+  //   await wait(() => {
+  //     expect(onCloseFunc).toHaveBeenCalled();
+  //   });
+  // });
```
tristanjasper: Unsure why the `onCloseFunc` callback isn't being detected as being called here. Intend to revisit if cannot find solution. Have tested in the browser and is working for me there.

packages/Confirmation/src/Confirmation.js
```diff
@@ -73,29 +75,27 @@ const Confirmation = props => {
...
   const handleCloseConfirm = () => {
-    if (isConfirmOpen) {
-      setIsConfirmOpen(false);
-      if (triggerRef.current) triggerRef.current.focus();
-      setTimeout(onClose, 250);
-    }
+    setIsConfirmOpen(prevIsConfirmOpen => {
+      if (prevIsConfirmOpen) {
+        if (triggerRef.current) triggerRef.current.focus();
+        setTimeout(onClose, 250);
```
jamiek-acl: You dont need the `setTimeout` anymore i thought?

tristanjasper: the `setTimeout` was already existing and is a delay to allow for the animation of popover closing. It's not related to the check of `prevIsConfirmOpen`

packages/Confirmation/test/confirmation.spec.js
jamiek-acl: Can you get this test working please (otherwise it will never get done).

packages/Confirmation/src/Confirmation.js
```diff
@@ -12,7 +12,7 @@ import { confirmStyles, confirmBodyStyles, confirmFooterStyles } from "./Confirm
 const propTypes = {
   buttonSize: PropTypes.oneOf(ShirtSizes.DEFAULT),
   /** Children should be a render prop in the form of a function to display trigger */
```
jamiek-acl: Obsolete comment

packages/Confirmation/src/Confirmation.js
nahumzs: ohhh you right, there was a` todo` to add onAnimationEnd and onAnimationStart to the popover, can we create an issue for that? and refer to this conversation?

tristanjasper: Sure will add to popover issues

nahumzs: nvm I created it https://github.com/acl-services/paprika/issues/352

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
```diff
@@ -6,15 +6,17 @@ const propTypes = {
   children: PropTypes.node,
   confirmId: PropTypes.string,
   icon: PropTypes.node,
```
jamiek-acl: Why is `icon` a separate prop? Cant it just be a child?

packages/Confirmation/test/confirmation.spec.js
nahumzs: maybe wrapped on window.requestAnimationFrame(()=> { 
  // run your test
})

tristanjasper: tried `async` `await`, `window.requestAnimationFrame` and 'setTimeout` but i'm finding none result in the `onClose` being being called... 🤔

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
tristanjasper: @mikrotron I recall we discussed this a while ago and settled on it together, do you recall a specific reason?

mikrotron: I think for the `<DropdownMenu>` we ended up moving away from having an `icon` prop... it's not a prop anymore for `<DropdownMenu.Trigger>`:
https://github.com/acl-services/paprika/blob/master/packages/DropdownMenu/src/components/Trigger/Trigger.js

But we DID need a `buttonType` prop, which accepted `"icon"` as a possible value.

jamiek-acl: So should we follow the same pattern here? I'm indifferent.

packages/Confirmation/src/Confirmation.js
```diff
@@ -73,29 +74,27 @@ const Confirmation = props => {
...
-      setIsConfirmOpen(false);
-      if (triggerRef.current) triggerRef.current.focus();
-      setTimeout(onClose, 250);
-    }
+    setIsConfirmOpen(prevIsConfirmOpen => {
+      if (prevIsConfirmOpen) {
+        if (triggerRef.current) triggerRef.current.focus();
+        setTimeout(onClose, 250);
+      }
+      return false;
```
mikrotron: why does it need to return `false`? If you leave this out it will implicitly return `undefined`, won't it? Which should be fine?

packages/Confirmation/src/Confirmation.js
```diff
@@ -73,29 +74,27 @@ const Confirmation = props => {
...
-  const renderTrigger = triggerComponent => {
-    // const triggerComponent = renderTriggerFunction(triggerProps);
+  const renderTrigger = () => {
     // wrapping the returned item in a function to avoid needing to tab twice
     // https://github.com/acl-services/paprika/issues/126
     return () =>
-      React.cloneElement(triggerComponent, {
+      React.cloneElement(children, {
+        isConfirmOpen,
+        onOpenConfirm: handleOpenConfirm,
```
mikrotron: 👍

packages/Button/src/Button.styles.js
```diff
@@ -230,7 +230,7 @@ const buttonStyles = props => `
   ${commonStyles}
   ${sizeStyles[props.size]}
   ${kindStyles(props)[props.kind]}
-  ${props.isFullWidth && !props.isSquare ? fullWidthStyles : ""}
+  ${props.isFullWidth ? fullWidthStyles : ""}
```
tristanjasper: @mikrotron I don't see that the `isSquare` prop is necessary anymore, or there another reason it is still existing here. The `isFullWidth` prop should be enough to apply the `fullWidthStyles`?

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
```diff
@@ -6,15 +6,17 @@ const propTypes = {
...
-  isConfirmOpen: PropTypes.bool.isRequired,
-  onOpenConfirm: PropTypes.func.isRequired,
+  isConfirmOpen: PropTypes.bool,
+  onOpenConfirm: PropTypes.func,
```
mikrotron: Why can't these remain `isRequired`?

packages/Confirmation/src/Confirmation.js
tristanjasper: isConfirmOpen should be boolean value, just by it's name, why not set it as a boolean value? Seems more logical to me

packages/Confirmation/test/confirmation.spec.js
```diff
@@ -18,28 +17,35 @@ function renderComponent() {
...
-  });
-
   it("should show confirmation panel by default when mounted ", () => {
-    expect(confirmationComponent).toBeVisible();
-  });
-
-  it("should trigger the confirm callback when confirm button is clicked", () => {
-    // expect(confirmationComponent).toBeVisible();
+    const { getByText } = renderComponent({ defaultIsOpen: true });
+    expect(getByText(/Confirm Delete/)).toBeVisible();
```
mikrotron:
```suggestion
    expect(getByText(/confirm delete/i)).toBeVisible();
```

packages/Confirmation/test/confirmation.spec.js
```diff
@@ -18,28 +17,35 @@ function renderComponent() {
...
-    // expect(confirmationComponent).toBeVisible();
+  it("should trigger the confirm callback when confirm button is clicked and on Close to not have been called", async () => {
+    const onCloseFunc = jest.fn();
+    const onConfirmFunc = jest.fn();
+    const { getByText } = renderComponent({
+      onConfirm: onConfirmFunc,
+      onClose: onCloseFunc,
+    });
+    fireEvent.click(getByText(/trigger/i));
+    fireEvent.click(getByText(/Confirm Delete/i));
```
mikrotron:
```suggestion
    fireEvent.click(getByText(/confirm delete/i));
```

packages/Confirmation/test/confirmation.spec.js
```diff
@@ -18,28 +17,35 @@ function renderComponent() {
...
 
-  it("should trigger the onConfirm callback when clicking the confirm button", () => {
-    // expect(confirmationComponent).toBeVisible();
+  it("should close the confirmation when clicking cancel", async () => {
+    const onCloseFunc = jest.fn();
+    const { getByText } = renderComponent({
+      onClose: onCloseFunc,
+    });
+    fireEvent.click(getByText(/trigger/i));
+    fireEvent.click(getByText(/Cancel/i));
```
mikrotron:
```suggestion
    fireEvent.click(getByText(/cancel/i));
```

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
```diff
@@ -1,40 +1,54 @@
 import React from "react";
 import PropTypes from "prop-types";
 import Button from "@paprika/button";
+import RawButton from "@paprika/raw-button";
```
tristanjasper: I will be following this PR up with a new proposal where we consolidate this `triggerButton` component with the trigger in the Dropdown component. At the moment I am just ensuring that they both use the same api for button type instead of using an `icon` prop here. Confirmation components `triggerButton` will now also support raw content as it's child.

packages/Confirmation/src/Confirmation.js
tristanjasper: Again, something we can aim to consolidate in a future pr with the dropdown component

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
tristanjasper: Yeah i've now updated the confirmation components `TriggerButton` component to use the same api as the dropdownMenus `trigger` component. In a future PR I shall consolidate so that both DropdownMenu and the Confirmation component use the same code for `Trigger` Component.

packages/Confirmation/src/Confirmation.js
mikrotron: Oh I see now, I didn't read it well before with the inline comments on.  

But I also now notice that there are side effects in your `setState` which I'm not so sure is a good practice.  (I'm not _sure_ it's a bad practice, it just makes my spidey sense tingle).

The `prevIsConfirmOpen` in the callback for `setState` is the same as `isConfirmOpen` before you call it though, right? So you _could_ just do that block first, then call `setState`?

packages/Button/src/Button.styles.js
mikrotron: It seems like we removed the `isSquare` prop, maybe when it was migrated from `acl-ui`:
https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.js

So yeah, this is a relic, thanks for cleaning it up.

mikrotron: This is a `Patch` update to `<Button>` though, which should be indicated in the description, please.

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
tristanjasper: 🤔   Previously I was getting missing prop type complaints when these were `required` when i was cloning these props onto the child - https://github.com/acl-services/paprika/blob/xd-7194-simplify-confirmation-component-api/packages/Confirmation/src/Confirmation.js#L96 and was a large part the reason i originally exposed these props to the consumer. However I don't see it complaining now so I have readded.

tristanjasper: Actually re-ran `yarn` and i still see the proptype errors when running unit tests and in storyboards, so setting back to not required.
image

tristanjasper: Perhaps another reason to use context over React cloning when possible.

packages/Confirmation/src/Confirmation.js
tristanjasper: You mean like we were doing before?
```
if (isConfirmOpen) {	    
      setIsConfirmOpen(false);
}
```
The reason I changed it to do it this was was because the `isConfirmOpen` value in state was being evaluated as `true` twice as this function is called on button click and popover close event. even though it has been set to false. This approach ensures there are no async problems.

packages/Button/src/Button.styles.js
tristanjasper: done

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
```diff
@@ -1,40 +1,54 @@
...
+  simple: Button,
+};
 
 const propTypes = {
   children: PropTypes.node,
   confirmId: PropTypes.string,
-  icon: PropTypes.node,
-  isConfirmOpen: PropTypes.bool.isRequired,
-  onOpenConfirm: PropTypes.func.isRequired,
+  buttonType: PropTypes.oneOf(ButtonTypes.ALL),
```
mikrotron: You are missing this line:
`ButtonTypes.ALL = Object.values(ButtonTypes);`

Which is causing a console warning.

PR #353, opened 2019-12-17
Os add sidepanel tests

slee1130:
### Purpose 🚀
_There were some testing missing in Sidepanel.spec.js so added some more scenarios to test.

### Notes ✏️
-Added sidepnal.group and sidepanel.trigger jest testing added
-Sidepanel.trigger testing could not use getByTestId since the test keeps failing if there is another way to test it please tell me I can fix it.
@oscarkwan 

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/os-add_sidepanel_tests

### Screenshots 📸

### References 🔗


packages/SidePanel/tests/SidePanel.spec.js
```diff
@@ -88,15 +87,96 @@ describe("SidePanel", () => {
...
+            <SidePanel.Header>With Header</SidePanel.Header>
+            <SidePanel.Overlay />
+          </SidePanel>
+          <SidePanel onClose={noop} isOpen>
+            <SidePanel.Header kind="primary">With Header</SidePanel.Header>
+            <SidePanel.Overlay />
+          </SidePanel>
+        </SidePanel.Group>
+      );
+      // data-pka-anchor
```
oscarkwan: Please remove comment

packages/SidePanel/tests/SidePanel.spec.js
```diff
@@ -88,15 +87,96 @@ describe("SidePanel", () => {
...
+      expect(getAllByTestId("heading")).toHaveLength(3);
```
oscarkwan: @slee1130 Use the `sidepanel` to determine number of sidepanels and not the heading.

packages/SidePanel/tests/SidePanel.spec.js
```diff
@@ -88,15 +87,96 @@ describe("SidePanel", () => {
...
+    });
+
+    it("throws the error when sidepanel has only one panel", () => {
```
oscarkwan: "throws an error when there is only one side panel in a group"

packages/SidePanel/tests/SidePanel.spec.js
```diff
@@ -88,15 +87,96 @@ describe("SidePanel", () => {
...
+            <SidePanel.Header>With Header</SidePanel.Header>
+            <SidePanel.Overlay />
+          </SidePanel>
+          <SidePanel onClose={noop} isOpen>
+            <SidePanel.Header kind="primary">With Header</SidePanel.Header>
+            <SidePanel.Overlay />
+          </SidePanel>
+        </SidePanel.Group>
+      );
+      fireEvent.click(getAllByTestId(/button/i)[0]);
```
oscarkwan: we should be using `sidepanel-header-close`

packages/SidePanel/tests/SidePanel.spec.js
slee1130: done

packages/SidePanel/tests/SidePanel.spec.js
slee1130: done

packages/SidePanel/tests/SidePanel.spec.js
```diff
@@ -88,15 +87,96 @@ describe("SidePanel", () => {
...
+    });
+
+    it("should tigger or render a button", () => {
+      const onClick = jest.fn();
+      const { getByRole } = render({
+        onClick,
+        children: <SidePanel.Trigger>Button</SidePanel.Trigger>,
+      });
+
+      expect(getByRole("button")).toBeInTheDocument();
```
oscarkwan: You have the same expect in line 169. Which means this test is testing the same thing as above.

Therefore, it makes this test kind of redundant. I think just keeping "renders with default props" is fine.

packages/SidePanel/tests/SidePanel.spec.js
slee1130: done

packages/SidePanel/tests/SidePanel.spec.js
slee1130: done

packages/SidePanel/tests/SidePanel.spec.js
slee1130: done

PR #354, opened 2019-12-18
Star 41 row select

nahumzs:
### Purpose 🚀
Adding support for Row Selection

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-41-4ow-select

### References 🔗


packages/DataTable/src/components/RowIndicator/RowIndicator.js
```diff
@@ -0,0 +1,63 @@
+import React from "react";
+import PropTypes from "prop-types";
+import CheckBox from "@paprika/checkbox";
+import * as styles from "./RowIndicator.styles";
+
+const propTypes = {
+  hasIndexIndicator: PropTypes.bool,
+  isSelected: PropTypes.func,
```
allison-c: Should this be either boolean or function?

nahumzs: nop, in this case unlinke most cases in paprika we want to use a function for check if the item is selected or not, this way we let the consumer to control the selection however they feel, and also we don't force them to reshape their object to add a selected item, they can use any property on their data to achieve this. As well they can implemented any kind of mechanism to decide what is and how is selected.

ex.

```js
const [consumerSelectedArray, setConsumerSelectedArray] = React.useState([]);

function handleSelect(row, rowIndex) {
  return consumerSelectedArray.includes(row.consumerDataId)}
}

<DataTable>
  <DataTable.ColumnDefinition cell={(...args) => <<DataTable.RowIndicator isSelected={handleSelected} {...args} />}
</DataTable>
```

PR #360, opened 2019-12-20
Ux 681 datepicker has error in formelement fix

tristanjasper:
### Purpose 🚀
Date Picker doesn't currently includes the `hasError` prop on the DatePicker input component. When the Date Picker component is used as as child of the Form Element component the `DatePicker.Input` is not an immediate child and so the `hasError` prop is not added to the input therefore not highlighting the input as red when `FormElement` has an error.

### Notes ✏️

Side changes
- Expanded DatePicker readme to include props
- Retained `hasError` prop on the `DatePicker.Input` to avoid breaking change

### Updates 📦

- [x] MINOR (backward compatible) change to DatePicker

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-681-datepicker-has-error-in-formelement-fix

### Screenshots 📸
Before
image

After
fixFormInputError

### References 🔗

### DIFF
--- a/packages/DatePicker/src/DatePicker.js
+++ b/packages/DatePicker/src/DatePicker.js
@@ -44,14 +44,18 @@ const propTypes = {
   /** Callback when date is selected or input. */
   onChange: PropTypes.func.isRequired,
 
-  /** Errors callback */
+  /** Internal errors callback */
   onError: PropTypes.func,
+
+  /** If there is an external error or not. Not required when wrapped with <FormElement>. */
+  hasError: PropTypes.bool,
 };
 
 const defaultProps = {
   children: null,
   dataFormat: "MM/DD/YYYY",
   date: null,
+  hasError: false,
   humanFormat: undefined,
   id: null,
   isDisabled: false,
@@ -73,6 +77,7 @@ function DatePicker(props) {
     isReadOnly,
     onChange,
     onError,
+    hasError,
   } = props;
 
   const formatDateProp = React.useCallback(
@@ -200,8 +205,6 @@ function DatePicker(props) {
     hideCalendar();
     handleChange(selectedDate);
   }
-
-  const hasError = extendedInputProps && extendedInputProps.hasError;
 
   const hasErrorValue = hasError || hasParsingError;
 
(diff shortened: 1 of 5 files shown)


packages/DatePicker/src/DatePicker.js
```diff
@@ -44,14 +44,18 @@ const propTypes = {
   /** Callback when date is selected or input. */
   onChange: PropTypes.func.isRequired,
 
-  /** Errors callback */
+  /** Internal errors callback */
```
tristanjasper: Trying to be more explicit here using "Internal" to mean errors coming from inside the component

packages/DatePicker/src/DatePicker.js
```diff
@@ -44,14 +44,18 @@ const propTypes = {
...
   onError: PropTypes.func,
+
+  /** If there is an external error or not. Not required when wrapped with <FormElement>. */
+  hasError: PropTypes.bool,
```
tristanjasper: whilst this is an error coming from the consuming application

packages/DatePicker/src/DatePicker.js
```diff
@@ -201,8 +206,6 @@ function DatePicker(props) {
     handleChange(selectedDate);
   }
 
-  const hasError = extendedInputProps && extendedInputProps.hasError;
```
allison-c: Shouldn't we also handle the `hasError` prop on `<DatePicker.Input />` ?

allison-c: Cause `hasError` is usually on `<Input />`

tristanjasper: Good point, Looks like we will need it on both now.

tristanjasper: Actually no the `hasError` prop is now automatically set on the Input at https://github.com/acl-services/paprika/blob/ux-681-datepicker-has-error-in-formelement-fix/packages/DatePicker/src/DatePicker.js#L237 so we don't need to include it as a prop on the `DatePicker.Input` this is a breaking api change though for DatePicker...

tristanjasper: discussed with @mikrotron Nicer to have a cleaner api, but for the time being we will retain the `hasError` prop on the `DatePicker.Input` component to avoid a breaking change.

Issue #369, opened 2019-12-31
<Modal> and <form> do not mix


alexzherdev:
# Bug Report
## Expected behavior
I need to implement a very common use case: a form in a modal (see example below). I'm expecting to be able to put my form fields inside `<Modal.Content>` and have a "submit" type button in `<Modal.Footer>`.

## Current behavior
This does not seem to be possible.
Since the submit button has to be a descendant of the form element, I have two options:

a)
```jsx
<form>
  <Modal>
    <Modal.Content><input type="text" /></Modal.Content>
    <Modal.Footer><Button type="submit"></Button></Modal.Footer>
  </Modal>
</form>
```
^ this doesn't work because the modal is rendered in a portal. In the resulting DOM, the form element is empty.

b)
```jsx
<Modal>
  <form>
    <Modal.Content><input type="text" /></Modal.Content>
    <Modal.Footer><Button type="submit"></Button></Modal.Footer>
  </form>
</Modal>
```
^ this doesn't work because `<Modal>` uses `extractChildren` and expects its subcomponents to be its direct children.
## Screenshots / Gifs / Codepens
![image](https://user-images.githubusercontent.com/93752/71609620-c0277700-2b3e-11ea-8310-1e30b29765d5.png)

## Additional context
Ok, I lied. In my particular case the form is AJAX-y, so technically the button doesn't _have_ to be inside the form element. **However**, I would argue that forms in modals are such a common case, we should definitely support a regular, non-AJAX HTML form.
Let's make sure we consider common scenarios like that when designing our APIs; I believe this was possible in ACL UI modals. We wanted to opt for a different composition model in Paprika but failed to notice it prohibits cases like this one.
(Also here's your regular plug for always trying to opt for context-based composition first—which would have made option b) possible. I believe switching from `extractChildren` to context is still possible for our modal, but would require changing the Overlay API. LMK if you need details)

alexzherdev:
1) Actually, this seems to be a real nuisance for AJAX forms as well.
If I want my form fields to be controlled, I'm basically forced to stuff all my form related code into the same component that renders the modal—so that the submit handler can access form state. I can't seem to find any way to break this apart into multiple components.
```jsx
const [value, setValue] = React.useState('');
function handleSubmit() {
  // submit `value` to the server
}
return (
  <Modal>
    <Modal.Content>
      <form>
        <input value={value} onChange={e => setValue(e.target.value)} />
      </form>
    </Modal.Content>
    <Modal.Footer>
[...]

PR #370, opened 2020-01-02
STAR-58-Modal - feature(Modal): ability to spread more props into the Modal container

nahumzs:
### Purpose 🚀
Partially address https://github.com/acl-services/paprika/issues/369 adding a workaround via styled-components and `as={}` prop

### Notes ✏️
This is a provisional fix, the full solution is to create a context to share state on the Modal.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-58-Modal-render

### Screenshots 📸

### References 🔗


packages/Modal/stories/Modal.stories.js
```diff
@@ -72,6 +72,23 @@ storiesOf("Modal", module)
...
+      onSubmit={event => {
+        alert("submit");
+        event.preventDefault();
+      }}
+    >
+      <Modal.Content>
+        <input type="text" deafultValue="your text" />
+      </Modal.Content>
+      <Modal.Footer>
+        <Button type="submit">Submit</Button>
```
alexzherdev: I think `isSubmit` prop is used for this. (clicking the button currently doesn't submit the form)

nahumzs: abstracted logic for the input type submits into the Button to make it work. https://codesandbox.io/s/quiet-water-8khug

packages/Modal/stories/Modal.stories.js
```diff
@@ -72,6 +72,31 @@ storiesOf("Modal", module)
...
+        event.preventDefault();
+      }}
+    >
+      <Modal.Content>
+        <input type="text" defaultValue="your text" />
+      </Modal.Content>
+      <Modal.Footer>
+        <Button
+          isSubmit
+          onClick={() => {
```
alexzherdev: This is not used I think

nahumzs: if the onClick is not included the onSubmit will not work when the button gets clicked, do you prefer to have a void/noop function and remove the console? unless we want to add a new event like onClickSubmit.

alexzherdev: That's surprising: we have a defaultProp for `onClick` which is `() => {}`, so it should work 🤔

nahumzs: your right, my bad, is working.

packages/Button/stories/examples/ButtonSubmit.js
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <ButtonStory>
+      <form
+        onSubmit={event => {
+          alert("submitted");
+          event.preventDefault();
+        }}
+      >
+        <input type="text" defaultValue="press enter while focus" />
+        <Button isSubmit isSemantic={false}>
```
DanilAgafonov: Excuse me, but... what is this?
image

What the reason to have `isSemantic={false}`?

As I understood, we render Button as `<span>` and then trying to emulate `<button>` behavior on `<span>`. But why?

nahumzs: It is for the same reason, style conflicts that are hard to fix from some applications in our ecosystem. In this case, the submit button behaves correctly as a button when it is render as span, but was lacking the ability to fire the onSubmit event, this addresses that. 

I guess eventually someday we will be able to deprecate the isSemantic prop.

alexzherdev: @DanilAgafonov 
image

PR #371, opened 2020-01-02
UX-698-FormElement-api-change-for-handling-children

tristanjasper:
### Purpose 🚀
We are using the formElement in results questionnaires for each question component.

Updated Description
We are now allowing the consuming application to handle the children that it would like to provide to the FormElement component. They will then have the choice to associate a label and `aria-description` id with a single input or otherwise if they so wish to.

Additionally native components and Paprika components all require different tags to be set on them, as well as in the case of the `ListBox` component a `refLabel` is available to be set as an attribute.

**REDUNDANT DESCRIPTION BELOW**
Each question has a number of different types of children, including plain divs, select options etc. We do not need formElement component to add `extendedProps` to these children. We only need to add `extendedProps` to a single input, derived from a paprika `Input` component or a native `input`, `select` or `textarea`.

It is expected that only a single child will be one of these that is passed to the formElement so this PR is doing two things

1. Ensuring the relevant `extendedProps` are only being applied to supported component types.
2. Ensures `extendedProps1` is only being set on a single child component.

### Updates 📦
- [x] MAJOR (breaking) change to FormElement
- [x] PATCH (bug fix) change to Button
- [x] PATCH (bug fix) change to Helpers

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-698-formElement-limit-supported-components

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/UX-698
This PR is to be made in conjunction with https://github.com/acl-services/acl-exception/pull/7108


packages/FormElement/src/FormElement.js
```diff
@@ -131,6 +131,14 @@ function FormElement(props) {
     "aria-invalid": hasError,
   };
 
+  const supportedComponents = ["DatePicker", "Input"];
```
tristanjasper: Currently we only support these "Paprika" components so we can be explicit about this. As we are using the `formElement` component as the root component for each question in results questionnaires (which gets passed many different types of children) it would be best to ensure the extendedProps only get added to the correct child type.

packages/FormElement/src/FormElement.js
```diff
@@ -164,6 +184,15 @@ function FormElement(props) {
                 "aria-readonly": isReadOnly,
               };
 
+              if (child.type.name === componentNameMap.LISTBOX) {
```
tristanjasper: only the Listbox component is currently making use of the `refLabel` prop. We need to ensure other components such as `Input` do not end up adding this to their generated `input` dom element so i'm being explicit here

packages/FormElement/src/FormElement.js
```diff
@@ -147,15 +165,17 @@ function FormElement(props) {
       <div css={isInline ? inlineContainerStyles : null}>
         {renderInstructions()}
         {extractedChildren.children.map(child => {
-          const extendedProps = isNativeElement(child)
-            ? {
+          let extendedProps = {};
+          if (!inputFound) {
```
tristanjasper: Currently formElement only really supports a single input component to `setErrors` etc on it. We really only want to  support adding these extendedProps to one supported component.

packages/FormElement/src/FormElement.js
```diff
@@ -132,6 +132,24 @@ function FormElement(props) {
     "aria-invalid": hasError,
   };
 
+  const componentNameMap = {
+    DATEPICKER: "DatePicker",
+    INPUT: "Input",
+    LISTBOX: "ListBox",
+  };
+
+  const supportedComponents = [componentNameMap.DATEPICKER, componentNameMap.INPUT, componentNameMap.LISTBOX];
```
allison-c: This part can be stored outside of the functional component?

tristanjasper: You mean th array? Yes it could be in some kind of helper and imported but for now I wanted to keep it simple until we other components need a similar lookup..

packages/FormElement/src/FormElement.js
alexzherdev: I would definitely say this should be the responsibility of the `Input` then.

packages/FormElement/src/FormElement.js
alexzherdev: I'll be honest I don't have a good alternative, but special casing components like that just seems really off to me. I guess this is a consequence of the overall API here, which again is based on inspecting your direct children and their types. 😞

packages/FormElement/src/FormElement.js
allison-c: Not only `Input` need to take care of it, `TextArea`, `Select`... all the similar components passing `moreProps` directly into DOM element have to take care of it. Is it more complex than doing it like this PR?

nahumzs:
What about if we destructure correctly the props in all our `Inputs` components, therefore they are not pushed to the children component.

if we know that **formElement** can passdown `a, b, c and d` props we do:

```
const { a, b, c, d, ...moreProps } = props;
```

we could even provide a function that encapsulated this destructuring part in the FormElement so controlled in one specific point. 

This idea might avoid changes to any API and will be transparent for the consumer.

I thought about using context, but then add complexity to each of the `inputs` where they will have to check if the context is available or not, kind of what we did with the L10n component. 

If we can prevent using `if` depending on the component in our code would be better IMHO.

tristanjasper:
@nahumzs I was thinking of creating explicit subcomponent, eg `FormElement.Input`, `FormElement.DatePicker` so that we are able to pass exactly the props we want that way to our  components. We would be able to create a context object for them all to access as well, thats kind of where I was heading next with this PR... Open to ideas though!

packages/DatePicker/src/DatePicker.js
```diff
@@ -32,7 +32,7 @@ const propTypes = {
   /** Date format used while displaying date. It should be human-friendly and spelled out, default is MMMM DD,YYYY */
   humanFormat: PropTypes.string,
 
-  /** ID for the <input>.  Not required when wrapped with <FormElement>. */
+  /** ID for the <input>. */
```
tristanjasper: We now need to set the id on the datepicker when using it in the formElement

packages/FormElement/src/FormElement.js
```diff
@@ -36,13 +36,10 @@ const propTypes = {
...
   isLabelVisuallyHidden: PropTypes.bool,
 
-  /** Should be read-only or not, default is false. */
-  isReadOnly: PropTypes.bool,
-
   /** Label text of this field. */
   label: PropTypes.string.isRequired,
 
-  /** Size of the label, child component, error, help and description (font size, min-height, padding, etc). */
+  /** Size of the label, error, help and description (font size, min-height, padding, etc). */
```
tristanjasper: `size` prop for the children (inputs etc) should be set explicitly

packages/FormElement/src/FormElement.js
```diff
@@ -66,117 +70,91 @@ function FormElement(props) {
...
-        id: ariaInstructionsId,
-      });
-    }
-    return null;
+  function renderContent() {
+    return getClonedElement(subComponentDisplayNames.Content, {
+      idForLabel,
+      refLabel,
+      ariaDescribedBy: `${ariaErrorId} ${ariaInstructionsId} ${ariaDescriptionId}`,
+    });
```
tristanjasper: This is the new `Content` subcomponent which will return a function to the consumer allowing them to use the props that can associate them with the label component

packages/FormElement/stories/examples/AccessibilityExample.js
```diff
@@ -37,7 +37,18 @@ const ExampleStory = () => {
...
-        <Input onChange={handleChange} value={value} placeholder="Form placeholder" />
+        <FormElement.Content>
+          {({ idForLabel, ariaDescribedBy }) => (
+            <Input
+              id={idForLabel}
+              onChange={handleChange}
+              value={value}
+              placeholder="Form placeholder"
+              aria-describedby={ariaDescribedBy}
+              hasError={Boolean(errorText.length)}
```
tristanjasper: Proposing we leave the `hasError` prop on the children to be set by the consumer rather than have the FormElement do it. That way if they have a number of children (more than one input for example) and they want to highlight only one they can.

packages/Input/src/Input.js
```diff
@@ -117,6 +117,9 @@ const Input = props => {
     <div css={inputStyles} {...styleProps} className={rootClasses}>
       {renderIcon()}
       <input
+        aria-invalid={hasError}
+        aria-disabled={isDisabled}
+        aria-readonly={isReadOnly}
```
tristanjasper: I believe we should be setting these attributes on our native input elements as well? @mikrotron

packages/FormElement/src/FormElement.js
```diff
@@ -66,117 +70,91 @@ function FormElement(props) {
...
   const ariaDescriptionId = React.useRef(uuidv4()).current;
   const ariaErrorId = React.useRef(uuidv4()).current;
   const ariaInstructionsId = React.useRef(uuidv4()).current;
-  const hasError = !!extractedChildren["FormElement.Error"] && !!extractedChildren["FormElement.Error"].props.children;
   const uniqueInputId = React.useRef(uuidv4()).current;
-  const inputId = isNil(id) || id === "" ? uniqueInputId : id;
+  const hasError =
+    !!extractedChildren[subComponentDisplayNames.Error] &&
+    !!extractedChildren[subComponentDisplayNames.Error].props.children;
+  const idForLabel = isNil(id) || id === "" ? uniqueInputId : id;
```
allison-c: maybe also rename `uniqueInputId ` to `uniquLabelId `

packages/FormElement/src/FormElement.styles.js
```diff
@@ -1,6 +1,6 @@
 import { css } from "styled-components";
 
-import tokens from "@paprika/tokens";
+// import tokens from "@paprika/tokens";
```
allison-c: not used?

packages/FormElement/src/components/Content/Content.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import PropTypes from "prop-types";
+import ContentContainerStyled from "./Content.styles";
+
+const propTypes = {
+  children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
+  idForLabel: PropTypes.string,
+  refLabel: PropTypes.shape({ current: PropTypes.instanceOf(Object) }),
+  ariaDescribedBy: PropTypes.string,
```
allison-c: Should these props be all required since Content is always inside FormElement

packages/FormElement/src/components/Description/Description.styles.js
```diff
@@ -1,12 +1,11 @@
...
-
-import tokens from "@paprika/tokens";
+import styled from "styled-components";
 import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
 
-const descriptionStyles = css`
-  ${stylers.lineHeight(-1)}
+const DescriptionStyled = styled.div`
+${stylers.lineHeight(-1)}
```
allison-c: The indentation has been removed

packages/FormElement/src/components/Instructions/Instructions.styles.js
```diff
@@ -1,12 +1,11 @@
-import { css } from "styled-components";
-
+import styled from "styled-components";
 import tokens from "@paprika/tokens";
 import stylers from "@paprika/stylers";
 
-const instructionsStyles = css`
-  ${stylers.lineHeight(-1)}
+const InstructionsStyled = styled.div`
+${stylers.lineHeight(-1)}
```
allison-c: same here

packages/Input/src/Input.js
allison-c: I think `disabled` and `readOnly ` are equal to aria-disabled and aria-readonly, but not 100% sure 😄

packages/FormElement/src/components/Content/Content.js
tristanjasper: image
I originally wanted them to be required but run into compiling warnings as they are added to the cloned element and not set by the consumer. Any ideas on how to avoid them?

allison-c: Oh I see, I forget they're empty initially, I don't have ideas 😄

packages/Input/src/Input.js
tristanjasper: I find some people mention `aria-disabled` is not needed if using `disabled` prop but don't find my opinion on the other attributes

packages/FormElement/src/FormElement.js
tristanjasper: It's not an id on the Label though, its the id of the input that gets used by the `htmlFor` attribute on the `label` element

packages/FormElement/src/components/Content/Content.js
tristanjasper: Yeah i've been running into that issue on other components after using `React.cloneElement` so would be great if someone had a suggestion for dealing with this!

packages/FormElement/src/FormElement.js
```diff
@@ -66,117 +70,91 @@ function FormElement(props) {
...
-    "FormElement.Description",
-    "FormElement.Error",
-    "FormElement.Help",
-    "FormElement.Instructions",
+    subComponentDisplayNames.Description,
+    subComponentDisplayNames.Error,
+    subComponentDisplayNames.Help,
+    subComponentDisplayNames.Instructions,
+    subComponentDisplayNames.Content,
   ]);
```
mikrotron: Just pass `Object.values(subComponentDisplayNames)` ?

packages/FormElement/stories/examples/AlternateLayouts.js
```diff
@@ -31,36 +42,64 @@ const ExampleStory = () => {
...
+      <br />
+      <FormElement hasRequiredLabel={hasRequiredLabel} label="Form Label">
         <FormElement.Instructions>
           <span>
             Example text for extra panel for questionnaires. Example text for extra panel for questionnaires Example
             text for extra panel for questionnaires. Example text for extra panel for questionnaires
           </span>
         </FormElement.Instructions>
-        <Input onChange={handleChange} value={value} placeholder="Form placeholder" />
+        <FormElement.Content>
```
alexzherdev: Oof I can't say I have a better alternative in mind (except perhaps the `<FormElement.Input>` idea we discussed), but having to add this wrapper in every single form element seems too verbose.
I'm curious if others feel the same

packages/FormElement/src/FormElement.js
```diff
@@ -66,117 +70,91 @@ function FormElement(props) {
...
   const uniqueInputId = React.useRef(uuidv4()).current;
-  const inputId = isNil(id) || id === "" ? uniqueInputId : id;
+  const hasError =
+    !!extractedChildren[subComponentDisplayNames.Error] &&
+    !!extractedChildren[subComponentDisplayNames.Error].props.children;
+  const idForLabel = isNil(id) || id === "" ? uniqueInputId : id;
   const refLabel = React.useRef(null);
 
+  const getClonedElement = (displayName, extraProps = {}) => {
+    return React.cloneElement(extractedChildren[displayName], extraProps);
```
mikrotron: Should there be a `if (!extractedChildren[displayName]) return null;` line here?

packages/FormElement/stories/examples/AlternateLayouts.js
mikrotron: Having to use it as a render prop every time seems a bit cumbersome... but sometimes you might want the control.  What if we handle both? 

If `Content` has a function as `children`, do this, but if its `children` is just a single `node`, just clone the `idForLabel` + `ariaDescribedBy` onto it? 

And if it's multiple nodes... uh... just render them? Maybe it would be nice to have a way to also "just render them" even if it's a single node... but not sure what the API could look like for that 🤔

mikrotron: ...maybe `<FormElement.Content shouldAugmentProps={false}>`? 
It could be `true` by default?

packages/FormElement/src/FormElement.js
```diff
@@ -66,117 +70,91 @@ function FormElement(props) {
...
     </div>
   );
 }
 
 FormElement.displayName = "FormElement";
-
 FormElement.propTypes = propTypes;
 FormElement.defaultProps = defaultProps;
-
+FormElement.SubComponentDisplayNames = subComponentDisplayNames;
```
mikrotron:
```suggestion
FormElement.subComponentDisplayNames = subComponentDisplayNames;
```

mikrotron: So that this doesn't seem like an actual component. It also aligns with the decision recently made here:

https://github.com/acl-services/paprika/pull/392

packages/FormElement/stories/examples/AlternateLayouts.js
tristanjasper: @mikrotron I think if you want to use the `idForLabel`, `ariaDescribedBy` or `labelRef` we should make the consumer make use of the `FormElement.Content` component to avoid confusion. If they do not need any of those props, as in the case they want to render multiple children, then we can just render them out without making use of `FormElement.Content`...wdyt?

packages/FormElement/src/components/Content/Content.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import PropTypes from "prop-types";
+import ContentContainerStyled from "./Content.styles";
+
+const propTypes = {
+  children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
+  idForLabel: PropTypes.string,
+  refLabel: PropTypes.shape({ current: PropTypes.instanceOf(Object) }),
```
mikrotron: Would be sweet to have this as one of our customPropTypes... 😄

packages/FormElement/src/components/Content/Content.js
```diff
@@ -0,0 +1,36 @@
...
+  ariaDescribedBy: PropTypes.string,
+};
+
+const defaultProps = {
+  idForLabel: "",
+  ariaDescribedBy: "",
```
mikrotron: Why not `null`?

packages/FormElement/src/components/Content/Content.js
```diff
@@ -0,0 +1,36 @@
...
+
+function Content(props) {
+  const { children, idForLabel, refLabel, ariaDescribedBy, ...moreProps } = props;
+
+  if (!children) {
+    return null;
+  }
+
+  return (
+    <ContentContainerStyled data-pka-anchor="formElement.content" {...moreProps}>
```
mikrotron:
```suggestion
    <ContentContainerStyled data-pka-anchor="form-element.content" {...moreProps}>
```

packages/FormElement/src/components/Content/Content.js
```diff
@@ -0,0 +1,36 @@
...
+      {typeof children === "function" ? children({ idForLabel, refLabel, ariaDescribedBy }) : children}
```
mikrotron: What if in the `else` case here we clone the `children` node (if there's only one) and add `idForLabel`, `refLabel`, `ariaDescribedBy`? Unless we're provided an explicit prop to flip that off?

packages/FormElement/src/FormElement.js
tristanjasper: Ok in the case of the `FormElement.Error` or `FormElement.Description` yes that would cause an error, although I won't do they check on that line, rather at the render function level..

packages/FormElement/src/components/Description/Description.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 
-import descriptionStyles from "./Description.styles";
+import DescriptionStyled from "./Description.styles";
```
mikrotron: I think the convention we settled on with Danil was to export as `styled`, so if there are multiple, you can just import that.  Then use like `<styled.Description>`.  Kinda trivial difference here, but when you have multiple being exported, it makes consuming easier.

packages/FormElement/src/components/Content/Content.js
tristanjasper: Because how do we know that that single component needs those props? For example only the `listBox` needs the `refLabel`. And perhaps there is only a single checkbox component, which also doesn't need the `idForLabel` etc. I would rather not be checking the type or displayName of the component in this component either as thats the point of this PR. To decouple the component type from what we can provide in the component...

packages/FormElement/src/components/Description/Description.js
mikrotron: Example:
https://github.com/acl-services/paprika/blob/master/packages/Modal/src/Modal.styles.js

packages/FormElement/stories/examples/AlternateLayouts.js
```diff
@@ -31,36 +42,64 @@ const ExampleStory = () => {
       </Heading>
       <Rule />
       <Tagline>Form Element with instructions component.</Tagline>
-      <Rule />
-      <FormElement label="Form Label">
+      <br />
```
mikrotron: `<Gap />`?
(Also from `storybook/assets/styles/common.styles`)

packages/Input/src/Input.js
mikrotron: Yeah, if it's a native html element, `disabled` and `readonly` are equivalent.  It doesn't _hurt_ have have the extra `aria` attributes, but the redundancy is not necessary.

packages/FormElement/src/FormElement.js
mikrotron: Maybe we should add `required` too if `hasRequiredLabel` is `true`?

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required

packages/FormElement/src/FormElement.js
```diff
@@ -36,13 +36,10 @@ const propTypes = {
   /** Should label be hidden, default is false. Note: this is discouraged because of accessibility requirements. */
   isLabelVisuallyHidden: PropTypes.bool,
 
-  /** Should be read-only or not, default is false. */
-  isReadOnly: PropTypes.bool,
-
   /** Label text of this field. */
   label: PropTypes.string.isRequired,
```
mikrotron: I know I said we should wait for another PR, and maybe we still should, but if this `label` prop was optional and we had another prop `legend` that would add a `<fieldset>` wrapper and put the `legend` in a `<legend>` then I think we'd be all set up for checkboxes and radios.

mikrotron: Or maybe we should keep this one as `isRequired` and add a `hasMultipleInputs` prop that does the same thing?

packages/FormElement/src/components/Description/Description.js
tristanjasper: Oh strange, I already had made that change a day ago, are you looking at current code?

packages/FormElement/stories/examples/AlternateLayouts.js
nahumzs: I feel the same, I don't think is a good idea to make the FormElement extend/clone their children components, the FormElement is a Component that will be used in multiple places and in many different scenarios, and doing all this (wrapper, cloning) seem odd to me and against the idea of composition. 

We are actually making the FormElement aware of all other input components instead of making it agnostic, which can be a red sign that we are doing something wrong. 

The only solid solution that I see is the FormElement use a renderProp function and pass down whatever we want to. Each component (input) is responsible to use whatever suite their case.

A renderProp work for 1 , 2 multiple scenario, yes is cumbersome but to me is the right pattern to use in this case. 

my two cents.

packages/FormElement/src/components/Description/Description.js
tristanjasper: Ok, i'll use that approach for the components i've updated

packages/FormElement/src/components/Content/Content.js
tristanjasper: Ok I've added

packages/Input/src/Input.js
tristanjasper: @mikrotron opinion on `aria-invalid`?

packages/FormElement/stories/examples/AlternateLayouts.js
nahumzs: Also, let's take into consideration that if we make this API incorrectly removing it from our ecosystem will be hard over time because it will be used multiple times per feature, is not like the sidepanel or modal which normally is 1 use per features.

alexzherdev: If so I'd seriously consider @tristanjasper's idea of producing thin wrappers like `<FormElement.Input>` that would hook into FormElement's context and remove the need for cloning or inspecting children.

packages/FormElement/src/components/Description/Description.js
tristanjasper: Gotta say i'm not all that keen on it though, I think it makes the Markup a bit less readable with `<styled.` written before every component. I'm not so sure it's good for every situation

packages/FormElement/src/FormElement.js
tristanjasper: This can be set by the consumer, I don't see the need to have this component figure out what the type of component the child is in order to know _if_ it should set `required` on it

packages/Input/src/Input.js
mikrotron: That one's good.  There's no native html equivalent.

packages/FormElement/src/FormElement.js
tristanjasper: Yeah if we do not include the `label` text then there would be no title at the top of the component, because that is what it is used for.. I think we should always show that in this component

packages/FormElement/src/FormElement.js
tristanjasper: Sure the only two components that were not covered were in the footer. So this covers that now

packages/FormElement/src/components/Content/Content.js
```diff
@@ -1,11 +1,12 @@
 import React from "react";
 import PropTypes from "prop-types";
+import { RefPropType } from "@paprika/helpers/lib/customPropTypes";
 import ContentContainerStyled from "./Content.styles";
 
 const propTypes = {
   children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
   idForLabel: PropTypes.string,
-  refLabel: PropTypes.shape({ current: PropTypes.instanceOf(Object) }),
+  refLabel: RefPropType,
```
mikrotron: 😍

packages/FormElement/src/components/Content/Content.js
mikrotron: Might be a good idea to have a custom propType that will check the `children` prop of `Content` and throw a console warning (in dev env) if the type is not a function – it will still seem like it's working just fine without the `id` and `aria-describedby` so consumers might not realize their implementation is not accessible.  But that can be done in another PR.

cc @nahumzs

packages/helpers/src/customPropTypes.js
```diff
@@ -37,3 +38,5 @@ export const InputValidTypes = {
 };
 
 InputValidTypes.ALL = Object.values(InputValidTypes);
+
+export const RefPropType = PropTypes.shape({ current: PropTypes.instanceOf(Object) });
```
alexzherdev: If this is a generic prop type for a ref, then this can be parameterized with the type of what goes under `current`. (also `PropType` in the name is redundant)
Like 
```js
export const RefOf = (type) => PropTypes.shape({ current: PropTypes.instanceOf(type) });
```

nahumzs:
```
PropTypes.oneOfType([
    PropTypes.func, 
    PropTypes.shape({ current: PropTypes.any })
])
```

packages/FormElement/README.md
```diff
@@ -28,7 +43,22 @@ import DatePicker from "@paprika/date-picker";
...
+        id={idForLabel}
+        onChange={handleChange}
+        value={value}
+        placeholder="Form placeholder"
+        aria-describedby={ariaDescribedBy}
+        aria-required={hasRequiredLabel}
+        hasError={false}
+        isDisabled={isDisabled}
+        isReadOnly={isReadOnly}
+        size={size}
```
mikrotron: This isn't quite right, according to the [README](https://github.com/acl-services/paprika/blob/master/packages/DatePicker/README.md). But, this doesn't need to document that component, so I'd just stick to the simplest case (still might need to show how to propagate `ariaDescribedBy` to the `<DatePicker.Input>` though?)

packages/FormElement/README.md
```diff
@@ -66,10 +96,17 @@ import MyInput from "./MyInput";
 - `label` (required)
 - `size`
 
+#### FormElement.Content
+
+- `children` (required)
```
mikrotron: I'm not sure how to do it, but I think it would be helpful to document the arguments to the `children` function (`idForLabel`, `refLabel`, `ariaDescribedBy`).

Might also be useful to have that here:
https://github.com/acl-services/paprika/blob/ux-698-formElement-limit-supported-components/packages/FormElement/src/components/Content/Content.js#L7

...but for now I think the README is good enough.

packages/FormElement/src/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [0.3.0] - 2020-01-17
```
mikrotron: I think you should also say what was removed (https://github.com/acl-services/paprika/pull/371/files#diff-70850fa02375fbeb5f7256649db70099L167) since that's the breaking change to the API.

nahumzs:
shouldn't CHANGELOG.md be at the same level as the README.md ? no on the src folder?

packages/helpers/src/customPropTypes.js
```diff
@@ -37,3 +38,5 @@ export const InputValidTypes = {
 };
 
 InputValidTypes.ALL = Object.values(InputValidTypes);
+
+export const RefOf = PropTypes.oneOfType([PropTypes.func, PropTypes.shape({ current: PropTypes.any })]);
```
mikrotron:
```suggestion
export const RefOf = (propType) => PropTypes.oneOfType([PropTypes.func, PropTypes.shape({ current: propType || PropTypes.any })]);
```

mikrotron: This is what you meant, right @alexzherdev ?

alexzherdev: yeah and `any` can go into the function signature as the default value

packages/FormElement/src/components/Content/Content.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { RefOf } from "@paprika/helpers/lib/customPropTypes";
+import * as styled from "./Content.styles";
+
+const propTypes = {
+  children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]).isRequired,
+  idForLabel: PropTypes.string,
+  refLabel: RefOf(PropTypes.func),
```
mikrotron:
```suggestion
  refLabel: RefOf(PropTypes.instanceOf(Element)),
```

mikrotron: It's not a function, it's the actual dom element, right?

https://stackoverflow.com/a/51127130

tristanjasper: 👍

Issue #365, opened 2019-12-30
Modal content focus outline hidden at the bottom by footer


alexzherdev:
# Bug Report
## Expected behavior
The focus outline is fully visible when modal's content is focused.

## Current behavior
When the modal is short (i.e. there is no vertical scrolling), the bottom part of the outline is hidden by the footer.

## Screenshots / Gifs / Codepens
![image](https://user-images.githubusercontent.com/93752/71596847-c3e2db80-2af5-11ea-840d-44482c92837b.png)

## Additional context
Looks like I can fix this by setting `z-index: 0` on the content element, but not sure that's the right way to go.

mikrotron:
@alexzherdev yeah this is not ideal, but we are planning to change it soon. We will use a custom styled focus indicator, but since this is not an input or control, we're probably going to use a more subtle design than the usual one which we have a token for.  @sunminsarahlee is working on it, and we'll update the component when she's got approval from the design team.

mikrotron:
Style will be `2px dotted #8a8a8a`.

Issue #172, opened 2019-08-20
Not show focus styling if user is not using a keyboard


oscarkwan:
# Feature request
## Prerequisites
Going to relay a message here from @mattcrest  on focus styling.

Basically, we don’t want focus styling to appear if we are *not* using the keyboard to navigate through our components. 

We would want this as a global change so for all our components this won't happen.

## Is your feature request related to a problem?
- We don't want focus styling to appear if you are not using the keyboard to interact

## Describe the solution you'd like
- a Global change that will affect all components going forward

![Screen Shot 2019-08-20 at 1 42 48 PM](https://user-images.githubusercontent.com/19940076/63386067-1564a800-c357-11e9-9a76-75ea6835b3ae.png)

oscarkwan:
I don't have a solution in mind at the moment, open to discussion

jamiek-acl:
How about a listener that adds a class to the body when the keyboard is used, and removes that class when the mouse is clicked?
Then in that class remove the focus style.

oscarkwan:
> How about a listener that adds a class to the body when the keyboard is used, and removes that class when the mouse is clicked?
> Then in that class remove the focus style.

🤔 🤔 Good suggestion

alexzherdev:
[:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) would be the proper solution, but browser support is still limited 😕 
https://stackoverflow.com/a/51093815 has a couple code samples, looks like @jamiek-acl's idea might work. Although I'm not a fan of the library operating on such a global level, but ‾\\\_(ツ)_/‾

mikrotron:
We adopted Jamie's approach for the _almost dead_ Starling site.  I don't think it creates any major issues, but I also have the same knee-jerk reservations about a global script.  It's also not so easy to remove, because it means we have to add a lot of conditions throughout the css of our components.

mikrotron:
The library used on the Starling site:
https://github.com/ten1seven/what-input

nahumzs:
some examples implementing @jamiek-acl ideas https://jmperezperez.com/outline-focus-ring-a11y/ https://medium.com/hackernoon/removing-that-ugly-focus-ring-and-keeping-it-too-6c8727fefcd2

seems like the way to go.

mikrotron:
Resolved by https://github.com/acl-services/paprika/pull/373

PR #372, opened 2020-01-02
STAR-59  Modal content focus indicator 🐐

mikrotron:
### Purpose 🚀
Add a new "subtle" focus indicator style to the `<Modal>` component's scrollable content container.

### Notes ✏️
- Added new token: `highlight.active.nonInteractive.outline`.
- Focus indicator token names are too long to remember, and sometimes additional rules are needed with them, so I've created new `stylers` (includes / mixins) with the following names in JS:
  - `stylers.focusRing` (also works as `stylers.focusRing()`)
  - `stylers.focusRing(true)` (the argument is for `inset` style)
  - `stylers.focusRing.bordered`
  - `stylers.focusRing.bordered(true)`
  - `stylers.focusRing.subtle` (this is the new one for "non-interactive" elements)
  - `stylers.focusRing.subtle(true)`
And in Sass:
  - `focus-ring`
  - `focus-ring(true)`
  - `focus-ring--bordered`
  - `focus-ring--bordered(true)`
  - `focus-ring--subtle`
  - `focus-ring--subtle(true)`

### Updates 📦

- [x] MINOR (backward compatible) change to `modal`, `tokens`, `stylers`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-59--modal-focus-indicator/?path=/story/modal--basic

### Screenshots 📸
Before
Screen Shot 2020-01-02 at 1 01 04 PM

After
Screen Shot 2020-01-02 at 12 59 56 PM

All `focusRing` styles:
Screen Shot 2020-01-03 at 1 34 41 PM

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-59
Resolves: https://github.com/acl-services/paprika/issues/365

### DIFF
--- a/packages/Stylers/src/includes.js
+++ b/packages/Stylers/src/includes.js
@@ -51,6 +51,45 @@ export const visuallyHidden = css`
   width: 1px !important;
 `;
 
+// Focus Indicators
+
+export const focusRing = (inset = false) => {
+  return inset === true
+    ? css`
+        box-shadow: ${tokens.highlight.active.withBorder.insetBoxShadow};
+        outline: none;
+      `
+    : css`
+        box-shadow: ${tokens.highlight.active.withBorder.boxShadow};
+        outline: none;
+      `;
+};
+
+focusRing.bordered = (inset = false) => {
+  return inset === true
+    ? css`
+        border-color: ${tokens.highlight.active.noBorder.borderColor};
+        box-shadow: ${tokens.highlight.active.noBorder.insetBoxShadow};
+        outline: none;
+      `
+    : css`
+        border-color: ${tokens.highlight.active.noBorder.borderColor};
+        box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
+        outline: none;
+      `;
+};
+
+focusRing.subtle = (inset = false) => {
+  return inset === true
+    ? css`
+        outline: ${tokens.highlight.active.nonInteractive.outline};
+        outline-offset: -2px;
+      `
+    : css`
+        outline: ${tokens.highlight.active.nonInteractive.outline};
+      `;
+};
+
 //
 // Forms
 //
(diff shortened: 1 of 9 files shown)


packages/Stylers/src/includes.js
```diff
@@ -47,10 +47,40 @@ export const visuallyHidden = css`
   overflow: hidden;
   padding: 0 !important;
   position: absolute !important;
+  white-space: nowrap;
```
mikrotron: This improves the way screen readers will read content in visually hidden elements.

Reference: https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe#.2g9wpybl9

oscarkwan:
We really need https://github.com/acl-services/paprika/issues/172 😅

packages/Stylers/src/includes.js
```diff
@@ -47,10 +47,40 @@ export const visuallyHidden = css`
...
   width: 1px !important;
 `;
 
+export const focusIndicator = css`
```
nahumzs: I like this abstraction, but don't understand what they mean or what is their intention, I don't see myself remembering this on top of my head easily, like I do with `stylers.spacer()` aren't the name a little complicated to remember? what about encapsulated this inside of a focus object.

I wouldn't know which one to use if I only see this code. 

export default focusIndicator = {
   inset: ``,
   normal: ``,
   subtle: ``,
   composite: ``,
   compositeInser: ``,
}

mikrotron: I'm not so sure it's any easier to remember `focusIndicator.inset` vs `insetFocusIndicator`.  But maybe... since the pattern is somewhat more consistent.  

Alternatively, it could be a function that takes an `options` argument... but that would still probably be about the same cognitive load.

`focusIndicator({ type: 'composite', inset: true })` (with defaults so the most common case could still be `focusIndicator()`. I'm not sure this could be replicated for the Sass version though, so it might have to be two arguments instead of an object.

Anyone prefer this API?

mikrotron: Okay, I've taken your feedback into account, and after discussing with Jamie have settled on an API I think I like.  Will update the description above with details.

nahumzs:
I'm ok with the api :)

PR #373, opened 2020-01-02
Dont put focus ring on Button or RawButton when navigating with keyboard

jamiek-acl:
### Purpose 🚀
Dont put focus ring on Button or RawButton when navigating with keyboard.
https://github.com/acl-services/paprika/issues/172

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/issue-172-no-focus-styling-keyboard

### Screenshots 📸
Notice the ring is not added when using the mouse, and it is added when using the keyboard.

RawButton:
focus-ring2

Button:
focus-ring

Story:
Screen Shot 2020-05-07 at 3 46 52 PM


nahumzs:
`what-input` ohh that's a cool library. This seems like actually solve the issue of the focus ring unless using mouse over, I don't have any counter about this.

oscarkwan:
ship it

packages/Button/src/Button.js
```diff
@@ -160,15 +161,24 @@ const Button = React.forwardRef((props, ref) => {
...
-      <ButtonIcon {...iconProps} isPending={isPending} data-pka-anchor="button.icon">
-        {isPending ? <RefreshIcon /> : icon}
-      </ButtonIcon>
-      {children}
-      <ButtonIcon {...iconProps} isDropdown data-pka-anchor="button.icon">
-        {isDropdown && <DownIcon />}
-      </ButtonIcon>
-    </span>
+    <React.Fragment>
+      <GlobalStyle />
```
alexzherdev: Hmm will 100 buttons create 100 global styles?

jamiek-acl: I guess so

packages/Button/src/Button.js
```diff
@@ -160,15 +161,24 @@ const Button = React.forwardRef((props, ref) => {
...
+      <span
+        css={buttonStyles}
+        data-pka-anchor="button"
+        {...buttonProps}
+        data-pka-type="button"
```
mikrotron: Why not use `data-pka-anchor`?

packages/Button/src/Button.styles.js
```diff
@@ -266,6 +266,13 @@ function getIconColor(props) {
   return iconColors[props.kind];
 }
 
+export const GlobalStyle = createGlobalStyle`
+  [data-whatinput="mouse"] [data-pka-type="button"]:focus {
```
DanilAgafonov: You don't need global styles here. And `[data-pka-type="button"]` too.

You can add this styles to `buttonStyles`:
```css
[data-whatinput="mouse"] &:focus {
  ...
}
```

DanilAgafonov: You don't need global styles here. And `[data-pka-type="button"]` too.

You can add this styles to `buttonStyles` or `commonStyles`:
```css
[data-whatinput="mouse"] &:focus {
  ...
}
```

packages/Button/src/Button.js
jamiek-acl: Because the consumer can pass it in and change it, like this: https://github.com/acl-services/paprika/blob/master/packages/Button/stories/examples/Basic.js#L25

packages/Button/src/Button.styles.js
jamiek-acl: I was sure this wouldn't work, but it did. Great suggestion!

mikrotron:
I feel like this technique threatens to violate WCAG 2.4.7 – Focus Visible (https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html) but after a quick discussion with @oscarkwan and @KenRiku we think it's okay. Focus is still shown for keyboard users, and like Oscar noticed, popular apps like Slack don't show focus indicators on buttons either (when using the mouse).

packages/Button/src/Button.styles.js
```diff
@@ -28,11 +28,17 @@ const commonStyles = `
     outline: none;
   }
 
+  [data-whatinput="mouse"] &:focus {
+    box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.1);
+    border-color: ${tokens.border.color};
+  }
```
mikrotron: Need to be a little more selective about this... it actually shouldn't be applied to `minor` or `link` buttons.  And the `box-shadow` probably shouldn't be applied to the `flat` buttons. 

Screen Shot 2020-01-03 at 10 32 02 AM
Screen Shot 2020-01-03 at 10 31 56 AM

mikrotron:
I did notice though, that for the `Ref` examples stories, where focus is set programmatically, the focus indicator is not shown. Not sure why... but this would be a violation of WCAG 2.4.7.

alexzherdev:
@mikrotron both RawButton's and Button's ref examples work for me 🤔

mikrotron:
> @mikrotron both RawButton's and Button's ref examples work for me 🤔

Oh, I see.  If you click on something (in the iframe), then load the story though... then you're still in "mouse mode" or whatever and it doesn't work. Well, it doesn't show the focus ring, it still focusses the element.

jamiek-acl:
> > @mikrotron both RawButton's and Button's ref examples work for me 🤔
> 
> Oh, I see. If you click on something (in the iframe), then load the story though... then you're still in "mouse mode" or whatever and it doesn't work. Well, it doesn't show the focus ring, it still focusses the element.

@mikrotron It's unclear what you mean here.  I loaded each `Button` story in its own window without a frame.  In all of them:

1. when using the keyboard the focus ring appears (that's good)
2. when using the mouse the focus ring does not appear (that's also good)

The only one that is a little suspicious is the `Buttons \ Examples \ Old Ref` (http://localhost:9009/iframe.html?id=button-examples--old-ref&viewMode=story) which puts a blue ring on the button onLoad, but that makes sense as i havent clicked anywhere with the mouse yet.

Please try it again, or let me know if I am missing something.

PR #375, opened 2020-01-03
Ux 699 listboxbrowser access pass 1

tristanjasper:
### Purpose 🚀
This PR ensures listBoxBrowser correctly shows focus on a single list box item at any time.
Additionally changes the border style of the column when focused and fixes the focus style box being cut off.

Addresses issues 1,  8 and focus outline cutting off.
of 
https://aclgrc.atlassian.net/wiki/spaces/UX/pages/662410769/Listbox+Browser+Component+Review+11+22+2019

### Notes ✏️

### Updates 📦
- [x ] PATCH (bug fix) change to _these packages_

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-699-listboxbrowser-acc-pass-1

### Screenshots 📸
Before:
accissues-before

After:
accissues

### References 🔗
https://aclgrc.atlassian.net/browse/UX-699


tristanjasper:
I have closed https://github.com/acl-services/paprika/pull/344 and created this new PR to use correct branch number

packages/ListBox/src/components/Content/Content.styles.js
```diff
@@ -0,0 +1,9 @@
+import styled from "styled-components";
+import tokens from "@paprika/tokens";
+
+export const ContentStyled = styled.div`
+  :focus {
+    outline: 2px ${tokens.color.blackLighten30} dotted;
```
tristanjasper: @jayde-acl Changed the dotted line colour to be as you requested

packages/ListBox/src/components/Content/Content.js
```diff
@@ -41,25 +42,32 @@ const handleBlur = (state, dispatch) => () => {
...
 export default function Content(props) {
   const [state, dispatch] = useListBox();
   const { refListBoxContainer } = state;
 
   /* NOTE no idea what ROLE should be this div when the ListBox is INLINE */
-  /* eslint-disable jsx-a11y/no-static-element-interactions */
   if (state.isInline) {
     return (
-      <div
+      <ContentStyled
```
tristanjasper: Currently this logic of removing the focus style of the list items when focus is moved away from the popover content is only happening when the listbox component with prop `isInline`. I ran into some issues when the listBox was simply in its basic configuration (not inline) but perhaps could be revisited if required in future.

nahumzs:
can you remove this http://storybooks.highbond-s3.com/paprika/ux-699-listboxbrowser-acc-pass-1/?path=/story/listbox-more-examples--columns-select story I don't think we need it now the listBoxBrowser is almost complete

tristanjasper:
> can you remove this http://storybooks.highbond-s3.com/paprika/ux-699-listboxbrowser-acc-pass-1/?path=/story/listbox-more-examples--columns-select story I don't think we need it now the listBoxBrowser is almost complete

Ok removed!

packages/ListBox/src/components/Content/Content.styles.js
mikrotron: Coming soon:
https://github.com/acl-services/paprika/pull/372/files#diff-35881852d236126a5c4a48174b177f8aR88

packages/ListBox/src/components/Content/Content.js
nahumzs: but is not affecting the current one correct?

tristanjasper: It is affecting the the listbox in the listbox browser (as both listboxes are inline there). Is that what you mean by current one?

packages/ListBox/src/components/Content/Content.styles.js
tristanjasper: Dotted line will look slightly different using that mixin/helper but we can switch it over when its available. 
image

PR #383, opened 2020-01-08
STAR-80 feature(ListBox): fixing stale state on ListBox

nahumzs:
### Purpose 🚀
In the current implementation, the onChange prop was store directly on the context state of the ListBox, this creates a stale state for the function which means will not get update it with the latest changes of the consumer state. 

This change helps to pass down the `prop.onChange` to the ListBox and use `useEffect` to properly handle the callback.

leaving some read about the issue:
https://dmitripavlutin.com/react-hooks-stale-closures/  
https://reactjs.org/docs/hooks-faq.html#why-am-i-seeing-stale-props-or-state-inside-my-function
https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

### Notes ✏️
- [x] added test for this specific issue.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-80-ListBox-scope-onChange-issue/?path=/story/listbox-single--onchange-listbox

### Screenshots 📸

### References 🔗


packages/ListBox/src/helpers/invokeOnChange.js
```diff
@@ -31,18 +31,14 @@ const invokeOnChange = (onChangeCallback = () => {}, eventType = "", event) => (
   }
 
   if (state.isMulti) {
-    onChangeCallback.apply(null, [
-      ...getSelectedOptionsMulti(state),
-      { dispatch, actionTypes: sanitizeActionTypes(), eventType, ...event },
-    ]);
+    const [selected, options, current] = getSelectedOptionsMulti(state);
```
nahumzs: this doesn't affect anything but it is better to remove the apply which doesn't do anything and creates confusion.

packages/ListBox/src/ListBox.js
```diff
@@ -62,8 +62,15 @@ export const defaultProps = {
 };
 
 export function ListBox(props) {
-  const [state] = useListBox();
-  const { children, height, placeholder, trigger: _trigger, footer, filter } = props;
+  const [state, dispatch] = useListBox();
+  const { children, height, placeholder, trigger: _trigger, footer, filter, onChange } = props;
+
+  React.useEffect(() => {
```
alexzherdev: TBH the fact you have to do this is concerning. This means that after I pass in a new handler, there is a non-trivial amount of time when the component still isn't aware it has to use the new handler (before the effect is run, which could be delayed). If I'm passing a new function as a prop, triggering a re-render of the component, the component must be aware of the new value in this same re-render, not afterwards. If it isn't, then the whole component prop-passing model is broken.

nahumzs: I'm not entirely understanding what you are describing. 

The onChange is somehow "complex" callback, can be trigger from several locations, by click on an option, by clicking enter/space, by clicking the clear button of the trigger or even clicking accept if the Listbox has a footer, even clicking `esc` on the Listbox.

Therefore I need to store the callback somewhere, that's why I'm storing it on the Context, because there are many moments that need to be executed.

What I'm not understanding is if you are concern about the fact that I'm storing the onChange on the context or by the fact I'm using a useEffect to dispatch an action and store the value then? 

Would then saving the initial onChange when initializing the reducer state would be enough?

nahumzs: Ok after thinking a little more came with another approach, I'm using a context to store the onChange prop, that way we don't need to rely on state at all, and I think @alexzherdev concerns should be address with this approach.

PR #385, opened 2020-01-08
Add data qa tags to confirmation buttons

tristanjasper:
### Purpose 🚀
Results system tests need to target a `data-qa-anchor` for the confirmation buttons. We need to add this qa tag and update confirmation component in visualizer data processing panel.

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/xd-2739-add-data-qa-tags-confirmation-buttons

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/XD-7241
https://aclgrc.atlassian.net/browse/XD-7239


packages/Confirmation/src/Confirmation.js
```diff
@@ -121,6 +121,7 @@ const Confirmation = props => {
               kind={confirmButtonType}
               size={buttonSize}
               onClick={handleOnConfirm}
+              data-qa-anchor="confirmation__confirm-button"
```
alexzherdev: Wouldn't this be `data-pka-anchor` (with the appropriate naming convention)? I think `data-qa-anchor` is something apps use, not something that should be in paprika.

mikrotron:
```suggestion
              data-pka-anchor="confirmation.confirm-button"
```

packages/Confirmation/src/Confirmation.js
```diff
@@ -130,6 +131,7 @@ const Confirmation = props => {
               kind={Button.Kinds.MINOR}
               size={buttonSize}
               onClick={handleCloseConfirm}
+              data-qa-anchor="confirmation__cancel-button"
```
mikrotron:
```suggestion
              data-pka-anchor="confirmation.cancel-button"
```

mikrotron:
```suggestion
              data-pka-anchor="confirmation.cancel-button"
```

packages/Confirmation/src/Confirmation.js
```diff
@@ -104,7 +107,7 @@ const Confirmation = props => {
   const I18n = useI18n();
 
   const popoverContent = (
-    <Popover.Content id={confirmId}>
+    <Popover.Content id={confirmId} data-pka-anchor={popoverContentAnchor}>
```
tristanjasper: Being able to add this custom data anchor allows us to target the confirmation component button even if there are multiple confirmation or popover components open (for whatever reason) as the content is rendered in a portal at the base of the dom

tristanjasper: reverted this change now as there should only be one portal content rendered at any one time, so we should be able to target the default popover content qa selector

PR #392, opened 2020-01-14
PD-13960 fix checkbox states

oscarkwan:
### Purpose 🚀
- fix checkboxStates import, before `import Checkbox, { checkedStates } from '@paprika/checkbox' was not working

### Notes ✏️
- Changed it so it is similar to `Button.Kinds`. So now we use `Checkbox.States` instead of exporting `checkboxStates`

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/PD-13960-fix-checkboxstates

### DIFF
--- a/packages/Checkbox/CHANGELOG.md
+++ b/packages/Checkbox/CHANGELOG.md
@@ -5,4 +5,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [0.2.0] - 2020-01-14
 ### Added
-- Fixed broken checkboxStates import to use Checkbox.States instead [@oscarkwan](https://github.com/oscarkwan).
+- Fixed broken checkboxStates import to use Checkbox.states instead [@oscarkwan](https://github.com/oscarkwan).
--- a/packages/Checkbox/CHANGELOG.md
+++ b/packages/Checkbox/CHANGELOG.md
@@ -4,5 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [0.2.0] - 2020-01-14
+### Removed
+- checkboxStates import which was broken [@oscarkwan](https://github.com/oscarkwan).
+
 ### Added
-- Fixed broken checkboxStates import to use Checkbox.states instead [@oscarkwan](https://github.com/oscarkwan).
+- Checkbox.states instead of checkboxStates [@oscarkwan](https://github.com/oscarkwan).
--- a/packages/Checkbox/CHANGELOG.md
+++ b/packages/Checkbox/CHANGELOG.md
@@ -1,0 +1,8 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [0.2.0] - 2020-01-14
+### Added
+- Fixed broken checkboxStates import to use Checkbox.States instead [@oscarkwan](https://github.com/oscarkwan).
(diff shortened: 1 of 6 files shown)


packages/Checkbox/src/Checkbox.js
```diff
@@ -74,8 +74,9 @@ const Checkbox = props => {
   );
 };
 
+Checkbox.States = checkboxStates;
```
alexzherdev: EnchantingExemplaryHoverfly-size_restricted

nahumzs:
Ok, I think we should settle in a way for declaring "constant" variables, Kind (Button), States(checkbox), are not consistent, what about using `types` ?

`Buttons.types.primary` `Checkbox.types.checked`

Also the fact that has an uppercase looks similar to a subcomponent which can be confusing, in react land.

oscarkwan:
> Ok, I think we should settle in a way for declaring "constant" variables, Kind (Button), States(checkbox), are not consistent, what about using `types` ?
> 
> `Buttons.types.primary` `Checkbox.types.checked`
> 
> Also the fact that has an uppercase looks similar to a subcomponent which can be confusing, in react land.

Hmm.. what about `Button.KINDS` or `Checkbox.STATES` ?

nahumzs:
Agree but then you don't have to learn STATE in Checkbox and KINDS in Button, you will have only to know that all of our components will have their ENUM (which are not technically enums) under types.

Maybe we can even call them enums, `Checked.enum.checked` `Button.enum.primary` I think is more consistent.
if we don't do something like this, each component will expose their constant variables in different ways :(, without consistency.

mikrotron:
What if we have a component that has both `KINDS` and `STATES` to expose? I think I like Oscar's suggestion, but am also fine with lowercase `states`, `kinds`, etc.  The property will be CAPS, like `Checkbox.states.CHECKED`.  

I feel like whether there's an intermediate `.enum.` or not, the consumer needs to RTFM to know what's available, so may as well be shorter.

oscarkwan:
> What if we have a component that has both `KINDS` and `STATES` to expose? I think I like Oscar's suggestion, but am also fine with lowercase `states`, `kinds`, etc. The property will be CAPS, like `Checkbox.states.CHECKED`.
> 
> I feel like whether there's an intermediate `.enum.` or not, the consumer needs to RTFM to know what's available, so may as well be shorter.

Ah right, i forgot its all caps at the end. I think lowercase is fine too 👍 i like that.

packages/Checkbox/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [0.2.0] - 2020-01-14
+### Added
```
mikrotron: The fact that you've `Added` the new `Checkbox.states` doesn't create the break though, it's the fact that you've `Removed` the `checkboxStates` export, right? Both should be included in the changelog, but it's the thing that breaks the component that I'd be most concerned with (followed closely by how to fix it).

nahumzs:
I was just supporting discoverability, if you know that all the constants are exposed to the same property `Component.enums` is quite easy to do `console.log(Component.enums)` and see what its available instead of `console.log(Component)` and figure out where are the constant values for that component. Just my two cents, I will let both of you decide about it.

About having KIND and STATE you just have to do:

`Component.enums = { {...KIND}, {...STATE} };` so you can have whatever you want.

alexzherdev:
IMO the discoverability should come from being able to peruse storybook/DS website, so from that point of view it doesn't matter much that they all share the same name?

oscarkwan:
Can we get a confirmation if we should go with
`Checkbox.states` ?

nahumzs:
not a fan of the decision, but I don't want to be a blocker.

PR #395, opened 2020-01-18
UX-633-DataGrid

nahumzs:
This PR has been split between functional code and [specs, stories, documentation](http://storybooks.highbond-s3.com/paprika/STAR-144-DataGrid-stories-readme-specs/?path=/docs/datagrid-docs--page) to make it easier to review. 

# DataGrid

⚠️ **Most important files ** ⚠️
If you have limited time please focus on reviewing the following file 🙏

- DataGrid.js `+628`
- useGridEventHandler.js `+458`

Thank you
---

It is a controlled component that allows consumers to have more flexibility rendering a vast amount of tabular data.

- [x] Uses [react-window](https://react-window.now.sh/) for creating a virtualized table
- [x] Options to have sticky header and columns
- [x] Ability to have custom cells by using a render prop
- [x] Allows for load more or infinite scroll design patterns
- [x] Support for screenreader
- [x] Support for keyboard navigation ↑, ↓, →, ←
- [x] API for keyboard events Enter, Space, Shift + Space

Pretty much this is a wrapper on top of react-windows that enhance it and add our `specific` functionality on it. 

## ⭐️ Documentation

`Please read` the documentation before reviewing the PR:

http://storybooks.highbond-s3.com/paprika/STAR-144-DataGrid-stories-readme-specs/?path=/docs/datagrid-docs--page

## 👩‍🔬 Technical notes
- The API was inspired and enhanced by other Grid libraries on the community.
- When implementing ↑, ↓, →, ← I tried first to use `useState` to keep track of the active cell, this turns out to be not a good fit from a performance perspective. 
React was re-rendering the DataGrid each time a new active cell was getting active making it hard to handle onKeyDown permanently without adding a debounce function. To improve it, I started using refs, a bit of vanilla javascript and `data` attributes, that way React would not have a need to react to anything. 
With this approach, React only gets notified when it need it like when a onScroll event happens and triggers react-window. 
The downside for this approach is that now I have to run `restoreHighlightFocus()` every time the table gets rerendered. https://github.com/acl-services/paprika/pull/395/files#diff-ca1f808a0997c8eee3f4afebca8e0e5cR331 which isn't a big of the deal and is transparent for the user and consumer.

- Also to improve some performance while navigating with the ↑, ↓, →, ← keys, I created an implicit API for the CELL component, this API exposes methods that trigger its internal state allowing me to rerender only the CELL that I want to and prevents the DataGrid from rerendering, this is useful for example to add specific style when the CELL is active. 

- `Sticky columns` is done by rendering another four `react-window` Grids around the main grid, this is a similar technique use on the Grid example for the react-virtualized library, and it is not supported nor will by react-window. https://github.com/bvaughn/react-window/issues/69#issuecomment-464189400

So DataGrid.js file has these four Grids. 

| Grid1StickyColumnHeader     |        Grid2StickyHeader      |
| Grid3StickyColumns               |        MainGridWithData       |

- To sync all the scrolling I'm using the react-windows onScroll event, which measn that both `Grid3StickyColumns` and `MainGridWithData` have to get sync between them depending which one is being scrolled. The downside of this approach is that we are getting some visual sync issue on DataGrids with a lot of Data, but I think eventually we can find a way to mitigate it. 

- On IE11 right now there is a scrollbar next to the StickyColumns (`Grid3StickyColumns`), on chrome/firefox/safari I was able to use `::-webkit-scrollbar` and `::-webkit-scrollbar-track-piece` to hide the scrollbar for `Grid3StickyColumns`

- The vertical and horizontal overflow (overflow-x, overflow-y) are always set to `overflow:scroll` to reduce complexity while calculating the ↑, ↓, →, ← navigation.

- Voice over is achieved by focussing on a visually-hidden div every time that the user uses ↑, ↓, →, ← or click on the cell. 

- Infinity Scrolling reacts to the number of rows before reaching the end of the scrollbar.

- Load More reacts when the user scrolls at the bottom of the table. (show / hide)

### Updates 📦
- [x] `NEW`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-633-DataGrid

### Screenshots 📸

Screen Shot 2020-02-14 at 11 50 27 AM
Screen Shot 2020-02-14 at 11 50 37 AM
Screen Shot 2020-02-14 at 11 50 45 AM
Screen Shot 2020-02-14 at 11 51 02 AM


packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+  }, [data.length, pageSize]);
+
+  React.useEffect(() => {
+    refPrevLastScrollHeight.current = refScrollGrid.current.scrollHeight;
+  }, []);
+
+  React.useEffect(() => {
+    // this is required to readjust the active highlight
+    // after any rerender
+    if (refPrevLastScrollHeight.current && refPrevLastScrollHeight.current < refScrollGrid.current.scrollHeight) {
```
nahumzs: When adding new rows to the table like you would do when clicking LoadMore, the onScroll event will not get notified, so to trick it I'm just tracking the previous height and if I notices has changed I move the scrollbar by 1px which triggers the onScroll event.

package.json
```diff
@@ -70,6 +70,7 @@
     "eslint-plugin-jsx-a11y": "^6.1.1",
     "eslint-plugin-react": "^7.11.1",
     "eslint-plugin-react-hooks": "^1.6.0",
+    "faker": "^4.1.0",
```
jamiek-acl: Where is this used?

package.json
```diff
@@ -95,7 +96,8 @@
     "style-loader": "^0.23.1",
     "styled-components": "^4.2.0",
     "verdaccio": "^3.11.6",
-    "wait-on": "^3.2.0"
+    "wait-on": "^3.2.0",
+    "workerize-loader": "^1.1.0"
```
jamiek-acl: and where is _this_ used?

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+import Cell from "./components/Cell";
+import useGridEventHandler from "./hooks/useGridEventHandler";
+import ColumnDefinition from "./components/ColumnDefinition";
+import * as styled from "./DataGrid.styles";
+import WhenScrollBarReachedBottom, { End } from "./components/WhenScrollBarReachedBottom";
+import InfinityScroll from "./components/InfinityScroll";
+
+const propTypes = {
+  children: PropTypes.node.isRequired,
+  data: PropTypes.arrayOf(PropTypes.shape({})),
```
jamiek-acl: No shape defined? So just an array of `object`s?

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+  onKeyDown: PropTypes.func,
+  onRowChecked: PropTypes.func,
+  onShiftSpaceBar: PropTypes.func,
+  onSpaceBar: PropTypes.func,
+  rowHeight: PropTypes.number,
+  width: PropTypes.number,
+};
+const defaultProps = {
+  data: [],
+  hasNoRecords: false,
```
jamiek-acl: I'd prefer `hasRecords`

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+  height: 600,
+  isIdle: false,
+  onClick: null,
+  onEnter: null,
+  onKeyDown: () => {},
+  onRowChecked: () => {},
+  onShiftSpaceBar: null,
+  onSpaceBar: null,
```
jamiek-acl: Our standard is `onPressXxx` not `onXxx` (`onEnter`, too)

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+    });
+  }, [ColumnDefinitions]);
+
+  const calculatedTableHeight = React.useMemo(() => {
+    return rowHeight * rowCount > height ? height : rowHeight * rowCount + scrollBarWidth;
+  }, [height, rowCount, rowHeight, scrollBarWidth]);
+
+  const calculatedTableWidth = React.useCallback(() => {
+    let width = 0;
+    ColumnDefinitions.forEach(ColumnDefinition => {
```
jamiek-acl: Maybe`.reduce`? If so, you could also do it in `stickyGridWidth`

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+  }, []);
+
+  const onChangeActiveCell = React.useCallback(({ columnIndex, rowIndex }) => {
+    const key = `${columnIndex}${rowIndex}`;
+
+    if (refPrevActiveCell.current && refPrevActiveCell.current in refsCell.current.keys) {
+      const prevCell = refsCell.current.keys[refPrevActiveCell.current];
+      if (prevCell) {
+        refsCell.current.keys[refPrevActiveCell.current].isActiveCell(false);
+      }
```
jamiek-acl: After declaring prevCell, I think you meant to use it:

    if (prevCell) {
      prevCell.isActivecell(false);
    }

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+    refContainer,
+    refGrid,
+    rowCount,
+    rowHeight,
+    scrollBarWidth,
+    stickyColumnsIndexes,
+  });
+
+  const a11yTextMessage = React.useCallback((value, column, rowIndex) => {
+    return `${value}. You are on row ${rowIndex}, column ${column}. Disregard the following information:`;
```
jamiek-acl: L10n

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+  }, []);
+
+  const handleScroll = React.useCallback(parameters => {
+    const { scrollLeft, scrollTop /* scrollUpdateWasRequested */ } = parameters;
```
jamiek-acl: remove comment (and the one on line 258)

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+
+    if (refScrollHeader.current) {
+      refScrollHeader.current.scrollTo({ left: scrollLeft, top: 0 });
+    }
+
+    // prevent rescrolling when this scrollbar gets sync with the one in the sticky column
+    if (refScrollHappenedBy.current === null) {
```
jamiek-acl: If it were me, i'd reverse the logic of this `if` block. Same with the one on line 261.

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+      refContainer.current && refContainer.current.querySelector(`.grid-${gridId} [role="row"]`).parentElement;
+
+    if (!scrollContainer) return;
+    // https://stackoverflow.com/questions/21064101/understanding-offsetwidth-clientwidth-scrollwidth-and-height-respectively
+    // https://davidwalsh.name/detect-scrollbar-width
+    setScrollBarWidth(() => scrollContainer.offsetWidth - scrollContainer.clientWidth);
+  }, [gridId, isIdle]);
+
+  function handleFocusGrid() {
+    const $isBlur = refContainer.current.querySelector(".grid--is-blur");
```
jamiek-acl: Not sure if it matters, but proper grammar is `is blurred` :)

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+    if ($isBlur) $isBlur.classList.remove("grid--is-blur");
+
+    if (gridShouldHaveFocus) {
+      setGridShouldHaveFocus(() => {
+        return false;
+      });
```
jamiek-acl: `() => false`

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+  const handleMouseLeave = React.useCallback(() => {
+    deemphasizeRow();
+  }, [deemphasizeRow]);
+
+  return (
+    <>
+      {hasNoRecords && (
+        <styled.BlockerItem gridId={gridId} $width={gridWidth} $height={height}>
+          <styled.Blocker $width={gridWidth} $height={height}>
+            <div>
```
jamiek-acl: Need this `div`?

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+              <Bubbly width={120} height={120} />
+              <p>No records found</p>
```
jamiek-acl: L10n

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+          </Grid>
+
+          {/** COLUMS */}
+
+          <Grid
+            className={`grid-${gridId}`}
+            columnCount={columnCount}
+            columnWidth={columnIndex => {
+              if (stickyColumnsIndexes.includes(columnIndex)) return 0;
+              return ColumnDefinitions[columnIndex].props.width;
```
jamiek-acl: Is one line nicer?

    return stickyColumnsIndexes.includes(columnIndex) ? 0 : ColumnDefinitions[columnIndex].props.width;

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+                  gridId={gridId}
+                  ref={handleRefCell({ columnIndex, rowIndex })}
+                  rowIndex={rowIndex}
+                  style={style}
+                />
+              );
+            }}
+          </Grid>
+        </styled.Flex>
+        <styled.FillerTopRigth rowHeight={rowHeight} scrollBarWidth={scrollBarWidth} />
```
jamiek-acl: Typo `FillerTopRight`

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+          <styled.Footer $width={gridWidth}>
+            <styled.RowCount>
+              Rows:{rowCount} Columns:{columnCount}
+            </styled.RowCount>
+          </styled.Footer>
+          {WhenScrollBarReachedBottom ? (
+            <End width={gridWidth} ref={refEnd}>
+              {WhenScrollBarReachedBottom}
+            </End>
+          ) : null}
```
jamiek-acl: Could get rid of these two `nulls` by doing `{whatever &&...` instead of `{whatever ?...`

packages/DataGrid/src/DataGrid.js
alexzherdev: I'm wondering why this prop is necessary at all, is it not equivalent to `data.length === 0`?
(otherwise I prefer `false` for default boolean values per discussion in https://github.com/acl-services/paprika/pull/399#pullrequestreview-347563964 and maybe this can be called `isEmpty` instead)

packages/DataGrid/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@paprika/helpers": "^0.2.7",
+    "@paprika/icon": "^0.2.8",
+    "@paprika/raw-button": "^0.2.5",
+    "@paprika/sidepanel": "^0.3.2",
+    "@paprika/spinner": "^0.2.9",
+    "@paprika/stylers": "^0.2.2",
+    "@paprika/tokens": "^0.1.9",
+    "nanoid": "^2.1.9",
+    "prop-types": "^15.7.2",
+    "react-virtualized-auto-sizer": "^1.0.2",
```
alexzherdev: not used

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  }}
+`;
+
+export const Cell = styled.div`
+  background: ${tokens.color.white};
+  border-bottom: 1px solid ${tokens.border.color};
+  border-left: 1px solid ${tokens.border.color};
+  box-sizing: border-box;
+  display: flex;
+  font-size: 16px;
```
alexzherdev: no token?

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
...
+};
+
+const outerElementType = React.forwardRef((props, ref) => <div role="rowgroup" ref={ref} {...props} />);
+const innerElementType = React.forwardRef((props, ref) => <styled.InnerElementType role="row" ref={ref} {...props} />);
+const outerElementTypeMainGrid = React.forwardRef((props, ref) => (
+  <styled.OuterElementTypeMainGrid role="rowgroup" ref={ref} {...props} />
+));
+
+const innerElementTypeMainGrid = React.forwardRef((props, ref) => (
+  <styled.InnerElementTypeMainGrid className="inner-element-type-main-grid" role="row" ref={ref} {...props} />
```
alexzherdev: className not used?

packages/DataGrid/src/DataGrid.js
nahumzs: Realizing I picked a misleading name for this prop, what actually does this props is to show a screen with bubbly saying no record found, similar to the screen `isIdle`. 

Maybe I could splitted this into different subcomponent like:

```js
<DataTable>
  <DataTable.NoRecordFound isVisible={data.length} />
  <DataTable.IsIdle isVisible={isIdle} />
</DataTable>
```

packages/DataGrid/src/DataGrid.js
nahumzs: `onPressShifSpaceBar` `onPressSpaceBar` 👍

packages/DataGrid/src/DataGrid.js
nahumzs: I actually implemented this with a reducer but I think is way simpler to read with forEach. https://youtu.be/qaGjS7-qWzg

.storybook/webpack.config.js
```diff
@@ -15,6 +15,9 @@ module.exports = {
       },
     ],
   },
+  output: {
+    globalObject: "self",
```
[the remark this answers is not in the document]
nahumzs: I'm using it on this story https://github.com/acl-services/paprika/pull/416/files#diff-985fed2a0717e1a1105479daf5a5bc53R1

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,628 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { VariableSizeGrid as Grid } from "react-window";
```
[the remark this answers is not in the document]
nahumzs: I picked that one at first, but you can't assign variable **width**
[columnWidth: number](https://react-window.now.sh/#/api/FixedSizeGrid)

Like on VariableSizeGrid [columnWidth: (index: number) => number
](https://react-window.now.sh/#/api/VariableSizeGrid)

package.json
nahumzs: https://github.com/acl-services/paprika/pull/416 it is use on the stories for the table

package.json
nahumzs: https://github.com/acl-services/paprika/pull/416  This PR is using it. (all stories, fixtures, etc)

packages/DataGrid/src/DataGrid.js
nahumzs: yeah, is the beauty the table is data agnostic the only thing that needs is an array of objects, where the objects are the rows and the attributes on those objects the columns

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
+import styled, { css } from "styled-components";
+import tokens from "@paprika/tokens";
+
+const experimentalGrey = "#e7e7e7"; /* experimental */
```
oscarkwan: Is this still experimental? If we are ready to merge / ready to review, then its probably not experimental no more?

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+export const Grid = styled.div.attrs(({ $width, isVisible }) => {
+  const _isVisible = isVisible ? { opacity: 0 } : { opacity: 100 };
```
oscarkwan: opacity should be `1` i don't think 100 is valid

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+})`
+  * {
+    box-sizing: border-box;
+  }
+  .grid--is-active {
+    box-shadow: ${tokens.highlight.active.withBorder.insetBoxShadow};
+  }
+
+  .grid--is-blur {
+    box-shadow: 0 0 0;
```
oscarkwan: `box-shadow: none`

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  [role="gridcell"] {
+    /* this hide the aria "voice over" focus message */
+    &:focus {
+      outline: 0;
+    }
+  }
+
+  border-top: 1px solid ${tokens.border.color};
+  background: {tokens.color.white};
+
```
oscarkwan:
```suggestion
```

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+      }
+
+      .${gridId}-header {
+        overflow: hidden !important;
+      }
+
+      .${gridId}-sticky-columns {
+        /*
+          The arrow navigatoion is border sensible, please avoid adding a border to this styled components
+          doing it will create a incorrect navigation while using the arrow keys.
```
oscarkwan: The arrow navigation is sensitive with borders. If a border is used, it will cause issues keyboard navigation.

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+        */
+        box-shadow: 3px 0px 0px 0px ${tokens.border.color};
+        overflow-x: hidden !important;
+        /* overflow: hidden !important; */
```
oscarkwan: ✂️

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+        z-index: 1;
+        ::-webkit-scrollbar {
+          width: 0px;
+        }
+        ::-webkit-scrollbar-track-piece {
+          -webkit-border-radius: 6px;
```
oscarkwan: token can be used

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  box-sizing: border-box;
+  display: flex;
+  font-size: 16px;
+  justify-content: flex-start;
+  line-height: 1;
+  position: relative;
+
+  &:focus {
+    outline: 1px solid transparent;
+    /* outline: 0; */
```
oscarkwan: ✂️

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  ${({ hasActiveRowShadow }) => {
+    return hasActiveRowShadow
+      ? css`
+          /* box-shadow: ${tokens.highlight.active.withBorder.insetBoxShadow}; */
+          &:after {
+
```
oscarkwan:
```suggestion
```

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  &:focus {
+    outline: 1px solid transparent;
+    /* outline: 0; */
+  }
+
+  ${({ hasActiveRowShadow }) => {
+    return hasActiveRowShadow
+      ? css`
+          /* box-shadow: ${tokens.highlight.active.withBorder.insetBoxShadow}; */
```
oscarkwan: ✂️

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+
+export const CellHeader = styled(Cell)`
+  background: ${experimentalGrey};
+  border-bottom: 1px solid ${tokens.border.color};
+  border-left: 1px solid ${tokens.border.color};
+  color: ${tokens.color.black};
+  display: block;
+  font-weight: 600;
+  overflow: hidden;
+  padding: 8px;
```
oscarkwan: token

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  width: 100%;
+`;
+
+export const InnerCell = styled.div`
+  display: block;
+  height: 100%;
+  overflow: hidden;
+  padding: 8px;
```
oscarkwan: token

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  width: 100%;
+`;
+
+export const InnerElementType = styled.div``;
+
+export const InnerElementTypeMainGrid = styled.div``;
+
+export const OuterElementTypeMainGrid = styled.div`
+  /*
+   The arrow navigatoion is border sensible, please avoid adding a border to this styled components
```
oscarkwan: The arrow navigation is sensitive with borders. If a border is used, it will cause issues keyboard navigation.

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  height: 1px;
+  left: 0;
+  overflow: hidden;
+  position: absolute;
+  top: 0;
+  whitespace: nowrap;
+  width: 1px;
+`;
+
+export const FillerTopRigth = styled.div`
```
oscarkwan: `FillerTopRigth` typo, -> `FillerTopRight`

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+    return `
+      ${displayBlock}
+      width: ${scrollBarWidth}px;
+      height: ${rowHeight}px;
+    `;
+  }}
+`;
+
+export const FillerBottomLeft = styled.div`
+  /* this is a small square filler on the right top corner of the table */
```
oscarkwan: Is this comment accurate? it says right top corner, but the styled component says bottom left..

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+  }}
+`;
+
+export const Flex = styled.div`
+  display: flex;
+`;
+
+export const RowCount = styled.div`
+  font-color: ${tokens.color.blackLighten40};
+  font-size: 13px;
```
oscarkwan: token

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+      height: ${$height}px;
+    `;
+  }}
+`;
+
+export const Footer = styled.div`
+  border: 1px solid ${tokens.border.color};
+  box-sizing: border-box;
+  font-size: 12px;
+  padding: 4px;
```
oscarkwan: tokens

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,258 @@
...
+    }
+  `;
+  }}
+`;
+
+export const Blocker = styled.div`
+  align-items: center;
+  background: transparent;
+  border: 1px solid ${tokens.border.color};
+  display: Flex;
```
oscarkwan: `display: flex`

packages/DataGrid/src/components/ColumnDefinition/ColumnDefinition.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  if ((propName === "cell" || propName === "header") && typeof props[propName] === "function") {
+    const nameOfRenderProp = propName === "cell" ? "cellA11yText" : "headerA11yText";
+
+    if (typeof props[nameOfRenderProp] === "undefined")
+      return new Error(`Accessibility is important 😇: \n
+        When using a render function as a value for the ${
+          props[propName]
+        } prop, you must provide an ${nameOfRenderProp} prop describing the cell content. Look on the ColumnDefinition with the following props: \n ${JSON.stringify(
```
oscarkwan:
```suggestion
        } prop, you must provide an ${nameOfRenderProp} prop describing the cell content. Look at the ColumnDefinition with the following props: \n ${JSON.stringify(
```

packages/DataGrid/src/components/ColumnDefinition/ColumnDefinition.js
```diff
@@ -0,0 +1,71 @@
...
+        props,
+        null,
+        2
+      )}.
+      `);
+
+    if (typeof props[nameOfRenderProp] !== "function")
+      return new Error(`Accessibility is important 😇: \n
+        The prop ${nameOfRenderProp} has to be a function returning a string. Look on the ColumnDefinition with the following props: \n ${JSON.stringify(
```
oscarkwan:
```suggestion
        The prop ${nameOfRenderProp} has to be a function returning a string. Look at the ColumnDefinition with the following props: \n ${JSON.stringify(
```

packages/DataGrid/src/components/RowIndicator/RowIndicator.styles.js
```diff
@@ -0,0 +1,48 @@
...
+  justify-content: center;
+  padding: 0;
+  width: 100%;
+`;
+
+export const Void = styled.div`
+  border: 1px solid ${tokens.color.blackLighten50};
+  border-radius: 50%;
+  height: 6px;
+  width: 6px;
```
oscarkwan: tokens

packages/DataGrid/src/components/RowIndicator/RowIndicator.styles.js
```diff
@@ -0,0 +1,48 @@
...
+export const RowIndexText = styled.div.attrs(({ hasFourDigitsOrMore }) => {
+  const style = hasFourDigitsOrMore
+    ? {
+        fontSize: "11px",
+        padding: "4px",
+      }
+    : {
+        fontSize: "13px",
+        padding: "6px",
+      };
```
oscarkwan: tokens

packages/DataGrid/src/components/columns/renderColumnExpand.js
```diff
@@ -0,0 +1,82 @@
...
+      {...options}
+      onClick={onClick}
+      width={26}
+      cellProps={() => ({
+        style: {
+          padding: 0,
+          display: "flex",
+          justifyContent: "center",
+          alignItems: "center",
+          boxShadow: "-1px 0px 0px 0px rgba(255,255,255,1)",
```
oscarkwan: `rgba(255,255,255,1)` same as `rgb(255,255,255)` why do you need `rgba`?

packages/DataGrid/src/DataGrid.js
nahumzs: won't do.

packages/DataGrid/src/DataGrid.js
nahumzs: yes, will let flex to justify and align the content. `(center)`

packages/DataGrid/src/DataGrid.styles.js
nahumzs: yes its, I updated the comment

packages/DataGrid/src/components/columns/renderColumnExpand.js
nahumzs: I always set the alpha, no difference. I can remove it :) or even use `white`

oscarkwan: i see, white is good then :D

packages/DataGrid/src/components/RowIndicator/RowIndicator.styles.js
nahumzs: I'm not 100% sure about the exact size for this, I think we will have a second-round for polishing the UI.

packages/DataGrid/src/components/RowIndicator/RowIndicator.styles.js
nahumzs: same here I don't think we have 6px and 13px and not sure 100% the UI for this part, we will polish it later, maybe whoever is working on the feature needing this can help. or myself in my next round

packages/DataGrid/src/DataGrid.styles.js
nahumzs: I think it is still experimental, there are some new addition still happening on the DataGrid we will need a polish UI part for this in another PR

alexzherdev:
Found a bug with the checkboxes where I can make one checkbox "stick".
dg

alexzherdev:
another issue, not sure if known. Here I scrolled down so that the focused cell is higher than the overscan threshold. Then nothing is happening, but I'm actually pressing the down key trying to get the selection to come down into the screen but it never does.
dg_scroll

packages/DataGrid/src/hooks/useGridEventHandler.js
```diff
@@ -0,0 +1,458 @@
...
+  const [gridId] = React.useState(() => `PKA${nanoid()}`);
+  const refContainerBoundClientRect = React.useRef(null);
+  const refScroll = React.useRef(null);
+  const refPrevCell = React.useRef(null);
+  const refHasHorizontalScrollBar = React.useRef(false);
+
+  const cell = React.useRef(null);
+  const prevCell = React.useRef(null);
+
+  const $getCell = React.useCallback(() => {
```
alexzherdev: Same question on the necessity of `useCallback`s.

packages/DataGrid/src/hooks/useGridEventHandler.js
```diff
@@ -0,0 +1,458 @@
...
+  onRowChecked,
+  onPressShiftSpaceBar,
+  onPressSpaceBar,
+  refContainer,
+  refGrid,
+  rowCount,
+  scrollBarWidth,
+  stickyColumnsIndexes,
+}) {
+  const [gridId] = React.useState(() => `PKA${nanoid()}`);
```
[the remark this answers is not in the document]
nahumzs: I had a combo with the guys of nanoId about this https://github.com/ai/nanoid/pull/159#issuecomment-565740494 settled on using state instead of ref.

packages/DataGrid/src/hooks/useGridEventHandler.js
```diff
@@ -0,0 +1,458 @@
...
+        $setRefs(columnIndex);
+        scroll(nextColumnIndex, 0);
+        if (rowIndex + 1 === rowCount) {
+          scrollToTheBottom();
+        }
+      }
+    },
+    // keydown will not trigger spacebar, keyup will, but we want to prevent scrolling.
+    " ": () => {},
+    // we move f (checked row) because this actions doesn't open a sidepanel and we want it
```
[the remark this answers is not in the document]
nahumzs: `f` is standard for selecting row accordingly the `a11y` team, about how to effectively communicate this to the user is something that the UX team it is still discussing.

packages/Icon/src/BubblySearchNoResults.js
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+
+function SvgBubblySearchNoResults({ title, titleId, ...props }) {
```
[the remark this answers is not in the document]
nahumzs: Yeah... I think this was published when we did the other one for the `<ActionBar />` which is the package for the filter part for the DataGrid. We should remove it, we should be agnostic of what to render, and the end-user should decide if they want to use bubbly or not.

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
```diff
@@ -0,0 +1,83 @@
...
+      onMouseLeave={handleMouseLeave}
+    >
+      {isActive || isActiveCell || isActiveRow || isValueNotUnchecked(isChecked({ rowIndex })) || !hasIndexIndicator ? (
+        <styled.Checkbox>
+          <CheckBox
+            tabIndex="-1"
+            size="small"
+            checkedState={isChecked({ rowIndex })}
+            onClick={handleCheck}
+            onChange={() => {}}
```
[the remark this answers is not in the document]
nahumzs: Yeah, this is part of the checkbox :/

packages/DataGrid/src/hooks/useGridEventHandler.js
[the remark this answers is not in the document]
alexzherdev: Oh that's enlightening. The part that `nanoid()` is not pure makes sense but is not immediately obvious. TIL 🥇

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,220 @@
...
+  background: transparent;
+  border: 1px solid ${tokens.border.color};
+  display: flex;
+  justify-content: center;
+  left: 0;
+  position: absolute;
+  top: 0;
+  z-index: 4;
+
+  ${({ $height, $width }) => {
```
tristanjasper: we should have a helper function for this scenario

packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -0,0 +1,220 @@
...
+      width: ${$width}px;
+      height: ${$height}px;
+    `;
+  }}
+`;
+
+export const Footer = styled.div`
+  border: 1px solid ${tokens.border.color};
+  box-sizing: border-box;
+  font-size: 12px;
```
tristanjasper: Are we not using `fontSize` helper?

packages/DataGrid/src/components/Cell/Cell.js
```diff
@@ -0,0 +1,72 @@
...
+
+const propTypes = {
+  a11yText: PropTypes.string.isRequired,
+  column: PropTypes.shape({ cellProps: PropTypes.func, cell: PropTypes.oneOfType([PropTypes.string, PropTypes.func]) })
+    .isRequired,
+  columnIndex: PropTypes.number.isRequired,
+  data: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
+  gridId: PropTypes.string.isRequired,
+  rowIndex: PropTypes.number.isRequired,
+  style: PropTypes.shape({}).isRequired,
```
tristanjasper: is this shape correct? Shouldn't it be more specific?

packages/DataGrid/src/components/Cell/Cell.js
```diff
@@ -0,0 +1,72 @@
...
+        return Number.parseInt(_rowIndex, 10) === rowIndex;
+      });
+    },
+    deemphasizeOnrow: _rowIndex => {
+      setIsRowActive(() => {
+        return !Number.parseInt(_rowIndex, 10) === rowIndex;
+      });
+    },
+    getIndexes: () => {
+      return { rowIndex, columnIndex };
```
tristanjasper: nit: do these return statements all need to be on their own line?

packages/DataGrid/src/components/Cell/Cell.js
```diff
@@ -0,0 +1,72 @@
...
+    <styled.Cell ref={ref} tabIndex={-1} style={{ ...style }} data-cell={`${gridId}.${columnIndex}.${rowIndex}`}>
+      <styled.GridCell role="gridcell">{a11yText}</styled.GridCell>
+      <styled.InnerCell
+        hasActiveRowShadow={isActiveRow}
+        {...cellProps}
+        aria-hidden="true"
+        {...options.attrs}
+        className={`${gridId}_cell`}
+      >
+        {typeof column.cell === "function" ? column.cell(options) : data[rowIndex][column.cell]}
```
tristanjasper: could use a helper function for the type check as it's being done in multiple places

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
```diff
@@ -0,0 +1,83 @@
...
+const defaultProps = {
+  hasIndexIndicator: true,
+  hasNumber: false,
+  isActiveCell: false,
+  isActiveRow: false,
+  row: null,
+  rowIndex: null,
+  onCheck: () => {},
+  isChecked: () => {
+    return "unchecked";
```
tristanjasper: can be on returned on one line

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
```diff
@@ -0,0 +1,83 @@
...
+  },
+};
+
+function isValueNotUnchecked(value) {
+  return value !== "unchecked";
```
tristanjasper: Can "unchecked" be mapped to a variable instead of comparing to the primitive string?

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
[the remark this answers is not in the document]
tristanjasper: I switched the handler to be on the `onChange` callback and it seems to work, still not completely understanding why it needs to be on an `onClick` callback?

packages/DataGrid/src/components/WhenScrollBarReachedBottom/WhenScrollBarReachedBottom.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default function WhenScrollBarReachedBottom(props) {
```
alexzherdev: I'd love for someone with better English than me to come up with better names for these two components. `Basement`? 😄 Just `Bottom`? `Footer`?
Then `End` would be e.g. `FooterWrapper`.

packages/DataGrid/src/components/Cell/Cell.js
nahumzs: I prefer the return statement because it's easier to either put a console.log or a debugger statement if you use the short version would force you to modify the code each time you want to add it. even 1 line, and in the end I think this is just preference which doesn't interfere with the code itself.

packages/DataGrid/src/components/Cell/Cell.js
nahumzs: Can be any property of the DOM style object in css, I guess it's correct.

packages/DataGrid/src/DataGrid.styles.js
nahumzs: Yes, but I think we don't have 12 is something like stylers.fontSize(-2) isn't? which I don't think is that descriptive. Maybe we need to review the tokens itself for the fontsize?

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,622 @@
...
+      refContainer.current && refContainer.current.querySelector(`.grid-${gridId} [role="row"]`).parentElement;
+
+    if (!scrollContainer) return;
+    // https://stackoverflow.com/questions/21064101/understanding-offsetwidth-clientwidth-scrollwidth-and-height-respectively
+    // https://davidwalsh.name/detect-scrollbar-width
+    setScrollBarWidth(() => scrollContainer.offsetWidth - scrollContainer.clientWidth);
+  }, [gridId, isIdle]);
+
+  function handleFocusGrid() {
+    const $isBlurred = refContainer.current.querySelector(".grid--is-blurred");
```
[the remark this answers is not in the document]
nahumzs: well is not really arbitrary, is returning a dom element which tends to be prefixed with a `$` to communicate their meaning. 
I guess on react that is not too common since normally you shouldn't need to deal with bare Dom elements.

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,622 @@
...
+        onBlur={handleBlurGrid}
+        onFocus={handleFocusGrid}
+        onKeyDown={handleKeyDownGrid}
+        onKeyUp={handleKeyUpGrid}
+        onMouseUp={handleMouseUpGrid}
+        onMouseOver={handleMouseOver}
+        onMouseLeave={handleMouseLeave}
+        ref={refContainer}
+        role="grid"
+        tabIndex={gridShouldHaveFocus ? 0 : -1}
```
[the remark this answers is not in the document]
nahumzs: That it's the intention to grant tab access to the `DataGrid`. if there is not an active cell the DataGrid should be able to be `"tabable"`. scenario:
```js
<Input  />
<ListBox />
<DataGrid />
```

user *tab* over the input -> *tab* listbox -> *tab* datagrid `(gridShouldHaveFocus ?)` if true the DataGrid as a component can receive focus **if not**, the focus will be redirected to the current active cell. 

Right now only make the DataGrid `tabable`, next step it's to assign an active cell if there is none one active.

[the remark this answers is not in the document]
nahumzs: I'm going to mark this as resolved for now, but please let me know if you still want me to change the name variable.

packages/DataGrid/src/components/columns/renderColumnExpand.js
```diff
@@ -0,0 +1,82 @@
...
+        <Maximize />
+      </div>
+    </div>
+  );
+}
+
+Expand.propTypes = propTypes;
+Expand.defaultProps = defaultProps;
+
+export default function renderColumnExpand(options = {}) {
```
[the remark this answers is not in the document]
nahumzs: The problem I'm facing here is to create reusable columns, I tried to encapsulate this as a component but the problem of doing that, its that will end creating a new component and making impossible for `DataGrid.js` to access the **props** on ColumnDefinition, 
Returning a component instead of instancing it will give me access to all props on Column definition that I can use on DataGrid file like `Column.props.cell`, `Column.props.width` etc. 

To me, this is similar when you create a `renderSomething` function inside your code.

If you can think of a better idea about how to achieve this, I can modify it.

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
```diff
@@ -0,0 +1,83 @@
...
+export default function RowIndicator(props) {
+  const [isActive, setIsActive] = React.useState(false);
+  const { rowIndex, isChecked, hasIndexIndicator, onCheck, isActiveCell, isActiveRow, hasNumber } = props;
+
+  function handleCheck() {
+    onCheck({ rowIndex });
+  }
+
+  function handleMouseOver() {
+    setIsActive(() => true);
```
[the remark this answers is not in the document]
nahumzs: just consistency, no oppose using `true / false` on booleans. Change it to a simple form.

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,595 @@
...
+const innerElementType = React.forwardRef((props, ref) => <sc.InnerElementType role="row" ref={ref} {...props} />);
+const outerElementTypeMainGrid = React.forwardRef((props, ref) => (
+  <sc.OuterElementTypeMainGrid role="rowgroup" ref={ref} {...props} />
+));
+
+const innerElementTypeMainGrid = React.forwardRef((props, ref) => (
+  <sc.InnerElementTypeMainGrid role="row" ref={ref} {...props} />
+));
+
+const DataGrid = React.forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
nahumzs: agree I will do it after we merge this pr, same goes for the `useGridEventHandler.js`

packages/DataGrid/src/hooks/useGridEventHandler.js
```diff
@@ -0,0 +1,458 @@
...
+  return $cell && $cell.dataset.cell;
+};
+
+export function timeDiff(t1, t2) {
+  if (isDevEnv()) {
+    console.log(`%c diff: ${t2 - t1}`, "color: yellow; background: #333;");
+  }
+}
+
+export default function useGridEventHandler({
```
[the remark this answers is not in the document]
nahumzs: will do on my next PR. with DataGrid.js file.

packages/DataGrid/src/DataGrid.styles.js
nahumzs: seems to me too much abstraction for something simple as return an object with width and height.

packages/DataGrid/src/DataGrid.styles.js
nahumzs: actually is a custom value for now. will keep 12px.

packages/DataGrid/src/components/WhenScrollBarReachedBottom/WhenScrollBarReachedBottom.js
nahumzs: I ended using Basement, I think is flexible enough. and oscar was 👍 as well...

packages/DataGrid/src/DataGrid.js
```diff
@@ -0,0 +1,595 @@
...
+    let width = 0;
+    if (stickyColumnsIndexes.length) {
+      stickyColumnsIndexes.forEach(index => {
+        width += ColumnDefinitions[index].props.width;
+      });
+    }
+    return width;
+  }, [ColumnDefinitions, stickyColumnsIndexes]);
+
+  const highlightRow = React.useCallback(({ rowIndex = null }) => {
```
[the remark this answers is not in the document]
nahumzs: will do in future PR

PR #397, opened 2020-01-21
PD-14210 add isPortal prop to popover

oscarkwan:
### Purpose 🚀
- add `isPortal` prop that allows consumer to render popover inline with the DOM or as a portal

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/PD-14210-add-isportal-prop

### DIFF
--- a/packages/Popover/src/components/Content/Content.js
+++ b/packages/Popover/src/components/Content/Content.js
@@ -88,31 +88,8 @@ const Content = React.forwardRef((props, ref) => {
     contentStyles.width = content.width;
   }
 
-  if (isPortal) {
-    /* eslint-disable jsx-a11y/mouse-events-have-key-events */
-    return ReactDOM.createPortal(
-      <ContentStyled
-        aria-hidden={!isOpen}
-        data-component-name="PopoverContent"
-        data-pka-anchor="popover.content"
-        ref={handleRef}
-        id={isEager ? content.ariaId : null}
-        isOpen={isOpen}
-        onBlur={handleBlur}
-        onMouseOut={handleMouseEvent}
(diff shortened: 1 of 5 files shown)


packages/Popover/src/components/Content/Content.js
```diff
@@ -87,8 +88,31 @@ const Content = React.forwardRef((props, ref) => {
     contentStyles.width = content.width;
   }
 
-  /* eslint-disable jsx-a11y/mouse-events-have-key-events */
-  return ReactDOM.createPortal(
+  if (isPortal) {
+    /* eslint-disable jsx-a11y/mouse-events-have-key-events */
+    return ReactDOM.createPortal(
+      <ContentStyled
```
alexzherdev: should definitely extract this in a variable so that there is no chance of missing one branch of the condition if there is an update to this part

packages/Popover/src/components/Tip/Tip.js
```diff
@@ -17,11 +17,31 @@ function Tip(props) {
...
   const isDark = React.useContext(ThemeContext);
-  const { content, tip, refTip, isOpen, portalElement } = React.useContext(PopoverContext);
+  const { content, tip, refTip, isOpen, isPortal, portalElement } = React.useContext(PopoverContext);
   const borderColor = isDark ? tokens.color.black : tokens.border.color;
   const backgroundColor = isDark ? tokens.color.black : tokens.color.white;
 
-  return ReactDOM.createPortal(
+  if (isPortal) {
+    return ReactDOM.createPortal(
+      <TipStyled
```
alexzherdev: same here

Issue #88, opened 2019-06-12
Using npmignore as white listing.


nahumzs:
# Repo upgrade
Not necessarily a bug, but we are using npmignore as whitelisting files we should find a better way to achieve this:

currently .npmignore structure:

```js
*

!lib/**
```

Refer to https://github.com/acl-services/paprika/pull/77#discussion_r291338658 for more context

nahumzs:
I think npmignore is working ok for us after a year.

PR #399, opened 2020-01-22
Star 106 Adding External link & NewTab Icon components

tristanjasper:
### Purpose 🚀
Creation of the `ExternalLink` component
Creation of the `NewTab` Icon component

### Notes ✏️
Porting basic version from acl-ui 3 repo
Adds text truncating feature

Link to older acl-ui version - https://github.com/acl-services/acl-ui/blob/master/src/components/ExternalLink/ExternalLink.js

### Updates 📦
- [x] MAJOR (breaking) change to externalLink component
- [x] PATCH (bug fix) change to Icon, Heading components

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/star-106-external-link-component

### Screenshots 📸
image

image

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-106


packages/ExternalLink/src/ExternalLink.js
```diff
@@ -0,0 +1,64 @@
...
+      target="_blank"
+      externalLinkIconSize={externalLinkIconSize}
+      {...moreProps}
+    >
+      <styled.ExternalLinkContentStyled hasUnderline={hasUnderline}>
+        {getContentText()}
+      </styled.ExternalLinkContentStyled>
+      <NewTabIcon
+        css={styled.ExternalLinkIconStyles}
+        color={tokens.textColor.icon}
```
tristanjasper: I'm not sure what might be the best way to keep these "styles" in the `externalLink.style.js` as they really are props being applied to the component... I'm wondering if there is a way to apply props as styles in styled components, anyone any experience with that?

packages/ExternalLink/src/ExternalLink.styles.js
```diff
@@ -0,0 +1,48 @@
...
+
+  ${props => {
+    return props.hasUnderline ? "text-decoration: underline" : externalLinkContentHoverFocus;
+  }};
+`;
+
+export const ExternalLinkIconStyles = `
+  display: inline-block;
+  margin-left: ${tokens.spaceSm};
+  position: absolute;
```
tristanjasper: Had to make this icon `position:absolute` as was seeing browser differences in how the component was evaluating the width and when it was applying the `text-overflow: ellipses`. Specifically Safari was showing differences. Keeping`postion:absolute` seems to make more consistent between browsers.

packages/ExternalLink/src/ExternalLink.js
```diff
@@ -0,0 +1,60 @@
...
+  hasUnderline: PropTypes.bool,
+
+  /** Link url for the target. */
+  href: PropTypes.string.isRequired,
+};
+
+const defaultProps = {
+  ariaText: null,
+  children: null,
+  hasUnderline: true,
```
AndreyChernykh: I'd prefer booleans with `false` by default. 
Here, for instance, `noUnderline: false` instead of `hasUnderline`. 

Compare: 

```js
<ExternalLink /> // underline by default
<ExternalLink hasUnderline /> // made no sense, since hasUnderline is true anyway
<ExternalLink hasUnderline={false} />
```
with
```js
<ExternalLink />
<ExternalLink noUnderline />
```

@nahumzs @alexzherdev @mikrotron

tristanjasper: Personally i agree with that. It was api from older component so i left it unchanged, but perhaps now is good time. Your suggestion makes more sense to me...

packages/ExternalLink/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [0.0.1] - 2020-01-22
+
+### Added
+
+- ExteralLink component created (some aspects ported from acl-ui version) [@tristanjasper](https://github.com/tristanjasper).
```
alexzherdev:
```suggestion
- ExtenralLink component created (some aspects ported from acl-ui version) [@tristanjasper](https://github.com/tristanjasper).
```

packages/ExternalLink/src/ExternalLink.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+function ExternalLink(props) {
+  const iconFontSize = 11;
+  const I18n = useI18n();
+  const { ariaText, children, hasUnderline, ...moreProps } = props;
+
+  const getContentText = () => {
+    return children || I18n.t("externalLink.view");
```
alexzherdev: Will this default value be used in reality?

packages/ExternalLink/src/ExternalLink.styles.js
```diff
@@ -0,0 +1,47 @@
...
+import { toInt } from "@paprika/stylers/lib/helpers";
+
+export const ExternalLinkStyled = styled.a`
+  border-radius: ${tokens.border.radius};
+  color: ${tokens.textColor.link};
+  display: inline-block;
+  max-width: 100%;
+  position: relative;
+
+  ${props => {
```
alexzherdev:
```js
${props => (
  `padding: 1px ${props.iconFontSize + toInt(tokens.spaceSm)}px 1px ${tokens.spaceSm};`
)}
```
does this look better?

packages/ExternalLink/stories/examples/LayoutExamples.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { Story } from "storybook/assets/styles/common.styles";
+import ExternalLink from "../../src";
+
+const ExampleStory = () => (
+  <Story>
+    <p>
+      This is default <code>ExternalLink</code> style:{" "}
+      <ExternalLink ariaText="aria text for link" href="http://acl-ui.aclgrc-s1.com" />
```
alexzherdev: Want to update these links to something more contemporary? 😄

packages/ExternalLink/src/ExternalLink.js
tristanjasper: I guess it could be, but shouldn't be encouraged in general. I'm just it says "view" and not "click here" :)

packages/ExternalLink/stories/examples/LayoutExamples.js
tristanjasper: Maybe. :) Ok Ok

packages/ExternalLink/src/ExternalLink.js
AndreyChernykh: I'd make `children` isRequired. 
Having them `null` make not much sense. If the content of the lint is empty, the link is pretty much useless anyway.

packages/ExternalLink/src/ExternalLink.js
mikrotron: As a counter, we like to prefix our boolean props with `is`, `has`, `should`, etc... and `hasNoUnderline` is a bit meh... I'm not a big fan of _negative_ booleans, typically.  So while _in general_ I agree that defaults should be `false`, I'm okay with exceptions too. 

Tristan is right though, now is the right time to revisit the API and make any changes we like.  I'm personally 50/50 on this one, so I'll let Tristan make the call.

packages/ExternalLink/src/ExternalLink.styles.js
mikrotron: How about:
```js
${ ({ iconFontSize }) => `padding: 1px ${iconFontSize + toInt(tokens.spaceSm)}px 1px ${tokens.spaceSm};` }
```

packages/ExternalLink/stories/examples/LayoutExamples.js
mikrotron: Don't waste a good rickroll opportunity!

packages/ExternalLink/src/ExternalLink.js
alexzherdev: IMO @AndreyChernykh 's point overweighs the naming concern (especially if we don't expect this prop to be used much anyway)

tristanjasper: Ok I'm going with `hasNoUnderline` with default value of `false` for now. I think having a default value of true doesn't make sense to me. 
Incidentally we have a prop `hasUnderline` on the `Heading` component as well, but its default value is `false` so I don't think we should change that.

packages/ExternalLink/src/ExternalLink.js
tristanjasper: Ok, i'll remove the fallback text, If they really want it to say "view" they can just pass that as the child....

alexzherdev: 👍 this also means this component doesn't have to depend on l10n

packages/ExternalLink/package.json
```diff
@@ -0,0 +1,29 @@
+{
```
alexzherdev: Don't forget `.npmignore`

packages/ExternalLink/src/ExternalLink.js
tristanjasper: Discussed with @mikrotron. The `color` prop can just be set as a `color` style instead on this component and it will be handled via the `currentColor` keyword, so the svg in the icon shall inherit the color from the style. This also means we don't need to apply styling props in this file, which is what i was trying to avoid. 👍

packages/ExternalLink/package.json
tristanjasper: we don't usually include that in each component, can you elaborate where/why?

packages/ExternalLink/src/ExternalLink.styles.js
tristanjasper: Going once, going twice. Sold.

packages/ExternalLink/src/ExternalLink.js
```diff
@@ -0,0 +1,50 @@
...
+      aria-label={ariaText || children}
+      onClick={handleSwallowClick}
+      rel="noopener noreferrer"
+      target="_blank"
+      iconFontSize={iconFontSize}
+      {...moreProps}
+    >
+      <sc.ExternalLinkContent hasNoUnderline={hasNoUnderline}>{children}</sc.ExternalLinkContent>
+      <NewTabIcon css={sc.ExternalLinkIconStyles} size={`${iconFontSize}px`} />
+    </sc.ExternalLink>
```
tristanjasper: Going with the `sc` prefix as I find this more readable overall.

packages/ExternalLink/package.json
alexzherdev: Hmm I think this was discussed before at some point... But personally I don't think there's enough reason to publish things like `src` and `stories`. If people disagree we could change this, but let's do so consistently.

mikrotron: Agreed.  I think most components do have the `.npmignore`, so that only the `lib` folder gets published for consumption on npm.  They should _all_ have it though.

mikrotron: You can copy this one:
https://github.com/acl-services/paprika/blob/master/packages/Button/.npmignore

tristanjasper: Oh, It looks like each component does have a `npmignore` file, i just don't see them in my ide, i'll add one

alexzherdev: FYI there is also #88 - for later.

tristanjasper: Ok, ive added `.npmignore` for this component as well as in `Checkbox`, `Dropdown`, `Confirmation` components

PR #401, opened 2020-01-23
UX-633 ActionBar component

allison-c:
### Purpose 🚀
`<ActionBar />` component
**Demo**: http://storybooks.highbond-s3.com/paprika/UX-633-Filters/?path=/story/action-bar--showcase
**Readme**: https://github.com/acl-services/paprika/tree/UX-633-Filters/packages/ActionBar

`<Filter />`
ezgif com-optimize (25)

`<Sort />`
ezgif com-optimize (26)

`<ColumnsArrangement />`
ezgif com-optimize (27)

### Notes ✏️
It might be confusing why all 3 components have the prop `columns`, but I choose to pass in them separately instead of put it on the main `<ActionBar />` component. It was because they require slightly different data, although they're all related to columns. For example, we need the `column.type` for sorting and filtering, but not columns arrangement. If the consumer want to use `<ColumnArrangement />` only, they should still be able to use it without column types data.

- [ ] Adding prop descriptions
- [ ] Adding tests once API design passed, in case some beaking change required
- [ ] Adding data display section in storybook
- [ ] Adding mock API call story

### Updates 📦
- [x] MAJOR (breaking) change to @paprika/action-bar

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-633-Filters

### Screenshots 📸

### References 🔗


packages/ActionBar/stories/ShowcaseApp/MyColumns.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import { ColumnsArrangement, useColumnsArragment } from "../../src";
+
+const defaultColumnsForArrangement = [
```
nahumzs: I would suggest that the defaultColumnArragement be an array of IDs if we pass the full objects when you are developing will be hard to move objects around to create a specific state you might want to test.

Ex.

its easier to modify `['goals', 'names', 'status', 'country']` and move it around than an entire object, now Imagine that you have a 100 rows in a json format, would be annoying to have to move that data structure around just specify what order you want for your columns.

I struggle with this when I was doing the POF for Assesst and Attributes

packages/ActionBar/stories/ShowcaseApp/MySort.js
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { Sort, useSort } from "../../src";
+
+export default function MySort({ columns }) {
```
nahumzs: 👏 this is way more consistent than our previous iteration awesome :) hooks to the rescue.

packages/ActionBar/stories/ShowcaseApp/MySort.js
```diff
@@ -0,0 +1,30 @@
...
+  const handleChange = fieldId => params => {
+    onChangeField({ ...params, id: fieldId });
+  };
+
+  return (
+    <Sort appliedNumber={appliedNumber} columns={columns} onAddField={onAddField} onApply={onApply}>
+      {sortedFields.map((field, index) => (
+        <Sort.Field
+          key={field.id}
+          {...field}
```
nahumzs: maybe would be good to explain or add a comment what does field has inside so whoever is reading this code on the future can have an idea why you are spreading the `{...field}`

packages/ActionBar/src/hooks/useSort/sort.js
```diff
@@ -0,0 +1,37 @@
+import moment from "moment";
+
+function compareString(a, b) {
+  return a.localeCompare(b);
```
AndreyChernykh: Regular `.localeCompare` might be slow. 

For sorting large arrays it is recommended to use `Intl.Collator` (but if the array size is small, then `localeCompare` should be OK).

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Collator

```js
const compareFunctionFallback = (a, b) => (a < b ? -1 : a > b ? 1 : 0);

function getCompareFunc() {
  if (!Intl || !Intl.Collator) {
    return compareFunctionFallback;
  }

  return new Intl.Collator(I18n.locale, {
    sensitivity: 'base',
  }).compare;
}

let alphaSortCompareFunction;

export function alphaSortAsc(a, b) {
  if (!alphaSortCompareFunction) {
    alphaSortCompareFunction = getCompareFunc();
  }

  return alphaSortCompareFunction(a, b);
}

export function alphaSortDesc(a, b) {
  if (!alphaSortCompareFunction) {
    alphaSortCompareFunction = getCompareFunc();
  }

  return alphaSortCompareFunction(b, a);
}
```

@nahumzs

packages/ActionBar/src/hooks/useSort/useSort.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  const handleChangeItem = React.useCallback(
+    ({ id: fieldId, direction, columnId }) => {
+      let change;
+      if (columnId) {
+        change = { columnId };
+      } else {
+        change = { direction };
+      }
```
AndreyChernykh:
```js
const change = columnId ? {columnId} : {direction};
```

packages/ActionBar/src/hooks/useSort/useSort.js
```diff
@@ -0,0 +1,83 @@
...
+  if (sortedFields.length > 0) {
+    sortedFields.forEach(field => {
+      sortedData = sort({
+        data,
+        columnId: field.columnId,
+        direction: field.direction,
+        columnType: columns.find(column => field.columnId === column.id).type,
+        momentParsingFormat: columns.find(column => field.columnId === column.id).momentParsingFormat,
+      });
+    });
```
AndreyChernykh: I don't quite understand how that works. If `sortedFields.length > 1` wouldn't `sortedData` be overridden ?

AndreyChernykh: Also, `columns.find(column => field.columnId === column.id)` can be extracted

packages/ActionBar/src/hooks/useSort/useSort.js
```diff
@@ -0,0 +1,83 @@
...
+        id: nanoid(),
+        columnId: columns[0].id,
+        direction: "ASCEND",
+      },
+    ]);
+  }, [columns]);
+
+  const handleDeleteItem = React.useCallback(
+    deletedId => {
+      setSortedFields(prevFields => [...prevFields].filter(filter => filter.id !== deletedId));
```
AndreyChernykh: There is no need to do `[...prevFields]`, because `filter` will return a new array anyway.

packages/ActionBar/stories/ControlledActionBar/MyFilter.js
```diff
@@ -0,0 +1,120 @@
...
+  const memorizedHandleChange = React.useCallback(handleFilterChange, [columns, setFilters]);
+
+  function handleAddFilter() {
+    setFilters(prevFilters => [...prevFilters, getDefaultFilter()]);
+  }
+
+  const memorizedAddFilter = React.useCallback(handleAddFilter, [setFilters]);
+
+  function handleDeleteFilter(deletedFilterId) {
+    setFilters(prevFilters => [...prevFilters].filter(filter => filter.id !== deletedFilterId));
```
AndreyChernykh: There is no need to do [...prevFields], because filter will return a new array anyway.

packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
```diff
@@ -0,0 +1,119 @@
...
+      <Popover.Trigger>
+        {(handler, attributes, isOpen) => (
+          <sc.Trigger
+            {...attributes}
+            onClick={handler}
+            hasColumnsHidden={columns.filter(column => column.isHidden).length > 0}
+            isOpen={isOpen}
+          >
+            <sc.Icon />
+            {getLabelText(columns.filter(column => column.isHidden).length)}
```
AndreyChernykh: `const hiddenColumns = columns.filter(column => column.isHidden)`

packages/ActionBar/src/hooks/useFilter/useFilter.js
```diff
@@ -0,0 +1,112 @@
...
+  function handleAddFilter() {
+    setFilters(
+      produce(draftFilters => {
+        draftFilters.push(getDefaultFilter());
+      })
+    );
+  }
+
+  function handleDeleteFilter(deletedFilterId) {
+    setFilters(prevFilters => [...prevFilters].filter(filter => filter.id !== deletedFilterId));
```
AndreyChernykh: There is no need to do [...prevFields], because filter will return a new array anyway.

packages/ActionBar/src/hooks/useFilter/testers.js
```diff
@@ -0,0 +1,41 @@
...
+  [Filter.rules.GREATER_THAN_OR_EQUAL_TO]: (value, testValue) => processNumber(value, testValue, (a, b) => a >= b),
+  [Filter.rules.LESS_THAN]: (value, testValue) => processNumber(value, testValue, (a, b) => a < b),
+  [Filter.rules.LESS_THAN_OR_EQUAL_TO]: (value, testValue) => processNumber(value, testValue, (a, b) => a <= b),
+  [Filter.rules.IS_EMPTY]: isBlank,
+  [Filter.rules.IS_NOT_EMPTY]: isNotBlank,
+  [Filter.rules.IS_BEFORE]: (value, testValue, { momentParsingFormat }) =>
+    processDate(momentParsingFormat, value, testValue, (a, b) => a.isBefore(b)),
+  [Filter.rules.IS_AFTER]: (value, testValue, { momentParsingFormat }) =>
+    processDate(momentParsingFormat, value, testValue, (a, b) => a.isAfter(b)),
+};
```
AndreyChernykh: @allison-c @nahumzs we need unit tests. Lots of them :)

packages/ActionBar/src/hooks/useFilter/useFilter.js
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable no-param-reassign */
+import React from "react";
+import nanoid from "nanoid";
+import produce from "immer";
+import Filter from "../../components/Filter";
```
AndreyChernykh: That is weird that the hook depends on the component. Maybe common dependencies, such as `defaultRulesByType`, can be extracted?

AndreyChernykh: @nahumzs @allison-c Also, I'm not sure why we separate hooks from components. We need to separate concerns, but not the form. Here, `Filter` and `useFilter` are logically connected. Imagine we had no hooks in React. Then, instead of `useFilter` you'd create another component, something like `FilterConsumer`, and would put in next to `Filter`. Then why we need a different treatment for `useFilter`?

AndreyChernykh: The same goes for `useSort` and `useColumnsArrangement`

packages/ActionBar/src/hooks/useFilter/useFilter.js
```diff
@@ -0,0 +1,112 @@
...
+        ? testers[filter.rule](
+            row[filter.columnId],
+            filter.value,
+            columns.find(column => filter.columnId === column.id)
+          )
+        : true;
+    };
+    return operator === logicalFilterOperators.AND ? filters.every(tester) : filters.some(tester);
+  });
+}
```
AndreyChernykh: Again, we have lots of logic here. We need lots of tests.

packages/ActionBar/src/hooks/useColumnsArragment/useColumnsArragment.js
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import produce from "immer";
+
+export default function useColumnsArragment(initialColumns) {
+  const [orderedColumns, setColumns] = React.useState(initialColumns);
+
+  function handleChangeOrder({ source, destination }) {
+    setColumns(
+      produce(draftColumns => {
```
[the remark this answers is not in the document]
AndreyChernykh: An additional improvement: 

```js
const setHiddenForAll => isHidden = produce(draftColumns => {
        draftColumns.forEach(column => {
          // eslint-disable-next-line no-param-reassign
          column.isHidden = isHidden;
        });
      })

export default function useColumnsArragment(initialColumns) {
.... 

 function handleHideAll() {
    setColumns(setHiddenForAll(true));
 } 

 function handleShowAll() {
    setColumns(setHiddenForAll(false));
  } 
```

packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
```diff
@@ -0,0 +1,119 @@
...
+export default function ColumnsArrangement(props) {
+  const { onChangeOrder, onChangeVisibility, columns, onHideAll, onShowAll } = props;
+  const I18n = useI18n();
+  const [searchTerm, setSearchTerm] = React.useState("");
+
+  const filteredColumns = searchTerm.length
+    ? columns.filter(column => column.label.match(new RegExp(searchTerm, "i")))
+    : columns;
+
+  function getLabelText(numberOfHiddenColumn) {
```
AndreyChernykh: `getLabelText()` can be moved outside of the component (optionally)

packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
```diff
@@ -0,0 +1,119 @@
...
+    const actualSource = columns.indexOf(columns.find(column => column.id === filteredColumns[source].id));
+    const actualDestination = columns.indexOf(columns.find(column => column.id === filteredColumns[destination].id));
+
+    onChangeOrder({ source: actualSource, destination: actualDestination });
+  };
+
+  function handleSearch(e) {
+    setSearchTerm(e.target.value);
+  }
+
```
AndreyChernykh: const hiddenColumns = columns.filter(column => column.isHidden);

packages/ActionBar/src/components/Sort/Sort.js
```diff
@@ -0,0 +1,118 @@
...
+          <sc.Icon />
+          {getLabelText(appliedNumber)}
+        </sc.Trigger>
+        <Popover.Content>
+          <Popover.Card>
+            <sc.FieldsPanel ref={fieldsRef} tabIndex={-1}>
+              {React.Children.count(children) === 0 ? (
+                <GenericNoAppliedPlaceholder>{I18n.t("actionBar.sort.no_sorts_applied")}</GenericNoAppliedPlaceholder>
+              ) : null}
+              {children}
```
AndreyChernykh: ) : children}

packages/ActionBar/src/hooks/useSort/sort.js
nahumzs: We should definitely try to implement this, we suffered with localeCompare when having thousands of rows.

packages/ActionBar/stories/ShowcaseApp/MyColumns.js
allison-c: How about this?
```js
<ColumnsArrangement
      orderedColumnIds={["goals", "names", "status", "country"]}
      onChangeOrder={() => {}}
      onHideAll={() => {}}
      onShowAll={() => {}}
      onChangeVisibility={() => {}}
    >
      <ColumnsArrangement.ColumnDefinition id="goals" label="Goals" isDisabled />
      <ColumnsArrangement.ColumnDefinition id="name" label="Name" isHidden />
      <ColumnsArrangement.ColumnDefinition id="status" label="Status" />
      <ColumnsArrangement.ColumnDefinition id="country" label="Country" />
</ColumnsArrangement>
```

packages/ActionBar/src/hooks/useSort/useSort.js
allison-c: If there's 2 fields for different columns, I'll sort by the first field first and then sort the data again by the second field. So the results compliance with both sorted fields.
If there's 2 fields for the same columns, the second one will override the first one.

packages/ActionBar/src/hooks/useFilter/useFilter.js
allison-c: My idea is to let consumers can use `Filter` as a controlled component but also an uncontrolled component if they use `useFilter`, same as `Sort` and `ColumnArrangement`. 
Maybe provide a `FilterController` component and a `Filter` component is a better option and easier to understand? 
We also take charge of the data filtering part in `useFilter` hook, so even though if we have a controlled filter and uncontrolled filter, we'll still need a `useFilter` hook to help the data filtering

alexzherdev:
I understand ActionBar is designed not to be DataGrid-specific, but is it a valid assumption that that is the main use case? At the very least, it's the _first_ use case we have for the ActionBar. As such, I'd love to see examples of that integration. I'm interested in the following:
- how easy is it to only define my columns once?
- what does it look like when I implement uncontrolled filtering? why might I need to control it?
- same for sorting
- are those controls easy to use when my grid is server-driven?
If these tasks require too much plumbing, then potentially we might consider creating a `DataGrid.ActionBar` that wraps ActionBar and simplifies the usage.

Other comments I had from looking at the stories:
- in both stories clicking Cancel in the popovers doesn't seem to cancel anything. I would expect it to undo any changes made after opening
- I don't think we should ever allow the same column to appear more than once in sorting? (e.g. I'm able to sort by Goals ascending and then by Goals descending)
- the filter input in columns arrangement could benefit from a clear button
- the layout for filters could _really_ benefit from being a table so that the fields are aligned. It really is a jarring look right now
image

johnpeele:
We are not quite ready to report our findings, but Fernanda and I conducted 6 usability tests on the ActionBar and DataGrid. There are a number of UX & UI suggestions that we have... Should we include those on this PR or wait and open a new PR?

Reading Alex's latest comments is what prompted me to comment - some of his points are exactly the same as our findings.

packages/ActionBar/src/hooks/useFilter/useFilter.js
nahumzs: Yeah, the idea was to use the hooks if you want to convert them into an uncontrolled components, that's the idea behind hook to reuse logic, I think its cleaner than creating another component that might only confuse our users.

alexzherdev: What I'm not liking here is, at least from seeing the stories, I don't really care what this hook returns. E.g.
```jsx
  const {
    orderedColumns,
    handleChangeVisibility,
    handleShowAll,
    handleHideAll,
    handleChangeOrder,
  } = useColumnsArragment(defaultColumnsForArrangement);

  return (
    <ColumnsArrangement
      columns={orderedColumns}
      onChangeOrder={handleChangeOrder}
      onHideAll={handleHideAll}
      onShowAll={handleShowAll}
      onChangeVisibility={handleChangeVisibility}
    />
```
All those variables only exist to immediately pass them on to the component, I don't need them otherwise. This could be made slightly better
```jsx
  const whatever = useColumnsArragment(defaultColumnsForArrangement);

  return (
    <ColumnsArrangement {...whatever} />
  );
```
but I'm not sure that's too much better than
```jsx
return <ColumnsArrangement.Default columns={defaultColumns} />;
```
?

allison-c:
> * sorting

> how easy is it to only define my columns once?

I agree it's not easy now, like I wrote in the PR description above, even though `columns` prop looks similar between 3 components, the data required in each column object is different. I'll discuss with Nahum.

> what does it look like when I implement uncontrolled filtering? why might I need to control it?
> same for sorting

The showcase example is an uncontrolled Action Bar

> are those controls easy to use when my grid is server-driven?

We have a mock server example in this branch, was for user testing https://github.com/acl-services/paprika/pull/449

> in both stories clicking Cancel in the popovers doesn't seem to cancel anything. I would expect it to undo any changes made after opening

This was a design decision, we'll take look with the user testing result.

> I don't think we should ever allow the same column to appear more than once in sorting? (e.g. I'm able to sort by Goals ascending and then by Goals descending)

Right, JP wants that, and ideally we can do what airtable does. I didn't get the chance to do it. Since assets and attributes will only support 1 sorting field, we decided to do it later.

> the filter input in columns arrangement could benefit from a clear button

👍 

cc @johnpeele 

Regarding to the follow-up changes, it depends on if we need to change API or not. I heard from @nahumzs he wants to merge this component by this week, so I assume someone is waiting for using it? If it's not that urgent, I can branch out from this branch and create follow-up PR, so the code review process will be easier

nahumzs:
@allison-c maybe we should wait, and discuss it a little more. 

About the complexity of integrating this component: 
Some of the changes in this PR were driven by the fact that I was implementing the A&A demo that was used for doing the user testing, and while was not hard to connect it, it was not straightforward at all, required lots of plumbing. 

The least I would like to do is to make hard to implement this from the consumer side, that its what we want to avoid in each of the components that we build in paprika. 

So let's branch out from this PR and implemented some of the feedback as you are suggesting allison, and wait as well from the usability report. 


FWIW:
Two other reasons to splitted the component from the DataGrid were:

- Consumers might want to add their own filters on the action bar like their own selects or specifics filter like we do on Mission control
- The coupling of the DataGrid and this component was intended at the beginning but was getting complex to manage and managing concern was not that clear and complex. This separation forces us to thread them as independent components and make maintainability easier.

Thank you for the feedback :)

alexzherdev:
> I agree it's not easy now, like I wrote in the PR description above, even though `columns` prop looks similar between 3 components, the data required in each column object is different. I'll discuss with Nahum.

Sorry, what I meant about columns is, I'd like to avoid having to define two arrays of columns, one for the action bar, one for the grid.

> The showcase example is an uncontrolled Action Bar
> 
> We have a mock server example in this branch, was for user testing #449
> 

> This separation forces us to thread them as independent components and make maintainability easier.

Ok, I think I should not have used the word integration.
My point is, I think a story like AAA from #449 (although a more realistic one—extracting columns from a server call is not something we're likely to do) should be a required part of this PR. Because the purpose of ActionBar (the way I see it) is to be able to work with **some sort of tabular component**. If you like, you can even forget about DataGrid—just do a plain HTML table generated from a JS array of objects. But have at least one story that renders
```jsx
<ActionBar />
<table />
```
and show (to reviewers and to yourselves) what the implementation of filters/sorting/columns would look like. How I would define columns, how I would get notified of a filter change etc.

I think this doesn't contradict at all the notion that ActionBar and DataGrid are independent components (which makes total sense to me btw). Remember issues with forms in Modals? As an example, imagine Form is also a component. It's totally logical Modals and Forms are independent components. But if I were developing a Modal, and I knew that Modals with forms in them are a very common case, I would strive to have a story reflecting that, just as an example of how Modal behaves in this case. As you know, we didn't do this, and now it's really difficult to put forms in Modals. That's where I'm coming from.

With regards to the timeline, if it's driven by Assets, we just had a priority change and we're going to be working on the API more right now, so I'd rather we didn't rush this.

packages/ActionBar/src/hooks/useFilter/useFilter.js
nahumzs:
```js
const whatever = useColumnsArragment(defaultColumnsForArrangement);
  return (
    <ColumnsArrangement {...whatever} />
  );
```

In that regard you are correct I don't know why I didn't just spread the Hook props on the component brain fart (maybe the pressure to delivering the demo for testing purposes). 

But I do think using hooks to abstract the logic it's the appropriate way to go for these cases, at the end, it's the main reason why they were created to be able to reuse logic. 

I recently read the updated version of the `reducer pattern` by Kent Dodds where he moves away from a render function to hooks and it is expecting it to use it in his downshift library. https://kentcdodds.com/blog/the-state-reducer-pattern-with-react-hooks

^ So that gives me confidence that the approach of using hooks it's the correct way to go.

packages/ActionBar/README.md
```diff
@@ -0,0 +1,153 @@
...
+const handleDeleteFilter = filterId => () => {
+  onDeleteFilter(filterId);
+};
+const handleChangeFilter = filterId => params => {
+  onFilterChange({ ...params, id: filterId });
+};
+
+return (
+  <Filter
+    appliedNumber={appliedNumberOfFilters}
```
nahumzs, line 58: I think something like 

`const props = useFilter(...)`

`<Filter {...props}  />`

is shorter and more efficient?

nahumzs: nip btw, I think for documentation is fine so it is clearer what we are doing.

PR #404, opened 2020-01-23
STAR-111 Apply subtle focus rings 🐐

mikrotron:
### Purpose 🚀
Apply new `stylers.focusRing.subtle()` styling to various components, as appropriate.

### Notes ✏️
- Addresses this error left in a previous PR:
  https://github.com/acl-services/paprika/pull/372/files#diff-35881852d236126a5c4a48174b177f8aR88

### Updates 📦
- [x] PATCH (bug fix) change to `Modal`
- [x] MINOR (backward compatible) change to `Takeover`
- [x] MINOR (backward compatible) change to `SidePanel`
- [x] MINOR (backward compatible) change to `Popover`
- [x] MINOR (backward compatible) change to `ProgressAccordion`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-111--apply-subtle-focusring

### Screenshots 📸
Modal
Screen Shot 2020-01-23 at 3 08 26 PM

Popover (isEager)
Screen Shot 2020-01-23 at 3 07 49 PM

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-111
_Actually_ resolves https://github.com/acl-services/paprika/issues/365


johnpeele:
@mikrotron Sorry for being late to the party, but where can I find the history of the design decisions related to this subtle focus ring? Is this an a11y-driven thing? Design driven? Customer request? etc etc

Thanks!

packages/ProgressAccordion/src/components/Responses/Responses.js
```diff
@@ -20,7 +22,7 @@ const Responses = ({ children }) => {
   if (!children) return null;
 
   return (
-    <dl data-pka-anchor="responses" css={responsesStyles}>
+    <dl data-pka-anchor="responses" css={responsesStyles} tabIndex="0">
```
allison-c: Just curious why this `dl` need to be focusable by tab?

mikrotron: Because it has `max-height`, so it can have a vertical scrollbar... then for keyboard access it needs to be tabbable.

mikrotron:
> @mikrotron Sorry for being late to the party, but where can I find the history of the design decisions related to this subtle focus ring? Is this an a11y-driven thing? Design driven? Customer request? etc etc
> 
> Thanks!

Took a bit of digging, but this was the instigator:

https://github.com/acl-services/paprika/pull/320#pullrequestreview-327777926

@DanilAgafonov also dug up this to support his case, and we even discussed it in UX Dev ShopTalk:

https://dequeuniversity.com/rules/axe/3.3/scrollable-region-focusable

But then some follow up discussion happened with Sarah + I and we decided to have a "subtle" focus indicator for non-interactive elements (you can't click, or enter data into a scrollable div).  

That led to https://github.com/acl-services/paprika/pull/372

And here we are.  I'm not sure if Sarah was planning to write this up as a guideline or not, but you could follow up with her on those details, or how she decided on the 2px dotted style (I don't think she made that design choice without consulting other designers though).

mikrotron:
@johnpeele Sarah + Jayde also told me there's a Confluence page detailing the design decisions (which you commented on 😁)

PR #405, opened 2020-01-24
STAR-110 Remove all IDs from SVG files

DanilAgafonov:
### Purpose 🚀
If we insert any icon to page more than one time, then we get ID collision. This PR eliminates this issue.

### Updates 📦
- [x] MINOR (backward compatible) change to `@paprika/icon`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-110-duplicate-svg-ids

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-110


DanilAgafonov:
I was looking for automated solution. But found only this issue: https://github.com/svg/svgo/issues/563

The only option for now - remove redundant `<defs>`/`<use>` manually.

packages/Icon/package.json
```diff
@@ -14,7 +14,7 @@
     "access": "public"
   },
   "scripts": {
-    "svgr": "npx @svgr/cli -d . src/svg --out-dir src/"
+    "svgr": "svgr -d . src/svg --out-dir src/"
```
DanilAgafonov: The reason of this change: `npx` use latest version of `@svgr/cli` from npmjs.com. It means that version lock (`"@svgr/cli": "^4.1.0"`) will be ignored. After this change Yarn will run `./node_modules/.bin/svgr`.

alexzherdev: good catch 💯

strarsis:
@DanilAgafonov: PR ready in `svgo`, see https://github.com/svg/svgo/issues/563#issuecomment-685920125.
You can try out the new `svgo` `dereferenceUses`.

PR #407, opened 2020-01-27
Ux 706 form component accessibility multiple children

tristanjasper:
### Purpose 🚀

We need to support groups of inputs or children in the formElement. In the case of multiple checkboxes or radios we should be using a `legend` element inside of a `fieldset` element.

- Ensures a `fieldSet` and `legend` element are used when new `isFieldSet` prop is true.
- Ensures only `required` label is shown if both `hasRequiredLabel` and `hasOptionalLabel` props are true.
- Allows html to be rendered as the text in Label component

Additional changes
- Checkbox and Radio component now receive a `tabIndex` prop which gets passed down to their respective `input` elements.

### Notes ✏️

### Updates 📦
- [x] MINOR (backward compatible) change to formElement, Checkbox, Radio
- [ ] PATCH (bug fix) change to Checkbox, Radio

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-706-form-component-accessibility-multiple-children

### Screenshots 📸
Provides storybook examples of usage with radios and checkboxes
image

Provides usage example for nested formElements
image

### References 🔗
https://aclgrc.atlassian.net/browse/UX-706


packages/Checkbox/src/Checkbox.js
```diff
@@ -13,24 +13,33 @@ const checkboxStates = {
 };
 
 const propTypes = {
+  /** Used for aria-describedby on the checkbox input  */
+  ariaDescribedBy: PropTypes.string,
```
tristanjasper: adding this new `ariaDescribedBy` prop in both Checkbox and Radio components as these are both typically used in a fieldset as a children of the formElement component. We can then directly apply the `ariaDescribedBy` prop to them, which is then applied to their internal `input` components.

packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -79,6 +79,7 @@ const checkboxStyles = css`
   ${boxSizingStyles};
   ${({ size }) => styles[size].baseFontSize};
   line-height: ${({ hasLabel }) => (hasLabel ? lineHeightValue(-1) : "0")};
+  margin: 0 0 ${tokens.space} 0;
```
tristanjasper: Adding a margin below checkbox and radio component by default as it will typically be used in this way.
image

packages/FormElement/src/FormElement.js
```diff
@@ -125,23 +131,30 @@ function FormElement(props) {
     });
   }
   return (
-    <div css={formElementStyles} isInline={isInline} size={size} isDisabled={isDisabled} {...moreProps}>
+    <sc.FormElement
+      as={hasFieldSet ? "fieldset" : "div"}
```
tristanjasper: Outer container now rendering as a `fieldset` element if `isFieldSet` prop is true

packages/FormElement/src/FormElement.js
```diff
@@ -125,23 +131,30 @@ function FormElement(props) {
...
+      isInline={isInline}
+      size={size}
+      isDisabled={isDisabled}
+      {...moreProps}
+    >
       <Label
-        hasOptionalLabel={hasOptionalLabel}
+        hasOptionalLabel={hasRequiredLabel ? false : hasOptionalLabel}
```
tristanjasper: Changed the logic here so that `required` label takes priority and will not show the `optional` field if it is `hasRequiredLabel` is true

packages/FormElement/src/components/Label/Label.js
```diff
@@ -9,32 +9,35 @@ const propTypes = {
...
   help: null,
+  id: null,
 };
 
 const Label = React.forwardRef((props, ref) => {
-  const { hasOptionalLabel, hasRequiredLabel, help, id, isInline, isVisuallyHidden, label } = props;
+  const { hasOptionalLabel, hasRequiredLabel, help, id, hasFieldSet, label, ...moreProps } = props;
   const I18n = useI18n();
 
+  const labelProps = hasFieldSet ? { as: "legend" } : { as: "label", htmlFor: id, ref };
```
tristanjasper: renders component container as a `legend` or `label` depending on on the `hasFieldSet` prop being passed in

packages/FormElement/stories/examples/AlternateLayouts.js
```diff
@@ -167,9 +173,81 @@ const ExampleStory = () => {
           Give me some help. <a href="wegalvanize.com">Learn more</a>.
         </FormElement.Help>
       </FormElement>
-    </FormElementStory>
 
-    // Additionally need to show with radio component example
+      <Rule />
+
+      <Tagline>Form Element nested Form Elements.</Tagline>
+      <br />
```
tristanjasper: Adding this example for the use case of nested form Elements, the styling of the inputs etc would be the responsibility of the consuming application
image

packages/FormElement/src/components/Label/Label.js
```diff
@@ -9,32 +9,35 @@ const propTypes = {
   hasOptionalLabel: PropTypes.bool.isRequired,
   hasRequiredLabel: PropTypes.bool.isRequired,
   help: PropTypes.node,
-  id: PropTypes.string.isRequired,
+  id: PropTypes.string,
   isInline: PropTypes.bool.isRequired,
   isVisuallyHidden: PropTypes.bool.isRequired,
-  label: PropTypes.string.isRequired,
+  labelText: PropTypes.node.isRequired,
```
tristanjasper: Allowing label text content to be rendered as html now

mikrotron: Why not keep the original prop name? Maintaining the same name means it's not a breaking change, and postfixing with `Text` is probably less intuitive, since it's now _anything_, not just text, it could be any html or jsx.

packages/FormElement/tests/FormElement.spec.js
```diff
@@ -51,13 +51,27 @@ describe("FormElement", () => {
...
     expect(getByTestId("form-element.label")).toHaveTextContent(/required/i);
   });
 
+  it("renders optional label if hasOptionalLabel is true", () => {
+    const { getByTestId } = renderComponent({ hasOptionalLabel: true });
+
+    expect(getByTestId("form-element.label")).toHaveTextContent(/optional/i);
+    expect(getByTestId("form-element.label")).not.toHaveTextContent(/required/i);
+  });
+
```
allison-c: Maybe also adding some tests for fieldset cases?

packages/FormElement/src/components/Label/Label.styles.js
```diff
@@ -1,21 +1,22 @@
...
 
 const labelStyles = css`
-  ${stylers.lineHeight(-1)}
   color: ${tokens.textColor.default};
+  display: inline-block;
   font-weight: bold;
-  margin: 0 0 ${tokens.spaceSm} 0;
-
-  ${({ isInline }) => isInline && `padding-right: ${stylers.spacer(1)};`}
+  margin: 0 ${toInt(tokens.space) * 4}px ${tokens.spaceSm} 0;
```
allison-c: ${stylers.spacer(4)}

packages/FormElement/src/components/Label/Label.js
tristanjasper: It's not a breaking change because the Label component is rendered by the formElement, not the consuming application. But sure, I can see why text would make you think of "string" only. I'll change it back though although something about just "label" seems to value as this is a "Label" component (which may also render a `legend` element incidentally)

nahumzs:
@tristanjasper can you destructure as well `tabIndex` from the props and forward it to the input?

tristanjasper:
> @tristanjasper can you destructure as well `tabIndex` from the props and forward it to the input?

The consumer can set the tabIndex on any child component it wants to... Can you elaborate?

packages/Checkbox/CHANGELOG.md
```diff
@@ -1,11 +1,21 @@
...
 
 ### Added
+
 - Checkbox.states instead of checkboxStates [@oscarkwan](https://github.com/oscarkwan).
+
+## [0.2.2] - 2020-01-31
+
+### Added
+
+- Added tabIndex prop which is forwarded to input [@tristanjasper](https://github.com/tristanjasper).
```
mikrotron: Not strictly needed, according to the process we agreed on (only breaking changes need to be added to changelog), but this _is_ appreciated, thanks! 🙇

packages/FormElement/CHANGELOG.md
```diff
@@ -13,3 +13,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Removed
 
 - FormElement no longer adds its extended props to the children components provided. This must now be done by the consumer [@tristanjasper](https://github.com/tristanjasper).
+
+## [0.4.0] - 2020-01-27
```
mikrotron: These aren't breaking changes though, so we won't bump up to 0.4.  When the version is less than 1.0, the middle number indicates a breaking change (this is a special case for <1).  So we should only bump to 0.3.2.

packages/FormElement/CHANGELOG.md
```diff
@@ -13,3 +13,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+
+### Added
+
+- FormElement component adds new isFieldSet prop to support rendering of fieldset and legend elements for multiple children
```
mikrotron: You went with `hasFieldSet` 😬

PR #409, opened 2020-01-31
UX-633 Adding isOpen to popover trigger render function params

allison-c:
### Purpose 🚀

Adding `isOpen` to `<Popover.Trigger />` render child function params to show different trigger status. This will be used in table navigation

### Notes ✏️

### Updates 📦
- [x] MINOR (backward compatible) change to `@paprika/popover`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-633-passing-isOpen-into-popover-trigger

### Screenshots 📸
Jietu20200130-165146-HD

### References 🔗

### DIFF
--- a/packages/Popover/src/components/Trigger/Trigger.js
+++ b/packages/Popover/src/components/Trigger/Trigger.js
@@ -50,7 +50,6 @@ function Trigger(props) {
   const { children, a11yText, ...moreProps } = props;
 
   if (typeof children === "function") {
-    console.log("isoPEN", isOpen);
     return (
       <React.Fragment>
         {React.cloneElement(children(handleTriggerEvent, { "aria-describedby": content.ariaId }, isOpen), {
--- a/packages/Popover/src/components/Trigger/Trigger.js
+++ b/packages/Popover/src/components/Trigger/Trigger.js
@@ -20,6 +20,7 @@ function Trigger(props) {
   const {
     content,
     isEager,
+    isOpen,
     onClick,
     onClose,
     onDelayedClose,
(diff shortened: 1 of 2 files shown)


packages/Popover/src/components/Trigger/Trigger.js
```diff
@@ -49,9 +50,12 @@ function Trigger(props) {
   const { children, a11yText, ...moreProps } = props;
 
   if (typeof children === "function") {
+    console.log("isoPEN", isOpen);
     return (
       <React.Fragment>
-        {React.cloneElement(children(handleTriggerEvent, { "aria-describedby": content.ariaId }), { ...moreProps })}
+        {React.cloneElement(children(handleTriggerEvent, { "aria-describedby": content.ariaId }, isOpen), {
```
allison-c: Added it as 3rd param to avoid breaking current usage like this:
```
        <Popover.Trigger>
          {(handler, attributes) => (
            <Icon
              {...attributes}
              onMouseOver={handler}
              onMouseOut={handler}
              onFocus={handler}
              onBlur={handler}
              tabIndex={0}
              role="img"
              aria-label="info"
            />
          )}
        </Popover.Trigger>
```

packages/Popover/src/components/Trigger/Trigger.js
```diff
@@ -49,9 +50,12 @@ function Trigger(props) {
   const { children, a11yText, ...moreProps } = props;
 
   if (typeof children === "function") {
+    console.log("isoPEN", isOpen);
```
nahumzs: remove?

nahumzs:
with this new feature now the popover would not close unless is toggle directly on the trigger button?

packages/Popover/src/components/Trigger/Trigger.js
allison-c: 😬  👀

allison-c:
> with this new feature now the popover would not close unless it is toggle directly on the trigger button?

No this PR didn't change any behavior, the screenshot is just an example that the trigger can render different text even though it's an uncontrolled popover

nahumzs:
Ohh I see, sidetracking,  I think we did a small error passing parameter one by one, instead of using one object that could be extended.

PR #413, opened 2020-02-06
STAR-136 Allow tabIndex override for Buttons 🐐

mikrotron:
### Purpose 🚀
Allow disabled `<Buttons>` (with `isSemantic={false}`) to be tabbable if `tabIndex` prop is supplied.

### Notes ✏️
- Turns out a native `<button>` with a `disabled` attribute cannot be focussed, even with `tabindex=0`.  So this approach will only work with `isSemantic={false}` `<Buttons>`.
- Also applied same code to `<RawButton>`.
- Also applied a symmetrical coding approach to `<Checkbox>` and `<Radio>`.
- Added a demonstration of `tabindex` propagation to Checkbox Group story in Storybook.

### Updates 📦
- [x] PATCH (bug fix) change to `Button`, `RawButton`, `Checkbox`, `Radio`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-136--button-tabindex-override

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-136
https://aclgrc.atlassian.net/browse/BOT-2067

### DIFF
--- a/packages/Button/src/Button.js
+++ b/packages/Button/src/Button.js
@@ -4,8 +4,7 @@ import RawButton from "@paprika/raw-button";
 import RefreshIcon from "@paprika/icon/lib/Refresh";
 import DownIcon from "@paprika/icon/lib/CaretDown";
 import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
-// support for IE11
-import "@paprika/helpers/lib/dom/closest";
+import "@paprika/helpers/lib/dom/closest"; // support for IE11
 import buttonStyles, { iconStyles } from "./Button.styles";
 
 import Kinds from "./ButtonKinds";
@@ -84,7 +83,7 @@ const defaultProps = {
   onClick: () => {},
   role: "button",
   size: ShirtSizes.MEDIUM,
-  tabIndex: 0,
+  tabIndex: null,
 };
 
 const buttonPropTypes = {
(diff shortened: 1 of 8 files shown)


packages/Button/src/Button.js
```diff
@@ -151,6 +150,8 @@ const Button = React.forwardRef((props, ref) => {
     if (!isButtonDisabled) onClick(event);
   };
 
+  const bestTabIndex = isButtonDisabled && tabIndex === null ? -1 : tabIndex || 0;
```
robyn3choi: Since it's already checking if tabIndex is `null`, what is the purpose of the `|| 0`?

packages/RawButton/src/RawButton.js
```diff
@@ -77,15 +77,7 @@ const RawButton = React.forwardRef((props, ref) => {
...
-  const getTabIndex = () => {
-    if (tabIndex >= 0) {
-      return tabIndex;
-    }
-    if (isDisabled) {
-      return -1;
-    }
-    return tabIndex;
-  };
+  const bestTabIndex = isDisabled && tabIndex === null ? -1 : tabIndex || 0;
```
robyn3choi: also here

packages/Button/src/Button.js
mikrotron: Oh! That will set it to `0` as a default, instead of `null`.  `tabindex=null` is not helpful, it should be `0` by default, but I want to differentiate between `0` as a default prop, and `0` that was specified by the developer, which is why I've changed the default to `null`.

PR #415, opened 2020-02-07
PD-14323 Fix sticky sidepanel footer width

allison-c:
### Purpose 🚀
Fix the sticky sidepanel footer to avoid the shifting

Bonus:

- Fixed console errors in yarn test
- Fixed style component conventions
- Fixed header text truncation issue

### Notes ✏️

### Updates 📦
MINOR (backward compatible) change to `@paprika/sidepanel`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/PD-14323-Fix-sticky-sidepanel-footer

### Screenshots 📸
Before:
http://storybooks.highbond-s3.com/paprika/master/?path=/story/sidepanel-footer--sidepanel-footer-issticky
Jietu20200207-143019-HD

After:
http://storybooks.highbond-s3.com/paprika/PD-14323-Fix-sticky-sidepanel-footer/?path=/story/sidepanel-footer--sidepanel-footer-issticky
good

### References 🔗


packages/L10n/src/locales-js/pt.js
```diff
@@ -73,9 +73,6 @@ const locales = {
         "arrowdown": "para baixo",
         "escape": "ESC"
       },
-      "externalLink": {
```
tristanjasper: glad someone pushed the changes to the these! 👍

packages/SidePanel/src/components/Footer/Footer.js
```diff
@@ -26,7 +29,14 @@ export default function Footer(props) {
   } = props;
 
   return (
-    <div data-pka-anchor="sidepanel.footer" css={footerCSS} {...{ isSticky, width }} {...moreProps}>
+    <div
+      data-pka-anchor="sidepanel.footer"
+      css={footerCSS}
```
tristanjasper: perhaps we should move away from using the `css` attribute approach by default and render the element as `<sc.sidePanelFooter>` instead and export that as `styled.div` in the css...

packages/SidePanel/src/components/Footer/Footer.styles.js
```diff
@@ -1,10 +1,32 @@
...
+    99% { width: 100%; }
+    100% { width: ${width}; }
+  `;
+}
+
 const fixedFooterStyles = css`
-  position: fixed;
-  width: ${props => props.width};
+  ${props => {
+    const animation = props.isOpen ? slideIn() : slideOut(props.width);
```
tristanjasper: nit: deconstruct `props.width` to a variable as it is used twice? looks a bit cleaner imo

packages/SidePanel/src/components/Footer/Footer.js
allison-c: 👍 I'm using the new convention in the new added things. btw why it's `sc` not `styled` ?

tristanjasper: Imo `sc` (stands for styled component) makes it more readable rather than having `styled` written everywhere thats all.. open to discussion though. You can always update these existing elements with the approach too, doesn't have to just be new ones :)

packages/SidePanel/src/components/Footer/Footer.styles.js
```diff
@@ -1,10 +1,32 @@
...
+function slideIn() {
+  return keyframes`
+    0% { width: 100%; }
+    100% { width: 100%; }
+  `;
+}
+
+function slideOut(width) {
+  return keyframes`
+    0% { width: 100%; }
```
tristanjasper: wait, shouldn't we be animating the position using `transform: translate` instead of animating the width value? Also couldn't' these be helper functions? 
I notice we have this function as well taht could also use something similar it perhaps? 
https://github.com/acl-services/paprika/blob/master/packages/SidePanel/src/components/Dialog/Dialog.styles.js#L18

tristanjasper: Actually why is the footer animated anyway? Isn't the the container dialog animated in? Why does this element need its own separate animation? Shouldn't it always be `width:100%` of the container Dialog compoennt?

allison-c: The issue with footer was because its `width` was `33%` all the time, but somehow when it's animating, it counts `33%` of the side panel, not `33%` of `body` 
image

I think that's why it was shifting, not because of `translate`

allison-c: In the dom it's like:
```
<dialog> //position fixed to container
 <header>
 <content>
 <footer>  //position fixed to container if it's a sticky footer
<dialog>
```
We're just animating the `dialog`, but when the `footer` is sticky, we have to animate it at the same time, otherwise it looks like shifting

nahumzs:
since your are on the sidepanel can you fix this error on the specs?
```sh
 PASS  packages/SidePanel/tests/SidePanel.spec.js (7.054s)
  ● Console

    console.error node_modules/jsdom/lib/jsdom/virtual-console.js:29
      Error: Uncaught [Error: <SidePanel.Group /> is intented to be use with two or more SidePanels]
          at reportException (/Users/nahum_zsilva/src/paprika/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:66:24)
          at invokeEventListeners (/Users/nahum_zsilva/src/paprika/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:209:9)
          at HTMLUnknownElementImpl._dispatch (/Users/nahum_zsilva/src/paprika/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:119:9)
          at HTMLUnknownElementImpl.dispatchEvent (/Users/nahum_zsilva/src/paprika/node_modules/jsdom/lib/jsdom/living/events/EventTarget-impl.js:82:17)
```

Some spec is using the SidePanel.Group with one SidePanel, either add another sidepanel or just use one Sidepanel. 🙏

packages/SidePanel/src/components/Footer/Footer.styles.js
tristanjasper: I just wonder if there is a way around this without needing to animate the footer as well as the sidepanel itself. Can the footer be absolute positioned at the bottom of a container div that is `height:100%` instead perhaps? Then it woudn't need to be `position:fixed`. Would need some time to play about with that but i think could work

allison-c: Do you think this is better? 
image
But this will be a breaking change since I added 2 wrappers

tristanjasper: I don't see what that does can you commit the code so i can see how it affects it?

tristanjasper:
image
 
It would be good if the scrollbar wasn't showing when there less lines in the content area but the footer is sticky

PR #417, opened 2020-02-13
PD-14397 push content hook

oscarkwan:
### Purpose 🚀
- Create option for sidepanel to push content when opening panel
- Improve animation speeds / smoothness

### Notes ✏️
https://www.loom.com/share/dcdfa262e52d4e94af36dc33ddc4d5e8

### Updates 📦
- [x] MINOR (backward compatible) change to _SidePanel_

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/PD-14397-push-body-sidepanel


packages/SidePanel/src/SidePanel.js
```diff
@@ -18,6 +18,9 @@ const propTypes = {
   /** The content for the SidePanel. */
   children: PropTypes.node.isRequired,
 
+  /** Function that provides the container DOM element to be pushed. */
+  getPushContentRef: PropTypes.func,
```
nahumzs: on the popover we are using the following convention when we want to receive an html node object, technically is not a react ref.

  /** Function that provides DOM element to use as target for positioning the popover. */
  getPositioningElement: PropTypes.func,

  /** Function that provides the scrolling DOM element that contains the popover. */
  getScrollContainer: PropTypes.func,

let's adjust the prop name?

oscarkwan: Sure `getPushContentContainer`?

nahumzs:
I have some concern that this pattern might not work correctly on legacy screen where having position absolutes with z-index really hight can appear a the top of the sidepanel or not responsive UI break while resizing it. 

But if the consumer want to pay that toll to have this pattern I 👌 with it.

nahumzs:
Can you add a simple test to have this pattern test? even a screener test with open and close capture might be enough.

packages/SidePanel/src/SidePanel.js
```diff
@@ -121,10 +129,26 @@ function SidePanel(props) {
...
+    pushContentRefStyle.transition = PUSH_REF_TRANSITION_STYLE;
+    pushContentRefStyle.transitionDelay = PUSH_REF_TRANSITION_DELAY_STYLE;
+
+    if (isOpen) {
+      pushContentRefStyle.marginRight = width;
+    } else {
+      pushContentRefStyle.transitionDelay = "0s";
+      pushContentRefStyle.marginRight = "0";
+    }
+  }, [isOpen, getPushContentRef, width]);
```
AndreyChernykh: Do we need a cleanup function?

packages/SidePanel/src/SidePanel.js
nahumzs: `getPushContentElement`? I think would be more obvious for the consumer?

oscarkwan:
> I have some concern that this pattern might not work correctly on legacy screen where having position absolutes with z-index really hight can appear a the top of the sidepanel or not responsive UI break while resizing it.
> 
> But if the consumer want to pay that toll to have this pattern I 👌 with it.

I think I am okay with it for now. If there is really absolute positioned elements that affect the layout when pushed, then the consumer should make the judgement call to not push the content

packages/SidePanel/stories/examples/Screener.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import Screener, { Steps } from "screener-storybook/src/screener";
+import SidePanelPushContentStory from "./SidePanelPushContentStory";
+
+const screenerScript = new Steps()
+  .snapshot("SidePanel - Push content expanded")
```
nahumzs: 😮 fancy!!!

PR #418, opened 2020-02-13
Ux 716 form element cleanup

tristanjasper:
### Purpose 🚀
Update storybook example for radio buttons usage to use Radio.Group

Allowing `value` and `name` to be set on radio component, as in the case of results it is scraped from the input on form submission.

~~Support rendering of "Instructions" title in the FormElement instructions sub component as this is a desired implementation in results and I expect could be for other consuming applications~~

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-716-form-element-cleanup

### Screenshots 📸
After
image

### References 🔗


packages/FormElement/src/components/Instructions/Instructions.js
```diff
@@ -1,21 +1,36 @@
...
+  const { children, hasTitle, ...moreProps } = props;
+  const I18n = useI18n();
 
   return (
     <sc.Instructions data-pka-anchor="form-element.instructions" {...moreProps}>
+      {hasTitle ? (
+        <Heading level={3} displayLevel={6} isLight>
+          {I18n.t("formElement.instructions")}
+        </Heading>
+      ) : null}
```
allison-c: Looks good to me! Isn't this already supported by `< Instructions />` component? Is this one of the design patterns?

allison-c: I mean developers can just pass the `<Heading>Instructions</Heading>` in the children of `< Instructions >`

tristanjasper: Not sure what you mean, currently there are no mocks for this sub component, and the component didn't have a title,  but I think it makes sense to include tbh. I don't really want to have the consumer start passing it it html in order to achieve this and we don't want to repeat that in all of our consuming applications

tristanjasper: Yes, we could but I don't think it's necessary for the consumer to do it if we can help it, I would think the title belongs in this component

tristanjasper: Perhaps we could allow customization of that title, or have the consumer pass it as a child, but i think just for convenience we can have the formElement component take of it

nahumzs:
is any reason that we need/want to add another prop?  it's possible to achieved the same from the consumer side in a really simple way. 
```jsx
const InstructionWithTitle = (props) =>  <FormElement.Instructions>Instructions {props.children}</FormElement.Instructions>

<FormElement>
  <InstructionWithTitle>My stuff</InstructionWithTitle>
</FormElement>
```
if the consumer really needs to have a reusable component is quite simple to achieve it over composition, I don't think we should add a prop to the component, that only toggle the Header `Instructions`

What do you think? @tristanjasper

tristanjasper:
> is any reason that we need/want to add another prop? it's possible to achieved the same from the consumer side in a really simple way.
> 
> ```js-jsx
> const InstructionWithTitle = (props) =>  <FormElement.Instructions>Instructions {props.children}</FormElement.Instructions>
> 
> <FormElement>
>   <InstructionWithTitle>My stuff</InstructionWithTitle>
> </FormElement>
> ```
> 
> if the consumer really needs to have a reusable component is quite simple to achieve it over composition, I don't think we should add a prop to the component, that only toggle the Header `Instructions`
> 
> What do you think? @tristanjasper

Well, i figured it would save on the each consumer having to pass the title as a child as i would expect the title could be a common use case, so it could be looked at from both angles. Its fine,  for now we can leave it to the consumer to pass it then.

packages/Radio/src/Radio.js
```diff
@@ -98,6 +102,7 @@ function Radio(props) {
     ref: inputRef,
     tabIndex,
     type: "radio",
+    value,
```
tristanjasper: In results questionnaires the value is scraped from the input as part of the form submission. Setting it on the input here allows us to retain this approach until the business logic for questionnaires submission is updated and changed.

packages/Radio/src/components/Group/Group.js
```diff
@@ -58,7 +58,7 @@ function Group(props) {
             isChecked: checkedIndex === index,
             isDisabled: isDisabled || child.props.isDisabled,
             canDeselect,
-            name,
+            name: child.props.name ? child.props.name : name,
```
tristanjasper: In results questionnaires the radio input also needs the name value set in order to scrape the value, so if it is present I am passing it onto the child radio component

tristanjasper: When the radio components are rendered as children of a group component they will all have the same `name` id set on them, which i believe is correct, however if there is a prop.child it will use that instead.

packages/Radio/src/Radio.js
allison-c: oh interesting, why not just pass it through `moreProps` like
```
const inputProps = {
    "aria-describedby": ariaDescribedBy,
    readOnly: true,
    onClick,
    checked: isChecked,
    disabled: isDisabled,
    id: radioId,
    name,
    onKeyDown: handleKeyDown,
    onKeyUp: handleKeyUp,
    ref: inputRef,
    tabIndex,
    type: "radio",
   ...moreProps,
}
```

tristanjasper: because `moreProps` is currently being spread to the container div on line 110, not the input so I don't think we want it on both.

packages/Radio/src/components/Group/Group.js
allison-c: `child.props.name ||  name,`

PR #419, opened 2020-02-13
STAR-146 Add styled-components linter

jamiek-acl:
### Purpose 🚀
Add styled-components linter. Linted files.

### Notes ✏️
Shouldn't change anything. 

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/star-146-styled-components-linter

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-146


package.json
```diff
@@ -9,6 +9,7 @@
     "jest:watch": "yarn jest --watch",
     "lint:fix": "yarn lint --fix",
     "lint": "eslint packages --report-unused-disable-directives",
+    "lint:css": "stylelint './packages/**/src/**/*.styles.js'",
```
nahumzs: `lint:jss` its not css per-se.

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -28,7 +28,7 @@ const itemStyles = css`
 
     colorString = isDestructive ? "orangeDarken10" : "colorBlack";
     return { color: tokens.color[colorString] };
-  }}}
+  }}
```
nahumzs: 🤔 this got remove but not opened `{` got removed?

jamiek-acl: Yeah there was one too many!

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
```diff
@@ -62,7 +61,7 @@ export const remove = css`
   display: inline-flex;
   font-size: 12px;
   height: 100%;
-  justify: flex-end;
+  justify-content: flex-end;
```
nahumzs: wondering if we should left out this fix since originally was written incorrectly, so that does mean is not being applied, with the fix now its apply.

.semaphore/semaphore.yml
```diff
@@ -15,11 +15,12 @@ blocks:
         commands:
           - checkout
           - cache restore client-node-modules-$SEMAPHORE_GIT_BRANCH-$(checksum yarn.lock),client-node-modules-$SEMAPHORE_GIT_BRANCH,client-node-modules-master
+          - nvm use 10.15.1
```
alexzherdev:
```suggestion
          - nvm use
```
this will just refer to the existing `.nvmrc`

packages/ListBoxBrowser/src/components/OptionsSelected/OptionsSelected.styles.js
jamiek-acl: You're right, it is not needed. I'll remove it altogether (along with another un-used style).

.semaphore/semaphore.yml
jamiek-acl: Nice, thanks!

PR #426, opened 2020-02-15
PD-14397 add custom width for pushing

oscarkwan:
### Purpose 🚀
- add another prop to be able to control push content width

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/PD-14397-add-custom-push-width

### DIFF
--- a/packages/SidePanel/src/SidePanel.js
+++ b/packages/SidePanel/src/SidePanel.js
@@ -24,6 +24,8 @@ const propTypes = {
   /** Y offset that is passed down from <SidePanel.Group> */
   groupOffsetY: PropTypes.number,
 
+  pushContentWidth: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+
   /** The width of the open panel. */
   width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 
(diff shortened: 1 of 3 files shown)


nahumzs:
Ok I now this in now wan to read, but seems like what you are doing is add a new feature to the SidePanel that is not even part of the original component, so I will request to create a new Subcomponent that can deal with this specific case and you can modify extend at your will.

So to me we should do something like:

```js
<SidePanel>
   <SidePanel.PushContent width={} getElementToPush={} />  {/* or other name */}
</SidePanel>
```

This way we prevent adding multiple props to our main use case for the sidepanel and adding complexity when learning the component.

oscarkwan:
> Ok I now this in not what you want to read, but it seems like what you are doing is adding a new feature to the SidePanel that is not part of the original component.
> 
> so I will request to create a new Subcomponent that can deal with this specific case and you can modify extend it at your will.
> 
> So to me we should do something like:
> 
> ```js
> <SidePanel>
>    <SidePanel.SplitPanel width={} getElementToPush={} />  {/* or another name */}
> </SidePanel>
> ```
> 
> This way we prevent adding multiple props to our main use case for the sidepanel and adding complexity when learning the component.
> 
> What do you think?

Sure, extracting it into another component could make sense!

nahumzs:
Closing for now, not relevant atm. cc @oscarkwan

PR #430, opened 2020-02-19
UX-718 slide from left sidepanel

oscarkwan:
### Purpose 🚀
- Add option to slide panel from the left side

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-718-sidepanel-slide-direction

### DIFF
--- a/packages/SidePanel/src/SidePanel.js
+++ b/packages/SidePanel/src/SidePanel.js
@@ -53,6 +53,9 @@ const propTypes = {
 
   /** Render the sidepanel inline */
   isInline: PropTypes.bool,
+
+  /** Slide direction of the sidepanel */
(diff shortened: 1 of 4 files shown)


nahumzs:
ok, what about if instead of adding slide direction, we just add a prop like `slideDirectionLeft` that way we are not introducing a breaking change. sliding from the right is always default and will keep working, ad slideDirectionLeft is by default false.

oscarkwan:
> ok, what about if instead of adding slide direction, we just add a prop like `slideDirectionLeft` that way we are not introducing a breaking change. sliding from the right is always default and will keep working, ad slideDirectionLeft is by default false.

Hm, But it shouldn't be breaking change if slideDirection has default prop of right?

PR #434, opened 2020-02-24
STAR-156 DateRangePicker component

DanilAgafonov:
### Purpose 🚀
Introduce DateRangePicker component

### Updates 📦
- [x] New component `@paprika/date-range-picker`
- [x] MINOR (backward compatible) change to `@paprika/date-input`
- [x] MINOR (backward compatible) change to `@paprika/icon`
- [x] MINOR (backward compatible) change to `@paprika/calendar`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-156-date-range-picker/?path=/story/forms-daterangepicker--showcase

### Screenshots 📸
image

### References 🔗


packages/DateInput/src/DateInput.js
```diff
@@ -147,15 +157,17 @@ const DateInput = React.forwardRef((props, ref) => {
     }
   }
 
-  function handleInputBlur() {
+  function handleInputBlur(e) {
+    onBlur(e);
```
mikrotron, line 161: Instead of exposing `onBlur` as part of the API for this component, what do you think of this technique used in the `<Biutton>` where we just peek into `moreProps` to see if it's being passed in?

https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.js#L146-L151

packages/DateRangePicker/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@paprika/date-input": "^0.1.0",
+    "@paprika/helpers": "^0.2.10",
+    "@paprika/icon": "^0.2.15",
+    "@paprika/popover": "^0.3.18",
+    "@paprika/stylers": "^0.2.6",
+    "@paprika/tokens": "^0.1.13",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "@paprika/l10n": "^0.3.13",
```
mikrotron, line 27:
```suggestion
    "@paprika/l10n": "^0.4.2",
```

packages/DateRangePicker/src/DateRangePicker.js
```diff
@@ -0,0 +1,243 @@
...
+    } = {},
+    "DateRangePicker.Popover": { props: popoverProps = {} } = {},
+  } = extractChildren(children, [
+    "DateRangePicker.Input",
+    "DateRangePicker.StartInput",
+    "DateRangePicker.EndInput",
+    "DateRangePicker.Popover",
+  ]);
+
+  function isAnyElementContainsFocus() {
```
mikrotron, line 78: The grammar of this is wrong. I still like it though – its intention is _clear_, which is the important thing, and the `is` prefix makes its boolean return value obvious.

packages/DateRangePicker/src/DateRangePicker.styles.js
```diff
@@ -0,0 +1,38 @@
...
+import OriginalDateInput from "@paprika/date-input";
+import tokens from "@paprika/tokens";
+import stylers from "@paprika/stylers";
+
+export const Wrapper = styled.div`
+  display: flex;
+`;
+
+export const DateInput = styled(OriginalDateInput)`
+  flex: 1 1 auto;
```
mikrotron, line 11:
```suggestion
  flex-grow: 1;
```

packages/DateRangePicker/src/DateRangePicker.styles.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const PopoverCardContent = styled.div`
+  display: flex;
+`;
+
+const CALENDAR_INTERNAL_PADDING = 9;
+
+export const CalendarsWrapper = styled.div`
+  display: flex;
+  margin: 0 calc(-${stylers.spacer(3)} - ${CALENDAR_INTERNAL_PADDING}px);
```
mikrotron, line 33:
```suggestion
  margin: 0 ${-1 * stylers.toInt(stylers.spacer(3)) - CALENDAR_INTERNAL_PADDING}px;
```

mikrotron: Better to avoid the CSS `calc()` if possible?

packages/DateRangePicker/src/createPropsCollector.js
```diff
@@ -0,0 +1,9 @@
+export default function createPropsCollector(displayName) {
```
mikrotron, line 1: Oh! This might be a hand utility to have in the `helpers` package?

packages/DateRangePicker/test/DateRangePicker.cypress.js
```diff
@@ -0,0 +1,54 @@
+import moment from "moment";
+
+describe("<DateRangePicker />", () => {
+  const targetStartDate = moment();
+  const targetEndDate = moment(targetStartDate)
+    .add(1, "month")
+    .date(1);
+
+  beforeEach(() => {
+    cy.visitStorybook("forms-daterangepicker-backyard-tests--cypress");
```
mikrotron, line 10: I added a `getStoryUrlPrefix()` helper to derive this URL from the component name via `storyTree`:
https://github.com/acl-services/paprika/blob/master/packages/DatePicker/test/DatePicker.cypress.js#L11

Alternatively, if you write the story in the new CSF style, you can add in `id` and specify a permalink (in case the component moves to another category).   

https://github.com/acl-services/paprika/blob/master/packages/SidePanel/stories/SidePanel.tests.stories.js#L15

packages/Icon/src/svg/Arrow-Right-B.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="480" height="480" viewBox="0 0 480 480">
```
mikrotron, line 1: Can we bring in `ArrowLeftB` too, just for completeness?

packages/Input/src/Input.styles.js
```diff
@@ -30,7 +30,15 @@ const inputStyles = css`
 
     background-color: ${tokens.color.white};
     border: 1px solid ${tokens.border.color};
-    border-radius: ${tokens.border.radius};
+    border-radius: ${({ hasBorderRadius }) => {
```
mikrotron, line 33: I think I would rather have the `<DateRangePicker>` override the styling of the `<Input>` instead of introducing this additional complexity here.  Similar to the way the `<ButtonGroup>` overrides the styling of the `<Buttons>` it is composed of:

https://github.com/acl-services/paprika/blob/master/packages/ButtonGroup/src/ButtonGroup.styles.js

packages/DateRangePicker/test/DateRangePicker.cypress.js
```diff
@@ -0,0 +1,54 @@
+import moment from "moment";
+
+describe("<DateRangePicker />", () => {
```
mikrotron, line 3: Curious about why these tests needed to by in Cypress, instead of Jest like the rest of them?

mikrotron: Also, minor point, but the containing folder should probably be `tests` instead of `test`.

mikrotron:
Looking good! This is a complex component, but you've kept the implementation relatively simple.

I noticed some odd behaviour though when: 
- I select an end date that is _before_ the start date
- I select a date from the opposite calendar to the corresponding input in focus (ie the end date input is focussed, but I pick from the left calendar, or vice versa).  

Was there a prototype or design document that this was based on? It might be good to have a designer also review the interaction (maybe @sunminsarahlee?).  If there are additional cosmetic adjustments that are requested, we could certainly defer those to a subsequent PR, but I'd like to make sure the interaction patterns are as intended.

DanilAgafonov:
@mikrotron

> Was there a prototype or design document that this was based on?

Yes: https://aclgrc.atlassian.net/wiki/spaces/UX/pages/814776926/Date+Range+Picker+Behaviour+in+review

Thanks for all comments! I will address your comments asap.

packages/DateRangePicker/src/DateRangePicker.styles.js
DanilAgafonov: Oh, cool. `stylers.toInt` was what I was looking for!

packages/DateRangePicker/src/createPropsCollector.js
DanilAgafonov: I think so, but it is a topic I want to discuss. Introducing this helpers is just my little experiment. I see 2 issues with it:
1. It doesn't work with our README.md generator.
2. Due to components created with `createPropsCollector` have no propTypes, there is no autocompletion  for consumers. I mean when someone will install `DateRangePicker` component and will try to start typing `<DateRangePicker.Input `, there will not be any suggestions for props in editor/IDE. And I think this not solvable, because editors/IDEs can inspect only statically declared components. But we generate props collector component dynamically with this helper.

Probably, I should refactor this PR to get rid of this helper and declare subcomponents statically.

packages/Icon/src/svg/Arrow-Right-B.svg
DanilAgafonov: Actually, I am a little bit confused all icon-related things.

We have `wasabicons` and `@paprika/icon`.

I remember someone mentioned in Slack that `@paprika/icon` shouldn't be consumed directly in Galvanize platform. Is true? Because I see a lot of usages in `workpapers` repo. If it actually shouldn't be consumed directly, should we keep brigades in touch? Can we technically (not just by convention) prevent direct usage of this package?

Due to the fact we should not use `@paprika/icon` in Galvanize apps, should I really add `ArrowLeftB` to this package?

cc: @nahumzs

packages/DateRangePicker/test/DateRangePicker.cypress.js
DanilAgafonov: It is hard to answer this question, because it is just like a feeling and not rational point. But I will try to explain.

1. It feels like there are a lot of visual things (not behavioral) (like floating elements). It can't be properly tested with Jest, because it requires real browser to render.
2. This tests are more like user scenarios. It feels that Cypress is more suitable for testing user scenarios.

But I have not strong opinion on it. So we can convert it to Jest if you think it is good idea.

DanilAgafonov:
> I select an end date that is before the start date – it clears the end date and closes the popover.

Hmmm, it is strange. I can't reproduce it.

> I choose a date on the right calendar for start, and on the left for end, I get into a state where the inputs can be populated with an end date before the start date, and more awkwardly, the popover remains open and clicking away does not close it. Pretty edge-casey, but it might indicate a bug that could sneak up in other situations?

Actually, there is no binding like "left calendar for start date, right calendar for end date". And actually I am not fan of this behavior. It always confuses me and feels a little bit illogical. I was talking to Jayde and John about it and tried to convince they avoid this behavior, but they gave a lot of comments why it will be better to keep it. Despite I don't like this behavior, I feel it reasonable to follow designer's recommendations.

But the fact that user can select end date prior to start date definitely is not ok. And I think we should fix it. But not in this PR.

nahumzs:
I found a small bug whenever the user want to switch dates using the long range year picker, the dropdown that its at the top of the calendar, see gif attached.

The bug: is not possible to select a date if you use that control.

packages/DateRangePicker/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@paprika/date-input": "^0.1.0",
+    "@paprika/helpers": "^0.2.10",
+    "@paprika/icon": "^0.2.15",
+    "@paprika/popover": "^0.3.18",
+    "@paprika/stylers": "^0.2.6",
+    "@paprika/tokens": "^0.1.13",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "@paprika/l10n": "^0.4.2",
```
nahumzs, line 27: you are missing core-js dependency, while will work without it because most of our apps already installed, it should be a peer-dependency, https://github.com/acl-services/paprika/blob/master/packages/Button/package.json#L17

packages/DateRangePicker/src/DateRangePicker.js
```diff
@@ -0,0 +1,238 @@
...
+
+const propTypes = {
+  /** Selected start date in moment object */
+  startDate: PropTypes.instanceOf(moment),
+
+  /** Selected end date in moment object */
+  endDate: PropTypes.instanceOf(moment),
+
+  /** Callback to fire when user select start or end date */
+  onDatesChange: PropTypes.func.isRequired,
```
nahumzs, line 26: I would suggest to uses `onChange` instead of onDateChange, onChange is what mostly all our component uses and easy to remember, if you feel strong about using your approach, then I would suggested switching to onChangeDates which followed our guidelines https://github.com/acl-services/paprika/wiki/Coding-Conventions#function-names

packages/DateRangePicker/src/DateRangePicker.js
```diff
@@ -0,0 +1,238 @@
...
+  children: null,
+};
+
+const DateRangePicker = ({ startDate, endDate, onDatesChange, children }) => {
+  const [shouldShowPopover, setShouldShowPopover] = React.useState(false);
+  const [currentFocus, setCurrentFocus] = React.useState(null);
+  const [possibleStartDate, setPossibleStartDate] = React.useState(startDate || moment());
+  const [possibleEndDate, setPossibleEndDate] = React.useState(endDate || moment().add(1, "months"));
+
+  const popoverContentRef = React.useRef(null);
```
nahumzs, line 43: super nip: normally we use `ref` as prefix instead of subfix but is totally `nip` suggestion

packages/DateRangePicker/src/DateRangePicker.js
```diff
@@ -0,0 +1,238 @@
...
+      startDateInputRef.current.focus();
+    } else if (type === END_DATE) {
+      endDateInputRef.current.focus();
+    } else {
+      hidePopover();
+    }
+  }
+
+  function onChangeStartDate(newDate) {
+    onDatesChange({ startDate: newDate, endDate });
```
nahumzs, line 123: yeah I think now that I see the response from this callback, would be beneficial to named `onChange` as keep consistency on the library:
```ts
 onChange({ startDate, endDate } : { starDate: moment.Moment, endDate: moment.Moment }): void
```

packages/DateRangePicker/src/DateRangePicker.js
```diff
@@ -0,0 +1,238 @@
...
+        />
+      </sc.Wrapper>
+
+      <Popover.Content ref={popoverContentRef}>
+        <Popover.Card>
+          {shouldShowPopover && (
+            <sc.PopoverCardContent>
+              <sc.CalendarsWrapper>
+                <sc.CalendarWrapper data-pka-anchor="daterangepicker.leftcalendar">
+                  <Calendar
```
nahumzs, line 199: I think was a good idea to split the `<Calendar />` to its own component reduced a lot of complexity while reading the code.

packages/DateRangePicker/src/createPropsCollector.js
nahumzs: Totally agree, props declaration should be as simple as possible abstractions should be avoided if possible, duplication for this kind of scenario is allowed so we can help static analyzer to find the better information for our users.

packages/DateRangePicker/stories/DateRangePicker.stories.js
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    const [{ startDate, endDate }, setDates] = React.useState({ startDate: null, endDate: null });
+
+    moment.locale("en");
+
+    return (
+      <L10n locale="en">
+        <DateRangePickerStory>
+          <DateRangePicker startDate={startDate} endDate={endDate} onDatesChange={setDates} {...dateRangePickerProps}>
```
nahumzs, line 40: I know this allowed the author to write more stories faster and with less duplication, but the downside is that if I'm a consumer I can't understand without digging into the code how to use this component, I would encourage maybe to add a simple story without spreading showcasing how to use it. What do you think? 

Also you can create and github issue asking someone else to write the stories we can asked in a starling day to ux-dev to help us with this. I think its time to merge and keep moving into new features :)

packages/Icon/src/svg/Arrow-Right-B.svg
nahumzs: I think is fine to add just what we need to paprika/icons, when the ArrowLeft its need it, we copy/paste it to paprika. 

The reason behind this confusion is that most of our iconset is a paid library, so whenever we want to use and icon in paprika (our open source library) we need to redo it so we don't fall into copyright issues. 

And you are correct we shouldn't use paprika icon on our highbond apps, we should use wasabicons (our paid iconset plus paprika icons mixed). 

At this moment don't have any idea how to prevent consumers to use paprika icon outside of paprika itself.

allison-c:
A small comment for the UI doesn't match the design. In the storybook, there's a border around the arrow icon. I can help you to polish some small details like this, let me know if you need me.

image

sunminsarahlee:
@DanilAgafonov I left this comment in ticket but I am leaving this message here again in case. There is design change in DateRange Picker. Please take a look this design and let me know if you have any question. We need to add grey heading colour in the calendar. Colour is #F0F0F0.
Screen Shot 2020-08-19 at 3 14 11 PM

PR #439, opened 2020-02-27
Star 165 takeover focus lock api

tristanjasper:
### Purpose 🚀
Ensuring focusLock api in takeover component is consistent with SidePanel and Modal component

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to SidePanel, Modal, Takeover

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/star-165-takeover-focusLock-api

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-165


packages/Modal/src/Modal.js
```diff
@@ -71,8 +71,8 @@ const Modal = props => {
       onClose={onClose}
       onAfterOpen={onAfterOpen}
       onAfterClose={onAfterClose}
-      {...overlayProps}
       focusLockOptions={focusLockOptions}
+      {...overlayProps}
```
tristanjasper: I would think these props should be applied to override the focusLockOptions....

packages/Takeover/src/Takeover.styles.js
```diff
@@ -3,6 +3,19 @@ import tokens from "@paprika/tokens";
...
+  flex-direction: column;
+  height: 100%;
+
+  &::before,
+  &::after {
+    content: "";
+    display: block;
+  }
+`;
+
```
tristanjasper: I _think_ these styles are correct, although the `modal` components styles also have a `flex-basis` style which I'm unsure is needed here - https://github.com/acl-services/paprika/blob/master/packages/Modal/src/Modal.styles.js#L48

packages/Takeover/src/components/FocusLock/FocusLock.js
```diff
@@ -0,0 +1,17 @@
...
+  returnFocus: true,
+};
+
+const FocusLock = props => <FocusLockLibrary {...props} />;
+
+FocusLock.propTypes = propTypes;
+FocusLock.defaultProps = defaultProps;
+FocusLock.displayName = "Takeover.FocusLock";
+
+export default FocusLock;
```
tristanjasper: I'm wondering if perhaps this component should be somehow be a shared import or even its own package, but not completely convinced on where it belongs as it is the same component across sidepanel, takeover and modal now

nahumzs: Totally fine, duplication is not intrinsically bad, for this scenario I think we are fine.

packages/Modal/src/Modal.js
nahumzs: why would be that? isn't we are having `<Modal.FocusLock  />` in charge of overriding the props?

packages/Takeover/README.md
```diff
@@ -1,31 +1,11 @@
-# Takeover
```
nahumzs: why are you removing the basic Readme example? any specific need to do this? the change log should be a different file

tristanjasper: Ah wrong file removed.

packages/Modal/src/Modal.js
tristanjasper: Ok, I see, I wasn't sure of the priority

PR #445, opened 2020-03-05
STAR-174 Fix Icon colour 🐐

mikrotron:
### Purpose 🚀
Ensure that all icons can be coloured with the CSS `color` property.

### Notes ✏️
- Removed generated artifacts (icon components / `js` files) from repo and added a `pretranspile` script to generate icons during build.
- Renamed `Filter.svg` to `Filter-Alt.svg` so it won't conflict with filter icon in `Wasabicons`. 
- Adjusted Storybook story so that long icon names break to multiple lines instead of overflowing.
- Also added a `size` and `color` knob to Storybook story.
- Sister PR in `wasabicons`: https://github.com/acl-services/wasabicons/pull/8

### Updates 📦
- [x] PATCH (bug fix) change to `Icon`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-174--fix-icon-color/?path=/story/icon--all-icons

### Screenshots 📸
Screen Shot 2020-03-05 at 3 01 17 PM

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-174
https://github.com/acl-services/wasabicons/pull/8


packages/Icon/stories/examples/AllIcons.js
```diff
@@ -1,20 +1,34 @@
...
+import tokens from "@paprika/tokens";
+import Heading from "@paprika/heading";
 import * as sc from "./AllIcons.styled";
 
-export default function AllIcons() {
+const iconSizes = {
+  small: stylers.spacer(2),
+  medium: stylers.spacer(3),
+  large: stylers.spacer(4),
+};
```
mikrotron: This aligns with Jonathan's new pitch for icon sizing that is still `WIP`.

packages/Icon/.svgrrc
```diff
@@ -1,7 +1,11 @@
 {
   "icon": true,
   "replaceAttrValues": { 
-    "#717171": "currentColor"
+    "#717171": "currentColor",
+    "#757575": "currentColor",
+    "#767676": "currentColor",
+    "#444": "currentColor",
+    "#3F3D3C": "currentColor"
```
mikrotron: This is the main fix.

packages/Icon/stories/examples/AllIcons.js
```diff
@@ -1,20 +1,34 @@
...
+
+export default () => {
+  const iconProps = {
+    size: select("size", iconSizes, stylers.spacer(3)),
+    color: color("color", tokens.textColor.icon),
+  };
+
   const req = require.context("../../src", false, /\.js$/);
```
allison-c: why is this still working? I thought you have to import from `../../lib` since you removed those js files from `src`

mikrotron: This script runs beforehand, and will create all those `.js` files in `/src`:
https://github.com/acl-services/paprika/blob/908f89ee9d59397a280264f74da96b308aa61469/packages/Icon/package.json#L17

alexzherdev: should those be gitignored then?

mikrotron: sure!

mikrotron: @alexzherdev 
Tokens ignores are in the root... should these also go in the same file? 
https://github.com/acl-services/paprika/blob/master/.gitignore#L36-L40
Or do you think both of  should be in a local `.gitignore`?

alexzherdev: both have pros and cons... I'd probably put them locally. Make sure that publishing still works. There is an interplay between `.gitignore` and `.npmignore` but AFAICT we should be fine https://docs.npmjs.com/using-npm/developers.html#keeping-files-out-of-your-package

PR #446, opened 2020-03-06
chore(DataGrid): UX-633-adding-npm-ignore missing npmignore file

nahumzs:
### Purpose 🚀
adding missing .npmignore files

### Notes ✏️

### Updates 📦
- [x] N/A

### Storybook 📕

### Screenshots 📸

### References 🔗


mikrotron:
If you're doing some ignore chores... what about moving these `git` ignore rules to a local `.gitignore`?
https://github.com/acl-services/paprika/blob/master/.gitignore#L36-L40

Also, I think we could use a `gitignore` in `l10n`, to ignore the generated `locale.js` files?

nahumzs:
right now I'm only focusing on this because directly apply to the DataGrid, I can do the Tokens ones is quite straight forward for the l10n is a little more laborious work because those files are already on git, so I will skip it for now. :)

PR #448, opened 2020-03-06
Trying to fix consumption of L10n component

jamiek-acl:
### Purpose 🚀
https://aclgrc.atlassian.net/browse/UX-728

The L10n component doesn't seem to work correctly, even though it looks like it's working in Storybook (but Storybook loads more than just the one story you are looking at, which is possibly why it works).

Here is the DatePicker in Projects\Certifications when my locale is `fr` (it may be falling back on some other L10n/context?)
before

Or if you make a simple app, it does not translate it correctly. Here is a CodeSandbox: https://codesandbox.io/s/cool-wind-4h9im

    import React from "react";
    import moment from "moment";
    import L10n from "@paprika/l10n";
    import DatePicker from "@paprika/date-picker";

      return <h1>hello</h1>;

      /*
      moment.locale("zh-cn"); // need to set moment locale for react-dates calendar

      return (
        <L10n locale="fr">
            <DatePicker onChange={() => {}}>
              <DatePicker.Input />
              <DatePicker.Popover />
            </DatePicker>
          </L10n>
      );
      */
    }

Outputs an error:

    ModuleNotFoundError
    Could not find module in path: '@paprika/l10n/lib/locales-js/en.js' relative to '/node_modules/@paprika/l10n/lib/locales/index.js'

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/testing-translation/?path=/story/datepicker--datepicker-with-locale
This doesn't seem to translate... but running Paprika's Storybook locally does translate correctly.

### Screenshots
A simple react app, using only `L10n` and `DatePicker` components:

**Before:**
Screen Shot 2020-03-09 at 9 57 01 AM

**After:**
after


nahumzs:
this will make codesandbox to work again awesome :D

jamiek-acl:
@nahumzs this isnt about codeSandbox, there are places in our app where the translations weren't working (Certifications \ DatePicker, Add Issue \ Date Picker).

PR #457, opened 2020-03-13
STAR-156 Extract DateInput as separate component

DanilAgafonov:
### Purpose 🚀
This PR done as preparation to build DateRangePicker. `DateInput` should be reused in `DatePicker` and `DateRangePicker`.

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/date-picker`
- [x] MINOR change to `@paprika/helpers`
- [x] New component `@paprika/date-input`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-156-extract-date-input

### References 🔗


DanilAgafonov:
On next Friday:
1. will add more Storybook stories
2. will add tests (extract some tests from `DatePicker`)
3. will add `README.md`
4. will add comments for `propTypes`

packages/DateInput/package.json
```diff
@@ -0,0 +1,29 @@
...
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/helpers": "^0.2.9",
+    "@paprika/icon": "^0.2.12",
+    "@paprika/input": "^0.2.15",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "@paprika/l10n": "^0.2.5",
```
jamiek-acl: the latest is 0.3.13

packages/DateInput/src/DateInput.js
```diff
@@ -0,0 +1,186 @@
...
+    denyConfirmation,
+    beforeConfirmation,
+    ...restProps
+  } = props;
+
+  const [hasParsingError, setHasParsingError] = React.useState(false);
+  const [inputtedString, setInputtedString] = React.useState(date ? date.format(dateFormat) : "");
+  const [hasFocus, setFocus] = React.useState(false);
+
+  // Effect
```
jamiek-acl: Need this comment?

packages/DateInput/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "moment": "^2.18.1",
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4"
```
nahumzs: missing styled-component as peer dependency isn't?

packages/DateInput/src/DateInput.js
```diff
@@ -0,0 +1,186 @@
...
+    setInputtedString("");
+    onChange(null);
+  }
+
+  function parseInput() {
+    let newDate = moment(inputtedString, dateFormat);
+
+    if (!newDate.isValid()) newDate = moment(inputtedString);
+
+    return newDate;
```
jamiek-acl: Small suggestion:
`return newDate.isValid() ? newDate : moment(inputtedString);`

packages/DateInput/test/DateInput.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    rerender: updatedProps => {
+      rerender(<DateInput {...updatedProps} onChange={handleChange} data-pka-anchor="dateinput" />);
+    },
+  };
+}
+
+describe("DateInput", () => {
+  it("should display empty in input by default", () => {
+    const { getByTestId } = render();
+
```
jamiek-acl: Seems like too much whitespace in these tests since they are only 2 or 3 lines

packages/DateInput/src/DateInput.js
```diff
@@ -0,0 +1,186 @@
...
+  hasError: PropTypes.bool,
+
+  /** Date format used while entering and parsing user input. */
+  dateFormat: PropTypes.string,
+
+  /** Selected date in moment object. */
+  date: isMomentObjectOrNull,
+
+  /** Possible date in moment object. This prop is used when user is typing date and we can't determinate exact date. */
+  possibleDate: isMomentObjectOrNull,
```
nahumzs: I'm a little confused, why would you want to pass-down a `possibleDate` to the component shouldn't this be something like onPossibleDate callback, like what you would receive after changing the date prop? 

Also **NO** for this PR but would be cool if we could have a `useDateInput` hook in some form of:

```
import { useDateInput } from "@paprika/date-input"
const { date, possibleDate } = useDateInput(defaultDate);
```
we should provide this uncontrolled option for all our components, our regular input would benefit having a similar pattern as well. 

cc @mikrotron @tristanjasper

packages/DateInput/src/DateInput.js
```diff
@@ -0,0 +1,186 @@
...
+  /** Callback when user inputs date. Will be called after every key up event. */
+  onChangePossibleDate: PropTypes.func,
+
+  /** Error callback. Will be called on blur or enter key press if inputted date can't be parsed. */
+  onError: PropTypes.func,
+
+  onClick: PropTypes.func,
+
+  /** Guard function. If it returns true - confirmation will be prevented. */
+  denyConfirmation: PropTypes.func,
```
nahumzs: out of curiosity in what scenario we would be using this prop?

packages/DateInput/src/DateInput.js
```diff
@@ -0,0 +1,186 @@
...
+  /** Callback when confirm */
+  beforeConfirmation: PropTypes.func,
+};
+
+const defaultProps = {
+  hasError: false,
+  dateFormat: "MM/DD/YYYY",
+  date: null,
+  possibleDate: null,
+  humanFormat: undefined,
```
nahumzs: why cannot be null moment conflict?

nahumzs:
can you expand on why this is a breaking change for the DatePicker?

packages/DateInput/src/DateInput.js
mikrotron: I suspect this was just moved over from the previous version.  I remember @allison-c explaining this to me when she initially added it, and it was important for the formatting, but I don't remember the exact details.  I don't think the purpose of this PR is to revisit the working logic though, so it might be best to leave it as is?

packages/DateInput/package.json
mikrotron: Looks like there is no `DateInput.styles.js`. I presume this component gets all the styling it needs from `Input`.

packages/DateInput/src/DateInput.js
mikrotron: Well, if it was set to `null` then the default value (from i18n) would not be set:
https://github.com/acl-services/paprika/pull/457/files#diff-1f6beb3c48ccc9dff8f817d5d81f1732R67

packages/DatePicker/src/DatePicker.js
```diff
@@ -1,30 +1,24 @@
...
 import Popover from "@paprika/popover";
-import useI18n from "@paprika/l10n/lib/useI18n";
+import DateInput from "@paprika/date-input";
 import useDebounce from "@paprika/helpers/lib/hooks/useDebounce";
-import usePrevious from "@paprika/helpers/lib/hooks/usePrevious";
 import isElementContainsFocus from "@paprika/helpers/lib/dom/isElementContainsFocus";
 
 import Calendar from "./components/Calendar";
-import DateInput from "./components/DateInput";
+import DateInputPropsExtractor from "./components/DateInput";
```
mikrotron: Have we been using this `PropsExtractor` naming convention elsewhere? I agree it's purpose is more clear than if it were `DatePickerInput`, and I'm not usually a fan of something as generic as `Container`, but maybe `PropsContainer` is a better description? `PropVessel` if you prefer something shorter/weirder 😁?

packages/DatePicker/src/helpers.js
```diff
@@ -1,5 +1,4 @@
 import React from "react";
-import momentPropTypes from "react-moment-proptypes";
 
 export function extractChildrenProps(children, target) {
```
mikrotron: This is also a nice, generic helper we could use elsewhere (I think). Can we also migrate it to the `helpers` package?

packages/DatePicker/src/DatePicker.js
```diff
@@ -201,19 +103,19 @@ function DatePicker(props) {
...
 
-  const hasInputError = extendedInputProps && extendedInputProps.hasError;
+  function handlePossibleDateChange(newPossibleDate) {
+    if (newPossibleDate.isSame(possibleDate, "year") && newPossibleDate.isSame(possibleDate, "month")) return;
+
+    setPossibleDate(newPossibleDate);
+  }
 
-  const hasErrorValue = hasError || hasParsingError || hasInputError;
+  const hasInputError = extendedInputProps && extendedInputProps.hasError;
```
mikrotron: Might be nice if extractChildrenProps returned `{}` instead of undefined, so we could skip the defensive check here.

packages/DateInput/test/DateInput.spec.js
DanilAgafonov: Oh, this is just my habit to write tests in AAA style (Arrange, Act, Assert). Can we leave at is? Probably, I can share tips about AAA in next ShopTalk.

packages/DateInput/src/DateInput.js
DanilAgafonov: @nahumzs 
> why would you want to pass-down a possibleDate to the component 

Oh, thanks! I just realized I do not use this prop inside DateInput. Will remove it and leave here only `onChangePossibleDate` callback.

packages/DateInput/src/DateInput.js
DanilAgafonov: @nahumzs 

"confirmation" in context of DateInput - is a moment when inputted date is converting from `dateFormat` to `humanFormat` and `onChange` callback is calling. Normally "confirmation" occurs on DateInput blur. But there is specific scenario:
1. we use DateInput inside DatePicker
2. we click in DateInput and calendar should be shown
3. If we click somewhere on blank space Calendar should be hidden and "confirmation" should be triggered. But if we click on shown Calendar, we should NOT trigger "confirmation".

But we have problem: DateInput does NOT know anything about shown Calendar. So I decided to pass this function from DatePicker to DateInput, to prevent "confirmation".

I am 100% sure that this is not ideal solution. Especially because we should call passed down `denyConfirmation` function via `requestAnimationFrame`. It is required to delay resolving `denyConfirmation`, because input blur handler will be called before calendar receive a focus. By using `requestAnimationFrame` we check should "confirmation" be called or not only after calendar will receive focus.

DanilAgafonov:
@nahumzs answering you question:
> can you expand on why this is a breaking change for the DatePicker?

Normally I mark commit with breaking changes with `!` sign and leave some description in commit message. Like here: https://github.com/acl-services/paprika/pull/457/commits/9032c23d29b9e730122cb4a3e5e1e638373397c8

In this case I renamed `dataFormat` props in DatePicker to `dateFormat`. I guess this was typo. By introducing this change we prevent end-users of DatePicker component from safe update. It is required to review application code (where DatePicker is used) and rename `dataFormat` to `dateFormat`.

packages/DatePicker/src/DatePicker.js
DanilAgafonov: Original idea of renaming this component: because we import actual DateInput here, we can't keep the same name for this component.

I just renamed it to `DateInputPropsContainer`. Thanks, @mikrotron.

As of I am not native speaker it is hard for to choose correct term. Another options that I can suggest is `DateInputPropsCollector`. WDYT about it?

packages/DatePicker/src/helpers.js
DanilAgafonov: @mikrotron Good idea. But I have question. After I extract this function to `helpers` package, should I bump version in `packages/helpers/package.json` and declare this version as minimal required version in `packages/DateInput/package.json`?

nahumzs: we already have a generic way for extracting the children on the helpers, I remember @allison-c  modify that one in this component because we didn't have it on our helper's functions. 

https://github.com/acl-services/paprika/blob/master/packages/helpers/src/extractChildren.js

nahumzs:
oohh I like that idea about the `!` didn't find anything on the conventional commit guide about that, its that a personal preference? 💯

DanilAgafonov:
@nahumzs This idea from conventional commit guide (https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-to-draw-attention-to-breaking-change).

packages/DatePicker/src/DatePicker.js
```diff
@@ -1,30 +1,24 @@
...
 import { calendarPopoverStyles } from "./DatePicker.styles";
 
-const INPUT_PARSE_ERROR = "INPUT_PARSE";
-
 const propTypes = {
   children: PropTypes.node,
 
   /** Date format used while entering and parsing user input. */
-  dataFormat: PropTypes.string,
+  dateFormat: PropTypes.string,
```
allison-c: We named it `dataFormat` because it's used when parsing the data from string to moment date, it's not a typo 😂 
Maybe people already discussed renaming it and I missed

DanilAgafonov: Oooohhh... Now I think I should revert changes related to renaming this prop. @allison-c @nahumzs WDYT? Should I?

DanilAgafonov: But it is very strange. I see comment just above this line: `/** `**Date format** `used while entering and parsing user input. */`. But prop itself called `dataFormat`.

allison-c: Let's ask @nahumzs 😂

allison-c: The description for `humanFormat` is similar `/** Date format used while displaying date. It should be human-friendly and spelled out, default is MMMM DD,YYYY */`

But I can see if people feel confusing, we probably can rename it..

packages/helpers/src/customPropTypes.js
```diff
@@ -70,3 +71,9 @@ InputValidTypes.ALL = Object.values(InputValidTypes);
 
 export const RefOf = (propType = PropTypes.instanceOf(Element)) =>
   PropTypes.oneOfType([PropTypes.func, PropTypes.shape({ current: propType })]);
+
+export function isMomentObjectOrNull(date) {
+  if (date === null) return true;
+
+  return momentPropTypes.momentObj(date);
```
alexzherdev: Any advantage here over using `moment.isMoment`? (and thus not needing a new dependency)

DanilAgafonov: Hmmm, I will check. But actually this code just moved from DatePicker helpers. But I agree it will be good not just move code, but to rethink it.

allison-c: `momentPropTypes.momentObj` will check if it is a valid moment object or not, `moment.isMoment` doesn't

allison-c: but maybe it's not necessary

allison-c: I used it because `react-dates` used it and we're using react-dates 😂

packages/DatePicker/src/DatePicker.js
nahumzs: to me seems more natural `dateFormat` as Danil is proposing.

But also we don't need to break the api we could just deprecate the prop and add dateFormat, that way we don't break the API and previous components can still keep working without problem.

I found a propType function that could help us with this https://github.com/thefrontside/deprecated-prop-type/blob/master/deprecated.js

cc @mikrotron

allison-c: This PR is already a breaking change for date picker, so I think we can break the api maybe?

packages/DatePicker/src/helpers.js
mikrotron: @DanilAgafonov Don't worry, `lerna` should take care of all the versioning for us when we publish it!

packages/DatePicker/src/DatePicker.js
mikrotron: I think either name will work fine – I prefer both to `Extractor`. 

I actually like `Collector` better personally, just because `Container` gets used for a lot of things, and so is almost meaningless.  But since this is internal to the component, it's not a big deal. You can choose what you like the best.

packages/DatePicker/src/DatePicker.js
mikrotron: I'm reluctant to maintain deprecated props for any of our components that are `< v1.0`.  We should probably start releasing some `v1.0` components, but if it's `0.x` then that suggests the API isn't quite stable, and we shouldn't worry too much about introducing breaking changes.

DanilAgafonov: @nahumzs I agree with Mike. I think we can introduce this breaking change without deprecation warnings.

-------

@allison-c 
> This PR is already a breaking change for date picker, so I think we can break the api maybe?

Actually, changing this prop name is the only one breaking change for DatePicker. There is no other breaking changes.

nahumzs: ok I'm good with that.

PR #459, opened 2020-03-13
L10N-2020-03-13-22785 - translations batch

alexandermccate:


packages/L10n/src/locales/ja.yml
```diff
@@ -4,6 +4,8 @@ ja:
     remove: 削除
     moreInformation: 詳しい情報
     close: 閉じる
+    dataGrid:
+      a11yTextMessage: "{{value}}. 行 {{rowIndex}}、列 {columnIndex}} の位置にいます。次の情報は無視してください。"
```
cristinajimenez: Missing opening bracket in: {columnIndex}}

alexandermccate: Good catch!

alexandermccate: @cristinajimenez fixed. needs an 🆗 now

PR #460, opened 2020-03-16
UX-702  Styling adjustments for Checkbox, Radio, + Heading 🐐

mikrotron:
### Purpose 🚀
Styling adjustments for `<Checkbox>`, `<Radio>`, + `<Heading>`.

### Notes ✏️
- Inline comments to justify changes.

### Updates 📦
- [x] MINOR (backward compatible) change to `checkbox`, `radio`, `heading`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-702--styling-adjustments-checkbox-radio-heading

### References 🔗
https://aclgrc.atlassian.net/browse/UX-702

### DIFF
--- a/packages/Checkbox/src/Checkbox.styles.js
+++ b/packages/Checkbox/src/Checkbox.styles.js
@@ -79,8 +79,11 @@ const checkboxStyles = css`
   ${boxSizingStyles};
   ${({ size }) => styles[size].baseFontSize};
   line-height: ${({ hasLabel }) => (hasLabel ? lineHeightValue(-1) : "0")};
-  margin: 0 0 ${tokens.space} 0;
   position: relative;
+
+  & + [data-pka-anchor="checkbox"] {
+    margin-top: ${tokens.space};
+  }
 
   input[type="checkbox"] {
     ${visuallyHidden};
--- a/packages/Checkbox/stories/examples/CheckboxGrouping.js
+++ b/packages/Checkbox/stories/examples/CheckboxGrouping.js
@@ -1,7 +1,12 @@
 import React from "react";
+import styled from "styled-components";
 import { Rule } from "storybook/assets/styles/common.styles";
 import Heading from "@paprika/heading";
-import Checkbox from "../../src/Checkbox";
+import PaprikaCheckbox from "../../src/Checkbox";
+
+const Checkbox = styled(PaprikaCheckbox)`
+  margin-bottom: 16px;
+`;
 
 const { CHECKED, UNCHECKED } = Checkbox.states;
 
@@ -16,13 +21,7 @@ const CheckboxExample = props => {
         Checkbox Grouping
       </Heading>
       <Rule />
-      <div
-        css={`
-          [data-pka-anchor="checkbox"] {
-            margin-bottom: 12px;
-          }
-        `}
-      >
+      <div>
         <Checkbox {...props} onChange={handleChange} checkedState={checkedState}>
           Slow-carb cold-pressed hexagon forage chillwave
         </Checkbox>
--- a/packages/Checkbox/stories/examples/CheckboxGrouping.js
+++ b/packages/Checkbox/stories/examples/CheckboxGrouping.js
@@ -2,10 +2,20 @@ import React from "react";
 import styled from "styled-components";
 import { Rule } from "storybook/assets/styles/common.styles";
 import Heading from "@paprika/heading";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
 import PaprikaCheckbox from "../../src/Checkbox";
 
+// Styling override example technique #1
 const Checkbox = styled(PaprikaCheckbox)`
-  margin-bottom: 16px;
+  margin-bottom: ${stylers.spacer(2)};
+`;
+
+// Styling override example technique #2
+const CheckboxGroup = styled.div`
+  ${Checkbox} {
+    border: 1px solid ${tokens.border.color};
+  }
 `;
 
 const { CHECKED, UNCHECKED } = Checkbox.states;
@@ -21,7 +31,14 @@ const CheckboxExample = props => {
         Checkbox Grouping
       </Heading>
       <Rule />
-      <div>
+      <CheckboxGroup
+        // Styling override example technique #3
+        css={`
+          [data-pka-anchor="checkbox"] {
+            padding: ${tokens.space};
+          }
+        `}
+      >
         <Checkbox {...props} onChange={handleChange} checkedState={checkedState}>
           Slow-carb cold-pressed hexagon forage chillwave
         </Checkbox>
@@ -38,7 +55,7 @@ const CheckboxExample = props => {
         <Checkbox {...props} isDisabled onChange={handleChange} checkedState={checkedState}>
           Disabled
         </Checkbox>
-      </div>
+      </CheckboxGroup>
     </div>
   );
 };
--- a/packages/Heading/src/Heading.js
+++ b/packages/Heading/src/Heading.js
@@ -66,11 +66,10 @@ function Heading(props) {
     role: isSemantic ? null : "heading",
     safeDisplayLevel: displayLevel ? safeValue(displayLevel) : null,
     safeLevel,
-    ...moreProps,
   };
 
   return (
-    <div data-pka-anchor="heading" css={headingStyles} {...elementProps}>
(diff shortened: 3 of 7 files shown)


packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -79,9 +79,12 @@ const checkboxStyles = css`
   ${boxSizingStyles};
   ${({ size }) => styles[size].baseFontSize};
   line-height: ${({ hasLabel }) => (hasLabel ? lineHeightValue(-1) : "0")};
-  margin: 0 0 ${tokens.space} 0;
   position: relative;
 
+  & + [data-pka-anchor="checkbox"] {
+    margin-top: ${tokens.space};
+  }
```
mikrotron: This way a single `<Checkbox>` will not have any margin. But adjacent `<Checkboxes>` will space themselves apart from their preceding sibling.

This is a bit of a compromise with the principle that components should not include margin – "external" styling that interferes with its surroundings. A containing component should be responsible for that kind of layout spacing.  However, especially for the `<Radio>` it's very common to use several of these components together and having the spacing work out of the box is convenient.

packages/Checkbox/stories/examples/CheckboxGrouping.js
```diff
@@ -1,7 +1,12 @@
 import React from "react";
+import styled from "styled-components";
 import { Rule } from "storybook/assets/styles/common.styles";
 import Heading from "@paprika/heading";
-import Checkbox from "../../src/Checkbox";
+import PaprikaCheckbox from "../../src/Checkbox";
+
+const Checkbox = styled(PaprikaCheckbox)`
+  margin-bottom: 16px;
+`;
```
mikrotron: This demonstrates how margin would be controlled by a container component. I'm of half a mind to insist we rely on this technique, but I also think that makes it less convenient for consumers.

packages/Checkbox/stories/examples/CheckboxGrouping.js
```diff
@@ -16,13 +21,7 @@ const CheckboxExample = props => {
         Checkbox Grouping
       </Heading>
       <Rule />
-      <div
-        css={`
-          [data-pka-anchor="checkbox"] {
-            margin-bottom: 12px;
-          }
```
mikrotron: An alternative, sloppier way to achieve the same spacing.

packages/Heading/src/Heading.js
```diff
@@ -49,29 +49,14 @@ const defaultProps = {
 const safeValue = n => (n < 1 || Number.isNaN(n) ? 6 : Math.min(n, 6));
 
 function getElementProps(safeDisplayLevel, safeLevel, props) {
-  const {
-    a11yText,
-    children,
-    displayLevel,
-    domRef,
-    hasDivider,
```
mikrotron: Need this `hasDivider` prop on the actual component for styling purposes.  In fact, we can rely on `moreProps` to propagate most of these props to the component.

packages/Heading/src/Heading.styles.js
```diff
@@ -64,5 +63,6 @@ export const headingStyles = props => `
   ${heading[`HEADING_LEVEL_${props.safeDisplayLevel || props.safeLevel}`]}
   ${props.isHidden ? stylers.visuallyHidden : ""}
   ${props.hasUnderline ? underline(props.safeDisplayLevel || props.safeLevel) : ""}
+  ${props.hasDivider ? "display: flex;" : ""}
```
mikrotron: Most of the time it's fine if the `<Heading>` is a `flex` instead of a `block`, but it's irritating to override to to be a `block` when it does cause issues, since it's reasonable for app developers to expect an `<h>` element to be a `block`.

packages/Checkbox/stories/examples/CheckboxGrouping.js
```diff
@@ -1,7 +1,12 @@
 import React from "react";
+import styled from "styled-components";
 import { Rule } from "storybook/assets/styles/common.styles";
 import Heading from "@paprika/heading";
-import Checkbox from "../../src/Checkbox";
+import PaprikaCheckbox from "../../src/Checkbox";
+
+const Checkbox = styled(PaprikaCheckbox)`
+  margin-bottom: 16px;
```
tristanjasper: $space * 2 ?

packages/Checkbox/stories/examples/CheckboxGrouping.js
tristanjasper: Are you sure about 16px vs 12px that it was before? I think we should only rely on it if it is a direct styling applicable to the directly nested children as is the case here.... But i do think this starts to get a little complicated to follow... Perhaps another alternative could be for the `div` that you removed the `css=` from could also be a styled component?

packages/Checkbox/stories/examples/CheckboxGrouping.js
```diff
@@ -16,13 +21,7 @@ const CheckboxExample = props => {
...
       </Heading>
       <Rule />
-      <div
-        css={`
-          [data-pka-anchor="checkbox"] {
-            margin-bottom: 12px;
-          }
-        `}
-      >
+      <div>
```
tristanjasper: iei this could be a styled component?

packages/Heading/src/Heading.js
tristanjasper: I think it starts to get confusing when we peel away certain props from `props` when there are this many only to return another object.... it starts to get confusing as to what is actually being used or not, as I see `level` isn't .... so it shouldn't need to be deconstructed in this function

I would prefer all props to be deconstructed in one place than in separate functions personally. (Ie just on line 69)

packages/Heading/src/Heading.js
```diff
@@ -49,29 +49,14 @@ const defaultProps = {
...
     "aria-label": a11yText || undefined,
-    safeLevel,
-    safeDisplayLevel,
-    isHidden,
-    hasUnderline,
-    isLight,
+    "aria-level": isSemantic ? null : safeLevel,
     ref: domRef,
     role: isSemantic ? null : "heading",
+    safeDisplayLevel,
```
tristanjasper: Why isn't lines 70 & 81 evaluated in this function instead of being passed in as parameters? That would make this a bit simpler

packages/Checkbox/stories/examples/CheckboxGrouping.js
tristanjasper: Also, Just wondering about if they were to be displayed `inline` perhaps we could include a `margin-left` in that situation, although i don't think that needs to be done in this pr

mikrotron: I made it `16px` so it was more obviously changed (this is just the story, remember).  

The reason I like this approach better than the `css` approach I had before was that we don't need to rely on the `pka-anchor` attribute to select the elements, we just use `styled-components` directly.  But really it's going to be up to the consumer how they do it in the app, this is just a suggestion.

mikrotron: I think for horizontal layouts, it'd be better to put the `<Checkbox>` components in a flexbox with columns.  Then the `<Checkbox>` can remain totally agnostic of its environement.

packages/Heading/src/Heading.js
mikrotron: I think this might be clearer if the indirection of `getElementProps` is just removed and more of this happens inline.

packages/Checkbox/stories/examples/CheckboxGrouping.js
mikrotron: I'll demonstrate a few different styling techniques, for inspiration. I don't think best practices have emerged, and for Storybook stories, I don't know that we need to enforce best practices.

packages/Heading/src/Heading.js
```diff
@@ -48,46 +48,29 @@ const defaultProps = {
...
+    "aria-level": isSemantic ? null : safeLevel,
+    as: isSemantic ? `h${safeLevel}` : "div",
+    hasDivider,
+    ref: domRef,
+    role: isSemantic ? null : "heading",
+    safeDisplayLevel: displayLevel ? safeValue(displayLevel) : null,
+    safeLevel,
+    ...moreProps,
+  };
+
```
tristanjasper: 💯  this looks much better and easier to follow to me! I wonder if we should spread the `...moreProps` on line 73 though just to be clear as it gets a bit lost on that object? No worries if not though this is already an improvement.

PR #469, opened 2020-03-27
UX-602 add button group

Nandez89:
### Purpose 🚀
This PR introduces the first version of the `<ButtonGroup>` component. 

### Notes ✏️
- For this version the idea is to roll it out first as an uncontrolled component and then extend it later when needed.
- Also content truncation only happens if the component is in full width. 

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/button`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UX-602-add-ButtonGroup

### Screenshots 📸
Some examples :) 
Screen Shot 2020-04-16 at 11 29 34 AM
Screen Shot 2020-04-16 at 11 29 45 AM
Screen Shot 2020-04-16 at 11 32 01 AM

### References 🔗
https://aclgrc.atlassian.net/browse/UX-602


packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,112 @@
...
+      }
+
+      setSelectedItems(newSelectedItemIds);
+      if (onChange) {
+        onChange(newSelectedItemIds);
+      }
+    }
+  };
+
+  const bestTabIndex = isButtonGroupDisabled && tabIndex === null ? -1 : tabIndex || 0;
```
mikrotron, line 72: Since this happens already in the `<Button>` component, we can omit it here and just pass `tabIndex` through to the `<ButtonItem>`.

FYI, this code may look a bit convoluted, but the point is that if the button is disabled, you can still allow it to be in the tab order by passing `tabIndex={0}`.  While this is not the usual case, sometimes for accessibility it's helpful to let screen reader users know the disabled button is there on the page as they tab through.

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,112 @@
...
+    size,
+    isSemantic,
+    ...moreProps,
+  };
+
+  if (isSemantic) {
+    buttonProps.disabled = isButtonGroupDisabled;
+  } else {
+    buttonProps.tabIndex = bestTabIndex;
+  }
```
mikrotron, line 89: This also can just be taken care of internally in the `<Button>` as long as we pass `isSemantic` through.

packages/ButtonGroup/src/ButtonGroup.styles.js
```diff
@@ -0,0 +1,32 @@
...
+export const Button = styled(ButtonItem)``;
+
+export const ButtonGroup = styled.div`
+  display: inline-flex;
+
+  ${Button} {
+    border-radius: 0;
+
+    &:first-child {
+      border-radius: ${tokens.spaceSm} 0 0 ${tokens.spaceSm};
```
mikrotron, line 19: Even though the result is the same, it would be better to use the more semantic token `button.borderRadius` here.

packages/ButtonGroup/src/components/ButtonItem/ButtonItem.styles.js
```diff
@@ -0,0 +1,39 @@
...
+    background-color: ${tokens.color.blueLighten30};
+    border-color: ${tokens.color.blue};
+  }
+`;
+
+export const CheckIcon = styled(Check)``;
+export const TimesIcon = styled(Times)``;
+
+export const ButtonItem = styled(Button)`
+  ${TimesIcon} {
```
mikrotron, line 22: Can you combine these rules as: 
`${TimesIcon}, ${CheckIcon} { ... }`
?

packages/ButtonGroup/stories/examples/Showcase.js
```diff
@@ -0,0 +1,40 @@
...
+const buttonGroupProps = () => ({
+  onChange: clickHandler,
+  hasIcons: boolean("hasIcons", false),
+  isDisabled: boolean("isDisabled", false),
+  isFullWidth: boolean("isFullWidth", false),
+  isSemantic: boolean("isSemantic", true),
+});
+
+const logSelected = selected => {
+  console.log(selected);
```
mikrotron, line 22: Instead of logging to the console, in the Storybook it's nicer to use the `action()` addon.  You can still pass the result, `selected`, to be output.

Nandez89:
Nice! 👍 thanks for the suggestions, I will have a look and make some improvements :)

packages/ButtonGroup/src/components/ButtonItem/ButtonItem.js
```diff
@@ -0,0 +1,41 @@
...
+  children: null,
+  hasIcon: false,
+  isActive: false,
+};
+
+const ButtonItem = props => {
+  const { children, label, isActive, hasIcon } = props;
+
+  return (
+    <sc.ButtonItem key={label} data-pka-anchor="button" {...props} kind="flat">
```
mikrotron, line 30: The `label` isn't actually being used, in your example anyway. You don't really need the `key` here either, you just need it in ButtonGroup, where you do a `map()`.

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,112 @@
...
+
+const ButtonGroup = props => {
+  const { children, hasIcons, isDisabled, isFullWidth, isSemantic, onChange, tabIndex, size, ...moreProps } = props;
+
+  const [selectedItems, setSelectedItems] = React.useState([]);
+  const isButtonGroupDisabled = isDisabled;
+
+  const handleClick = event => {
+    if (!isButtonGroupDisabled) {
+      const clickedId = event.target.textContent;
```
mikrotron, line 54: This works okay in your example where the content of each button is just a string, but theoretically you can have a `node` there, which could get weird. I wonder if it would be better to have a required `value` prop for each `ButtonItem`, kind of like a checkbox or radio control.  That would need to be unique, and could be used for the key, and for the onClick API. What do you think?

mikrotron: ... it could also be different than the visible label too, and more useful for the application developer that might care about some data value more than the (localized) label on the button.

packages/ButtonGroup/src/components/ButtonItem/ButtonItem.js
```diff
@@ -0,0 +1,41 @@
...
+};
+
+const defaultProps = {
+  children: null,
+  hasIcon: false,
+  isActive: false,
+};
+
+const ButtonItem = props => {
+  const { children, label, isActive, hasIcon } = props;
```
mikrotron, line 27: Probably you should peel off `moreProps` here and propagate to the actual `<Button>` below.

packages/ButtonGroup/src/components/ButtonItem/ButtonItem.js
mikrotron: Also, you don't need the `pka-anchor` and `kind` here if they're already in ButtonGroup.  But actually, I'm not sure which place makes the _most_ sense 🤔

packages/ButtonGroup/src/ButtonGroup.js
Nandez89: Nice idk I think it makes sense, I'm just having a hard time visualizing how can node be more than just the button with the label. I guess technically we are allowing the user to pass in anything, but maybe that is the restriction we need instead?

If we go down the path you suggest would we need to define an `onClick` for each ButtonItem to bubble up the value prop? or how would that value be captured on the event?

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,106 @@
...
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+
+import * as sc from "./ButtonGroup.styles";
+import ButtonItem from "./components/ButtonItem";
+
+const propTypes = {
+  /** The content of each of the toggle buttons in the group. */
+  children: PropTypes.node,
+
+  /** T show or hide the icons for selected or not. */
```
oscarkwan, line 13: Typo here i think.

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,105 @@
...
+  children: null,
+  hasIcons: false,
+  isDisabled: false,
+  isFullWidth: false,
+  isSemantic: true,
+  onChange: () => {},
+  size: ShirtSizes.MEDIUM,
+};
+
+const ButtonGroup = props => {
```
nahumzs, line 43: nip: can we use `function ButtonGroup` when declaring const function might be hard for the reader to distinguish if it's the beginning of the component or just a declaration inside of it.

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,105 @@
...
+  const { children, hasIcons, isDisabled, isFullWidth, isSemantic, onChange, size } = props;
+
+  const validChildren = children.filter(child => child.type.displayName === ButtonItem.displayName);
```
nahumzs, line 46: I think this way of extracting children will fail in different user scenarios as:
```jsx
function renderButtonItems()
  return (<React.Fragment>
     < ButtonGroup.Item>1< ButtonGroup.Item>
     < ButtonGroup.Item>2< ButtonGroup.Item>);
</React.Fragment>)
```

You can use extractChildren function from the helpers and you should have better coverage. 

https://github.com/acl-services/paprika/blob/master/packages/helpers/src/extractChildren.js

how to use it:
https://github.com/acl-services/paprika/blob/master/packages/SidePanel/src/SidePanel.js#L107

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,105 @@
...
+      const newSelectedItems = [...selectedItems];
+
+      if (itemUsedToBeSelected) {
+        newSelectedItems.splice(itemIndex, 1);
+      } else {
+        newSelectedItems.push(clickedId);
+      }
+
+      setSelectedItems(newSelectedItems);
+      if (onChange) {
```
nahumzs, line 65: `if(onChange && typeof onChange === "function")`

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,105 @@
...
+const ButtonGroup = props => {
+  const { children, hasIcons, isDisabled, isFullWidth, isSemantic, onChange, size } = props;
+
+  const validChildren = children.filter(child => child.type.displayName === ButtonItem.displayName);
+
+  const initiallySelectedItems = validChildren.filter(item => item.props.isActive).map(item => item.props.value);
+
+  const [selectedItems, setSelectedItems] = React.useState([...initiallySelectedItems]);
+
+  const handleClick = clickedId => {
```
nahumzs, line 52: just for consistency with our other components can we do it in a curry format like:

const handleClick = (clickedId) => ()  => {
  { /* ... your code... */ } 
}

so later on line: 81 we can do:

`onClick={handleClick(item.props.value)}`

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,105 @@
...
+    <sc.ButtonGroup {...props}>
+      {validChildren.map(item => {
+        const buttonProps = {
+          isDisabled,
+          ...item.props,
+
+          hasIcon: hasIcons,
+          isFullWidth,
+          isSemantic,
+          onClick: () => handleClick(item.props.value),
```
nahumzs, line 81: if you adopted the curry function now you can just do: `onClick={handleClick(item.props.value)}`

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,105 @@
...
+          size,
+        };
+
+        return (
+          <sc.Button
+            {...buttonProps}
+            isActive={selectedItems.includes(item.props.value)}
```
nahumzs, line 88: nip: here you used includes but before you used index of, nothing big deal just curious any reason why no used the same approach in both cases?

nahumzs:
Thank you!! Daniel the component on storybook looks really polish 💥 🙏

packages/ButtonGroup/src/ButtonGroup.js
alexzherdev: I can see this has already been refactored some on the way from ACL UI so...
I still don't have 100% 🦅 👁️on this, but I'm ~90% sure that we're again iterating over children when it's not necessary. Seems like everything here can be achieved with context. @Nandez89 an example of what I mean: https://github.com/acl-services/paprika/pull/134

nahumzs: ^ yeap context can also be used here. thought its a small component with not too many variations.

oscarkwan:
Can you make it so the icons align centered with the text?

Screen Shot 2020-04-16 at 7 21 56 PM

packages/ButtonGroup/src/ButtonGroup.js
alexzherdev: Well it was you who brought up the fragment thing 😄 that won’t work with `extractChildren` but it will work with context. Just in general the more people are aware of this the better

packages/ButtonGroup/src/ButtonGroup.js
alexzherdev: 👍 but the first check is redundant

johnpeele:
Hahaha, I just realized that Oscar made the same comment I was about to make. I made a screen recording of a suggested fix before I read all the comments 😆

https://www.loom.com/share/23cb8913843148bda1b1698a880be79f

cc @oscarkwan

Nandez89:
👍 Nice, all great feedback, thanks everyone for having a look. I will coordinate with Mike depending on how fast we need these to start making the changes. definitely learning a lot from doing this collaboration.

alexzherdev:
Is it expected that the active buttons have the same kind of outline as one that indicates focus? This means when I tab onto an active button, I can't tell that it's focused. (in the screenshot, "two" is focused)
image

packages/ButtonGroup/src/ButtonGroup.js
Nandez89: Hi @nahumzs, it took a while, but I'm back at addressing some of these suggestions... I wanted to confirm what is the preferred approach here. 
Just got a bit confused since there are already a bunch of components that use `const name` as the approach.

In this line the idea is to declare the beginning of a component so it should be declared using the `function ButtonGroup(props){...`?

packages/ButtonGroup/src/ButtonGroup.js
Nandez89: mmm to be honest, at the time of doing it was probably me just following the logic on the Toggle component from previous library for the indexOf. But then for this `isActive` flag just reads cleaner using `includes`.

When trying to use `includes` on both though... I did notice we do need the `indexOf` the item not only to see if it's included but also to be able to do the splice if the item was already selected. So I'd say we could avoid looping the array twice. But happy to do it either way... I don't thing this lists will be significantly big to cause issues for that. 🤷  :)

packages/ButtonGroup/src/ButtonGroup.js
Nandez89: These are both interesting cases: 
On the extractChildren approach would be good to use and I its more explicit code for readability, and avoiding having to manually loop over the child seems like a good trade off too... 
I still don't have too much practice with the use of the Context though... so I find it a bit obscure when I'm trying to understand how the context will filter only certain allowed children to be passed in. 

I will look into it more and if I get caught in a loop (🥁) I'll hand it over to Mike to help me out

nahumzs: Hey @Nandez89 take a look at this I wrote it after this threat :) the option number two it's the one that talks about these specific cases [SubComponent Pattern](https://github.com/acl-services/paprika/wiki/React-SubComponent-Pattern#2-composing-your-parent-component-with-different-subcomponents) there its a codesandbox example at the bottom so you can see the code working. 

in short for this case is better to use Context, I take back my initial suggestion, give it a try if you find it hard, we can go with this solution and refactor it later :)

packages/ButtonGroup/src/ButtonGroup.js
nahumzs: no issue was just curiosity everything 👍

packages/ButtonGroup/src/ButtonGroup.js
nahumzs: Yes exactly, a lot of content in the community declare components with `const MyComponent = (props) => {}` and its 100% valid. 

To make a visual improvement in paprika it is better to use the `function MyComponent` declaration whenever you start a component. Its just a visual breaking point to enhance the visual experience that's all.

Nandez89: Sounds good, 💯 I will apply.

packages/ButtonGroup/src/ButtonGroup.js
alexzherdev: I'm not sure what the utility is in filtering children in this case. If you stick something other than a buttongroup item into a buttongroup then... you should know you're not gonna have something that makes sense or renders well? 🤷‍♂️

Nandez89:
> Hahaha, I just realized that Oscar made the same comment I was about to make. I made a screen recording of a suggested fix before I read all the comments 😆
> 
> https://www.loom.com/share/23cb8913843148bda1b1698a880be79f
> 
> cc @oscarkwan

Hey guys I did try the suggested solution for the alignment issue, but that ended up breaking truncation for the full width...so in the end I did some of the more minor fixes and I shared the findings with Mike and he's gonna continue looking into that and the other changes requested I couldn't get to. 

Thanks again everyone for the support and guidance and I'll follow up on future updates to continue to learn. 🙌

packages/ButtonGroup/src/components/ButtonItem/ButtonItem.js
```diff
@@ -0,0 +1,81 @@
...
+    setSelectedItems,
+  } = React.useContext(ButtonGroupContext);
+
+  React.useEffect(() => {
+    if (defaultIsActive) {
+      setSelectedItems(prevItems => {
+        return isMulti ? [...prevItems, value] : [value];
+      });
+    }
+  }, []);
```
mikrotron, line 47: An appropriate way to push the `defaultIsActive` prop up to the context for initial selected state?

alexzherdev: If going this way, better to use a layout effect to remove the possibility of the user seeing the wrong UI.
Have you considered having a `defaultValue` on the button group instead?

mikrotron: I think having `defaultIsActive` on the actual `<Item>` is a more intuitive + desirable API, which is why I wanted to go this direction.

alexzherdev: My thinking is that if it's likely that the set of values is not hardcoded but rather comes from a data source, then it's more work to distribute those values across the individual items, rather than just pass it over to the root component.

packages/ButtonGroup/stories/examples/Sandbox.js
```diff
@@ -0,0 +1,56 @@
...
+    <Tagline>Use the knobs to tinker with the props.</Tagline>
+    <Rule />
+    <p>
+      <button>Pre</button>
+    </p>
+    <ButtonGroup {...props}>
+      <ButtonGroup.Item value="one" kind="primary">
+        One
+      </ButtonGroup.Item>
+      <ButtonGroup.Item value="two" isDisabled defaultIsActive>
```
alexzherdev, line 40: a couple more realistic stories would help see the appropriateness of the API. I.e.:
- the overall set of items can be hardcoded (e.g. "weeks/months/quarters") or dynamic (fetched from the server and coming in as an array)
- the default values can be hardcoded (e.g. always want to start with "weeks" on) or dynamic (if e.g. the last value used is stored as a preference)

I'm not sure what the actual real use cases are (the confluence page just has the one) so feel free to ignore this if you know better 👍

packages/ButtonGroup/src/index.js
```diff
@@ -0,0 +1,3 @@
+import ButtonGroup from "./ButtonGroup";
+
+export { ButtonGroup as default };
```
alexzherdev, line 3:
```suggestion
export { default } from "./ButtonGroup";
```

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -0,0 +1,165 @@
...
+  function getButtonRefs() {
+    return Array.from(groupRef.current.querySelectorAll("[data-pka-anchor='button-group.button']"));
+  }
+
+  function getEnabledIndexes() {
+    const buttonRefs = getButtonRefs();
+    return buttonRefs.map((item, index) => (isItemDisabled(item) ? null : index)).filter(index => index !== null);
+  }
+
+  React.useLayoutEffect(() => {
```
oscarkwan, line 74: Was there problems using just `React.useEffect` here?

oscarkwan:
Docs Page? @mikrotron or do later?

packages/ButtonGroup/src/ButtonGroup.js
mikrotron: Just that it would render without the selection for a split second. Same reason Alex recommended `useLayoutEffect` here:
https://github.com/acl-services/paprika/pull/469#discussion_r422331013

mikrotron:
> Docs Page? @mikrotron or do later?

Later.  This PR is big enough without a bunch of stories too.

packages/ButtonGroup/stories/examples/Sandbox.js
mikrotron: As Oscar mentioned, we should have a `Docs` page with some use cases like this.  But I don't actually see any problem with the items coming from an array... that array will just need to be `map()`ed to `<ButtonGroup.Items>`.  If the selected state is included in each item, that will inform the `defaultIsActive` props.  If the selection comes separately, then something like `defaultIsActive: value === savedValue` can be added to each.

alexzherdev: Not gonna die on this hill for this particular component but still think those should be part of the original PR in the same spirit as https://github.com/acl-services/paprika/pull/401#issuecomment-607571444

---

```jsx
<ButtonGroup defaultValue={selectedItems}>
  {items.map((item) => (
    <ButtonGroup.Item>{item}</ButtonGroup.Item>
  ))}
</ButtonGroup>

<ButtonGroup>
  {items.map((item) => (
    <ButtonGroup.Item defaultIsActive={selectedItems.includes(item)}>{item}</ButtonGroup.Item>
  ))}
</ButtonGroup>
```
🤷‍♂️

mikrotron: Yeah, that looks cleaner in the first example, but where did the `selectedItems` come from?
Maybe you'd have to do something like:
```js
const selectedItems = items.filter(item -> item.isSelected)
```
Or 
```js
const selectedItems = items.map(item -> item.value)
```
So it still seems like a toss up to me.  The reason I prefer the current approach though is that fundamentally the selected state of each item is most closely associated with that item.   

Also, if you consider `radio` or `checkbox` `<inputs>`, which this component functionally resembles, this is a closer match to that API.

alexzherdev: Alright, sounds good to me 👍

packages/Button/src/Button.js
```diff
@@ -154,13 +155,17 @@ const Button = React.forwardRef((props, ref) => {
   }
 
   const buttonProps = {
+    tabIndex,
+    "data-pka-anchor": "button",
+    "aria-pressed": isActive,
+    ...moreProps,
+    "data-has-forced-focus": hasForcedFocus || null,
+    isActive,
```
jamiek-acl, line 163: Hmm why did `isActive` get pulled out of moreProps (line 103), only to be added back in here?

mikrotron: Oh, I need it for `aria-pressed` as well (above on L160).

jamiek-acl: ah, didnt see that sneaky line in there

packages/Button/src/Button.styles.js
```diff
@@ -4,9 +4,52 @@ import { css, keyframes } from "styled-components";
...
+  text-decoration: none;
+  text-shadow: none;
+
+  &:focus {
+    box-shadow: none;
+    text-shadow: none;
+    transform: none;
+  }
+
+  &,
```
jamiek-acl, line 22: The lone `&` is to make these styles also apply to the "base" component?

packages/ButtonGroup/src/ButtonGroup.styles.js
```diff
@@ -0,0 +1,36 @@
+import tokens from "@paprika/tokens";
+import styled, { css } from "styled-components";
+
+const Button = "[data-pka-anchor='button-group.button']";
```
jamiek-acl, line 4: maybe `buttonSelector` would be a bit more clear

packages/ButtonGroup/src/components/Item/Item.styles.js
```diff
@@ -0,0 +1,54 @@
...
+
+  &:focus {
+    ${stylers.z(3)}
+  }
+
+  &:hover {
+    ${stylers.z(2)}
+  }
+
+  ${({ isActive }) => (isActive ? activeStyles : "")}
```
jamiek-acl, line 44: You use this pattern a lot.  How do you feel about this alternative:

    ${({ isActive }) => (isActive && activeStyles)}

packages/Button/src/Button.styles.js
mikrotron: Correct.  Saves repeating the CSS rule.

packages/ButtonGroup/src/components/Item/Item.styles.js
mikrotron: It is shorter, so I feel that it is superior 😄

packages/Button/src/Button.styles.js
```diff
@@ -29,23 +72,25 @@ const commonStyles = `
...
         border-color: transparent;
       }
-  
+
       &[kind="flat"] {
         box-shadow: none;
       }
     }
+
+    ${({ isActive }) => (isActive ? activeStyles : "")}
```
jamiek-acl, line 93: Missed this guy

packages/Button/src/Button.styles.js
```diff
@@ -308,5 +316,5 @@ export const iconStyles = props => css`
       `
     : ""}
 
-  ${props.isDropdown ? `margin:0 0 0 ${tokens.spaceSm};` : ""}
+  ${props.isDropdown ? `margin: 0 0 0 ${tokens.spaceSm};` : ""}
```
jamiek-acl, line 319: and this one too

PR #475, opened 2020-04-08
Star 188 listbox issue 472

### LINKED ISSUE #472: ListBox - search filter not cleared after close via Trigger
# Bug Report

### Observed
When a search string is entered into the search filter field and then the ListBox is closed by clicking the Trigger element or pressing `tab` to move focus away, then when the ListBox is reopened the search filter is cleared, but the search is still applied to the ListBox options list. When the ListBox is initially closed by clicking away (blur) or pressing `esc`, the search filter will be cleared, but the search is no longer applied and the complete list of options is shown.  

###  Expected
The behaviour should be consistent when clicking the Trigger / tabbing away with clicking away / pressing `esc`. Or else the search filter field should persist until the user clears it manually.  

### ScreenCap
listbox - search bug

nahumzs:
### Purpose 🚀
This fixed #472 where the filter was not resetting properly when a onBlur were happening, also fix an issue where using a tab on the trigger was not closing the popover. 

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-188-Listbox-issue-472

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/L10n/build-translations.js
+++ b/packages/L10n/build-translations.js
@@ -13,7 +13,9 @@ const removeTranslationFiles = () => {
     }
 
     files.forEach(file => {
-      fs.unlinkSync(`${translationsPath}/${file}`);
+      if (file !== ".gitkeep") {
+        fs.unlinkSync(`${translationsPath}/${file}`);
+      }
     });
   });
 };
--- a/packages/ListBox/src/components/Filter/Filter.js
+++ b/packages/ListBox/src/components/Filter/Filter.js
@@ -40,9 +40,19 @@ const Filter = React.forwardRef((props, ref) => {
   const applyFilterType = useListBox.types.applyFilter;
   const I18n = useI18n();
 
+  const reset = React.useCallback(() => {
+    window.requestAnimationFrame(() => {
+      applyFilter(dispatch, applyFilterType)([], false);
+      setTextSearch("");
+    });
+  }, [applyFilterType, dispatch]);
+
   React.useImperativeHandle(ref, () => ({
     clear: () => {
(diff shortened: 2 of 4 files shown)


packages/L10n/build-translations.js
```diff
@@ -13,7 +13,9 @@ const removeTranslationFiles = () => {
     }
 
     files.forEach(file => {
-      fs.unlinkSync(`${translationsPath}/${file}`);
+      if (file !== ".gitkeep") {
```
nahumzs, line 16: we need to keep .gitkeep on the localization folder in order to version the folder in git and use it to process the locale generator script.

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -40,10 +40,20 @@ const Filter = React.forwardRef((props, ref) => {
...
+    });
+  }, [applyFilterType, dispatch]);
+
   React.useImperativeHandle(ref, () => ({
     clear: () => {
       setTextSearch(() => "");
     },
+    reset: () => {
+      reset();
+    },
```
allison-c, line 56: just curious can this be `reset: reset` ?

nahumzs: indeed :) can.

mikrotron: ... or just `reset,`?

alexzherdev: or just  ` `! wait, that doesn't work...

PR #479, opened 2020-04-09
feature(ListBoxBrowser) STAR-189 Adding support for hidding the root …

nahumzs:
### Purpose 🚀
Lets the listboxbrowser be use as one column instead of two.

### Updates 📦
- [x] MINOR (backward compatible) change to `@paprika/listbox-browser`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-189-one-column

### Screenshots 📸

Breadcrumb scaling
breadcrumb scaling

### References 🔗


packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -95,6 +95,10 @@ const propTypes = {
     initial value.
   */
   defaultSelectedView: PropTypes.func,
+  /** 
+    In the case you want to use the ListBoxBrowser with one column you can hide the root column
+   */
+  isVisibleRoot: PropTypes.bool,
```
mikrotron, line 101: `isRootVisible` sounds better in English, I think. But maybe `hasLeftColumn` would be more obvious?

packages/ListBoxBrowser/src/ListBoxBrowser.styles.js
```diff
@@ -1,19 +1,31 @@
...
+  ${({ isVisibleRoot }) => {
+    return isVisibleRoot
+      ? ""
+      : css`
+          display: inline;
+          [data-pka-anchor="listbox-content-inline"]:last-child {
+            border: 0;
+            border-radius: ${tokens.border.radius};
+          }
+        `;
```
mikrotron, line 22: Can we define this above as `multiColumnStyles` (or something like that) and then collapse line 13 to a single, more readable line?

packages/ListBoxBrowser/src/ListBoxBrowser.js
nahumzs: I know but didn't we agree to start always with isVisible => then what?

packages/ListBoxBrowser/src/helpers.js
```diff
@@ -16,7 +16,8 @@ function getOptions(indexes, list) {
     });
   }
 
-  return [list[indexes].value];
+  const options = [list[indexes].value].filter(chunk => chunk);
+  return options.length ? options : null;
```
mikrotron, line 20:
```suggestion
  return options || null;
```

mikrotron: Or just?
```js
return [list[indexes].value].filter(chunk => chunk) || null
```

packages/ListBoxBrowser/src/ListBoxBrowser.styles.js
nahumzs: to me is more understandable this way I can see actually what it's happening instead of having to go up on abstraction, and understand what multiColumStyles means, I think this comes to preference, I'm not a real fan of small abstraction or condense this type of logic. 

to me this way you pretty much are seeing what it's happening and what it's being modifying this style. but again its preference.

packages/ListBoxBrowser/src/ListBoxBrowser.js
nahumzs: taking hasLeftColumn then

packages/ListBoxBrowser/src/ListBoxBrowser.js
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
-import { container, flex } from "./ListBoxBrowser.styles";
+import * as sc from "./ListBoxBrowser.styles";
```
robyn3choi, line 3: why `*` instead of the actual names?

mikrotron: We ended up at this conclusion for readability + convenience. It's convenient to have one `import` that you don't have to mess with as you build out / refactor the component.  And it's more readable to identify styled components with a `sc.` prefix on them (and then you can export something without a `styled` prefix / suffix).  So in the `styles.js` you can `export Button`, and then in the JSX, you can use `<sc.Button>`.

Issue #442, opened 2020-03-03
Modals API Inconsistency - Part 2


mikrotron:
### Problem
Modal components `<Modal>` and `<Takeover>` do not have generic dialog props that the `<SidePanel>` does, namely `zIndex` and `a11yText`.  

### Solution
Add those two props to both  `<Modal>` and `<Takeover>`.

nahumzs:
does the following has been addressed?

> There is an inconsistency with the SidePanel lacking a <SidePanel.Content /> subcomponent. Takeover and Modal accept a <Modal.Content /> or <Takeover.Content /> subcomponent, the SidePanel shouldn't impose or modify the padding in their container div that is the responsibility for the subcomponent Content.

mikrotron:
> does the following has been addressed?
> 
> > There is an inconsistency with the SidePanel lacking a <SidePanel.Content /> subcomponent. Takeover and Modal accept a <Modal.Content /> or <Takeover.Content /> subcomponent, the SidePanel shouldn't impose or modify the padding in their container div that is the responsibility for the subcomponent Content.

No, but I've moved that to a new issue (https://github.com/acl-services/paprika/issues/495) and will close this one now.

PR #492, opened 2020-04-22
STAR-199  Add zIndex props to layered components 🐐

mikrotron:
### Purpose 🚀
Establish a consistent API pattern for `zIndex` props (like the `<Popover>`) for all "layered" Paprika components.

### Notes ✏️
- Propagated the `zIndex` prop for `<SidePanel>` to it's `<Overlay>` element as well.
- Added a `zIndex` prop to the `<Overlay>` component (which is shared by `<Takeover>` + `<Modal>`.
- Added `zIndex` to `<Takeover>` and `<Modal>`.
- `<DropdownMenu>`, `<Confirmation>` and `<DatePicker>` already propagate `moreProps` to their internal `<Popover>` components (including a `zIndex`).
- Added an explicit `zIndex` to `<DropdownMenu>` that will be cloned to its `<Confirmation>` for convenience.
- Added / adjusted several `data-pka-anchor` attributes.
- Also added `a11yText` props to `<SidePanel>` and `<Takeover>`.
- **BREAKING CHANGE**: eliminated exports of `<Overlay.Wrapper>` and `<Overlay.Backdrop>` (low impact) 

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/overlay`
- [x] MINOR (backward compatible) change to `@paprika/modal`, `@paprika/takeover`
- [x] PATCH (bug fix) change to `@paprika/sidepanel`, `@paprika/dropdown-menu`, `@paprika/confirmation`, `@paprika/popover`, `@paprika/date-picker`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-199--add-zindexes

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-199
https://github.com/acl-services/paprika/issues/442


packages/Confirmation/src/Confirmation.js
```diff
@@ -10,50 +10,52 @@ import TriggerButton from "./components/TriggerButton";
 import { confirmStyles, confirmBodyStyles, confirmFooterStyles } from "./Confirmation.styles";
 
 const propTypes = {
+  body: PropTypes.node,
```
mikrotron, line 13: I started making changes to this `Confirmation.js` file, then realized it was unnecessary.  Rather than revert some code tidy up, I left it in here, but none of it introduces any changes.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -11,32 +10,37 @@ import Item from "./components/Item";
 import { contentStyles } from "./DropdownMenu.styles";
 
 const propTypes = {
-  /** Alignment of the dropdown menu */
-  align: PropTypes.oneOf(AlignTypes.ALL),
+  /** Alignment of the Popover */
+  align: Popover.propTypes.align,
```
mikrotron, line 14: `eslint` warns me about this, but it seems like a good way to keep these propTypes in sync.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -129,6 +133,7 @@ function DropdownMenu(props) {
         getPositioningElement: () => document.getElementById(triggerId.current),
         offset: popoverOffset,
         onClose: handleCloseMenu,
+        zIndex,
```
mikrotron, line 136: Ensure the `Confirmation` Popover also gets put on the same `z-index` layer.

packages/DropdownMenu/src/DropdownMenu.js
```diff
@@ -159,11 +164,12 @@ function DropdownMenu(props) {
       align={align}
       offset={popoverOffset}
       isOpen={isOpen}
+      edge={edge}
+      zIndex={zIndex}
+      {...moreProps}
```
mikrotron, line 169: Moved this above `onClose` to unsure there's no way that can be overridden by the consumer.

packages/Overlay/src/Overlay.js
```diff
@@ -5,8 +5,7 @@ import FocusLock from "react-focus-lock";
 import LockBodyScroll from "@paprika/helpers/lib/components/LockBodyScroll";
 import Portal from "@paprika/helpers/lib/components/Portal";
 import tokens from "@paprika/tokens";
-import Wrapper from "./Wrapper";
-import Backdrop from "./Backdrop";
+import * as sc from "./Overlay.styles";
```
mikrotron, line 8: Aligning with our more current practice of exporting styled components from a `.styles.js` file and importing all to a `sc` object.

packages/Overlay/src/index.js
```diff
@@ -1,3 +1 @@
 export { default } from "./Overlay";
-export { default as Backdrop } from "./Backdrop";
-export { default as Wrapper } from "./Wrapper";
```
mikrotron, line 3: This is a breaking change to the API 💥 
No Paprika components were consuming these however, and it seems unlikely that our apps would be.

packages/Popover/src/Popover.js
```diff
@@ -89,7 +90,7 @@ const defaultProps = {
   getPositioningElement: null,
   getScrollContainer: null,
   shouldKeepFocus: false,
-  zIndex: 1,
+  zIndex: zValue(1),
```
mikrotron, line 93: I would like to revisit all of the `default` values for `zIndex` props and align to a seven-layer model (for example the ListBox.Popover has a default of an arbitrary `100`). 

However, some thought needs to go into that, so for now I am not introducing any changes.

packages/SidePanel/src/components/Overlay/Overlay.js
```diff
@@ -1,44 +1,44 @@
...
-  const { onClose, hasOutsideClick, ...moreProps } = props;
   const handleClick = () => {
     if (hasOutsideClick) {
       onClose();
     }
   };
 
-  const vh = visuallyHidden;
+  const overlayProps = {
+    a11yText: I18n.t("close"),
```
mikrotron, line 34: Using this (`aria-label`) instead of a `visuallyHidden` `<span>` in the `children`.

packages/Takeover/src/Takeover.js
```diff
@@ -1,35 +1,44 @@
...
+  zIndex: PropTypes.number,
 };
 
 const defaultProps = {
-  onAfterClose: () => {},
+  a11yText: null,
   onClose: () => {},
+  onAfterClose: () => {},
   onAfterOpen: () => {},
+  zIndex: zValue(5),
```
mikrotron, line 37: Why 5? 🤷‍♂️

packages/Takeover/stories/Takeover.stories.js
```diff
@@ -80,21 +82,22 @@ storiesOf("Takeover", module)
...
+
   .add("with nested Popover", () => (
     <TakeoverStory>
       <Takeover.Content>
         <p>
           This example demonstrates how the Popover handles focus when created inside of a Takeover. Click the icons
           below to see how the popovers behave.
         </p>
-        <Popover>
+        <Popover zIndex={11}>
```
mikrotron, line 93: Technically `6` would work here.  Again, I think we need a deliberate set of guidelines for what goes on what level... but because the Takeover is already at a certain level, a Popover "inside" a Takeover may need to be on a higher... uh... plane?

packages/Takeover/src/Takeover.js
mikrotron: Also, I don't consider this a breaking change, but it's possible that it _could_ affect how an app renders a view with the Takeover.

packages/DropdownMenu/src/DropdownMenu.js
alexzherdev: Yeah the warning is for when you remove proptypes from production builds. In a certain configuration of that babel plugin, this line would start throwing.
We don't do that, but it'd be interesting to see what impact on the bundle size that would have.

mikrotron: If propTypes are removed, then this would be removed too, so it wouldn't throw an error? 
If an error is thrown in a forest that doesn't exist, does anyone hear the build fail? 🤔

alexzherdev: Oh I think you're right 🤦 I'm not too familiar with situations like this.

PR #494, opened 2020-04-24
STAR-156 Extract calendar component

DanilAgafonov:
### Purpose 🚀
Work on DateRangePicker component is in progress. But we can't finish it due Calendar is a part of DatePicker component.

The ideas of this PR:
1. extract Calendar as first-class citizen
2. but current version of Calendar allow to select only one date, so we want to introduce another new component to select date range on calendar view

Names of this components is a topic to discuss, but currently we decided to use: `SingleDateCalendar` and `DateRangeCalendar`.

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/l10n` (see commits for comments)
- [x] New `@paprika/calendar` component

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-156-extract-calendar-component

### References 🔗


DanilAgafonov:
For next approach:
1. README.md
2. Find out why calendar resets showing month to current one after selecting date in not current month
3. Implement `DateRangeCalendar`

mikrotron:
In Storybook, this component looks amazing 😍! It seems to work perfectly, and I'm very happy with it.  I have only done a quick look through at the code, and what did concern me is how similar the two calendar components are – a lot of the code is duplicated.  Do you think it would be possible to compose the `<SingleDate>` version out of the `<DateRange>` version? 

Since it is unlikely that anyone is going to consume these components directly in their app, I'm not too concerned about this _right now_ and since it works, it may be reasonable to merge this now, and get the `<DateRangePicker>` completed first.  But I'm interested to know what you think we can do to make this more maintainable in the future, and maybe capture it in a GitHub issue which we can revisit later.

packages/L10n/src/locales/en.yml
```diff
@@ -33,8 +33,9 @@ en:
       confirm: Save
       decline: Decline
     datePicker:
-      calendar_header_format: MMMM YYYY
```
mikrotron, line 36: Is this the breaking change?

alexzherdev:
Didn't have a chance to give this a review, but curious why put two components in the same package? Having to use that `lib` export is not ideal.

DanilAgafonov:
@mikrotron 
> a lot of the code is duplicated. Do you think it would be possible to compose the <SingleDate> version out of the <DateRange> version?

Yes, I am 100% agree, there is a lot of duplications. But it is hard for me to imagine how one component can be composed out of another. The only possibility to DRY the code I see is extract some logic to custom hooks. And I think hooks the only option that can be used in React in such cases. (citation from react docs: "Hooks allow you to reuse stateful logic without changing your component hierarchy.")

@alexzherdev 
> why put two components in the same package?
The main reason: there is to many shared code. Especially styles. And both components uses same set of dependencies. If we split them, then it would be harder to maintain them (in case if we will need to change styles or update deps).

packages/L10n/src/locales/en.yml
DanilAgafonov: Yes.

Note: every time when I introduce breaking change, I do commit with `!` in message and write some comment why this change considered as breaking.

Example:
image

mikrotron:
> The main reason: there is too much shared code. Especially styles. And both components uses same set of dependencies. If we split them, then it would be harder to maintain them (in case if we will need to change styles or update deps).

Reasonable. But what about exporting a default, I suggest the `SingleDate` version, as `<Calendar>`, and then the `DateRange` version as `<Calendar.Range>`? Or it could be the other way around.  What do you think? We do something similar with `<Button>` where the most common use case is just `<Button>` but then we support other common variants as `<Button.Icon>` and `<Button.Close>`.

allison-c:
Code looks good to me!

>Reasonable. But what about exporting a default, I suggest the SingleDate version, as <Calendar>, and then the DateRange version as <Calendar.Range>? Or it could be the other way around. What do you think? We do something similar with <Button> where the most common use case is just <Button> but then we support other common variants as <Button.Icon> and <Button.Close>.

Personally, I prefer the current implementation, `react-dates` is also using this pattern, if the consumer really needs to reduce the bundle size, they'll need to import from `lib` . I like it because it's more clear than making `DateRangeCalendar ` as a sub-component of `SingleDateCalendar`, I think they should be on the same level

PR #496, opened 2020-04-27
STAR-219 update table header token and datagrid

oscarkwan:
### Purpose 🚀
- update token

### Updates 📦
- [x] MINOR (backward compatible) change to _DataGrid, Tokens_

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-219-update-color-table

### DIFF
--- a/packages/Tokens/src/tokens.yaml
+++ b/packages/Tokens/src/tokens.yaml
@@ -141,7 +141,7 @@ table:
   row:
     background-color: $color--white
   row-even:
-    background-color: '#f5f5f5'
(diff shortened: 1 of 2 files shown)


packages/Tokens/src/tokens.yaml
```diff
@@ -137,7 +137,7 @@ font-family:
 #
 table:
   header:
-    background-color: '#edebe8'
+    background-color: $color--black-lighten-70
   row:
     background-color: white
```
mikrotron, line 142: While you're at it, can you update this to `$color--white` too?

PR #499, opened 2020-04-28
Ux 741 textarea support uncontrolled

tristanjasper:
### Purpose 🚀
Allow the text area component to support setting of defaultValue (Required for current results questionnaires implentation)

### Notes ✏️
Ensures only a `defaultValue` or `value` can be applied to the textarea component at any one time.

Additional changes:
- Convert to functional component
- Added new uncontrolled storybook example
- Added new spec test to ensure changing defaultValue has no effect on the content after it is mounted.

### Updates 📦

- [x] MINOR (backward compatible) change to 'textarea'

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-741-textarea-support-uncontrolled

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/UX-741


packages/Textarea/src/Textarea.js
```diff
@@ -30,75 +36,83 @@ const defaultProps = {
...
+    hasError,
+    inputRef,
+    isDisabled,
+    isReadOnly,
+    maxHeight,
+    size,
+    ...moreProps
+  } = props;
+
+  if (moreProps.defaultValue) {
```
tristanjasper, line 84: Here, I am ensuring that only either a `defaultValue` or `value` can be set on the textarea component at any one time, prevents React warnings, as well as ensures `defaultValue` works correctly (pre this pr it was still applying `value=null` to the component even if `defaultValue` is present.

packages/Textarea/src/Textarea.js
```diff
@@ -8,97 +8,109 @@ const propTypes = {
...
+      textarea.style.height = 0;
+      textarea.style.height = `${textarea.scrollHeight + 2}px`;
     }
-  }
+  };
 
-  componentDidUpdate(prevProps) {
-    if (this.props.canExpand && this.props.value !== prevProps.value) {
-      this.resize();
+  React.useEffect(() => {
```
alexzherdev, line 49: Will this effect be called in the uncontrolled scenario?

packages/Textarea/src/Textarea.js
```diff
@@ -8,97 +8,109 @@ const propTypes = {
...
+  value: null,
 };
 
-class Textarea extends React.Component {
-  componentDidMount() {
-    if (this.props.canExpand) {
-      this.resize();
-      window.addEventListener("resize", this.resize);
+function Textarea(props) {
+  let textarea = null;
```
alexzherdev, line 40: Using a local variable for this feels a bit weird. Could be a ref.

packages/Textarea/src/Textarea.js
```diff
@@ -8,97 +8,109 @@ const propTypes = {
...
-    if (this.textarea && this.textarea.style) {
-      this.textarea.style.height = 0;
-      this.textarea.style.height = `${this.textarea.scrollHeight + 2}px`;
+  React.useEffect(() => {
+    if (props.canExpand) {
+      resize();
     }
+  }, [props.value]);
+
+  const setRef = node => {
```
alexzherdev, line 66: Note that porting this from how it was written in the old class component is not exactly equivalent. This will be called twice on every update, first time with `null`, then with the actual DOM element. So if any consumers passing in `inputRef` didn't expect it to be called with `null`, they might get a crash.
https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

packages/Textarea/src/Textarea.js
```diff
@@ -8,97 +8,109 @@ const propTypes = {
...
-
-  resize = () => {
-    if (this.textarea && this.textarea.style) {
-      this.textarea.style.height = 0;
-      this.textarea.style.height = `${this.textarea.scrollHeight + 2}px`;
+  React.useEffect(() => {
+    if (props.canExpand) {
+      resize();
     }
+  }, [props.value]);
```
alexzherdev, line 64: `canExpand` is also a dependency of this effect.
Which actually indicates a slight issue with this component that it stops the data flow https://overreacted.io/writing-resilient-components/#dont-stop-the-data-flow-in-rendering
If I type a bunch of text in an expandable textarea and then pass in `canExpand={false}` it will not revert to an unexpanded look.

PR #500, opened 2020-04-29
UX-742-simplify-uncontrolled-logic in input

tristanjasper:
### Purpose 🚀
Simplify the uncontrolled value state in the input component
Additionally added new spec tests

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-742-simplify-input-uncontrolled-logic

### Screenshots 📸

### References 🔗


packages/Input/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
...
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [0.2.23] - 2020-04-29
+
+### Added
+
+- Simplified uncontrolled api removing state for value [@tristanjasper](https://github.com/tristanjasper).
```
jamiek-acl, line 11: What's this CHANGELOG business, are we doing this now? Arent git commits the changelog?

tristanjasper: I believe we _are_ doing this now for changes we make, think it was discussed in a workshop a while ago..

tristanjasper:
@jamiek-acl @nahumzs Do we have any more clarity on if we think an `onChange` should still be triggered by an uncontrolled component? Jamie mentioned Nahum suggested we should be using 
```
React.useEffect(() => {
   props.onChange(valueTextInput)
}, [valueTextInput]) 
```
But I don't see how that works in this scenario, @nahumzs can you expand on that thought?

nahumzs:
@tristanjasper if we go with the React example about uncontrolled components https://reactjs.org/docs/uncontrolled-components.html we should be able to do this:

`<Input ref={refInput} defaultValue="myInitialValue" />` it's this something possible with our current component?

`refInput.current.value` should retrieve the value of the input.

packages/Input/tests/input.spec.js
```diff
@@ -1,13 +1,51 @@
...
+    const onChangeFunc = e => {
+      changeValue = e.target.value;
+    };
+    const { getByTestId, rerender } = renderComponent({ value: "init value", onChange: onChangeFunc });
+    expect(getByTestId("input").value).toBe("init value");
+    fireEvent.change(getByTestId("input"), { target: { value: "input change" } });
+    expect(changeValue).toBe("input change");
+    expect(getByTestId("input").value).toBe("init value");
+    rerender({ value: "input value controlled change" });
+    expect(getByTestId("input").value).toBe("input value controlled change");
```
jamiek-acl, line 35: I dont think lines 34 and 35 are necessary; they are the same as lines 29 and 30.

packages/Input/tests/input.spec.js
```diff
@@ -1,13 +1,51 @@
...
-    const { container } = render(<Input onChange={() => {}} placeholder="First Name" size={ShirtSizes.LARGE} />);
+  it("should be visible and render large text", () => {
+    const { container } = renderComponent({ size: ShirtSizes.LARGE });
     expect(container.firstChild.classList.contains("form-input--large")).toBe(true);
   });
-});
 
-// TODO: add more tests
+  it("should change value when value changes", async () => {
+    let changeValue = null;
```
jamiek-acl, line 25: Can you make variables for `init value` and `input change`, and use those throughout? May make it easier to follow:

    const initialValue = 'init value';
    const updatedValue = 'input change';
    ...

packages/Input/tests/input.spec.js
```diff
@@ -1,13 +1,51 @@
...
-  it("should be visible", () => {
-    const { container } = render(<Input onChange={() => {}} placeholder="First Name" size={ShirtSizes.LARGE} />);
+  it("should be visible and render large text", () => {
+    const { container } = renderComponent({ size: ShirtSizes.LARGE });
     expect(container.firstChild.classList.contains("form-input--large")).toBe(true);
   });
-});
 
-// TODO: add more tests
+  it("should change value when value changes", async () => {
```
jamiek-acl, line 24: Maybe add `controlled` into the test name

packages/Input/tests/input.spec.js
```diff
@@ -1,13 +1,51 @@
...
+    const { getByTestId, rerender } = renderComponent({ value: "init value", onChange: onChangeFunc });
+    expect(getByTestId("input").value).toBe("init value");
+    fireEvent.change(getByTestId("input"), { target: { value: "input change" } });
+    expect(changeValue).toBe("input change");
+    expect(getByTestId("input").value).toBe("init value");
+    rerender({ value: "input value controlled change" });
+    expect(getByTestId("input").value).toBe("input value controlled change");
+  });
+
+  it("should set input value with defaultValue and call the inputRef function passing node when input changes", async () => {
```
jamiek-acl, line 38: maybe add `uncontrolled` into the test name

packages/Input/tests/input.spec.js
```diff
@@ -1,13 +1,51 @@
...
+    let node = null;
```
jamiek-acl, line 39: Please use variables here, too, instead of string "init default value"... makes it easier to read

PR #501, opened 2020-04-29
Ux 741 textarea uncontrolled 2

tristanjasper:
### Purpose 🚀

- Fixes text area resize when uncontrolled
- Other Minor fixes for Textarea component based on comments in https://github.com/acl-services/paprika/pull/499/files

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to `Textarea`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-741-textarea-uncontrolled-2

### Screenshots 📸

### References 🔗


packages/Textarea/src/Textarea.js
```diff
@@ -55,18 +55,13 @@ function Textarea(props) {
     return function cleanup() {
       window.removeEventListener("resize", resize);
     };
-  });
+  }, []);
```
tristanjasper, line 58: As discussed in https://github.com/acl-services/paprika/pull/499/files#r417552311

packages/Textarea/src/Textarea.js
```diff
@@ -55,18 +55,13 @@ function Textarea(props) {
...
     if (props.canExpand) {
       resize();
     }
-  }, [props.value]);
-
-  const setRef = node => {
-    textarea = node;
-    props.inputRef(node);
-  };
+  }, [props.value, props.canExpand]);
```
alexzherdev, line 64: Oh actually I meant this effect - if the component is uncontrolled, and typing doesn't result in a re-render this will not run. And the textarea will not expand.

packages/Textarea/src/Textarea.js
```diff
@@ -105,7 +108,11 @@ function Textarea(props) {
         data-pka-anchor="textarea"
         disabled={isDisabled}
         readOnly={isReadOnly}
-        ref={setRef}
+        onChange={handleChange}
+        ref={node => {
```
alexzherdev, line 112: Is this to address the comment on the previous PR?
Granted I'm not sure if anyone uses `inputRef` so may not be worth discussing. But note that this change actually does nothing. You used to have this function in a variable, now you have it inline. But point is, it's a new function each render anyway (which is the reason React will call it with `null`).
I think this can be partially made better by memoizing
```jsx
const setRef = React.useCallback(() => {
  ...
}, [props.inputRef]);
```
This will preserve referential integrity (unless `inputRef` changes) but honestly not sure it's worth it. I just did a search in the entire github org and no one seems to be using it.

packages/Textarea/src/Textarea.js
```diff
@@ -37,7 +37,7 @@ const defaultProps = {
 };
 
 function Textarea(props) {
-  let textarea = null;
+  let textarea = React.useRef(null);
 
   const resize = () => {
     if (textarea && textarea.style) {
```
alexzherdev, line 43: This now has to use `textarea.current` (also there are assignments to `textarea` elsewhere that need to be changed)

packages/Textarea/src/Textarea.js
tristanjasper: 👌 Perhaps I'll leave it like it is for now. If we see problems we can revisit? Or you still think worth making that change?

alexzherdev: maybe revert to what it was before (with the `const`) so that this PR doesn't have the unnecessary diff?

packages/Textarea/src/Textarea.js
tristanjasper: Using `textarea.current` on any of these values breaks the resize functionality for me, I'me not even sure that `textarea` needs to be a ref?. Can you let me know whats wrong with it just being a variable of `null`?

packages/Textarea/src/Textarea.js
```diff
@@ -37,56 +37,64 @@ const defaultProps = {
 };
 
 function Textarea(props) {
-  let textarea = null;
+  const textarea = React.useRef(null);
```
alexzherdev, line 40: I feel like I've seen people suffixing ref names with `Ref` elsewhere (so `textareaRef`) although could be wrong

tristanjasper: you could be right! :) changed

PR #505, opened 2020-04-29
STAR-223 Eliminate Takeover.Overlay proptype warnings 🐐

mikrotron:
### Purpose 🚀
Eliminate `<Takeover.Overlay>` proptype warnings and alleviate the need to include a redundant `isOpen` prop.

### Notes ✏️
- Remove propTypes from `<Takeover.Overlay>` and rename file to `OverlayPropsCollector.js`.
- Since props for the `<Overlay>` are passed from the `<Takeover>` API, the `...moreProps` was moved so it cannot override them.  `moreProps` will only be needed for propagating `classNames`, `qa-id` attributes, or additional HTML attributes.

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/takeover`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-223--eliminate-takeover-proptype-warnings

### Screenshots 📸
Screen Shot 2020-04-29 at 4 11 46 PM

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-223


packages/Takeover/src/components/OverlayPropsCollector.js
```diff
@@ -0,0 +1,5 @@
+export default function OverlayPropsCollector() {
```
tristanjasper, line 1: interesting component name. makes sense i guess, i'd imagine we'll see more components that serve this purpose so we'd want to use the same naming convention for the others in the future.

mikrotron: Yeah, we had a naming discussion in a PR for the `<DatePicker>` a while back. I think it was @DanilAgafonov's PR...

alexzherdev: Can y'all remind me why we need this at all? So far only seeing the displayName rehashing.

mikrotron: It's all for the API.  We do this:

https://github.com/acl-services/paprika/blob/126258c17692d4be62acd346d28979ac4ffc3b93/packages/Takeover/src/Takeover.js#L57

And this:

https://github.com/acl-services/paprika/blob/126258c17692d4be62acd346d28979ac4ffc3b93/packages/Takeover/src/Takeover.js#L68

Then we can pass props to the underlying `<Overlay>` by doing `<Takeover.Overlay someProp={someValue} />` instead of the undesirable `<Takeover someOverlayProp={someValue} />` which can get pretty ugly as it scales.  

Historically I think maybe the `<Listbox>` was the first to adopt the pattern – we wanted to pass a number of props to the underlying `<Popover>` and didn't want a `popoverProps` prop.

alexzherdev: Thanks. 👍 
FWIW I think this is unnecessary. The same should be possible by just having `Takeover.Overlay = Overlay`, and then this
https://github.com/acl-services/paprika/blob/126258c17692d4be62acd346d28979ac4ffc3b93/packages/Takeover/src/Takeover.js#L67-L89
would instead be
```jsx
React.cloneElement(overlayExtracted, { "data-pka-anchor": "takeover.overlay", ... }, state => (
  ...
))
```
Seems like a less roundabout way to achieve the same.

mikrotron: I like "less roundabout ways", thanks for the suggestion. I will give it a try.

alexzherdev: Surfacing Overlay this way has some interesting semver implications though. I.e., now if Takeover bumps their Overlay version across a major, it becomes a major bump for Takeover itself, because the API of Overlay is a subset of Takeover's API.

.yarnrc
```diff
@@ -1,2 +1,4 @@
---install.audit true
---add.audit true
+# Disabling temporarily
+# https://github.com/yarnpkg/yarn/issues/7025
+# --install.audit true
+# --add.audit true
```
mikrotron, line 4: The 503 error is getting more persistent and pretty annoying. Sounds like the payload size of our growing repo could be the reason why.

tristanjasper:
LGTM

packages/Takeover/src/Takeover.js
```diff
@@ -37,7 +39,10 @@ const defaultProps = {
   zIndex: zValue(5),
 };
 
-const Takeover = props => {
+const TakeoverOverlay = Overlay;
+delete TakeoverOverlay.propTypes.isOpen;
```
allison-c, line 43: LGTM, just curious is it good to do something like this? isn't more clear to define all propTypes again in this case for consumers? or define a common prop type and reuse? for example:
```
const BasicOverlayPropTypes = {
  isOpen:PropTypes.bool,
  children: PropTypes.func,
  ...
}
```
and use like:

```
Overlay.propTypes.= {
  ...BasicOverlayPropTypes,
 isOpen: PropTypes.bool.isRequired
}

TakeoverOverlay.propTypes.= BasicOverlayPropTypes
```

mikrotron: Sorry, I didn't notice this comment earlier! I think this is `basicPropTypes` composition approach makes sense in some cases, but here I'm just using the `propTypes` for a prop collector, it's not even a real, functional component, so the extra complexity added to `<Overlay>` (which shouldn't really care about how `<Takeover>` is using it) feels unnecessary.

PR #511, opened 2020-05-06
UX-743-add aria-invalid on textarea

tristanjasper:
### Purpose 🚀
Added `aria-invalid` prop to text area for accessibility

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to textarea

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-743-textarea-aria-invalid

### Screenshots 📸

### References 🔗


mikrotron:
Cool. Can we do the same thing to the `<Select>` while we're at it?

tristanjasper:
> Cool. Can we do the same thing to the `<Select>` while we're at it?

Will do in a separate PR

PR #513, opened 2020-05-08
L10N-2020-05-07-1632 - translations batch

alexandermccate:


packages/L10n/src/locales/zh.yml
```diff
@@ -92,3 +92,59 @@ zh:
       explore: 探索更多选项
       selected: 选定选项
       selectOne: 选择一项或多项...
```
cristinajimenez, line 94: Could maybe change "..." for "…" (could be done next batch)

PR #514, opened 2020-05-08
ISSUE-172 Make buttons have focus ring when they are focused programa…

jamiek-acl:
…tically

### Purpose 🚀
Simplify the logic that decides when the buttons should have a focus ring.

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/issue-172-no-focus-styling-keyboard-simplified

### DIFF
--- a/packages/Button/src/Button.js
+++ b/packages/Button/src/Button.js
@@ -177,15 +177,13 @@ const Button = React.forwardRef((props, ref) => {
     kind,
   };
 
-  const hasForcedFocusAttribute = { "data-has-forced-focus": hasForcedFocus || null };
-
   return (
     <span
       css={buttonStyles}
       data-pka-anchor="button"
       {...buttonProps}
       as={isSemantic ? "button" : RawButton}
-      {...hasForcedFocusAttribute}
+      data-has-forced-focus={hasForcedFocus || null}
     >
       <ButtonIcon {...iconProps} isPending={isPending} data-pka-anchor="button.icon">
         {isPending ? <RefreshIcon /> : icon}
--- a/packages/Button/src/Button.js
+++ b/packages/Button/src/Button.js
@@ -152,7 +152,7 @@ const Button = React.forwardRef((props, ref) => {
     return cn;
   }
 
-  function getOnBlur() {
+  function handleBlur() {
     if ("onBlur" in moreProps) {
       moreProps.onBlur();
     }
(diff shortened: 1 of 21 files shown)


packages/Button/src/Button.js
```diff
@@ -146,9 +146,24 @@ const Button = React.forwardRef((props, ref) => {
 
   const bestTabIndex = isButtonDisabled && tabIndex === null ? -1 : tabIndex || 0;
 
+  function getClassName() {
```
mikrotron, line 149: Instead of tacking on a custom className, can we keep this component agnostic of classes and use a custom data attribute instead? `data-show-focus`?

packages/Button/src/Button.js
```diff
@@ -146,9 +146,24 @@ const Button = React.forwardRef((props, ref) => {
...
+    let cn = "className" in moreProps ? moreProps.className : "";
+    cn += hasForcedFocus ? " has-forced-focus" : "";
+    return cn;
+  }
+
+  function getOnBlur() {
```
mikrotron, line 155: I think `handleBlur` would more closely follow our naming conventions.

packages/Button/src/Button.js
jamiek-acl: k but i'd like to keep the words `has` and `forced` in there: `data-has-forced-focus`

packages/Button/src/Button.js
```diff
@@ -169,8 +177,16 @@ const Button = React.forwardRef((props, ref) => {
     kind,
   };
 
+  const hasForcedFocusAttribute = { "data-has-forced-focus": hasForcedFocus || null };
```
mikrotron, line 180: Why the indirection? Can't this go right into line 188?

packages/Button/src/Button.styles.js
```diff
@@ -27,7 +27,7 @@ const commonStyles = `
     outline: none;
   }
 
-  [data-whatinput="mouse"] &:focus {
+  [data-whatinput="mouse"] &:focus&:not([data-has-forced-focus="true"]) {
```
mikrotron, line 30:
```suggestion
  [data-whatinput="mouse"] &:focus:not([data-has-forced-focus="true"]) {
```

packages/Button/src/Button.js
jamiek-acl: because i dont want to pollute the dom with `data-has-forced-focus="false"`, which will be 99.9% of cases

jamiek-acl: oh i understood that wrong... if it line breaks nicely, sure i can do that

PR #523, opened 2020-05-13
Star 182 uploader component take3 - Breaking change

jamiek-acl:
### Purpose 🚀
I added a UI and changed the API. This is basically a new component.

Github issues were created to address a few missing pieces (a11y, tests, etc)

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/star-182-uploader-component-take3

### Screenshots 📸
Screen Shot 2020-05-19 at 11 13 29 AM

dnd

### References 🔗


packages/Uploader/src/Uploader.js
```diff
@@ -3,11 +3,17 @@ import PropTypes from "prop-types";
...
-import ProgressBar from "./components/ProgressBar";
-import types from "./types";
+import FileList from "./components/FileList";
+import DropZone from "./components/DropZone";
+import types from "./statuses";
 import useDragAndDropEvents from "./useDragAndDropEvents";
 import useProcessFiles from "./useProcessFiles";
 
+// TODO: there is a flicker when go over dropzone
+// TODO: why is restarting the 2nd file waiting until the 1st file is complete?
```
nahumzs, line 13: are these true still or should we remove it?

packages/Uploader/src/components/DropZone/DropZone.js
```diff
@@ -0,0 +1,41 @@
...
+  const isDraggingOver = uc.isDraggingOver;
+  const uploadIconColor = isDraggingOver ? tokens.color.purpleDarken10 : tokens.color.blackLighten50;
+
+  /* eslint-disable jsx-a11y/anchor-is-valid */
+  const body = isDraggingOver ? (
+    I18n.t("uploader.drop_files")
+  ) : (
+    <React.Fragment>
+      {I18n.t("uploader.drop_files_here_or")}&nbsp;
+      <a
```
mikrotron, line 20: This should be a `<Button kind="link">`. I know you've flagged `a11y` as a `TODO`, but this is a quick refactor, and hard for me to let a practice like this slip through.

packages/Uploader/src/components/DropZone/DropZone.js
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import tokens from "@paprika/tokens";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import UploadIcon from "@paprika/icon/lib/Upload";
+import { UploaderContext } from "../../Uploader";
+import * as div from "./DropZone.styles";
```
mikrotron, line 6: The convention we seem to have settled on was to import styled components like this as `sc` (vs `div`).

packages/Uploader/src/components/DropZone/DropZone.js
```diff
@@ -0,0 +1,41 @@
...
+        {I18n.t("uploader.choose_from_computer")}
+      </a>
+    </React.Fragment>
+  );
+  /* eslint-enable jsx-a11y/anchor-is-valid */
+
+  return (
+    <uc.FileInput>
+      <div.DropZone isDraggingOver={isDraggingOver}>
+        <UploadIcon size="32px" color={uploadIconColor} />
```
mikrotron, line 36: Since you're using `tokens` already for the `color`, you could use one here for the size too... or else omit it and set the `font-size` for `[data-pka-ancor='icon']` to a token in the `DropZone` styles. Could also set the colour there, dependent on `isDraggingOver` if you want to really consolidate styling in one place.

packages/Uploader/src/components/DropZone/DropZone.js
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import tokens from "@paprika/tokens";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import UploadIcon from "@paprika/icon/lib/Upload";
+import { UploaderContext } from "../../Uploader";
+import * as div from "./DropZone.styles";
+
+export default function DropZone() {
+  const I18n = useI18n();
+  const uc = React.useContext(UploaderContext);
```
mikrotron, line 10: You could destructure here instead of using a somewhat ambiguous `uc` name.

packages/Uploader/src/components/File/File.js
```diff
@@ -0,0 +1,121 @@
...
+};
+
+const defaultProps = {
+  error: "",
+  progress: 0,
+};
+
+function File({ error, fileKey, name, progress, size, status }) {
+  const uc = React.useContext(UploaderContext);
+  const { cancelFile, restartFileUpload } = uc;
```
mikrotron, line 31:
```suggestion
  const { cancelFile, restartFileUpload } = React.useContext(UploaderContext);
```

packages/Uploader/src/components/File/File.js
```diff
@@ -0,0 +1,121 @@
...
+                }}
+                size="small"
+              >
+                <RetryIcon />
+              </Button.Icon>
+            </Popover.Trigger>
+            <Popover.Content>
+              <Popover.Card>{I18n.t("uploader.restart_upload")}</Popover.Card>
+            </Popover.Content>
+          </Popover>
```
mikrotron, line 61: Should these have `<Popover.Tips>`?

packages/Uploader/src/Uploader.js
jamiek-acl: @nahumzs they are true; just wanted to get this PR out there and take a break from this component

packages/Uploader/src/components/DropZone/DropZone.js
jamiek-acl: I originally had it as a button, but it gets ugly on narrow window size - it won't split the text mid-button.

packages/Uploader/src/components/File/File.js
```diff
@@ -0,0 +1,121 @@
...
+  error: "",
+  progress: 0,
+};
+
+function File({ error, fileKey, name, progress, size, status }) {
+  const uc = React.useContext(UploaderContext);
+  const { cancelFile, restartFileUpload } = uc;
+  const I18n = useI18n();
+
+  function getUploadedAmount() {
```
tristanjasper, line 34: return on one line with arrow function?

packages/Uploader/src/components/File/File.js
```diff
@@ -0,0 +1,121 @@
...
+
+  function renderIcon() {
+    switch (status) {
+      case statuses.ERROR:
+      case statuses.CANCEL:
+        return (
+          <Popover isDark isEager>
+            <Popover.Trigger>
+              <Button.Icon
+                kind="minor"
```
tristanjasper, line 49: could use variable from `Button.Kinds` over a string

packages/Uploader/src/components/File/File.js
```diff
@@ -0,0 +1,121 @@
...
+                onClick={() => {
+                  restartFileUpload(fileKey);
+                }}
+                size="small"
```
tristanjasper, line 53: could use variable from ShirtSize variable  from customPropTypes over a primitive string

packages/Uploader/src/components/File/File.styles.js
```diff
@@ -0,0 +1,92 @@
...
+
+export const Info = styled.div`
+  align-items: flex-end;
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: ${tokens.spaceSm};
+`;
+
+export const Name = styled.div`
+  flex: 1 1 auto;
```
tristanjasper, line 41: thought we weren't writing shorthand approach anymore?

packages/Uploader/src/useProcessFiles.js
```diff
@@ -91,90 +92,108 @@ export default function useProcessFiles({
...
 
-      if (files.length && !isSameList(files, uploadingFileList) && !isDisabled) {
-        setUploadingFileList(() => files);
-        setIsDisabled(() => true);
-        setisCompleted(() => null);
-        onChange(files);
-        files.forEach(file => {
-          if (file.isValid && file.status !== types.SUCCESS) {
-            uploadToServer({ file, endpoint, onProgress, onSuccess, onError, headers });
+    function onProgress({ file, percent }) {
```
tristanjasper, line 164: shouldn't it be `onProgressChange`?

packages/Uploader/src/components/DropZone/DropZone.js
mikrotron: > it won't split the text mid-button

Good...?

Seems to me like breaking "Choose files..." into two lines would look worse than breaking the line before it anyway?

packages/Uploader/src/components/File/File.js
mikrotron: ...or just declare as a `const`?

packages/Uploader/src/components/File/File.styles.js
```diff
@@ -0,0 +1,92 @@
...
+  padding: ${stylers.spacer(2)};
+`;
+
+export const Left = styled.div`
+  flex: 1;
+`;
+
+export const Right = styled.div`
+  display: flex;
+  flex: 0 0 ${stylers.spacer(4)};
```
mikrotron, line 19:
```suggestion
  flex-shrink: 0;
  width: ${stylers.spacer(4)};
```

mikrotron: ☝️ I almost never use `flex-basis`, and am not even sure what the difference would be in this case vs `width`.  Do you?

packages/Uploader/src/components/File/File.styles.js
```diff
@@ -0,0 +1,92 @@
...
+        color: ${tokens.color.orange};
+      `;
+    }
+  }}
+`;
+
+export const ProgressBarWrapper = styled.div`
+  background: ${tokens.color.blackLighten40};
+  border-radius: ${tokens.border.radius};
+  height: 4px;
```
mikrotron, line 70:
```suggestion
  height: ${tokens.spaceSm};
```

packages/Uploader/src/components/FileList/FileList.js
```diff
@@ -0,0 +1,52 @@
...
+      return file.error.message;
+    }
+
+    return null;
+  }
+
+  return (
+    files &&
+    files.map(file => (
+      <File
```
mikrotron, line 38: For your follow-up `a11y` PR, it would be nice to use `<ul>` + `<li>` to make this a semantic list.  Or, alternatively, `role=list` + `role=listitem`.

packages/Uploader/src/helpers.js
```diff
@@ -115,8 +115,25 @@ export function getFiles({ event, maxFileSize, okFileTypes, endpoint }) {
...
 
   return [];
 }
+
+export function getNumberWithUnits(number) {
+  if (number > 1024) {
+    if (number > 1024 * 1024) {
+      if (number > 1024 * 1024 * 1024) {
+        if (number > 1024 * 1024 * 1024 * 1024) {
+          return `${(number / (1024 * 1024 * 1024 * 1024)).toFixed(3)}TiB`;
```
mikrotron, line 129: `TiB` should be localized, no?

packages/Uploader/src/helpers.js
```diff
@@ -1,6 +1,6 @@
 import uuidv4 from "uuid/v4";
 import superagent from "superagent";
-import types from "./types";
+import statuses from "./statuses";
```
mikrotron, line 3: 👍

packages/Uploader/src/useProcessFiles.js
mikrotron: If you wanna update function names, maybe `upload` → `onUpload` and `restartFileUpload` → `onRestartUpload`?

packages/Uploader/src/components/DropZone/DropZone.js
```diff
@@ -0,0 +1,41 @@
...
+  /* eslint-disable jsx-a11y/anchor-is-valid */
+  const body = isDraggingOver ? (
+    I18n.t("uploader.drop_files")
+  ) : (
+    <React.Fragment>
+      {I18n.t("uploader.drop_files_here_or")}&nbsp;
+      <a
+        href="#"
+        onClick={() => {
+          uc.refInput.current.click();
```
mikrotron, line 23: The `href` is causing navigation without an `e.preventDefault()` in the `onClick` handler.  Also prevented by using a `<Button>`.

packages/Uploader/src/components/DropZone/DropZone.js
mikrotron: Oh, I see that it's longer, "Choose from your computer... " 🤔

packages/L10n/src/locales/en.yml
```diff
@@ -7,25 +7,26 @@ en:
...
+      drop_files: Drop files to upload
+      drop_files_here_or: Drop files to upload here or
+      choose_from_computer: choose from your computer
+      cancel_upload: Cancel upload
+      restart_upload: Restart upload
       errors:
-        file_size: exceeds the maximum {{size}} file size.
-        file_type: "{{type}} is incorrect file type."
+        size: File must be smaller than {{maxFileSize}}
+        type: "File must be one of the following types: {{supportedTypes}}"
```
mikrotron, line 17: I see `&#x2F;` in the UI...

packages/Uploader/stories/Uploader.stories.js
```diff
@@ -2,117 +2,103 @@ import React from "react";
...
-      <p>Will accept dropped files only if they are drop at the FileInput area.</p>
-      <Uploader isBodyDroppable={false} endpoint="http://localhost:9000/upload.php">
-        <Testing />
+      <Uploader {...props} isBodyDroppable={false}>
+        <Uploader.DropZone />
+        <Uploader.FileList />
       </Uploader>
     </Story>
   ))
   .add("Allow only one file per upload", () => (
```
mikrotron, line 83: The "Choose..." button doesn't seem to work for this story.

packages/Uploader/src/helpers.js
nahumzs: I don't think you can translate `Mebibyte` can you?  Can you translate that acronym? mebibyte would be translate to `mega binary byte` so would make sense to translate TiB to `tera binary byte`? which would make really stranger to read. But to be honest I haven't seen this word translated it.

nahumzs: are you kidding me french?  `Megabyte` => `mégaoctet`

packages/Uploader/src/components/DropZone/DropZone.js
jamiek-acl: I've never liked "size" and "colour" props since they can just be targeted with css. However, since they are there, I'm gonna use 'em!

I can't use the `space` token for this, but can use `stylers/spacer`.

packages/Uploader/src/components/DropZone/DropZone.js
jamiek-acl: `uc` stands for `UploaderContext`! I like being deeper in the code and coming across `uc.whatever` and then I know where it gets assing, so I don't have to look up in the rest of the code to see where it is getting assigned a value.  But i'm not married to it, so i'll destructure that and another thing I see.

Maybe we should be destructuring I18n's, too.
`const { t } = useI18n();`

packages/Uploader/src/components/File/File.js
jamiek-acl: Yes, good call

packages/Uploader/src/components/File/File.js
jamiek-acl: Yeah I dont know why I did that, lol. Fixing.

packages/Uploader/src/useProcessFiles.js
jamiek-acl: Hmm not sure how i feel about these renaming suggestions.
`onProgress` goes with `onSuccess` and `onError`.

And when I want the upload to start, it doesn't feel right to fire `onUpload();`. Same for restarting.

Thoughts?

packages/Uploader/src/components/DropZone/DropZone.js
jamiek-acl: Resize using anchor (pretty):
nice-resize-anchor

Resize using `<Button type="link">` (ugly):
bad-resize-button

packages/Uploader/src/components/File/File.styles.js
jamiek-acl: @mikrotron Nah I think width and flex-basis do the same thing when flex-direction is `row`.
But I think it makes sense to use `flex-basis` instead of `width` when using flexbox.

packages/Uploader/src/components/DropZone/DropZone.js
jamiek-acl: Haha, thanks. I was stuck on this for too long and preventDefault escaped my mind.

packages/L10n/src/locales/en.yml
jamiek-acl: Screen Shot 2020-05-28 at 3 06 50 PM


@mikrotron Fixed

packages/Uploader/src/components/File/File.styles.js
nahumzs: > The first thing that comes to mind when reading your question is that flex-basis doesn't > always apply to width.
>
> When flex-direction is row, flex-basis controls width.
> 
> But when flex-direction is column, flex-basis controls height.

nahumzs: sound a complex question to answer https://stackoverflow.com/a/34355447/196038 :P

tristanjasper:
Wonder if we can prevent the popover close on click for the restart action, kind of annoying that it closes and reopens instantly because you are hovering over it... 

popover close on click

jamiek-acl:
> Wonder if we can prevent the popover close on click for the restart action, kind of annoying that it closes and reopens instantly because you are hovering over it...

I made an issue to look at this: https://github.com/acl-services/paprika/issues/543

PR #524, opened 2020-05-13
Ux 614 fix input evaluating value false

tristanjasper:
### Notes ✏️
Fix bug where empty string as value is causing a warning, set onChange callback to not required

### Updates 📦
- [x] PATCH (bug fix) change to input

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-614-fix-input-evaluating-value-false

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/UX-746

### DIFF
--- a/packages/Input/src/Input.js
+++ b/packages/Input/src/Input.js
@@ -33,7 +33,7 @@ const propTypes = {
   /** If true it makes the input read only. */
   isReadOnly: PropTypes.bool,
 
-  /** Callback to be executed when the input value is changed */
+  /** Callback to be executed when the input value is changed. Should not be used with defaultValue prop */
   onChange: PropTypes.func,
 
   /** Callback to be executed when the input value is cleared */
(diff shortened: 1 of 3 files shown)


packages/Input/src/Input.js
```diff
@@ -34,7 +34,7 @@ const propTypes = {
   isReadOnly: PropTypes.bool,
 
   /** Callback to be executed when the input value is changed */
-  onChange: PropTypes.func.isRequired,
+  onChange: PropTypes.func,
```
tristanjasper, line 37: this isn't required if we are using the input in an uncontrolled manner

packages/Input/src/Input.js
```diff
@@ -108,7 +109,7 @@ const Input = props => {
     ...moreProps
   } = props;
 
-  if (moreProps.value) {
+  if (moreProps.value || moreProps.value === "") {
```
tristanjasper, line 112: `""` as `value` was evaluated as false, and so when a different `value` is provided there would be warning that the input component is being changed from controlled to uncontrolled. 

image

PR #529, opened 2020-05-20
SNP-14 Fix confirmation component accessibility

allison-c:
### Purpose 🚀

Fixed: the confirmation component doesn't announce the popover content before.

### Notes ✏️

I noticed that it's announcing the popover content when focusing on the popover trigger. That's not correct. It should announce the popover content when the content popup
test it here: http://storybooks.highbond-s3.com/paprika/SNP-14-Fix-confirmation-a11y/?path=/story/confirmation--confirmation

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/confirmation` and `@paprika/button`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/SNP-14-Fix-confirmation-a11y

### Screenshots 📸

### References 🔗


packages/Confirmation/src/Confirmation.js
```diff
@@ -82,7 +83,7 @@ const Confirmation = props => {
   }, [confirmButtonRef, defaultIsOpen]);
 
   React.useEffect(() => {
-    popoverKey = uuid();
+    popoverKey = nanoid();
```
oscarkwan, line 86: Just wondering, whats the difference?

allison-c: it's very small and the id is shorter, I noticed we start using it in new components

packages/Button/src/Button.styles.js
```diff
@@ -42,6 +42,10 @@ const commonStyles = `
     }
   }
 
+  [data-whatinput="mouse"] &[data-has-forced-focus="true"]:focus {
+    box-shadow: ${tokens.highlight.active.noBorder.boxShadow}; 
+  }
+
```
jamiek-acl, line 48: i think this could be nested under line 30, and cleaned up a bit (the `:focus` is re-used`):

    [data-whatinput="mouse"] &:focus {
      box-shadow: ${tokens.highlight.active.noBorder.boxShadow}; 

      &:not([data-has-forced-focus="true"]) {
        box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.1);
        border-color: ${tokens.border.color};

        &[kind="minor"],
        &[kind="link"] {
          box-shadow: none;
          border-color: transparent;
        }

        &[kind="flat"] {
          box-shadow: none;
        }
      }
    }

allison-c: Sure! But I guess I'll still need `[data-has-forced-focus="true"]`, otherwise it might be overrided by rawbutton component, it has `[data-whatinput="mouse"] &:focus {  box-shadow: none; }`

PR #531, opened 2020-05-22
STAR-232 Card Component

karencumlat:
### Purpose 🚀
Created the ```<Card>``` component. The Card component is a customizable content container. It comes with subcomponents that allow for quick and easy creation of various card types for different use cases. Additionally any other child elements can be passed into the Card.

**Sub-Components**
The subcomponents include:

- ```<Card.Header></Card.Header>```
The header sub-component allows you to add a header section to the card.

- ```<Card.Content></Card.Content>```
The content sub-component allows you to add a container that can section of a part of the card.

- ```<Card.Title></Card.Title>```
The title sub-component allows you to add a title to the card.

- ```<Card.Seperator></Card.Seperator>```
The seperator sub-component allows you to add space to the card, you can adjust the size of the seperator with its size property (small, medium, and large).

- ```<Card.Text></Card.Text>```
The text sub-component allows you to add a text body to the card.

- ```<Card.Metadata></Card.Metadata>```
The metadata sub-component allows you to add metadata to the card.

- ```<Card.Footer></Card.Footer>```
The footer sub-component allows you to add a footer section to the card.

**Properties**
The properties include:

- ```size```
The size property allows the user to set a defined width and height for the card to create a set of uniform cards (small, medium, and large).

- ```isFullWidth```
The isFullWidth property allows the user to set the width of the card to the width of its parent container (100%).

- ```isAutoHeight```
The isAutoHeight property allows the users to set the height of the card component to be based on the amount of children elements within it.

- ```isDisabled```
The isDisabled property allows the user to set the entire Card to be disabled for the user.

- ```isActive```
The isActive property allows the user to set the Card to be Active which highlights/focuses it for the user.

- ```onClick```
The Callback property can take a function to be executed when the card is clicked or activated.

### Notes ✏️
`<Avatar />` component was also created. For now, it is included on components folder of `<Card />`.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-232-Card

### Screenshots 📸

Deluxe Card

Basic Card

### References 🔗


packages/Card/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/helpers": "^0.2.11",
+    "@paprika/stylers": "^0.2.7",
+    "@paprika/tokens": "^0.1.14",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
+    "styled-components": "^4.2.0"
```
nahumzs, line 26: Do we have any l10n component need from aria perspective @oscarkwan ? if we do let's add L10n as peer if not ignore my comment.

nahumzs:
Please add an .npmignore ex: file https://github.com/acl-services/paprika/blob/master/packages/Button/.npmignore

packages/Card/src/Card.js
```diff
@@ -0,0 +1,78 @@
...
+  isAutoHeight: PropTypes.bool,
+
+  /** If the card is in an "active" or "selected" state. */
+  isActive: PropTypes.bool,
+
+  /** If the card is disabled. */
+  isDisabled: PropTypes.bool,
+
+  /** Callback to be executed when the card is clicked or activated by keyboard. Typically required. */
+  onClick: PropTypes.func,
```
nahumzs, line 30: 🤔 I see the intention, but it seems to me that the Card shouldn't be responsible to handle this behaviour. 

What you want to do instead, is to compose the `<Card />` component with another component that we have for this kind of the scenarios the `RawButton`. So if I were a consumer I will do something like this:

```js
import RawButton from "@paprika/raw-button"

function SomeThing() {
  return (
    <RawButton onClick={() => {}}>
       <Card>{...content}</Card>
    </RawButton>
  )
}
```

thoughts?

nahumzs: the `isDisabled` behaviour will also be delegated to the `raw-button`

packages/Card/src/Card.js
```diff
@@ -0,0 +1,78 @@
...
+  children: null,
+  size: ShirtSizes.MEDIUM,
+  isFullWidth: false,
+  isAutoHeight: false,
+  isActive: false,
+  isDisabled: false,
+  onClick: () => {},
+};
+
+const Card = props => {
```
nahumzs, line 46: please use `function Card(props) {}`

While arrow function will do the work, it's better for readability to use the `function Name() {}` to mark a point where the component starts.

packages/Card/src/Card.styles.js
```diff
@@ -0,0 +1,81 @@
+import styled from "styled-components";
+import tokens from "@paprika/tokens";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+
+// Common
+
+const commonStyles = styled.div`
+  background: ${tokens.color.white};
+  border: 1px solid #f0f0f0;
```
nahumzs, line 9: we have `${tokens.border.color}`

packages/Card/src/Card.styles.js
```diff
@@ -0,0 +1,81 @@
...
+  box-shadow: ${tokens.shadow};
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+  position: relative;
+`;
+
+// Sizes
+
+const sizeStyles = {
```
nahumzs, line 20: 👏 great use of computed properties

packages/Card/src/Card.styles.js
```diff
@@ -0,0 +1,81 @@
...
+// States
+
+const disabledStyles = `
+  box-shadow: none;
+  color: ${tokens.color.blackDisabled};
+  cursor: not-allowed;
+  text-decoration: none;
+  text-shadow: none;
+
+  &:focus {
```
nahumzs, line 56: 🤔 why do we have two focus are we overriding something? I think @oscarkwan might have a better understanding for this.

packages/Card/src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import * as sc from "./Avatar.styles";
+
+const propTypes = {
+  /** Avatar content. It can be initial as a string or icon */
+  children: PropTypes.node,
+  /** Background color of the Avatar */
+  color: PropTypes.oneOf(["black", "blue", "green", "orange", "pink", "red"]),
```
nahumzs, line 10: should not be a defined colour for the background, the colour can be any string, we will not validate that the string passes down its a valid colour that its responsibility of the consumer.

packages/Card/src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,51 @@
...
+  /** Randomize background color */
+  isRandomBackground: PropTypes.bool,
```
nahumzs, line 12: this shouldn't be an option, in an advanced scenario the colour is created depending on the name of the user, so if you have this kind of avatar in a chat you have less probability to have two users with the same initial and same background colour.

packages/Card/src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,51 @@
...
+  /** Size of Avatar */
+  size: PropTypes.oneOf(ShirtSizes.LIMITED),
+};
+
+const defaultProps = {
+  children: "C",
```
nahumzs, line 18: the default value should be `""` empty

packages/Card/src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,51 @@
...
+    size,
+  };
+
+  if (isRandomBackground) {
+    const colors = ["black", "blue", "green", "orange", "pink", "red"];
+    const randomColor = colors[Math.floor(Math.random() * colors.length)];
+    avatarProps.color = randomColor;
+  }
+
+  return (
```
nahumzs, line 40: we might need to accept the background and foreground for the text or icon. So dark and light contrasts can exist without a problem.

packages/Card/src/Card.styles.js
```diff
@@ -0,0 +1,81 @@
...
+  box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
+  border-color: ${tokens.highlight.active.noBorder.borderColor};
+`;
+
+export const cardStyles = styled(commonStyles)`
+  ${props => sizeStyles[props.size]}
+  ${props => (props.isActive ? activeStyles : "")}
+  ${props => (props.isFullWidth ? fullWidthStyles : "")}
+  ${props => (props.isAutoHeight ? autoHeightStyles : "")}
+  ${props => (props.isDisabled ? disabledStyles : "")}
```
jamiek-acl, line 80: Another way to do these that is a little easier on the eyes seems to have some traction:

    ${props => (props.isDisabled && disabledStyles)}

packages/Card/src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,51 @@
+import React from "react";
```
jamiek-acl, line 1: Maybe a question for @nahumzs shouldn't Avatar be it's own component?

packages/Card/src/components/Avatar/Avatar.js
jamiek-acl: @nahumzs the acl-ui Avatar component took in a string, and it decided on what colours to use, and what to display. That is not the plan anymore?

packages/Card/src/components/Content/Content.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+const defaultProps = {
+  children: null,
+};
+
+function Content(props) {
+  const { children } = props;
+
+  return <sc.Content data-pka-anchor="content">{children}</sc.Content>;
```
jamiek-acl, line 17: Maybe the anchor should be more specific like BEM: `card--content`

packages/Card/src/components/Footer/Footer.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import PropTypes from "prop-types";
+import * as sc from "./Footer.styles";
+
+const propTypes = {
+  /** Body content of the button (an icon). */
```
jamiek-acl, line 6: What does `(an icon)` mean? The footer doesn't have to be an icon

packages/Card/src/components/Footer/Footer.styles.js
```diff
@@ -0,0 +1,18 @@
...
+
+export const footerStyles = styled.div`
+  background: ${tokens.color.blackLighten70};
+  border-bottom-left-radius: ${tokens.card.borderRadius};
+  border-bottom-right-radius: ${tokens.card.borderRadius};
+  bottom: 0;
+  box-sizing: border-box;
+  ${stylers.fontSize(-1)};
+  left: 0;
+  min-height: 40px;
```
jamiek-acl, line 13: Maybe 

    import { spacer } from '@paprika/stylers/lib/helpers';
    ...
    min-height: ${spacer(5)};`

packages/Card/src/components/Text/Text.js
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import PropTypes from "prop-types";
+import * as sc from "./Text.styles";
+
+const propTypes = {
+  /** Body content of the button (an icon). */
```
jamiek-acl, line 6: Copy/paste `(an icon)`

packages/Card/src/components/Text/Text.js
```diff
@@ -0,0 +1,47 @@
...
+  children: PropTypes.node,
+  isTruncate: PropTypes.bool,
```
jamiek-acl, line 8: I think this should be `shouldTruncate`

packages/Card/src/components/Text/Text.styles.js
```diff
@@ -0,0 +1,12 @@
+import styled from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+
+export const textStyles = styled.p`
+  color: ${tokens.color.black};
+  ${stylers.fontSize(-2)};
+  height: 55px;
```
jamiek-acl, line 8: Should this be 56, which is divisible by 8? and therefore use the `spacer` helper

packages/Card/src/components/Title/Title.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import PropTypes from "prop-types";
+import * as sc from "./Title.styles";
+
+const propTypes = {
+  /** Body content of the button (an icon). */
```
jamiek-acl, line 6: Copy paste. Please check all the other files, Im going to stop marking them.

packages/Card/src/components/Title/Title.styles.js
```diff
@@ -0,0 +1,10 @@
+import styled from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+
+export const titleStyles = styled.h1`
+  color: ${tokens.color.black};
+  ${stylers.fontSize(0)};
+  margin: 0px;
+  padding: 0px;
```
jamiek-acl, line 9: Small thing but you can just say `0`, dont need `px`

.storybook/config.js
```diff
@@ -41,6 +41,9 @@ const stories = [
   // CodeViewer
   require.context("./components/CodeViewer", false, /CodeViewer.story.js/),
 
+  // Card
```
jamiek-acl, line 44: @mikrotron why do we have to do this? This is usually automated

packages/Card/README.md
```diff
@@ -0,0 +1,46 @@
...
+<Card size="medium">
+    <Card.Header>
+        <Avatar isRandomBackground>E</Avatar>
+    </Card.Header>
+    <Card.Content>
+        <Card.Title>Example</Card.Title>
+        <Card.Metadata>metadata</Card.Metadata>
+        <Card.Seperator size="small" />
+        <Card.Text>This is an example of a medium sized Card</Card.Text>
+        <Card.Seperator size="small" />
```
jamiek-acl, line 26: You spelled `Separator` wrong in the docs 😆 😭

packages/Card/src/components/Avatar/Avatar.js
oscarkwan: I think mentioned in the PR description that for now it will live here. extracting it laters

oscarkwan: Can we put a TODO: Comment here to extract it as its own package later?

packages/Card/README.md
```diff
@@ -0,0 +1,46 @@
+<!-- autogenerated don't modify -->
+
+<!-- autogenerated don't modify -->
+<!-- content -->
+
+## Card
+
+The <Card> component is a customizable content container. It comes with subcomponents that allow for quick and easy creation of various card types for different use cases. Additionally any other child elements can be passed into the Card.
```
oscarkwan, line 8:
```suggestion
The `<Card>` component is a customizable content container. It comes with subcomponents that allow for quick and easy creation of various card types for different use cases. Additionally any other child elements can be passed into the Card.
```

packages/Card/stories/Card.mdx
```diff
@@ -0,0 +1,200 @@
...
+    </Card.Header>
+    <Card.Content>
+        <Card.Title>Card</Card.Title>
+        <Card.Metadata>100,000 total</Card.Metadata>
+    </Card.Content>
+</Card>
+```
+#### Result
+
+<Card size="small"><Card.Header><Avatar><Icon /></Avatar></Card.Header><Card.Content><Card.Title>Card</Card.Title><Card.Metadata>100,000 total</Card.Metadata></Card.Content></Card>
```
oscarkwan, line 76: Screen Shot 2020-05-25 at 1 24 46 PM

The avatar is cut off a little here

packages/Card/stories/Card.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+
+<Card size="small" isFullWidth><Card.Header><Avatar><Icon /></Avatar></Card.Header><Card.Content><Card.Title>Card</Card.Title><Card.Metadata>100,000 total</Card.Metadata></Card.Content></Card>
+
+### Making the Card height auto
+
+You can change the Card to be the full height based of the children elements passed into the card by adding the isAutoHeight property to the `<Card/>` component and setting it to true.
+
+```js
+<Card size="small" isAutoHeight='true'>
```
oscarkwan, line 185: `isAutoHeight` is a boolean property, so can just do `isAutoHeight`

packages/Card/src/components/Avatar/Avatar.styles.js
```diff
@@ -0,0 +1,56 @@
...
+  `,
+  red: `
+  background: ${tokens.color.chartColor01};
+  `,
+};
+
+export const avatarSizeStyles = {
+  small: `
+    border-radius: 10px;
+    height: 30px ;
```
oscarkwan, line 29: extra space

packages/Card/src/components/Avatar/Avatar.styles.js
```diff
@@ -0,0 +1,56 @@
...
+    width: 30px;
+    ${stylers.fontSize(2)} 
+  `,
+
+  medium: `
+    border-radius: 12px
```
oscarkwan, line 35: semi colon

packages/Card/src/components/Avatar/Avatar.styles.js
```diff
@@ -0,0 +1,56 @@
...
+    height: 40px ;
```
oscarkwan, line 36: space

packages/Card/src/components/Separator/Separator.js
```diff
@@ -0,0 +1,32 @@
...
+const defaultProps = {
+  size: ShirtSizes.SMALL,
+};
+
+const Separator = props => {
+  const { size } = props;
+
+  const separatorProps = {
+    size,
+  };
```
oscarkwan, line 20: Seems a bit weird here, what exactly are you trying to do here?

packages/Card/src/Card.js
oscarkwan: If we are going for this `onClick` approach, we will need to create a story to showcase this so we can test for accessibility and other things

packages/Card/src/Card.styles.js
oscarkwan: I think they are just trying to extract out the styles. But yeah I would also like to know more on what you are trying to achieve here. I couldn't find a story with a disabled card style so i couldn't really test this part

.storybook/config.js
mikrotron: Gives us control over the ordering.  Probably a better way, this is just a temporary workaround.

packages/Card/stories/examples/BasicCardContent.js
```diff
@@ -0,0 +1,90 @@
...
+      questions: "10",
+      creator: "Cory McBain",
+      updatedBy: "Charles Bradley",
+    },
+  ];
+
+  return (
+    <Story>
+      <div style={{ display: "flex", justifyContent: "flex-start", flexWrap: "wrap" }}>
+        {card.map(asset => {
```
jamiek-acl, line 69: General tip... if you named your array `cards` then you could use the word `card` instead of `asset` (which doesn't help people understand).

    cards.map(card => {...

packages/Card/stories/examples/BasicCardContent.js
```diff
@@ -0,0 +1,90 @@
...
+import { Story } from "storybook/assets/styles/common.styles";
+import Card from "../../src";
+
+export default () => {
+  const card = [
+    {
+      title: "Master view",
+      attrTypes: "25",
+      questions: "100",
+      creator: "Cory McBain",
```
jamiek-acl, line 11: When the title, creator, or updater are longer than one line, it doesn't handle it gracefully.  Perhaps it should add `...`

packages/Card/src/components/Separator/Separator.js
```diff
@@ -0,0 +1,27 @@
...
+      &nbsp;
+    </sc.separatorStyles>
+  );
+};
+
+Separator.displayName = "Separator";
+Separator.propTypes = propTypes;
+Separator.defaultProps = defaultProps;
+
+export default Separator;
```
jamiek-acl, line 27: So this has 3 sizes or whatever.  Let's say small is 10px, medium is 20px, large is 30px.
What happens when a designer inevitably wants 24px of separation?

And are we supposed to use it like `<br />`s?  That doesn't feel right:

    ...content...
    <Separator />
    <Separator />
    <Separator />
    ...more content...

jamiek-acl: And what about when a card has more content than the one beside it? i dont think this would be possible with so much control over the content:

Screen Shot 2020-05-25 at 3 26 45 PM

The first card would be cut off the “Last updated by” line.
You support flexibility in the header and footer, but not the body.  In my opinion the Card's API should be simpler, not micro-managing as much:

    <Card>
      <Header />
      <Body>
        <Title />
        ...any nodes that i want...
      </Body>
      <Footer />
    </Card>

I bring this up for discussion with @acl-services/ux-devs, not requesting immediate changes.

mikrotron: I agree.  I don't think a `<Separator>` component is the right approach. We can add `margin-bottom` via the parent (`<Content>`) component, or the `<Content>` (or `<Body>` in Jamie's example) could contain another container that the consumer supplies and would have min-height (to match heights) or margin or whatever it wants.

oscarkwan: Agreed, separator not the right approach.

packages/Card/src/components/Separator/Separator.js
KaanDarcey: Was attempting to keep things consistent with the way we set up the other components. However, it was kinda unnecessary so i removed it.

packages/Card/src/components/Avatar/Avatar.js
nahumzs: Yes it is, but I think that shouldn't be part of the avatar component that should be part of another logic right? maybe like:
```js
import Avatar from "@paprika/avatar";

const colors = Avatar.getColors("my name");
<Avatar background={colors.background} color={colors.color} />
```

with this approach, you can either use it in a chat environment or in Assets and Attributes where in the future we might want to let the user set custom colours for their cards.

.storybook/config.js
nahumzs: yes, I have on my list to remove this, and dynamically load it and sort it by a json or something else.

packages/Card/stories/examples/BasicCardContent.js
nahumzs: Maybe we can add this in next iteration? what do you think?

packages/Card/src/components/Avatar/Avatar.js
alexzherdev: In Assets, at the time when an Asset Type is created, a colour is assigned to it. Then that colour is always shown when the Asset Type is displayed. So we can't pick a random colour. (Although it totally makes sense if the component can do both)

packages/Card/src/components/Separator/Separator.js
nahumzs: I actually think the separator it's the correct approach. But with the ability to add custom height. My proposal would be something like:

`<Separator space={Separator.small} />` you only need one separator.

or 

`<Separator space={45} />`

But definitely like oscar said once, won't going to die in this hill. 

So please either let Karen and Kaan if we should remove this component or add my modification.

But lets take a decision so Jamie can start using this component :)

oscarkwan: If we only need to use 1 Separator, then i'm okay for now. Can we add that prop in instead of removing it completely. Should be quick to fix?

jamiek-acl: Wont a single separator still have problem with spacing between two sections of text?  In English the two items will fit correctly, but in German they wont (since the text could be a line longer, the space would need to be smaller).

packages/Card/stories/examples/BasicCardContent.js
jamiek-acl: Sure, I won't lose sleep over it :)

packages/Card/src/Card.styles.js
```diff
@@ -0,0 +1,59 @@
...
+
+// States
+
+const activeStyles = `
+  box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
+  border-color: ${tokens.highlight.active.noBorder.borderColor};
+`;
+
+export const cardStyles = styled(commonStyles)`
+  ${props => sizeStyles[props.size]}
```
nahumzs, line 55: this could be one function:
```js
${props => {
  const { isActive, isAutoHeight, fullWidthStyles } = props;
  const size = sizeStyles[props.size];
   const active = isActive ? activeStyles : "";
  const fullWidth ? fullWidthStyles : "";
  const autoHeight ? autoHeightStyles : "";
  
  return `
       ${size};
       ${active};
       ${fullWidth};
       ${autoHeight};
   `
}}
```

packages/Card/src/components/Separator/Separator.js
nahumzs: ohh I understand now what do you mean Jamie, make sense to remove the Separator for now and let the consumer deal with this depending on their needs. make sense?

jamiek-acl: It makes sense to me

nahumzs:
screener, it is failing in one sidepanel's story, I don't think it is a big of a deal, also I don't see how this PR can affect the Sidepanel.  
But can you help me checking on it @mikrotron and help me to identify if that's actually a false positive or a real issue?

mikrotron:
> screener, it is failing in one sidepanel's story, I don't think it is a big of a deal, also I don't see how this PR can affect the Sidepanel.
> But can you help me checking on it @mikrotron and help me to identify if that's actually a false positive or a real issue?

Yeah, I bet this is just a timing issue.  The changed shot has the close button with the focus highlight, which would happen after a delay, so maybe it's going to be an intermittent issue?

packages/Card/src/Card.styles.js
```diff
@@ -0,0 +1,59 @@
+import styled from "styled-components";
+import tokens from "@paprika/tokens";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+
+// Common
```
tristanjasper, line 5: not sure this or https://github.com/acl-services/paprika/pull/531/files#diff-b3b863891964f1b45c757d2de0e060e0R18 is necessary

packages/Card/src/components/Avatar/Avatar.styles.js
```diff
@@ -0,0 +1,34 @@
+import styled from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+
+export const avatarSizeStyles = {
+  small: `
+    border-radius: 10px;
+    height: 30px;
```
tristanjasper, line 8: width and height could use the same variable set in this file. Same for medium below

packages/Card/src/components/Footer/Footer.styles.js
```diff
@@ -0,0 +1,19 @@
+import styled from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+import { spacer } from "@paprika/stylers/lib/helpers";
+
+export const footerStyles = styled.div`
+  background: ${tokens.color.blackLighten70};
+  border-bottom-left-radius: ${tokens.card.borderRadius};
```
tristanjasper, line 8: perhaps set the border radius here the the same as you did at https://github.com/acl-services/paprika/pull/531/files#diff-a13b744a53b71e448868586e137c4388R5

packages/Card/src/components/Text/Text.js
```diff
@@ -0,0 +1,47 @@
...
+  const textProps = {
+    children,
+    shouldTruncate,
+  };
+
+  function truncateText(text, num) {
+    if (text.length <= num) {
+      return text;
+    }
+    return text.slice(0, num).concat("...");
```
tristanjasper, line 28: Not sure if it is but do have the `lodash.truncate` package already available and is being used in our Collapsible text component, may not be necessary though, https://github.com/acl-services/paprika/blob/master/packages/CollapsibleText/src/CollapsibleText.js#L29

packages/Card/src/components/Text/Text.styles.js
```diff
@@ -0,0 +1,13 @@
+import styled from "styled-components";
+import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
+import { spacer } from "@paprika/stylers/lib/helpers";
+
+export const textStyles = styled.p`
+  color: ${tokens.color.black};
+  ${stylers.fontSize(-2)};
+  height: ${spacer(8)};
+  margin: 0px;
```
tristanjasper, line 10:
```suggestion
  margin: 0;
```

packages/Card/src/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,47 @@
...
+    size,
+    color,
+  };
+
+  return (
+    <sc.Avatar
+      style={{ backgroundColor: avatarProps.backgroundColor, color: avatarProps.color }}
+      data-pka-anchor="avatar"
+      {...avatarProps}
+    >
```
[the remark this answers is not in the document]
nahumzs: we are goiing to split the Avatar package in another PR. Don't worry about this too much

packages/Card/src/Card.styles.js
nahumzs: do you mean the comment?

nahumzs:
Let's merge it @karencumlat looks good!

Issue #349, opened 2019-12-16
Create new <SidePanel.Content> component


oscarkwan:
# Bug Report

If app consumer wants to have a side panel without padding, it is not possible without overwriting the css of the `<Dialog>`. Therefore we need to expose a `<SidePanel.Content>` component to the consumer but right now it is embedded into the `<SidePanel>`

PR #533, opened 2020-05-26
STAR-234 SidePanel.Conent sub component

KaanDarcey:
### Purpose 🚀
Resolves issue #349 

To create a ```<SidePanel.Content>``` sub component that allows you to set content within the side panel without any preset padding.

### Notes ✏️
The SidePanel component has the Dialog set with padding by default. Therefore to override the existing padding the Content component makes use of negative margin to offset it parents padding.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-234

### Screenshots 📸
Without ```<SidePanel.Content>``` :
Screen Shot 2020-05-26 at 11 46 53 AM

With ```<SidePanel.Content>``` :
Screen Shot 2020-05-26 at 11 47 07 AM

### References 🔗


packages/SidePanel/src/components/Content/Content.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from "styled-components";
+import stylers from "@paprika/stylers";
+
+export const contentStyles = styled.div`
+  margin-left: ${stylers.spacer(-3)};
```
tristanjasper, line 5: I just wonder if this is the best solution to apply negative margins on this component, rather than just setting the padding to 0 on the parent itself, and having a prop like `hasZeroPadding` to set it to zero padding? I know we want to avoid breaking api changes but it might be a good time to?

packages/SidePanel/stories/examples/Examples.js
```diff
@@ -232,3 +232,46 @@ export const ZIndexSidePanel = () => {
...
+        <SidePanel isOpen={isOpen} onClose={toggle} zIndex={99}>
+          <SidePanel.Overlay />
+          <SidePanel.Trigger onClick={toggle}>Open SidePanel</SidePanel.Trigger>
+          <SidePanel.Header>Header</SidePanel.Header>
+          <SidePanel.Content>
+            <p style={{ marginTop: 0 }}>
+              The <code>zIndex</code> prop of this <code>&lt;SidePanel&gt;</code> is also <code>99</code>.
+            </p>
+            <p>
+              Because the content is renderered as a <code>&lt;Portal&gt;</code> at the end of the DOM, it will be
```
oscarkwan, line 258:
```suggestion
              Because the content is rendered as a <code>&lt;Portal&gt;</code> at the end of the DOM, it will be
```

packages/SidePanel/src/components/Content/Content.styles.js
nahumzs: good point, I wonder he is trying to not break the current implementation, but I think we should not use margins at all, use the parent container and set it to padding 0, and the the `<Content />` can have padding=8 16 whatever its need. 

How do you feel about breaking the current implementation @mikrotron @oscarkwan

mikrotron: Break it. The version is `< 0`, that means it's not yet stable.  We shouldn't be too reluctant to make breaking _improvements_.

nahumzs:
@KaanDarcey can you please remove the padding from the main container and move it inside of `<SidePanel.Content />` 🙏 

The expected results its If a consumer code `<SidePanel>whatever</SidePanel>` shouldn't be any gap or space between the edge `no padding no margin`.

packages/SidePanel/src/components/Content/Content.styles.js
alexzherdev: this component is so WIP we had to give it a negative version

nahumzs:
@KaanDarcey can you implement `<SidePanel.Content />` in all our stories? so they don't expand 100% without space?

PR #534, opened 2020-05-26
STAR-235 DropdownMenu scroll on overflow

karencumlat:
### Purpose 🚀
I added a new prop `maxHeight` to stop the `<DropdownMenu/>` to overflow off the screen and enable scrolling. 

If the height of the menu prevents all the menu items from being displayed, the menu can be scrollable.
### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-235-DropdownMenu-scroll-onOverflow
### Screenshots 📸
withScroll
woScroll

### References 🔗


packages/DropdownMenu/stories/examples/DropdownMenuScroll.js
```diff
@@ -0,0 +1,65 @@
...
+          Groot. We are Groot. We are Groot. We are Groot. We are Groot. I am Groot. I am Groot. We are Groot. I am
+          Groot. I am Groot. I am Groot. We are Groot. We are Groot. We are Groot. We are Groot. I am Groot. I am Groot.
+          We are Groot. I am Groot. We are Groot. I am Groot. I am Groot. I am Groot. I am Groot. We are Groot. We are
+          Groot. I am Groot. We are Groot. We are Groot. We are Groot. I am Groot. We are Groot. I am Groot. We are
+          Groot. I am Groot. I am Groot. We are Groot. We are Groot. I am Groot. I am Groot. We are Groot. I am Groot. I
+          am Groot. I am Groot. We are Groot. We are Groot. I am Groot. We are Groot. I am Groot. I am Groot. We are
+          Groot. We are Groot. We are Groot. I am Groot. I am Groot. I am Groot. I am Groot. I am Groot. We are Groot. I
+          am Groot. We are Groot. We are Groot. I am Groot. I am Groot. We are Groot. I am Groot. We are Groot. I am
+          Groot. I am Groot. We are Groot. I am Groot. I am Groot. We are Groot. I am Groot. I am Groot. We are Groot. I
+          am Groot. We are Groot. I am Groot. I am Groot. I am Groot.
```
tristanjasper, line 58: 🥇 😱

packages/DropdownMenu/src/DropdownMenu.styles.js
```diff
@@ -5,6 +5,8 @@ import stylers from "@paprika/stylers";
 export const contentStyles = css`
   margin-left: -${tokens.spaceLg};
   margin-right: -${tokens.spaceLg};
+  overflow-x: hidden;
```
tristanjasper, line 8: not sure if this is necessary, did you find a particular situation that you needed it?

tristanjasper: I'm seeing a bug (on master as well, so perhaps not related but perhaps fixing in this pr when trying to focus the edit option in the dropdown via keyboard navigation...
image
If not fixed here perhaps create another ticket (or let me know and i can create one) so that we are tracking that issue

tristanjasper:
I think the main issue i'm seeing here is with keyboard navigation, as it appears to jump down quite a way and doesn't keep the selected one in view. Would be worth investigating a solution for this as it does affect UX quite a bit.
overflow-keyboard-navigation

nahumzs:
> I think the main issue i'm seeing here is with keyboard navigation, as it appears to jump down > quite a way and doesn't keep the selected one in view. Would be worth investigating a > solution for this as it does affect UX quite a bit.

I think this one and the previous one can be filed in a github issue and work on them in a different instance?

packages/DropdownMenu/src/DropdownMenu.styles.js
karencumlat: For the `overflow-x:hidden`, if the menu item uses a long word it overflows. 
overflowx

There's a `max-height` for the `<DropdownMenu/>` so I'm not sure how to approach it. Either use `overflow-x:hidden`
overflowxHidden

or `overflow-wrap:break-word`
overflowWrap

nahumzs: Let's remember its a menu, no really a content holder, I think to apply overflow: hidden should be good for now.

tristanjasper: Agree, go with `overflow: hidden` for now as I imagine could run into issues if there is a max-height on the dropdown.... We can always revisit this issue

tristanjasper:
> > I think the main issue i'm seeing here is with keyboard navigation, as it appears to jump down > quite a way and doesn't keep the selected one in view. Would be worth investigating a > solution for this as it does affect UX quite a bit.
> 
> I think this one and the previous one can be filed in a github issue and work on them in a different instance?

Yep, lets create an issue for this one

PR #535, opened 2020-05-27
Snp 182 fix listbox

allison-c:
### Purpose 🚀

the hook `useChildrenLengthChange ` is wrong for updating the `options` in the state, cause sometimes there might be options got updated, but remained the same number of them. This probably not the best way to do, but we want to solve the issue in people scheduler as soon as possible

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/SNP-182-fix-listbox

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/ListBox/src/components/Trigger/Trigger.js
+++ b/packages/ListBox/src/components/Trigger/Trigger.js
@@ -52,7 +52,17 @@ const defaultProps = {
 export default function Trigger(props) {
   const [state, dispatch] = useListBox();
   const onChangeContext = React.useContext(OnChangeContext);
-  const { placeholder, hasClearButton, onClickFooterAccept, children, isHidden, ...moreProps } = props;
+  const {
+    placeholder,
+    hasClearButton,
+    onClickFooterAccept,
+    children,
+    isHidden,
+    onClickClear,
+    // eslint-disable-next-line react/prop-types
+    onFooterClickAccept,
+    ...moreProps
+  } = props;
   const {
     isDisabled,
     formElementLabelDescribedBy,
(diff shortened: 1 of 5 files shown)


tristanjasper:
LGTM

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -52,7 +52,17 @@ const defaultProps = {
...
-  const { placeholder, hasClearButton, onClickFooterAccept, children, isHidden, ...moreProps } = props;
+  const {
+    placeholder,
+    hasClearButton,
+    onClickFooterAccept,
+    children,
+    isHidden,
+    onClickClear,
+    // eslint-disable-next-line react/prop-types
+    onFooterClickAccept,
```
allison-c, line 63: Not sure why this has been passed in but it's not used anywhere

packages/ListBox/src/hooks/hooks.js
```diff
@@ -7,17 +8,22 @@ function waitForPopoverAnimation(func) {
...
   React.useEffect(() => {
-    if (Object.keys(state.options).length !== React.Children.count(children)) {
-      const options = getDataOptions(children);
+    const options = getDataOptions(children);
 
-      dispatch({
-        type: useListBox.types.updateOptions,
-        payload: options,
-      });
+    if (Object.keys(state.options).length === Object.keys(options).length) {
```
nahumzs, line 16: ohhh nice, way better approach allison 💯 ... so now we don't need to do the approach that we did for `Load more` when we have to add an `isHidden` on the option. Because now it is tracking the probs as well.

nahumzs: It shouldn't impact too much the performance because it will be run only when the children change.

PR #536, opened 2020-05-27
Tw 154 Button Group and Raw Button support imperative focus method

tristanjasper:
### Purpose 🚀
Update ButtonGroup component to include an imperative api focus method which can be used by the formElement to get focus on label click which then shows shows an outline on the first available item

Additionally: 

- Update `raw-button` component to support an imperative api method of `focus` which when called ensure that a focus event occurs and forces the outline to be shown. - Required for ListBox integration with formElement
- Updated formElement to include a new `onClickLabel` prop which allows calling a child ref focus method.

Fixes 

### Notes ✏️
`Button` component will need to be updated as `raw-button` is a dependency of `Button` component, which intern is a dependency of ButtonGroup component.

### Updates 📦
- [x] PATCH (bug fix) change to RawButton, ButtonGroup, Button, FormElement

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/tw-154-button-group-support-labelref

### Screenshots 📸
After:
focusapi

### References 🔗
https://aclgrc.atlassian.net/browse/TW-154


packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -90,6 +93,24 @@ function ButtonGroup(props) {
...
 
+  React.useLayoutEffect(() => {
+    const $label = refLabel && refLabel.current;
+
+    if (!$label) return;
+
+    function handleClickLabel() {
+      const enabledIndexes = getEnabledIndexes();
+      focusButton(enabledIndexes[0]);
+      console.log("focusing first button", document.activeElement);
```
tristanjasper, line 104: currently finding the button focusing but the imperative api `focus` function isn't being triggered. Investigating whether a dedicated ref is needed for each `buttonGroup.Item` in order the call the `focus()` api on that component.

nahumzs:
What about to add a onClickLabel on the `<FormElement />` and instead passing a refLabel you just get the ref from the <ButtonGroup /> like



```js
  const refButtonGroup = React.useRef(null);

    <FormElement label="Form Label" onClickLabel={() => {refButtonGroup.focus()}}>
      <FormElement.Content>
          {() => <ButtonGroup ref={refButtonGroup}>{buttonGroupOptions}</ButtonGroup>}
      </FormElement.Content>
    </FormElement>
```

the imperative handler focus function for ButtonGroup could be something like

```js
return {
   focus(index = 0) {
       const $findIndexElement = getTheRefElementSomewhere(index);
      $findIndexElement.current.focus()
   }
}
```
This approach would be less invasive than forcing to received a **refLabel**.

packages/RawButton/src/RawButton.js
```diff
@@ -46,18 +46,28 @@ const RawButton = React.forwardRef((props, ref) => {
...
 
   React.useImperativeHandle(ref, () => ({
     focus: () => {
+      setHasForcedFocus(true);
       rawButtonRef.current.focus();
     },
   }));
 
   const handleClick = event => {
+    setHasForcedFocus(false);
```
nahumzs, line 59: I don't think you need react state at all. you can simply use regular dom and `setAttribute`

like 
```js
// adding
$dom.setAttribute("data-has-forced-focus", true)
// removing
$dom.removeAttribute("data-has-forced-focus);
```

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -90,6 +93,28 @@ function ButtonGroup(props) {
...
   }
 
+  React.useLayoutEffect(() => {
+    const $label = refLabel && refLabel.current;
+
+    if (!$label) return;
+
+    function handleClickLabel() {
+      const enabledIndexes = getEnabledIndexes();
+      const firstEnabledIndex = enabledIndexes[0];
```
mikrotron, line 103:
```suggestion
      const firstEnabledIndex = getEnabledIndexes()[0];
```

mikrotron: A little terser; just as readable (in my opinion).

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -152,7 +177,11 @@ function ButtonGroup(props) {
 
   return (
     <sc.ButtonGroup {...props} ref={groupRef} onKeyDown={handleKeyDown}>
-      <ButtonGroupContext.Provider value={contextValue}>{children}</ButtonGroupContext.Provider>
+      <ButtonGroupContext.Provider value={contextValue}>
+        {React.Children.map(children, (element, index) => {
+          return React.cloneElement(element, { ref: childRefs[index] });
```
mikrotron, line 182: I went to a fair bit of effort to avoid `cloneElement()` so this component would support children like this:

https://github.com/acl-services/paprika/blob/master/packages/ButtonGroup/stories/examples/Sandbox.js#L36-L47

(with wrapping fragments, `<divs>`, etc)

mikrotron: I think you can probably avoid the need for a `childRefs[]` array, and instead utilize `focusButton()`

https://github.com/acl-services/paprika/blob/master/packages/ButtonGroup/src/ButtonGroup.js#L88-L91

packages/RawButton/src/RawButton.js
mikrotron: Using state is consistent with the `<Button>` though:

https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.js#L118

And is arguably more readable? Maybe @jamiek-acl had other reasons for using state?

nahumzs: ??? but state it's not even needed here, what readability has to do with use proper javascript? just curious.

tristanjasper: I have changed it (and was also going to do so in the Button component), i am not completely clear on the pros versus cons of this approach of using a state change though. @nahumzs can you expand on that?

mikrotron: I'm not _sure_ but if you do `$dom.setAttribute()` that will manipulate the DOM right now, vs setting state, which will leave it to React do decide when to manipulate the DOM.  I know manipulating the DOM is slow, so if we can leave it to React, won't it make the best (most efficient) decision about when it should really do it?

nahumzs: `"data-has-forced-focus"` it's not a state its a `css class / attribute target` that you are adding to modify your element visually,  you don't need to add a state to add that attribute in your dom element you can do it with simple javascript. 

> DOM is slow is not slow 

You already have the ref to the dom element you want to add and remove the attribute doesn't see the perf issue in this such small interaction. And In the end, React also needs to update the Dom. 

But I'm not hard on this to me was just simpler comment in the spirit to write simpler code and avoid useless state variables.

packages/ButtonGroup/src/ButtonGroup.js
tristanjasper: @mikrotron Unfortunately not from what I can tell, that `focusButton()` function is just calling focus on the Dom Element, not the react component that rendered it. So currently I'm finding creating dedicated Refs for each component rendered is the only way to achieve that...

packages/RawButton/src/RawButton.js
tristanjasper: @nahumzs Ok, Just so i understand this for moving forward, are you saying that if a component has to have a class or any attribute that only changes the component visually _and_ there is a ref available you prefer we use `setAttribute` over a state variable?
I don't think we could use that approach for setting a class though.

packages/ButtonGroup/src/ButtonGroup.js
mikrotron: Ah, right, then it's focus highlight remains invisible... 😖 
I still think it's important to avoid `cloneElement()` though.  You may need to put the `childRefs[]` on `state`, then pass a setter on `context` to the items.  Then they can utilize a hook to set their refs and pass them back up.  I do something similar for `defaultIsActive`.

packages/RawButton/src/RawButton.js
nahumzs: It's just for this case and it is not that I prefer a thing over the other one. :)

alexzherdev: thoughts:
> "data-has-forced-focus" it's not a state its a css class / attribute target that you are adding to modify your element visually

I think something like this absolutely is a state. "Modify your element visually" is not an indicator of whether something is state or not. In fact, _everything_ that governs what an element looks like on the screen is state in React _by default_ (if it's not a prop but you catch my drift). That's literally what state is for, to be able to change what's presented on the screen. You set the state => the component re-renders, potentially painting a different set of pixels.

Sidestepping this flow has to have justification. Because after you modify the DOM directly, React no longer knows what's in there (its vdom doesn't match the real one). And React docs even [say](https://reactjs.org/docs/integrating-with-other-libraries.html#integrating-with-jquery-chosen-plugin)
> It is important that if you modify the DOM outside of React flow, you must ensure React doesn’t have a reason to touch those DOM nodes.

I can think of just a few cases when that could be necessary:
- high-performance stuff
- integrating with third-party libs
- perhaps some weird animation shenanigans.

This component is literally a button. In my eyes, not using state here doesn't simplify, it complicates. Having or not having a ref to the element is inconsequential.

> if you do $dom.setAttribute() that will manipulate the DOM right now, vs setting state, which will leave it to React do decide when to manipulate the DOM. I know manipulating the DOM is slow

Yeah, like Nahum said React would basically do the same operation for you anyway. Also, I've seen people misunderstand the time scale of what it takes for a browser to do stuff, perhaps due to vague language in the docs and articles. I believe, unless it results in some complex CSS being applied, setting a single attribute on a DOM element is a matter of milliseconds. Unless you're doing this hundreds and thousands of times in a row, this is literally indistinguishable.

tristanjasper:
> What about to add an onClickLabel on the `<FormElement />` and instead of passing a refLabel you just get the ref from the `<ButtonGroup />` like
> 
> ```js
>   const refButtonGroup = React.useRef(null);
> 
>     <FormElement label="Form Label" onClickLabel={() => {refButtonGroup.current.focus()}}>
>       <FormElement.Content>
>           {() => <ButtonGroup ref={refButtonGroup}>{buttonGroupOptions}</ButtonGroup>}
>       </FormElement.Content>
>     </FormElement>
> ```
> 
> the imperative handler focus function for ButtonGroup could be something like
> 
> ```js
> return {
>    focus(index = 0) {
>        const $findIndexElement = getTheRefElementSomewhere(index);
>       $findIndexElement.current.focus()
>    }
> }
> ```
> 
> This approach would be less invasive than forcing to received a **refLabel**.

Sure, that could work. I went this route to keep it consistent with how we are handling the same issue in ListBox with Trigger - https://github.com/acl-services/paprika/blob/master/packages/ListBox/src/components/Trigger/Trigger.js#L87

I'll create an issue to refactor this.

packages/RawButton/src/RawButton.js
tristanjasper: Ok, Personally I vote we take the approach that is most consistent and in this case I feel using state is certainly a more consistent pattern that is easier for most devs to follow, so will go with that for now.

mikrotron:
> Sure, that could work. I went this route to keep it consistent with how we are handling the same issue in ListBox with Trigger - https://github.com/acl-services/paprika/blob/master/packages/ListBox/src/components/Trigger/Trigger.js#L87

I like the idea of being consistent with other components, in this case `ListBox.Trigger`.  Refactoring them both to work the same way is probably outside the scope of this PR, but it will be a breaking change for both, so it might make sense to follow Nahum's suggestion now and create the GitHub issue as a follow up (which could reference the approach from this PR).

packages/RawButton/src/RawButton.js
nahumzs: I think you are misinterpreting my point or I didn't explain it correctly. And agree it doesn't matter if it's a ref or not. 

My point is that in this particular example there it's not moment when react need to know about the attribute, the OUTPUT of the component will not change with our without the attribute visually speaking.

On the markup the attribute will be added or removed. But visually the output will be always the same.

This visual representation is changing only because there is a class declare **outside of the component** that changes because of that attribute. 

To me, there is no point (apart from doing it in a "react way") to add a state.

> React doesn’t have a reason to touch those DOM nodes.

 ^ and react doesn't need to know about it, it's not gaining anything from knowing it. 

> This component is literally a button. In my eyes, not using state here doesn't simplify, it > complicates

In my eyes, otherwise, actually adding a state add complexity for something that can be solved with two lines of code. But I agree that this can be seen as complexity for some people since it's not the direct way of doing it, or the "react way". But that doesn't mean it is not completely valid and in my eyes even better. :) 

But I guess for the sake of "consistency" with other components we can use state even if add more code.

tristanjasper:
> > Sure, that could work. I went this route to keep it consistent with how we are handling the same issue in ListBox with Trigger - https://github.com/acl-services/paprika/blob/master/packages/ListBox/src/components/Trigger/Trigger.js#L87
> 
> I like the idea of being consistent with other components, in this case `ListBox.Trigger`. Refactoring them both to work the same way is probably outside the scope of this PR, but it will be a breaking change for both, so it might make sense to follow Nahum's suggestion now and create the GitHub issue as a follow up (which could reference the approach from this PR).

Created issue - https://github.com/acl-services/paprika/issues/544

packages/ButtonGroup/src/components/Item/Item.js
```diff
@@ -31,9 +31,13 @@ const Item = props => {
     size,
     selectedItems,
     setSelectedItems,
+    setItemRefs,
   } = React.useContext(ButtonGroupContext);
 
+  const itemRef = React.useRef(null);
+
   React.useEffect(() => {
+    setItemRefs(itemRefs => [...itemRefs, itemRef]);
```
mikrotron, line 40:
```suggestion
    setItemRefs(prevItemRefs => [...prevItemRefs, itemRef]);
```

PR #546, opened 2020-06-03
STAR-237 Automagic component categories 🐐

mikrotron:
### Purpose 🚀
Make it easy to organize Storybook's primary navigation (list of components) in order and by category.

### Notes ✏️
- Story order / categories are controlled by this object: 
  https://github.com/acl-services/paprika/blob/610da23aabfcc7771a7d50548c18de4527c69aee/.storybook/storyTree.js#L2-L79
- The categories chosen were arbitrarily chosen for now, but can be updated easily by changing this one config file.
- A `getStoryName()` helper is also provided for the `stories.js` files to use, which is how they are categorized automatically.
- Shuffled some component stories around to adhere to a newly established pattern for organization (folders for `Examples`, `Backyard/Tests`, `Backyard/Sandbox`, etc).
- Had to migrate to the new `main.js` style of Storybook configuration
  https://github.com/acl-services/paprika/pull/546/files#r434752981
- **BONUS**: Updated `README` to include `npm` badges for a bunch of new components.

### To Do (in follow up PR) 📋
- Recategorized based on results of user research 🔬
- Cypress tests will break if stories are moved currently, but I think there is a way to setup permalinks for stories that we could use in Cypress to resolve this issue.  

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/listbox-browser` (eslint fix)

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-237--component-categories

### Screenshots 📸
Screen Shot 2020-06-03 at 10 47 23 AM

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-237


packages/ListBoxBrowser/src/components/CustomListBox/CustomListBox.js
```diff
@@ -57,7 +57,7 @@ export default function CustomListBox(props) {
 
   return (
     <ListBox key={id} height={height} isMulti={isMulti} isInline onChange={onChange}>
-      <ListBox.Trigger isHidden></ListBox.Trigger>
+      <ListBox.Trigger isHidden />
```
mikrotron, line 60: Not sure why `eslint` started complaining about this, but I guess it counts as a `PATCH` update 🙄

packages/Radio/src/Radio.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
```
mikrotron, line 1: This really shouldn't count as an update... but `lerna publish` may consider it one anyway.

.nvmrc
```diff
@@ -1 +1 @@
-v10.15.1
+v10.17
```
mikrotron, line 1: Needed for the latest version of Storybook.

.storybook/main.js
```diff
@@ -0,0 +1,51 @@
...
+    const path = `../packages/${component}/stories/${component}`;
+    stories.push(
+      `${path}.stories.mdx`,
+      `${path}.stories.js`,
+      `${path}.examples.stories.js`,
+      `${path}.backyard.stories.js`,
+      `${path}.tests.stories.js`
+    );
+  });
+});
```
mikrotron, line 16: This is how the order of the stories is controlled.  Going forward we should adhere to a consistent structure of `stories` files so that we can get the stories within a component ordered the way we want.

.storybook/config.js
```diff
@@ -1,85 +0,0 @@
...
-  require.context("../packages/SidePanel", true, /SidePanel.stories.mdx/),
-  require.context("../packages/SidePanel", true, /SidePanel.stories.js/),
-  require.context("../packages/SidePanel", true, /SidePanel.examples.stories.js/),
-  require.context("../packages/SidePanel", true, /SidePanel.Tests.stories.js/),
-
-  // DataGrid
-  require.context("../packages/DataGrid", true, /DataGrid.stories.(js|mdx)$/),
-
-  // Remaining
-  require.context("../packages", true, /\.stories\.(js)$/),
```
mikrotron, line 82: The parameters of `require.context()` _MUST_ be literals / regexes.  This made it impossible to dynamically load with this approach and necessitated the switch from Storybook's OLD `config.js` (including `webpack.config.js` / `presets.js` / `addons.js`) to the NEW `main.js` (including `preview.js` / `manager.js`) style of configuration.

package.json
```diff
@@ -46,14 +46,14 @@
...
-    "@storybook/addons": "^v6.0.0-alpha.29",
-    "@storybook/react": "^v6.0.0-alpha.29",
+    "@storybook/addon-a11y": "^v6.0.0-beta.20",
+    "@storybook/addon-actions": "^v6.0.0-beta.20",
+    "@storybook/addon-cssresources": "v6.0.0-beta.20",
+    "@storybook/addon-docs": "^v6.0.0-beta.20",
+    "@storybook/addon-knobs": "^v6.0.0-beta.20",
+    "@storybook/addon-links": "^v6.0.0-beta.20",
+    "@storybook/addons": "^v6.0.0-beta.20",
+    "@storybook/react": "^v6.0.0-beta.20",
```
mikrotron, line 56: Linvin' on the _edge_!

packages/Card/stories/Card.screener.stories.js
```diff
@@ -1,12 +1,12 @@
 import React from "react";
+import { getStoryName } from "storybook/storyTree";
 import Screener from "./examples/Screener";
 
 export default {
-  title: " | Card/Automation Tests",
+  title: `${getStoryName("Card")}/Backyard/Tests`,
```
mikrotron, line 6: Hey buddy, why don't you use a `const` here, instead of inlining the `getStoryName()`? This is a `js` file, not a weird `mdx` file that requires you to `export` every declaration!

mikrotron: Oh! Right, good call, buddy.

packages/Collapsible/stories/Collapsible.stories.mdx
```diff
@@ -1,9 +1,14 @@
 import { Meta, Story } from "@storybook/addon-docs/blocks";
 import { Props, Description } from "@storybook/addon-docs/blocks";
 import { Gap } from "storybook/assets/styles/common.styles";
+import { getStoryName } from "storybook/storyTree";
 import Readme from "../README.md";
 import Collapsible from "../src";
 
-<Meta title=" | Collapsible/ Docs" component={Collapsible} parameters={{ options: { showPanel: false } }} />
+<Meta
+  title={`${getStoryName("Collapsible")}/Docs`}
```
mikrotron, line 9: Inlining the `getStoryName()` here makes more sense, since `mdx` is kinda weird about declaring variables.

packages/Card/stories/Card.screener.stories.js
alexzherdev: I see no problem with inlining here 😄

packages/Confirmation/stories/screener.stories.js
```diff
@@ -1,11 +0,0 @@
-import React from "react";
-import { storiesOf } from "@storybook/react";
-import { Story } from "storybook/assets/styles/common.styles";
-
-import ConfirmationExample from "./examples/ConfirmationExample";
-
-storiesOf("Confirmation / screener", module).add("confirmation test", () => (
```
mikrotron, line 7: This is not the only example where I've consolidated `stories.js` files, there were a few places it seemed simpler to do so, but here especially.

.storybook/main.js
```diff
@@ -0,0 +1,51 @@
...
+stories.push("./components/CodeViewer/CodeViewer.story.js");
+stories.push("../packages/**/*.stories.js");
```
mikrotron, line 18: I've refactored all existing component stories to utilize the `storyTree`, but even if you don't, or if you add `stories.js` files for a n existing component that don't match the above `5` patterns, they will still be added to the Storybook because of this catchall.  

Components that aren't in the `storyTree` will appear either above existing stories (for `storiesOf()` stories) or below all of them (for CSF stories).

packages/DataGrid/stories/DataGridstories.js
```diff
@@ -1,7 +0,0 @@
-import DataGrid from "../src";
-
-export default {
-  title: "DataGrid",
-  component: DataGrid,
-  includeStories: [], // or don't load this file at all
-};
```
mikrotron, line 7: I'm not sure why this file existed at all?

.storybook/main.js
allison-c: I think this will be more readable, but just my own opinion
```jsx
const path = `../packages/${component}/stories`;
    stories.push(
      `${path}/${component}.stories.mdx`,
      `${path}/${component}.stories.js`,
      `${path}/${component}.examples.stories.js`,
      `${path}/${component}.backyard.stories.js`,
      `${path}/${component}.tests.stories.js`
    );
```

.storybook/main.js
```diff
@@ -0,0 +1,51 @@
...
+    stories.push(
+      `${path}.stories.mdx`,
+      `${path}.stories.js`,
+      `${path}.examples.stories.js`,
+      `${path}.backyard.stories.js`,
+      `${path}.tests.stories.js`
+    );
+  });
+});
+stories.push("./components/CodeViewer/CodeViewer.story.js");
```
allison-c, line 17: just curious why CodeViewer is outside of storyTree

.storybook/storyTree.js
```diff
@@ -0,0 +1,97 @@
...
+  return `${getCategory(component)} | ${component}`;
+}
+
+function getCategory(component) {
+  let found;
+  storyTree.forEach(catList => {
+    if (catList.components.includes(component)) found = catList.category;
+  });
+  return found;
+}
```
allison-c, line 91: Maybe a shorter way?
```jsx
storyTree.find(catList => catList.components.includes(component)). category
```

.storybook/main.js
mikrotron: It's not a Paprika component, it's a Storybook helper component.  It's `path` does not include `packages` so the regular loop won't work.  Might be worth thinking it through a little more, but for now it's a one-off exception, so I wasn't too concerned about it.

Issue #545, opened 2020-06-02
Card - combine `isAutoHeight` and `size` props


jamiek-acl:
I feel that the `isAutoHeight` prop should be removed, and an `auto` setting can be passed into the existing `size` prop.

I recently used the `Card`, didn't want small/medium/large size, and tried to pass in a `className` to manually set the size to `auto`.  The className didn't get added to the component (maybe it should?). I didn't realize there was a separate `isAutoHeight` prop.

PR #549, opened 2020-06-04
STAR-249 Combine Card `isAutoHeight` and `size` props

karencumlat:
### Purpose 🚀
Combine `<Card/>` `isAutoHeight` and `size` props

### Notes ✏️
- Add `...moreProps`
- Remove `isAutoHeightProp`
- Add `auto` value to size prop `size="auto"`
```
auto: `
height: auto;
width: auto;
`
```
- Set `size` prop default value to `auto`

### Updates 📦

### Storybook 📕

### Screenshots 📸
card-auto-sample

card-auto-size

### References 🔗
Resolves Issue #545


packages/Card/src/Card.js
```diff
@@ -13,35 +13,35 @@ const propTypes = {
...
+  className: null,
+  size: "auto",
   isFullWidth: false,
-  isAutoHeight: false,
   isActive: false,
 };
 
 function Card(props) {
-  const { children, size, isFullWidth, isAutoHeight, isActive } = props;
+  const { children, className, size, isFullWidth, isActive } = props;
```
mikrotron, line 38: Instead of adding a `className` prop, we can keep this component agnostic of classes entirely, and allow other attributes (like `data-test-id`, etc) by simply adding `...moreProps` to the destructure list, then adding that same `...moreProps` to the `cardProps` object below.  This is a standard pattern in Paprika, you can find lots of examples in other components.

mikrotron:
I don't have enough context about this component to suggest the best solution, but specifying the height feels a bit wrong to me.  I think the `height:auto` should be the only case, except maybe offering a `maxHeight` prop (maybe).  Ideally the `<Card>` doesn't care about aligning to other `<Cards>`, the parent component cares about that.  And if the `<Card>` is wrapped by a `flex` container with `align-items: stretch`, then it should align, right?  But the height will really depend on the content, and every time we use the `<Card>` the content could be quite different, and currently, with heights fixed to the `size` prop, there's likely to be a lot of scrolling or empty white space.  Anyway, I'd need to discuss a bit with you to understand the use cases better before really insisting on any particular approach.

kcumlat:
> I don't have enough context about this component to suggest the best solution, but specifying the height feels a bit wrong to me. I think the `height:auto` should be the only case, except maybe offering a `maxHeight` prop (maybe). Ideally the `<Card>` doesn't care about aligning to other `<Cards>`, the parent component cares about that. And if the `<Card>` is wrapped by a `flex` container with `align-items: stretch`, then it should align, right? But the height will really depend on the content, and every time we use the `<Card>` the content could be quite different, and currently, with heights fixed to the `size` prop, there's likely to be a lot of scrolling or empty white space. Anyway, I'd need to discuss a bit with you to understand the use cases better before really insisting on any particular approach.

I remember when we started creating the `<Card>` component, we had the issue on alignment when we have a card group with different card content but should have the same height. For example, 
card-siblings

Though this example is based on Asset types and the only basis of having fixed height or size. But yes, not all `<Card>` content will be the same and the `maxHeight` would be a better solution. Would be great to discuss more about the component.

PR #550, opened 2020-06-04
Star 242 refractor css

KaanDarcey:
### Purpose 🚀
To resolve issue #98 by adjusting all current paprika components to follow the same standard styling patterns. 

No longer will styling be applied by css inline to the component, ex:

 ```<Component css={style} />```. 

Instead styling will be imported and used as follows, ex:

```
import * as sc from "./Card.styles";

  return (
    <sc.cardStyles {...cardProps} data-pka-anchor="card">
      {children}
    </sc.cardStyles>
  );
```

Additionally when using a prop functions inside of a css tag function it should destructure the props in the arrow function parameters ex:
```
export const cardStyles = styled(commonStyles)`
  ${props => {
    const size = sizeStyles[props.size];
    const active = props.isActive && activeStyles;
    const fullWidth = props.isFullWidth && fullWidthStyles;
    const autoHeight = props.isAutoHeight && autoHeightStyles;

    return `
      ${size};
      ${active};
      ${fullWidth};
      ${autoHeight};`;
  }}
`;
```
### Notes ✏️
- Removed  ```-webkit-appearance``` from Select component (Select.styles.js) as it conflicts with lint standards. Could be a problem for Safari browser.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-242-refractor-css

### Screenshots 📸
N/A

### References 🔗
Issue #98


packages/Checkbox/src/Checkbox.styles.js
```diff
@@ -75,7 +75,7 @@ const styles = {
   },
 };
 
-const checkboxStyles = css`
+export const checkboxStyles = styled.div`
```
mikrotron, line 78: Since this is no longer just the styles, it's an actual component, I'd recommend we change the naming convention (this comment applies to all the components in this PR).  In this case, probably `CheckboxWrapper` (Note the capital `C`), since it's not the _actual_ checkbox.

packages/Counter/src/Counter.styles.js
```diff
@@ -69,10 +69,17 @@ const counterStyles = css`
...
+    const hasIndicator = props.hasIndicator && indicatorStyles;
+    const color = counterColorStyles[props.color];
+    const size = counterSizeStyles[props.size];
+
+    return `
+      ${hasIndicator};
+      ${color};
+      ${size};`;
+  }}
+`;
```
mikrotron, line 85:
```suggestion
export const counterStyles = styled.span`
  ${commonStyles};
  ${props => `    
    ${props.hasIndicator && indicatorStyles}
    ${counterColorStyles[props.color]}
    ${counterSizeStyles[props.size]}
  `}
`;
```

mikrotron: Maybe that would be simpler? It assumes you remove `styled(commonStyles)` from line `74` too.

packages/FormElement/src/FormElement.styles.js
```diff
@@ -16,10 +16,20 @@ const inlineFormElementStyles = css`
...
+    const size = FontSizes[props.size];
+    const isInline = props.isInline && inlineFormElementStyles;
+    const isDisabled = props.isDisabled && `opacity: 0.5;`;
+
+    return `
+    ${size};
+    ${isInline};
+    ${isDisabled};`;
+  }}
+`;
```
mikrotron, line 35: Why the change? I liked it better before 😕

mikrotron:
This is a big chore, but thanks for taking it on! I didn't comment all the way through, but there are two recurring points I think should be addressed: 
1. The naming convention (the exports are now components, not styles, so they can be named like a component).
2. The `commonStyles` refactoring in several components has made the code longer and more indirect with no clear benefit that I can see.

packages/Checkbox/src/Checkbox.styles.js
KDarcey: Will make sure to adjust all of the names where needed

packages/FormElement/src/FormElement.styles.js
KDarcey: The primary reasoning was to maintain consistency with how I styled the other components, that being said I could understand why the previous version in this instance may be preferred as the code is certainly shorter. This is definitely something I could adjust if needed.

packages/Counter/src/Counter.styles.js
KDarcey: I'll want to adjust the other components to follow the same styling pattern, but it shall be done 👍

packages/Counter/src/Counter.styles.js
```diff
@@ -43,7 +43,7 @@ const counterColorStyles = {
   `,
 };
 
-const indicatorStyles = css`
+const indicatorStyles = `
```
mikrotron, line 46: I think it's a good idea to keep the `css` helper here, mainly because it will improve the syntax highlighting in our editors.

packages/Card/src/Card.js
```diff
@@ -38,9 +38,9 @@ function Card(props) {
   };
 
   return (
-    <sc.cardStyles data-pka-anchor="card" {...styleProps} {...moreProps}>
+    <sc.CardWrapper data-pka-anchor="card" {...styleProps} {...moreProps}>
```
mikrotron, line 41: I'm not a fan of the `Wrapper` suffix for naming in general – it's too vague to be helpful most of the time – but sometimes it really _is_ necessary.  But in this case I think simply `<sc.Card>` fits nicely, is a bit less typing, and a bit easier to read (will, at least when the component name also gets long).  

Bit of a hassle to make this adjustment throughout, but I think it's worth it to get it done right and then move on.  Thanks!

packages/ExternalLink/src/ExternalLink.styles.js
```diff
@@ -3,21 +3,23 @@ import tokens from "@paprika/tokens";
...
+  ${props => `
   align-items: center;
   border-radius: ${tokens.border.radius};
   color: ${tokens.textColor.link};
   display: inline-flex;
   max-width: 100%;
   position: relative;
 
-  ${({ iconFontSize }) => `padding: 1px ${iconFontSize + toInt(tokens.spaceLg)}px 1px ${tokens.spaceSm};`}
+  ${`padding: 1px ${props.iconFontSize + toInt(tokens.spaceLg)}px 1px ${tokens.spaceSm};`}
```
mikrotron, line 15: I think when the prop is _only_ being used in this one spot it makes more sense _not_ to wrap the whole styling block in a function, but keep it scoped to the one line, as it was.  

Gets to be a bit of a judgement call then about when there's only _two_ lines... or _three_ lines... when to wrap the whole block... what do you think?

packages/FormElement/src/FormElement.styles.js
```diff
@@ -8,18 +8,23 @@ const FontSizes = {
 };
 
 export const SectionsContainer = styled.div`
-  ${({ isInline }) => isInline && `flex-grow: 1;`}
+  ${props => `
```
mikrotron, line 11: I preferred the tidyness of the previous version where this could all be in one line.  I notice that you're consistently avoiding the destructuring of `props`, was there another reason besides consistency?  Was that approcah recommended by someone?

packages/Heading/src/Heading.styles.js
```diff
@@ -52,17 +40,28 @@ const underline = level => {
...
+  color: inherit;
+  font-weight: 700;
+  letter-spacing: 0;
+  ${stylers.lineHeight(-1)};
+  margin: ${stylers.spacer(2)} 0;
+
+  &:focus {
+    outline: none;
+  }
+  ${props => `
```
mikrotron, line 60: Heres' an example fo a block with a bunch of props being used where it totally makes sense to wrap the whole block once in a function, instead of for each line.  And could make it more readable _not_ to destructure props at the beginning, since you'd have a list of 4 or 5.  But I still am not crazy about the way this approach scales down... 🤔

packages/Input/src/Input.styles.js
```diff
@@ -21,122 +21,122 @@ const iconStyles = {
...
-      display: none;
-    }
-  }
+export const InputWrapper = styled.div`
+  ${props => `
+    line-height: 1;
+    position: relative;
 
-  /* Sizes */
+    input.form-input__input {
```
mikrotron, line 29: You don't need to do it in this PR, it's a bit out of scope, but thought I'd call it out that this "className" approach is a legacy left over from migrating the component from `acl-ui`, and one we'd like to avoid going forward by simply exporting another styled component for this element and applying it in the component.

packages/Input/src/Input.styles.js
```diff
@@ -21,122 +21,122 @@ const iconStyles = {
...
-      box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
-      outline: none;
-    }
-
-    &::-ms-clear {
-      display: none;
-    }
-  }
+export const InputWrapper = styled.div`
+  ${props => `
```
mikrotron, line 25: Oh, this is probably not the only place, but I think there should be a `css` helper here too to get the IDE syntax highlighting to work properly.

packages/ListBox/src/components/Box/Box.styles.js
```diff
@@ -7,20 +7,16 @@ const isInlineCSS = `
...
-    const widthCSS = props.triggerWidth ? `width: ${props.triggerWidth}px;` : "";
-    return `
-    ${widthCSS}
-    ${inlineCSS}
-    `;
-  }}
+  ${props => `
+    ${props.isInline ? isInlineCSS : ""};
+    ${props.triggerWidth ? `width: ${props.triggerWidth}px;` : ""};
+  `}
```
mikrotron, line 21: 👍

packages/ListBox/src/components/Label/Label.styles.js
```diff
@@ -14,6 +14,8 @@ export const labelStyles = `
   width: 100%;
 `;
 
-export const Label = styled.span`
-  ${props => (props.isPlaceholder && !props.hasImplicitAll ? placeholderStyles : "")}
+export const LabelWrapper = styled.span`
+  ${props => `
```
mikrotron, line 18: The template string is unnecessary here.

packages/ListBox/src/components/Popover/Popover.js
```diff
@@ -21,9 +21,9 @@ export default function Popover(props) {
   const { children, zIndex } = props;
 
   return (
-    <PopoverStyled {...props} isOpen={state.isOpen} maxWidth={state.triggerWidth} offset={0} zIndex={zIndex}>
+    <sc.PopoverStyled {...props} isOpen={state.isOpen} maxWidth={state.triggerWidth} offset={0} zIndex={zIndex}>
```
mikrotron, line 24: `<sc.PopoverStyled>` seems redundant to me. `<sc.Popover>` tells me the same thing with fewer words.

packages/Pill/src/Pill.styles.js
```diff
@@ -1,101 +1,126 @@
-import { css } from "styled-components";
+import styled from "styled-components";
 import stylers from "@paprika/stylers";
 import tokens from "@paprika/tokens";
+import RawButton from "@paprika/raw-button";
 
 const pillColorStyles = {
-  black: css`
+  black: `
```
mikrotron, line 7: Probably not a big deal in these cases, you could leave it as is, but the `css` helper was there for that same syntax highlighting reason.

packages/Switch/src/Switch.styles.js
```diff
@@ -8,36 +8,33 @@ const KnobPositionStyles = {
     transform: translateX(12px);
   `,
   medium: css`
-    transform: translate(16px, 0);
+    transform: translate(20px, 0);
   `,
   large: css`
-    transform: translate(20px);
+    transform: translate(28px);
```
mikrotron, line 14: Not sure why these changed... maybe a conflict resolution with integration after you made styling changes to the Switch? Be careful not to undo your previous changes 😬

packages/Switch/src/Switch.js
```diff
@@ -38,8 +37,7 @@ function Switch(props) {
   const { a11yText, canPropagate, isChecked, isDisabled, size, onChange, ...moreProps } = props;
 
   return (
-    <RawButton
-      css={switchStyles}
+    <sc.switchStyles
```
mikrotron, line 40: I think you missed this one... should be `<sc.switchWrapper>` – or, as I'd recommend, just `<sc.Switch>`.

packages/Tabs/src/components/List/List.js
```diff
@@ -37,9 +37,9 @@ export default function List(props) {
   if (a11yText) moreProps["aria-label"] = a11yText;
 
   return (
-    <div {...moreProps} css={tabsListStyles} role="tablist" ref={ref => setTabListRef(ref)} data-pka-anchor="tabs.list">
+    <sc.tabsListStyles {...moreProps} role="tablist" ref={ref => setTabListRef(ref)} data-pka-anchor="tabs.list">
```
mikrotron, line 40: `<sc.TabsList>`

packages/Tabs/src/components/Panel/Panel.js
```diff
@@ -23,9 +23,9 @@ const Panel = props => {
   }
 
   return (
-    <div {...moreProps} css={panelStyles} role="tabpanel" data-pka-anchor="tabs.panel">
+    <sc.panelStyles {...moreProps} role="tabpanel" data-pka-anchor="tabs.panel">
```
mikrotron, line 26: `<sc.Panel>`

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -44,26 +43,24 @@ export default function Tab(props) {
 
   if (href && !isDisabled) {
     return (
-      <a
+      <sc.linkStyles
```
mikrotron, line 46: `<sc.TabLink>`

packages/Tabs/src/components/Tab/Tab.js
```diff
@@ -44,26 +43,24 @@ export default function Tab(props) {
...
       >
         {children}
-      </a>
+      </sc.linkStyles>
     );
   }
 
   return (
-    <RawButton
+    <sc.tabStyles
```
mikrotron, line 60: `<sc.Tab>`

packages/Toast/src/Toast.styles.js
```diff
@@ -29,7 +29,7 @@ const closeButtonColors = {
   [Kinds.LOCKED]: tokens.color.yellowLighten30,
 };
 
-const fixedStyles = css`
+const fixedStyles = `
```
mikrotron, line 32: Keep the `css`

mikrotron:
Okay, I went through the _whole_ PR this time, and again a few recurring themes:
1. I think you should avoid the `Wrapper` suffix pretty much everywhere. The `sc.` prefix serves a similar purpose of identifying that this is a styled-components wrapper.
2. I think after reading through and thinking about it, we should avoid wrapping the _whole_ styling block in a `props => `...`` function block and instead only wrap the part that relies on `props`.  When that's only one line, probably best to just do it inline as it was before, but if it's two or more, maybe it's a good time to then use a single function wrapper for both lines (or more).
3. Please try to use the `css` helper (before styling blocks wrapped in a template literal).  This helps the IDE recognize that this can be formatted like CSS.  (Not necessary when you use `styled.div` or `styled(SomeComponent)` though.

I think by now you have realized that this was a bit of a trap and that there are hazards to using JavaScript two write CSS – there are _too many ways_ to do the same thing, and establishing conventions and best practices is hard 😅 

But I appreciate your work on this big PR, and maybe you can discuss the conventions that have emerged and any lingering ones that we are still debating on Tuesday during our next UX Dev ShopTalk meeting. Thanks!  🙏

packages/ExternalLink/src/ExternalLink.styles.js
alexzherdev: Does line 7 also need `css` for syntax highlighting?

I'm ambivalent but just sharing some links. Overall this approach seems similar to this, correct? https://twitter.com/NikkitaFTW/status/1262423045874089990
And FWIW I found another one in the comments: https://twitter.com/Saeris/status/1262441528078688256

mikrotron: I am interested in your ambivalent feelings... or do you really mean _indifferent_?

I actually really like the suggestion from @saeris... so this would look more like:

```jsx
export const ExternalLink = styled.a(props) => css`
  ...
`;
```

or 

```jsx
export const ExternalLink = styled.a({ someProp, someOtherProp }) => css`
  ...
`;
```
```

Also, yes, line 7 _should_ include the `css` helper... I didn't comment on every instance throughout the PR, it comes up for nearly every component.

nahumzs:
This PR will be split in multiple PRs.

PR #551, opened 2020-06-04
L10N-2020-06-04-16609 - translations batch

alexandermccate:


packages/L10n/src/locales/es.yml
```diff
@@ -7,25 +7,32 @@ es:
...
+        processing: En proceso
+        uploading: Cargando
+        success: Éxito
+        cancel: Cancelar
+        error: Error
+      progress:
+        cancelled: Cancelado/a
+        complete: Completado/a
+        idle: Ocioso
+        uploading: Carga de {{progressWithUnits}} de {{sizeWithUnits}}
```
agherreras, line 29: Source says: "Uploading {{progressWithUnits}} of {{sizeWithUnits}}"
I would use something different. For example "Cargando {{progressWithUnits}} de {{sizeWithUnits}}"
An example: "Cargando 1 de 5"
What do you think, @cristinajimenez?

cristinajimenez: Totally agree.

jamiek-acl: @agherreras and @cristinajimenez not sure if it is relevant, but in English it will say something like: `Uploading 143kb of 2.5mb` (not `Uploading 1 of 5`)

cristinajimenez: Thanks @jamiek-acl for explaning. In that case I would use 
`Cargados {{progressWithUnits}} de {{sizeWithUnits}}`
Agree @agherreras ?

agherreras: Yes.
In this case we might also want to update other languages?
Portuguese also uses the gerund form of the verb.

cristinajimenez: Maybe, but this is a stylistic choice, I don't think the gerund would be incorrect either. Could be fixed next batch if needed.

alexandermccate: Cool. Thanks for all the feedback! Will double check.

alexandermccate: Let's use Carregando (pt) and Cargando (es) for consistency with previous stylistic choice. Is it necessary to change it to "Cargados"?

agherreras: I like more "Cargados", since it indicates that amount of data has already being uploaded, which I think is accurate, but since source says "Uploading" instead of "Uploaded", I guess we can go with "Cargando"...

cristinajimenez: I agree with Alberto, I would use "Cargados", it describes better what the action will do.

alexandermccate: Yes, let's not overdo it :-)

alexandermccate: "Uploading" is used in source rather than "Uploaded", since the action is in progress. Can we keep it simple and stick to the source, translating it as Carregando (pt) and Cargando (es)?

image

cristinajimenez: Sure

alexandermccate:
Fixes applied and checking with translators regarding `Uploading 143kb of 2.5mb`.

PR #554, opened 2020-06-07
STAR-250 New ResizeObserver component 🐐

mikrotron:
### Purpose 🚀
`<ResizeDetector>` is a new provider component that implements a simple DOM wrapper element and provides callbacks for resize events and custom hooks with dimension values for consumers.

### Notes ✏️
- `onResize` is a debounced callback that will be fired when the wrapper element's height/width is changed.
- `useObservedDimensions` hook will provide the updated width/height to a consumer.
- `onBreak` is also a debounced callback that will be fired with the wrapper's `width` crosses one of two thresholds: `breakpointSmall` or `breakpointLarge`.
- `useBreakpoints` hook will provide the updated `size` (`small` / `medium` / `large`) of the wrapper.
- Sizes + breakpoints are fixed to a limited scope for simplicity:
  `small` : `0px` – `breakpointSmall` |  `medium` : `breakpointSmall` – `breakpointLarge`  | `large` : `breakpointLarge` – `∞`
- Breakpoints can be passed as `props` and `null` values are allowed.
- Wrapper DOM element is `inline-block`, but be default behaves like a `block` with `width: 100%`.  This can be adjusted with `isFullWidth={false}` or `isFullHeight={true}`.
- BONUS: Fixed the `hasNoUnderline` prop on `<ExternalLink>` ([fourth example is broken](https://paprika.highbond.com/?path=/story/buttons-externallink-examples--layout-examples))

### Storybook Documentation 🤓
- Also added a `<Resizer>` utility component for use in Storybook stories.
- Created an `<ExampleStory>` helper for this component, but could be extracted as a general Storybook helper.
- Added Example stories in an Examples folder, but also used in a rich MDX `Docs` page.
- Added a `Showcase` story with `knobs`, `actions`, and a `<CodeViewer>`.

### To Do 📋
- Cypress Tests (in a follow-up task)

### Updates 📦
- [x] NEW package `@paprika/resize-detector`
- [x] PATCH (bug fix) change to `@paprika/external-link`,

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-250--new-resize-observer/?path=/story/utilities-resizedetector-docs--page

### Screenshots 📸
In Action!
resizeobserver

Showcase Story (knobs)
Screen Shot 2020-07-15 at 5 21 44 PM

Example Story
Screen Shot 2020-07-15 at 5 21 22 PM

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-250


packages/ResizeObserver/src/ResizeObserver.js
```diff
@@ -0,0 +1,95 @@
...
+  isFullHeight: PropTypes.bool,
+  onBreak: PropTypes.func,
+  onResize: PropTypes.func,
+};
+
+const defaultProps = {
+  breakpointLarge: 768,
+  breakpointSmall: 360,
+  children: null,
+  debounceDelay: 200,
```
mikrotron, line 23: Not sure about a default value for this?

packages/ResizeObserver/src/ResizeObserver.js
```diff
@@ -0,0 +1,95 @@
...
+      <ResizeContext.Provider value={{ width, height, size }}>{children}</ResizeContext.Provider>
+    </sc.ResizeObserver>
+  );
+}
+
+ResizeObserver.displayName = "ResizeObserver";
+ResizeObserver.propTypes = propTypes;
+ResizeObserver.defaultProps = defaultProps;
+ResizeObserver.useObservedDimensions = useObservedDimensions;
+ResizeObserver.useBreakpoints = useBreakpoints;
```
mikrotron, line 93: Not sure about this convention of attaching the custom hooks I want to export right to the component... I like it, but can't think of an example where I've seen it before.

oscarkwan: We use this pattern usually i believe.

`import { useColumnsArrangement } from '@paprika/action-bar'`

Any reason for you to not do it this way instead?

mikrotron: Well, it means my import in the consumer is a little neater, I can just do 
`import ResizeObserver  from '@paprika/action-bar';`  

And I can export from `ResizeObserver.js` instead of from `index.js` (which I don't usually look at).  

Neither are deal-breakers, but since both approaches work, I kind of prefer the way I'm proposing here.

packages/ResizeObserver/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@paprika/resize-observer",
```
alexzherdev, line 2: I wouldn't reuse an existing well-known name here. `ResizeObserver` API exists in modern browsers _and_ it has no visual user-facing component. `Resizer` works well enough?

packages/ResizeObserver/src/ResizeObserver.js
alexzherdev: I feel like we could go lower by default.

packages/ResizeObserver/src/ResizeObserver.js
```diff
@@ -0,0 +1,95 @@
+import React from "react";
+import PropTypes from "prop-types";
+import useResizeObserver from "use-resize-observer/polyfilled";
+import debounce from "lodash.debounce";
+import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
+import * as sc from "./ResizeObserver.styles";
+
+const propTypes = {
+  breakpointLarge: PropTypes.number,
```
alexzherdev, line 9: Is it obvious enough that these values refer to widths?

packages/ResizeObserver/src/ResizeObserver.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function useBreakpoints() {
+  const { size } = React.useContext(ResizeContext);
+  return { size };
+}
+
+function getSize(width, breakpointSmall, breakpointLarge) {
+  let size = ShirtSizes.MEDIUM;
+  if (breakpointSmall && width <= breakpointSmall) {
```
alexzherdev, line 46: The bounds comparisons seem inconsistent. E.g. for small = 100, large = 200:
```
0 to 100 is small
101 to 199 is medium
200+ is large.
```
I'd expect:
```
0 to 99 small
100 to 199 medium
200+ large
```

packages/ResizeObserver/src/ResizeObserver.js
```diff
@@ -0,0 +1,95 @@
...
+  useResizeObserver({
+    ref: refContainer,
+    onResize: debounce(handleResize, debounceDelay, { maxWait: MAX_WAIT }),
+  });
+
+  React.useLayoutEffect(() => {
+    const { width, height } = refContainer.current.getBoundingClientRect();
+    setDimensions({ width, height });
+    setSize(getSize(width, breakpointSmall, breakpointLarge));
+  }, [children, breakpointSmall, breakpointLarge]);
```
alexzherdev, line 80: 🤨 what are you trying to achieve by including children here?

packages/ResizeObserver/src/ResizeObserver.js
mikrotron: `30` seems barely perceptible.  We could go with that be default and if there's too much of a performance hit with heavy components, consumers would be responsible to increase the delay. Sound more reasonable?

.storybook/components/Resizer/Resizer.story.js
```diff
@@ -0,0 +1,44 @@
...
+  width: 100%;
+`;
+
+storiesOf("Storybook | Resizer", module).add("Demo", () => (
+  <Story>
+    <Heading level={1} displayLevel={2} isLight>
+      Resizer
+    </Heading>
+    <Tagline>
+      <big>Demo</big>
```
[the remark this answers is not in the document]
mikrotron: Yeah, okay, okay. It's just Storybook, but it is deprecated...

packages/ResizeObserver/package.json
mikrotron: Okay, fair point.  I already use `<Resizer>` for the Storybook helper, and that name suggests that the component _does_ something besides just reporting... I also considered `ResizeListener` and `ResizeSensor`.  Any other suggetions?

packages/ResizeObserver/src/ResizeObserver.js
mikrotron: Hmm, I think some explanation will be required to know how the breakpoints work no matter what.  But we might want height breakpoints in the future, and then might need to introduce a breaking change to the API for distinguishing – although I think there's a good chance we will never need that either.  `width` is a far more common use case, and I'd like to keep it simple where possible.  `breakpointLargeWidth` and `breakpointSmallWidth` are kinda ugly, but they ARE more clear... so I am conflicted about this 🤔

packages/ResizeObserver/src/ResizeObserver.js
mikrotron: Yeah, I was thinking that since the `small` breakpoint breaks _to_ small, it should break to the small size at that point.  This is already a bit inconsistent with the way conventional media query breakpoints work, but I thought a `medium` breakpoint would be more ambiguous. But in this case it's probably still reasonable for people to have that mental model and, like you, expect the "break" to the next size up to happen at the breakpoint, so I will adjust it.

packages/ResizeObserver/src/ResizeObserver.js
mikrotron: Well, if the `children` change, then it should reset the size... but now I realize that since the component would already be mounted, the useResizeObserver would already be working and the setSize is going to be called by it already.

packages/ResizeObserver/package.json
alexzherdev: `ResizeDetector` sounds good to me 👍

packages/ResizeObserver/src/ResizeObserver.js
nahumzs: React suggest that custom hooks should always start with use, I guess adding the `ResizeObverser` kind of break this principle, the reason they suggested is because helps static analyzers to do a proper job on the hooks. However, In this case, I know there is nothing to analyze but somehow is not a regular way of exporting hooks.

That's why I end up exporting the hook for the Action bar as a property of the import instead of adding it directly to the ActionBar as you are proposing. 

I think while is not more beautiful, is more "standard" <- whatever that means 

> Custom Hooks are more of a convention than a feature. If a function’s name starts with ”use” and it calls other Hooks, we say it is a custom Hook. The useSomething naming convention is how our linter plugin is able to find bugs in the code using Hooks.
https://reactjs.org/docs/hooks-overview.html

packages/ResizeDetector/stories/storyHelpers.js
```diff
@@ -0,0 +1,79 @@
...
+            Source
+          </ExternalLink>
+        )}
+      </HeaderFlex>
+      {tagline && <Tagline>{tagline}</Tagline>}
+      <Rule />
+      {children}
+    </Story>
+  );
+};
```
mikrotron, line 40: If everyone thinks this is a good pattern, we could extract it to `/.storybook/components` and use it everywhere.

packages/ExternalLink/src/ExternalLink.js
```diff
@@ -32,7 +32,7 @@ function ExternalLink(props) {
 
   return (
     <sc.ExternalLink
-      aria-label={ariaText || children}
+      aria-label={ariaText || null}
```
mikrotron, line 35: This would not have worked if `children` was a node instead of a string.  But it's also unnecessary, since the content of the `<a>` will be used as the accessible name by AT automatically if there's no `aria-label`.

packages/ExternalLink/src/ExternalLink.styles.js
```diff
@@ -1,17 +1,30 @@
...
+`;
+
 export const ExternalLink = styled.a`
   align-items: center;
   border-radius: ${tokens.border.radius};
   color: ${tokens.textColor.link};
   display: inline-flex;
   max-width: 100%;
   position: relative;
+  text-decoration: none;
```
mikrotron, line 22: Maybe it was only Storybook, but this is needed to actually remove the underline when `hasNoUnderline={true}`.

packages/ExternalLink/stories/ExternalLink.stories.js
```diff
@@ -10,4 +10,4 @@ storiesOf(storyName, module)
   .addDecorator(withKnobs)
   .add("Showcase", Showcase);
 
-storiesOf(`${storyName}/Examples`, module).add("Layout Examples", LayoutExamples);
+storiesOf(`${storyName}/Examples`, module).add("Screener Examples", LayoutExamples);
```
mikrotron, line 13: This is a quick, lazy way to start covering this story with Screener for now.

packages/ExternalLink/stories/examples/Showcase.js
```diff
@@ -9,7 +9,7 @@ const externalLinkProps = () => ({
   ariaText: text("ariaText", ""),
   hasNoUnderline: boolean("hasNoUnderline", false),
   href: text("href", "http://www.wegalvanize.com"),
-  width: number("Inline width px value", "120"),
+  maxWidth: number("Inline maxWidth", "120"),
```
mikrotron, line 12: `maxWidth` makes a better example than `width`, otherwise you can end up with a bit empty gap between the (short) hyperlink label and the external link icon.

packages/ResizeDetector/README.md
```diff
@@ -0,0 +1,93 @@
+<!-- autogenerated don't modify -->
+
+# @paprika/resize-observer - 0.0.1
```
mikrotron, line 3: When we publish we should remember to apply a `MINOR` bump, so we initially publish the `0.1.0` version.

cc @nahumzs

packages/ResizeDetector/stories/storyHelpers.js
nahumzs: this is good, but I guess we need a page where we can see what component are available for building our stories, right now only a few of us know about this component, and it's easy to forget about them.  (of course for another pr or for future initiatives)

packages/ExternalLink/src/ExternalLink.js
nahumzs: no for this, but the way I did it on the DataGrid its that if the user passes it down a string, its use automatically as a a11ytext, otherwise if it's a node or function I fire a `console.warn` asking for the developer to provide an `allyLabel` prop to describe the content on the node.

packages/ResizeDetector/README.md
nahumzs: you can put it on 0.1.0 I always remember whenever its a new component to select the option on lerna as custom version and I input `0.1.0` and since there its not any version with that number neither on github or npm will use it.

.storybook/components/CodeViewer/CodeViewer.js
```diff
@@ -1,4 +1,5 @@
 import React from "react";
```
nahumzs, line 1: this could be a small packages 😅 `@paprika/syntax-highlight` hehe

packages/ResizeDetector/stories/storyHelpers.js
mikrotron: There's a section now in the Storybook Navigation for ... `Storybook` components.  Only `<Codeviewer>` and `<ResizeDetector>` there so far, but that's where I think we could show off and document these types of components.

.storybook/components/Resizer/Resizer.js
```diff
@@ -0,0 +1,32 @@
...
+const propTypes = {
+  children: PropTypes.node,
+  initHeight: PropTypes.number,
+  initWidth: PropTypes.number,
+  onResizeStop: PropTypes.func,
+};
+
+const defaultProps = {
+  children: null,
+  initHeight: 360,
```
nahumzs, line 14: for another iteration, seems like would be beneficial to have the viewport size instead of random values ... :)

packages/ExternalLink/src/ExternalLink.js
mikrotron: Yeah, I think that's a good pattern, and I've seen it in the Takeover too:
https://github.com/acl-services/paprika/blob/master/packages/Takeover/src/Takeover.js#L75-L85

mikrotron: Maybe this would be useful to have in the `helpers` package?

PR #563, opened 2020-06-15
SNP-354 Adding missing attributes in listbox initial state

allison-c:
### Purpose 🚀
Adding missing attributes in listbox initial state

### Notes ✏️

`<ListBox>` initial state is missing some values. When the `originalState` is missing `lastKnownSelectedOptions`, will cause an error in reducer while calling `selectedOptions: state.lastKnownSelectedOptions.slice(0)`

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/SNP-354-fix-listbox

### Screenshots 📸

### References 🔗


packages/ListBox/src/store/Provider.js
```diff
@@ -66,12 +66,14 @@ function initializeState(props) {
     formElementId,
     formElementLabelDescribedBy,
     size,
+    hasFilter: false,
+    lastKnownSelectedOptions: selectedOptions.slice(0),
+    onChangeFn: null,
```
jamiek-acl, line 71: Can you please add these alphabetically? The others are (with the exception of `formElementId` and `formElementLabelDescribedBy`... maybe fix up those two too please)

allison-c: 👍 done!

PR #566, opened 2020-06-17
Star 248 datagrid performance

nahumzs:
### Purpose 🚀

### Notes ✏️

### Updates 📦
- [x] MINOR (backward compatible) change to Datagrid (this is a break change)

### Storybook 📕
this branch has the same performance implementation:
http://storybooks.highbond-s3.com/paprika/STAR-241-borders-scrollbars/iframe.html?id=data-table-datagrid-examples--stress-test&viewMode=story

### Screenshots 📸

Tested manually a11y voice over still working correctly 

^ 🦀  lol, the gif didn't capture the voice over, but it's working, tested manually

### References 🔗


packages/DataGrid/src/DataGrid.Rows.js
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable react/prop-types */
```
nahumzs, line 1: I extracted all the four grid's rows from the DataGrid: `HeaderSticky, Header, StickyGrid and MainGrid` into this file **DataGrid.Rows.js**.

The idea is to apply React.memo as it is suggested on React-window library to improve the performance of the DataGrid [info](https://react-window.now.sh/#/examples/list/memoized-list-items); also the use of `memonize-one` is required to memonize the data passdown to the React.memo components.

packages/DataGrid/src/DataGrid.js
```diff
@@ -105,19 +105,18 @@ const DataGrid = React.forwardRef((props, ref) => {
...
-  const overscanColumnCount = 2;
+  const overscanRowCount = 7;
+  const overscanColumnCount = 5;
 
   const rowCount = data.length;
   const {
     "DataGrid.ColumnDefinition": extractedColumnDefinitions,
     "DataGrid.Basement": Basement,
-    "DataGrid.InfiniteScroll": InfiniteScroll,
+    "DataGrid.InfiniteScroll": extractedInfiniteScroll,
```
nahumzs, line 119: rename infiniteScro for extractedInfintiveScroll to no be confused with the import of the `Infinitive` component.

packages/DataGrid/src/DataGrid.js
```diff
@@ -244,33 +242,36 @@ const DataGrid = React.forwardRef((props, ref) => {
...
-  const handleScroll = React.useCallback(parameters => {
-    const { scrollLeft, scrollTop } = parameters;
+  const handleScroll = React.useCallback(
+    parameters => {
+      const { scrollLeft, scrollTop } = parameters;
 
-    if (refScrollHeader.current) {
-      refScrollHeader.current.scrollTo({ left: scrollLeft, top: 0 });
-    }
+      if (refScrollHeader.current && stickyColumnsIndexes.length) {
```
nahumzs, line 249: if there it is not sticky columns we can omit this run.

packages/DataGrid/src/DataGrid.js
```diff
@@ -320,12 +311,18 @@ const DataGrid = React.forwardRef((props, ref) => {
   function handleFocusGrid() {
     const $isBlurred = refContainer.current.querySelector(".grid--is-blurred");
     if ($isBlurred) $isBlurred.classList.remove("grid--is-blurred");
-
-    if (gridShouldHaveFocus) {
-      setGridShouldHaveFocus(false);
-    }
   }
 
+  const itemData = createItemData(
```
nahumzs, line 316: Creates a memonize `itemData` prop which will be passdown to each of the react-window grids once the React.memo row needs to be consumed `DataGrid.rows.js`

packages/DataGrid/src/DataGrid.js
```diff
@@ -376,7 +373,7 @@ const DataGrid = React.forwardRef((props, ref) => {
     if ($isActive) $isActive.classList.toggle("grid--is-blurred");
   }, []);
 
-  const handleMouseUpGrid = React.useCallback(
+  const handleMouseDownGrid = React.useCallback(
```
nahumzs, line 376: this adds faster feedback to the user instead of waiting to mouse down and mouse up (a regular click), once you mouse down this will be executed. 

This might be a breaking change on some features using mouseup to test their features, please replace mouseup with `mousedown` from now on. cc: @oscarkwan and @mikrotron

packages/DataGrid/src/DataGrid.js
```diff
@@ -500,12 +497,12 @@ const DataGrid = React.forwardRef((props, ref) => {
...
         onKeyDown={handleKeyDownGrid}
         onKeyUp={handleKeyUpGrid}
-        onMouseUp={handleMouseUpGrid}
+        onMouseDown={handleMouseDownGrid}
         onMouseOver={handleMouseOver}
         onMouseLeave={handleMouseLeave}
         ref={refContainer}
         role="grid"
-        tabIndex={gridShouldHaveFocus ? 0 : -1}
+        tabIndex={0}
```
nahumzs, line 505: I remove gridShouldHaveFocus because it was preventing from clickling the first time the Datagrid, also I think there should be a better discussion about how to handle the `tab` process if the Datagrid it's part of a form. `a11y` discussion. cc @mikrotron 

We can improve and bring back this feature later.

packages/DataGrid/src/DataGrid.js
alexzherdev: I can't find the reference, but I feel like I read somewhere that firing an action on mouse down is not accessible 🤔

mikrotron: Correct.  Although it's a 2.1 criteria...
https://www.w3.org/WAI/WCAG21/quickref/#pointer-cancellation

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
```diff
@@ -1,79 +1,64 @@
...
-      onMouseLeave={handleMouseLeave}
-    >
-      {isActive || isActiveCell || isActiveRow || isValueNotUnchecked(isChecked({ rowIndex })) || !hasIndexIndicator ? (
+    <sc.RowContainer>
+      {hasNumber && (isActiveCell || isActiveRow) ? (
+        indicator
+      ) : (
         <sc.Checkbox>
-          <CheckBox tabIndex="-1" size="small" checkedState={isChecked({ rowIndex })} onChange={handleCheck} />
+          <input type="checkbox" checked={isChecked} onChange={() => {}} onMouseDown={handleCheck} />
```
nahumzs, line 46: So this is an interesting one, our checkbox requires having an onChange method, but for this case, it's irrelevant since the user can check the checkbox either, with  `enter, space or focusing on the input and click on it directly. 

Also, the state is controlled by the consumer so is not really a need to know the state of the input only setting if its checked or not. 

Later I can try to replace it with our checkbox, but not part of this PR

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
```diff
@@ -1,79 +1,64 @@
...
         </sc.Checkbox>
-      ) : (
-        indicator
       )}
     </sc.RowContainer>
   );
 }
 
+function areEqual(prev, next) {
```
nahumzs, line 53: require it to prevent the checkbox rerender each time that the datagrid append or remove new rows.

packages/DataGrid/stories/DataGrid.collapse.stories.js
```diff
@@ -152,7 +154,7 @@ export function App() {
           }}
           cellA11yText={({ row }) => row.objective}
           cellProps={({ row }) => {
-            return { style: { textIndent: `${row.indent * 16}px`, cursor: "pointer" } };
+            return { style: { textIndent: `${(row.indent || 0) * 16}px`, cursor: "pointer" } };
```
nahumzs, line 157: fixing a small bug, don't know how was working and its working on s3

🤷

packages/DataGrid/src/DataGrid.js
nahumzs: what does it mean is 2.1 criteria do apply in this context?

packages/DataGrid/src/DataGrid.js
```diff
@@ -289,16 +290,6 @@ const DataGrid = React.forwardRef((props, ref) => {
     [refScrollGrid]
   );
 
-  const handleRefCell = React.useCallback(({ columnIndex, rowIndex }) => {
```
nahumzs, line 292: `handleRefCell` was used to have access to the cell imperative API, but was really heavy, I think there can be a better alternative to change just one cell or a number of cell without having to store the ref or each cell. Maybe recoil can work better for this specific case.

Again this is a feature no currently use by any DataGrid, so is fine to remove it and thinking about again.

packages/DataGrid/src/components/Cell/Cell.js
```diff
@@ -21,36 +21,13 @@ const propTypes = {
   hasZebraStripes: PropTypes.bool.isRequired,
 };
 
-const Cell = React.forwardRef((props, ref) => {
+export default function Cell(props) {
   const { style, gridId, columnIndex, rowIndex, column, data, a11yText, hasZebraStripes } = props;
-  const [isActiveCell, setIsActiveCell] = React.useState(false);
-  const [isActiveRow, setIsRowActive] = React.useState(false);
-
-  React.useImperativeHandle(ref, () => ({
```
nahumzs, line 29: adding this imperative API for the cell so would be possible to do row hover or mouseover/mouseout event on the table turn to be heavy, so I decided to remove it, I'm not aware of any DataGrid using these features, so as I commented previously I think a better option for this thing of interaction is to look into [recoiljs](https://recoiljs.org/).

packages/DataGrid/src/DataGrid.js
mikrotron: WCAG 2.1 is the newest standard from W3C from 2018.  Most legislation still references WCAG 2.0 (level AA) from 2008 though, and we are specifically targeting that standard for HighBond, even though at some point we know the legislation (and requirements demanded by customers) will be updated to WCAG 2.1.  So we have some leeway here, but probably _best_ to be compliant with WCAG 2.1 when we can (especially for Paprika, and other shared components, like the `<GlobalNav>`).

packages/DataGrid/src/DataGrid.Rows.js
```diff
@@ -0,0 +1,124 @@
...
+  const { style: styleProps = {}, ...moreProps } = typeof headerProps === "function" ? headerProps({ header }) : {};
+
+  return (
+    <CellHeader role="columnheader" style={{ ...style, ...styleProps }} {...moreProps}>
+      {typeof header === "function" ? header() : header}
+    </CellHeader>
+  );
+}, areEqual);
+
+export const StickyHeaderRow = React.memo(({ data: RowData, columnIndex, style }) => {
```
AndreyChernykh, line 113: I wonder if this memoization would work correctly. `RowData` is an object, for instance. Will it keep have the same reference in between re-renders?

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
AndreyChernykh: What if the grid cells would contain another interactive elements, such as dropdowns? (it is a made up example; just curious)

packages/DataGrid/src/components/columns/renderColumnIndicator.js
```diff
@@ -4,30 +4,29 @@ import ColumnDefinition from "../ColumnDefinition";
...
-            isChecked={() => isAllChecked}
-            {...propsHeader}
-          />
-        ) : null;
+      header={() => {
+        return hasHeaderChecker ? <RowIndicator onCheck={onCheck} isChecked={isAllChecked} isHeaderChecker /> : null;
       }}
-      cell={propsCell => <RowIndicator hasNumber={hasNumber} isChecked={isChecked} onCheck={onCheck} {...propsCell} />}
+      cell={propsCell => (
+        <RowIndicator
```
AndreyChernykh, line 23: What is a row indicator? What is the purpose of this component.

packages/DataGrid/stories/DataGrid.collapse.stories.js
```diff
@@ -59,7 +59,9 @@ const insert = (arr, index, newItems) => [
   ...arr.slice(index),
 ];
 
-const remove = ({ data, start, end }) => [...data.slice(0, start), ...data.slice(end + 1, data.length)];
+const remove = ({ data, start, end }) => {
+  return [...data.slice(0, start), ...data.slice(end + 1, data.length)];
+};
```
AndreyChernykh, line 64: I'm not 100% sure what is the purpose of this method. But I suspect that the same effect can be done with `splice` instead; and it will be more efficient (you wouldn't have to create multiple copies of an array) 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

```js
const remove = ({ data, start, end }) => {
   const deleteCount = end - start;
   return data.splice(start, deleteCount);
});
```

Probably, you don't really need `remove` at all, and can just use `splice` directly without wrapping it.

packages/DataGrid/src/components/RowIndicator/RowIndicator.js
nahumzs: I can add an example tomorrow about this.

PR #568, opened 2020-06-18
STAR-255-Switch-component-sizes

KDarcey:
### Purpose 🚀
To adjust the sizes of the Paprika Switch component so that it is consistent with the UI Kit.

### Notes ✏️
N/A

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-255-Switch-component-sizes

### Screenshots 📸
Before:
image

After:
image

### References 🔗
N/A

### DIFF
--- a/packages/Switch/src/Switch.styles.js
+++ b/packages/Switch/src/Switch.styles.js
@@ -8,10 +8,10 @@ const KnobPositionStyles = {
     transform: translateX(12px);
   `,
   medium: css`
-    transform: translate(20px, 0);
+    transform: translate(16px, 0);
   `,
   large: css`
-    transform: translate(28px);
+    transform: translate(20px);
   `,
 };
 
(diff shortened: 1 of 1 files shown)


jamiek-acl:
hey @KaanDarcey please note in the details of this PR you had this marked as a `major - breaking change`, but i changed it to `patch - bug fix`. we use these to determine versions of the releases.

packages/Switch/src/Switch.styles.js
```diff
@@ -8,10 +8,10 @@ const KnobPositionStyles = {
     transform: translateX(12px);
   `,
   medium: css`
-    transform: translate(20px, 0);
+    transform: translate(16px, 0);
```
jamiek-acl, line 11: Normally we use `$space` token, or `stylers.spacer` helper... but these seem pretty arbitrary so i think how you did it is fine.

KDarcey: I could change that. Should i replace all of the width and height values with spacers as opposed to px as well?

jamiek-acl: I dont think so, as some would be like 2 spacer + 1/2 spacer

PR #569, opened 2020-06-18
star-257 Create enum helper

tristanjasper:
### Purpose 🚀
Consolidate all packages enums to one central helper location

This is an early wip pr so comments on this approach welcome!

### Notes ✏️

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗
Parent ticket: https://aclgrc.atlassian.net/browse/STAR-256


packages/Button/src/Button.js
```diff
@@ -4,12 +4,10 @@ import "what-input";
 import RawButton from "@paprika/raw-button";
 import RefreshIcon from "@paprika/icon/lib/Refresh";
 import DownIcon from "@paprika/icon/lib/CaretDown";
-import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
```
nahumzs, line 7: Two things on top of my mind:
1.- enums / types / constants should be their own package.
2.- ShirtSize and ButtonKinds are not descriptive and when use it on prototypes documentation react docgen sometimes is not capable to list exactly what to write on it, also are not memorable neither easier to remember in specific situations.

My proposal:

As a user I would like to see something like this:
```js
import { button } from "@paprika/constants";

<Button kind={button.kind.MEDIUM} />
```

```js
import { datagrid } from "@paprika/constants";

<DataGrid borderTypes={datagrid.borderType.HORIZONTAL} />
```

**Namespace** format for the constant: `component` / ` prop`  / `constant value`

Doing this is easier to know where to look for the value you don't have to imagine or search on the code to figure out the value even is easier to do:

```js
import { listBox } from "@paprika/constants";
console.log(listbox.size)
```

To see if what are the value available or even just log listBox and see all the available constants.


With this approach, I think our constants are flexible enough to escalate properly and be easy to remember.

packages/Button/src/Button.styles.js
```diff
@@ -147,7 +146,7 @@ const sizeStyles = {
 // Kinds
 
 const kindStyles = props => ({
-  [Kinds.DEFAULT]: css`
+  [ButtonKinds.DEFAULT]: css`
```
nahumzs, line 149: using my previous example this could be rewrite as:

```js
import { button } from "@paprika/buttons"
const { kind } = button;
const kindStyles = props => ({
  [kind.DEFAULT]: css``,
  [kind.PRIMARY]: css``,
  {...etc}
})
```

packages/helpers/src/enums.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const ShirtSizes = {
+  XSMALL: "xsmall",
+  SMALL: "small",
+  MEDIUM: "medium",
+  LARGE: "large",
+  XLARGE: "xlarge",
+};
+
+ShirtSizes.DEFAULT = [ShirtSizes.SMALL, ShirtSizes.MEDIUM, ShirtSizes.LARGE];
```
nahumzs, line 18: personally I would discourage using this kind of declaration on our propTypes because the consumer will not know what `ShirtSize` means unless goes and look into the code which doesn't seem to be a please developer experience. 

instead, I would put each individual value on the propTypes so the user knowns how to use it.

DanilAgafonov:
I think it would be better to keep packages separated as much as possible. In this context it feels like not a good idea to have one package that contains mentions of a list of different packages (like `button`, `datagrid`, etc. in `@paprika/constants`)

From my point of view, it would be better to have `tokens.js` (or `constants.js`) in each package. In this case we can use it like:

```js
import { Kinds as ButtonKinds } from "@paprika/button/lib/tokens";

<Button kind={ButtonKinds.MEDIUM} />
```

----

In addition we can re-export tokens in `index.js` of each package like:
```js
export { default } from "./Button";
export * from "./tokens";
```

After that we can use it like:

```js
import Button, { Kinds as ButtonKinds } from "@paprika/button";

<Button kind={ButtonKinds.MEDIUM} />
```

----

@nahumzs @tristanjasper WDYT? Maybe I missed some point why we want to extract this constants/tokens out of packages to one central package?

nahumzs:
@DanilAgafonov good point, but I think the main part its discoverability if we stable way of shipping the types, will be easier for our developer to understand how to use them, I think what we can do it exposes the types as part of each package and use the constant package for internal use only for paprika authors example:
```js
// file @paprika/constants
export const sizes = {
   medium, small, large
}

```js
// types.js (inside of listbox package)

import * as constants from @paprika/constants;  // internal package

// compose your own packages types from a base of constants.
export const small = constants.size.small
export const medium = constants.size.medium
export const large = constants.size.large 

// listbox.js
import * as types from "./types.js"
Listbox.types = types;

// -----

// userFile.js
import ListBox from "@paprika/listbox"

<ListBox size={ListBox.types.small} />

```

Taking that approach forces the author and the user to understand the logic behind how to use the types, if we don't have a central place for the types which we can reuse, then each package will have their own types and will be hard for the user to predict how to use our types.

Also, the user can expect that doing `console.log(ListBox.types)` will print the available types.

This solution is easy to use and escalate pretty well. 

What do you think?

tristanjasper:
Sure, I think these approaches do makes sense so I'll refactor to find a middle ground that works.

tristanjasper:
> @DanilAgafonov good point, but I think the main part its discoverability with a stable way of shipping the types. Doing that will be easier for our consumers on how to understand the use of our `types`.
> 
> I think what we can do it exposes the types as part of each package and use the constant package for internal use only for paprika authors example:
> 
> ```js
> // file @paprika/constants
> export const sizes = {
>    medium, small, large
> }
> 
> // types.js (inside of listbox package)
> 
> import * as constants from @paprika/constants;  // internal package
> 
> // compose your own packages types from a base of constants.
> export const small = constants.size.small
> export const medium = constants.size.medium
> export const large = constants.size.large 
> 
> // listbox.js
> import * as types from "./types.js"
> 
> ListBox.propTypes = propTypes;
> ....
> Listbox.types = types;
> 
> // -----
> 
> // userFile.js
> import ListBox from "@paprika/listbox"
> 
> <ListBox size={ListBox.types.small} />
> ```
> 
> Taking that approach forces the author and the user to understand the logic behind how to use the types, if we don't have a central place then each package will have their own types and will be hard for the consumer to predict how to use our types.
> 
> Also, the user can expect `console.log(ListBox.types)` will print the available types.
> 
> This solution is easy to use and escalate pretty well.
> 
> What do you think?|
> 
> Ps. I just found out that the checkbox is using `Checkbox.states.CHECKED` this is exactly what I want to prevent, having a not orthodox way of doing thing one component uses states another one use buttonKind, is just no uniform if we stick to `AnyComponent.types.SOME_VALUE` format its easier for the developer.

@nahumzs Regarding the checkbox, what do you suggest `Checkbox.states.CHECKED` should be as this is different to `kinds` or `sizes`. Do you have another name other than "states" you think works better?

tristanjasper:
Closing as an alternative approach will be taken

PR #571, opened 2020-06-19
STAR-254 Migrate Avatar component

kcumlat:
### Purpose 🚀
Migrate `<Avatar>` component from `<Card>`

### Notes ✏️
An Avatar component represents an object or entity and displays initials or icon.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-254-migrate-avatar
### Screenshots 📸
SizesColorsAvatar

### References 🔗


packages/Avatar/src/Avatar.js
```diff
@@ -0,0 +1,47 @@
...
+
+const defaultProps = {
+  children: null,
+  size: "medium",
+};
+
+function Avatar(props) {
+  const { backgroundColor, children, size, color } = props;
+
+  const avatarProps = {
```
nahumzs, line 25: little confuse about this one, why do we need to split the props into two?

packages/Avatar/src/helpers.js
```diff
@@ -0,0 +1,27 @@
+import Tokens from "@paprika/tokens";
+
+export default function getAvatarColors(text) {
+  const COLORS = [
+    { background: Tokens.color.chartColor01, font: Tokens.color.white },
```
nahumzs, line 5: nip:

const color = Tokens.color;
then:

```js
    { background: color.chartColor02, font: color.white },
    { background: color.chartColor03, font: color.black },
     ...
```

jamiek-acl: @nahumzs I think you mean `nit` 😉

packages/Avatar/stories/examples/SizesColors.js
```diff
@@ -0,0 +1,50 @@
...
+  "jellybeans",
+  "kayak",
+  "lamp",
+];
+
+export default () => {
+  return (
+    <Story>
+      <Heading level={3}>Colors</Heading>
+      <div style={{ display: "flex", flexWrap: "wrap" }}>
```
nahumzs, line 26: for use only on stories, instead of style prop you can make use of the css prop provided by styled-components.

```js
<div css={`display: flex; flexWrap: wrap;`}>
```

packages/Avatar/src/Avatar.js
```diff
@@ -0,0 +1,47 @@
...
+    backgroundColor,
+    children,
+    size,
+    color,
+  };
+
+  return (
+    <sc.Avatar
+      style={{ backgroundColor: avatarProps.backgroundColor, color: avatarProps.color }}
```
nahumzs, line 34: why are we using style here? shouldn't this be props pass down to `<sc.Avatar />`

packages/Avatar/src/helpers.js
alexzherdev: nip the duplication in the bud!

packages/Avatar/stories/examples/SizesColors.js
alexzherdev: would this be syntax highlighted btw?

packages/Avatar/src/helpers.js
jamiek-acl: actually tristan gave very similar feedback and i told him "no way"

packages/Avatar/src/Avatar.js
kcumlat: didn't pass the props in the styles.js. Will update this one.

packages/Avatar/src/Avatar.js
kcumlat: was trying to replicate the format of other components.
Would this make more sense?
```
  const { backgroundColor, children, size, color } = props;

  const styleProps = {
    backgroundColor,
    size,
    color,
  };

    <sc.Avatar data-pka-anchor="avatar" {...styleProps}>
      {children}
    </sc.Avatar>
```
or `<sc.Avatar data-pka-anchor="avatar" {...props} />;`

packages/Avatar/src/Avatar.js
nahumzs: you can add a `$` like $backgroundColor and $color if those props are pass-down the HTML element. 

backgroundColor and color used to be valid attributes in previous HTML versions.

packages/Avatar/src/helpers.js
nahumzs: so let's ignore this comment then.

packages/Avatar/src/Avatar.js
nahumzs: const { backgroundColor, color, ...moreProps  } = props;
```js
<sc.Avatar data-pka-anchor="avatar" $background={backgroundColor} $color={color}  {...moreProps} />
```

^ what about?

kcumlat: Seems like `<div>` doesn't accept attribute since it's a container and can only be styled using css

packages/Avatar/src/Avatar.js
nahumzs: ex
https://codesandbox.io/s/avatar-inrdr?file=/src/App.js

packages/Avatar/src/Avatar.js
```diff
@@ -0,0 +1,42 @@
...
+function Avatar(props) {
+  const { backgroundColor, children, size, color } = props;
+
+  const styleProps = {
+    backgroundColor,
+    size,
+    color,
+  };
+
+  return (
```
nahumzs, line 31: https://codesandbox.io/s/avatar-inrdr?file=/src/App.js look at this example you don't need to split the props. :)

PR #572, opened 2020-06-19
STAR-243 Spread props on checkbox input

allison-c:
### Purpose 🚀

### Notes ✏️

### Updates 📦
- [x] MINOR (backward compatible) `@paprika/checkbox`
- [x] PATCH (bug fix) change to `@paprika/helper` `@paprika/card`

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/Checkbox/src/Checkbox.js
```diff
@@ -57,6 +59,9 @@ const Checkbox = props => {
 
   const checkboxId = React.useRef(uuid()).current;
   const inputRef = React.useRef(null);
+  const extendedInputProps = extractChildrenProps(children, CheckboxInputPropsCollector);
+
+  console.log(extendedInputProps);
```
nahumzs, line 64: do you want to left the console.log?

allison-c: No haha, thanks

packages/Card/stories/examples/DeluxeCard.js
```diff
@@ -14,7 +14,7 @@ export default () => {
           <Card.Title>Asset type name</Card.Title>
           <Input />
           <div style={{ width: "100%", height: "24px" }} />
-          <Checkbox>Use a display name</Checkbox>
+          <Checkbox onChange={() => {}}>Use a display name</Checkbox>
```
nahumzs, line 17: onChange shouldn't be a required value, maybe you can remove it and add a default prop with that empty function. 

There are cases where the checkbox value is provided by an upper state and might not be handle by the user.

mikrotron:
Looks great, I like this approach of using the `<Checkbox.Input>` to capture and propagate props! Would be really lovely if you could do the same thing for the `<Radio>` too so that they both have the same API 😄

PR #575, opened 2020-06-22
RAPRM-373 Propagate Takeover classnames 🐐

mikrotron:
### Purpose 🚀
Implement the propagation of `classNames` for root `<Takeover>` element, as well as `<Takeover.Overlay>` + `<Takeover.Content>`.

### Notes ✏️
- Now propagating all props for Takeover and all Takeover subcomponents via `moreProps`.
- Had to add a `displayName` property to `<Overlay>` for `extractChildren` to work properly.
- Added custom classNames to all subcomponents in the Takeover "Basic" story.
- **BONUS**: avoid errors when the `a11yText` prop is omitted and the children for `Takeover.Header` component is a node by displaying a console.error instead.

### Updates 📦

- [x] MINOR (backward compatible) change to `@paprika/takeover`
- [x] PATCH (bug fix) change to `@paprika/overlay`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/RAPRM-373--propagate-takeover-props

### References 🔗
https://aclgrc.atlassian.net/browse/RAPRM-373

### DIFF
--- a/packages/Takeover/src/Takeover.js
+++ b/packages/Takeover/src/Takeover.js
@@ -93,7 +93,7 @@ export default function Takeover(props) {
       data-pka-anchor="takeover"
       role="dialog"
     >
-      {headerExtracted && <sc.Header onClose={onClose} {...headerExtracted.props} />}
+      {headerExtracted && <sc.Header {...headerExtracted.props} onClose={onClose} />}
       {contentExtracted && (
         <sc.ContentWrapper {...contentExtracted.props} role="region" tabIndex="0">
           {contentExtracted}
--- a/packages/Takeover/src/Takeover.js
+++ b/packages/Takeover/src/Takeover.js
@@ -1,8 +1,8 @@
 import React from "react";
 import PropTypes from "prop-types";
-import Overlay from "@paprika/overlay";
 import { zValue } from "@paprika/stylers/lib/helpers";
 import extractChildren from "@paprika/helpers/lib/extractChildren";
+import Overlay from "@paprika/overlay";
 import FocusLock from "./components/FocusLock";
 import Header from "./components/Header";
 import Content from "./components/Content";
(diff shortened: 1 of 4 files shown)


packages/Takeover/src/Takeover.js
```diff
@@ -46,17 +46,12 @@ export default function Takeover(props) {
   const { a11yText, isOpen, onClose, onAfterClose, onAfterOpen, zIndex, ...moreProps } = props;
 
   const {
+    "Takeover.Content": contentExtracted,
     "Takeover.FocusLock": focusLockExtracted,
-    "Takeover.Overlay": overlayExtracted,
     "Takeover.Header": headerExtracted,
-    "Takeover.Content": contentExtracted,
+    Overlay: overlayExtracted,
```
tristanjasper, line 52: Why do we need this "Overlay" extraction again? 🤔

packages/Takeover/src/Takeover.js
```diff
@@ -66,24 +61,41 @@ export default function Takeover(props) {
...
 
-  const ariaLabel = a11yText || (headerExtracted ? headerExtracted.props.children : null);
+  function getAriaLabel() {
+    if (a11yText) return a11yText;
+    if (!headerExtracted || typeof headerExtracted.props.children !== "string") {
+      console.error(
+        "Accessibility is important 😇\nThe Takeover requires either an a11yText prop or a Takeover.Header with a string for children."
+      );
+      return null;
+    }
```
tristanjasper, line 82: Perhaps this if block could be better placed outside of this function, maybe a generic displayAccessibilityErrors function (Eventually that could even be its own helper)

I would only expect the actual label text to be returned here to keep it clean...

packages/Takeover/src/Takeover.js
```diff
@@ -66,24 +61,41 @@ export default function Takeover(props) {
...
+    return headerExtracted.props.children;
+  }
 
   return React.cloneElement(takeoverOverlay, overlayProps, state => (
-    <sc.Wrapper state={state} role="dialog" aria-label={ariaLabel} aria-modal="true" data-pka-anchor="takeover">
+    <sc.Wrapper
+      state={state}
+      {...moreProps}
+      aria-label={getAriaLabel()}
+      aria-modal="true"
```
tristanjasper, line 92:
```suggestion
      aria-modal
```

packages/Takeover/src/Takeover.js
mikrotron: Ideally we could have a custom propType that validates earlier.  But anything we do is really "nice to have" and better than it was.  For now I just need to get this through to finish up a PR.

PR #583, opened 2020-06-25
STAR-254 Fix Avatar colors naming convention

kcumlat:
### Purpose 🚀

- To update the naming convention of **calculateAvatarColours** to **getAvatarColors** function on `helpers.js`
- To fix the documentation of  the `<Avatar >` component and reflect the same naming convention
- To update how functions in the `helpers.js` are being exported

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-254-avatar-colors

### Screenshots 📸

### References 🔗


packages/Avatar/README.md
```diff
@@ -5,22 +5,27 @@
...
 ```js
-import Avatar, { getAvatarColors } from '@paprika/avatar';
+import { getAvatarColors } from "@paprika/avatar/lib/helpers";
 
-const color = getAvatarColors(text);
+const color = getAvatarColors("text");
 
-<Avatar backgroundColor="black" color="white" size="small">A</Avatar>
+<Avatar backgroundColor="color.backgroundColor" color="color.fontColor" size="small">
```
nahumzs, line 26: `<Avatar backgroundColor="color.backgroundColor" color="color.fontColor" size="small">`

it should be with `{` `}` no `"`
`<Avatar backgroundColor={color.backgroundColor} color={color.fontColor} size="small">`

nahumzs:
I don't think we have a story for screener using the `getAvatarColors` and testing it right? or do we?

kcumlat:
> I don't think we have a story for screener using the `getAvatarColors` and testing it right? or do we?

The example [Avatar Colors](http://storybooks.highbond-s3.com/paprika/STAR-254-avatar-colors/?path=/story/basic-avatar-examples--sizes-and-colors) uses the `getAvatarColors`

packages/Avatar/README.md
```diff
@@ -5,22 +5,27 @@
...
-Avatar using getAvatarColors() to use consistent avatar colors
+Avatar using getAvatarColors() to use consistent colors
 
 ```js
-import Avatar, { getAvatarColors } from '@paprika/avatar';
+import { getAvatarColors } from "@paprika/avatar/lib/helpers";
 
-const color = getAvatarColors(text);
+const color = getAvatarColors("text");
```
mikrotron, line 24: To better illustrate the intention here, I think the `"text"` should match the content of the `<Avatar>` (ie `"A"`).  But maybe an actual first and last name would even better illustrate the intention, instead of just "A".

mikrotron:
Beyond the scope of this PR, but... couldn't the component have a default value of `null` for `color` and `backgroundColor` and when passing those props to the styled component, we could simply call `getAvatarColors(children)` (assuming `typeof children === 'string', otherwise just use `black` or something).  In _most_ cases, this improves the DX by allowing you to use the component like `<Avatar>Horatio Bolivar</Avatar>` and that's it!  But the consume could still override the color props and access `getAvatarColors`, so they still have all the same advanced options? What do you think?

packages/Avatar/README.md
nahumzs, line 24:
```suggestion
const color = getAvatarColors("Terry Fox");
```

kcumlat:
> Beyond the scope of this PR, but... couldn't the component have a default value of `null` for `color` and `backgroundColor` and when passing those props to the styled component, we could simply call `getAvatarColors(children)` (assuming `typeof children === 'string', otherwise just use `black`or something). In _most_ cases, this improves the DX by allowing you to use the component like`Horatio Bolivar`and that's it! But the consume could still override the color props and access`getAvatarColors`, so they still have all the same advanced options? What do you think?

Totally, that could improve the DX. Will look into that.

packages/Avatar/README.md
```diff
@@ -9,21 +9,33 @@
...
 
+By default, Avatar component sets the background and text color using the `getAvatarColor`function with the children props value as the parameter.
+
 ```js
-<Avatar backgroundColor="black" color="white" size="small">
-  A
-</Avatar>
+
+<Avatar>A</Avatar>;
```
mikrotron, line 17:
```suggestion
<Avatar>Terry Fox</Avatar>;
```

packages/Avatar/README.md
```diff
@@ -9,21 +9,33 @@
...
 
 ```js
 import Avatar from "@paprika/avatar";
 import { getAvatarColors } from "@paprika/avatar/lib/helpers";
 
 const color = getAvatarColors("Terry Fox");
 
-<Avatar backgroundColor={color.backgroundColor} color={color.fontColor} size="small">
+<Avatar backgroundColor={color.backgroundColor} color={color.fontColor}>
+  T
```
mikrotron, line 29:
```suggestion
  Terry Fox
```

packages/Avatar/README.md
```diff
@@ -9,21 +9,33 @@
...
+</Avatar>;
+```
+
+Set custom size, background and text color
+
+```js
+import Avatar from "@paprika/avatar";
+
+<Avatar backgroundColor="black" color="white" size="small">
   A
```
mikrotron, line 39:
```suggestion
  Terry Fox
```

packages/Avatar/README.md
kcumlat: For now, `<Avatar>` is setup to take the values pass on  `children` prop (string or icon). For example ```<Avatar>Terry Fox</Avatar>``` 
avatar
Since it is brought up, I guess it makes sense (and improves DX) to define how to display the `children` prop values. If we are going to approach the `Avatar` like the above example. Are we going to set the initial to _uppercase_ and _one letter only_ (first character)?

mikrotron: Oh, interesting 🤔 

If the `typeof children === 'string'` it's possible we could re-use the logic from the `acl-ui` `<Avatar>`:
https://github.com/acl-services/acl-ui/blob/master/src/components/__beta__/Avatar/Avatar.js#L10-L31

But if it's a `node` (Icon)` I guess we'd have to just return that. This is different than the way the `acl-ui` version works though.

PR #587, opened 2020-06-26
STAR-258 Button + ButtonGroup accessibility + styling 🐐

mikrotron:
### Purpose 🚀
1. The `<Button>` should not be interpreted as a "toggle button" if the `isActive` prop is not explicitly provided.
2. The `<ButtonGroup>` `focus()` method should focus on the last focused item, not necessarily the _first_ item.

### Notes ✏️
- Also, the styling of various button states has been adjusted to match the intended design (it was a bit off after the introduction of the `whatinput` logic).

### Updates 📦
- [x] PATCH (bug fix) change to `button` / `button-group` / `raw-button`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-258--button-buttongroup-a11y-styling/?path=/story/commands-button--variations

### References 🔗
https://aclgrc.atlassian.net/browse/STAR-258

### DIFF
--- a/packages/Button/src/Button.styles.js
+++ b/packages/Button/src/Button.styles.js
@@ -4,6 +4,28 @@ import { css, keyframes } from "styled-components";
 import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
 import Kinds from "./ButtonKinds";
 
+// Borders
+
+const borderColors = {
+  [Kinds.DEFAULT]: tokens.border.color,
+  [Kinds.PRIMARY]: tokens.color.green,
+  [Kinds.SECONDARY]: tokens.color.purple,
+  [Kinds.DESTRUCTIVE]: tokens.color.orange,
+  [Kinds.FLAT]: tokens.border.color,
+  [Kinds.MINOR]: "transparent",
+  [Kinds.LINK]: "transparent",
+};
+
+const borderHoverColors = {
+  [Kinds.DEFAULT]: tokens.border.hoverColor,
+  [Kinds.PRIMARY]: tokens.color.greenDarken10,
+  [Kinds.SECONDARY]: tokens.color.purpleDarken10,
+  [Kinds.DESTRUCTIVE]: tokens.color.orangeDarken10,
+  [Kinds.FLAT]: tokens.border.hoverColor,
+  [Kinds.MINOR]: "transparent",
+  [Kinds.LINK]: "transparent",
+};
+
 // States
 
 const disabledStyles = css`
@@ -51,17 +73,12 @@ const activeStyles = css`
 
 const mouseStyles = css`
   [data-whatinput="mouse"] &:not([data-has-forced-focus="true"]):focus {
-    border-color: ${tokens.border.color};
+    border-color: ${({ kind }) => borderColors[kind]};
     box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.1);
+    ${({ kind }) => [Kinds.FLAT, Kinds.MINOR, Kinds.LINK].includes(kind) && "box-shadow: none;"}
 
-    &[kind="minor"],
-    &[kind="link"] {
-      box-shadow: none;
-      border-color: transparent;
-    }
-
-    &[kind="flat"] {
-      box-shadow: none;
+    &:hover {
+      border-color: ${({ kind }) => borderHoverColors[kind]};
     }
   }
 `;
@@ -141,12 +158,10 @@ const kindStyles = props => ({
 
     background-color: ${tokens.color.white};
     background-image: linear-gradient(${tokens.color.blackLighten90}, ${tokens.color.blackLighten70});
-    border-color: ${tokens.border.color};
     color: ${tokens.color.black};
 
     &:hover {
       background: ${tokens.color.blackLighten70};
-      border-color: ${tokens.border.hoverColor};
     }
 
     ${props.isDisabled ? disabledStyles : ""}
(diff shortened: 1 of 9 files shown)


packages/RawButton/src/RawButton.js
```diff
@@ -62,48 +67,45 @@ const RawButton = React.forwardRef((props, ref) => {
...
 
   const bestTabIndex = isDisabled && tabIndex === null ? -1 : tabIndex || 0;
 
   return (
     <span
+      aria-pressed={isActive}
+      data-pka-anchor="raw-button"
+      tabIndex={bestTabIndex}
+      aria-label={a11yText}
+      {...moreProps}
```
tristanjasper, line 98: Just a thought, I wish there was a good way to know where the spreading of `moreProps` should be in the order of applied attributes here.... would be very easy for someone to move this around and not know if it breaks things.... 🤔

packages/Button/src/Button.styles.js
```diff
@@ -46,11 +68,18 @@ const disabledTextStyles = css`
...
 const activeStyles = css`
-  border-color: ${tokens.highlight.active.noBorder.borderColor};
-  box-shadow: ${tokens.highlight.active.noBorder.boxShadow};
+  ${stylers.focusRing.bordered()}
+`;
 
-  &:hover:not([disabled]):not([aria-disabled="true"]) {
-    border-color: ${tokens.highlight.active.noBorder.borderColor};
+const mouseStyles = css`
+  [data-whatinput="mouse"] &:not([data-has-forced-focus="true"]):focus {
```
tristanjasper, line 75: perhaps this could be setup as a mixin/definition in helpers? `[data-whatinput="mouse"] &:not([data-has-forced-focus="true"]):focus` rather than having to write this out every time....

tristanjasper:
Perhaps you can add a link to the mock or screen shot of what the styling should be? I'm not familiar with the `whatinput` logic that you mention

packages/ButtonGroup/src/ButtonGroup.js
```diff
@@ -74,11 +74,9 @@ const ButtonGroup = React.forwardRef((props, ref) => {
...
   React.useImperativeHandle(ref, () => ({
     focus: () => {
-      const firstEnabledIndex = getEnabledIndexes()[0];
-      const firstEnabledItemRef = itemRefs[firstEnabledIndex];
-      if (firstEnabledItemRef) {
-        firstEnabledItemRef.current.focus();
-        setFocusIndex(firstEnabledIndex);
+      if (currentFocusIndex !== null) {
+        const currentFocusRef = itemRefs[currentFocusIndex];
+        if (currentFocusRef) currentFocusRef.current.focus();
```
tristanjasper, line 79: 👍  yeah this works well and does feel pretty natural interacting with it.

packages/Button/src/Button.styles.js
```diff
@@ -46,11 +68,18 @@ const disabledTextStyles = css`
...
+    border-color: ${({ kind }) => borderColors[kind]};
+    box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.1);
+    ${({ kind }) => [Kinds.FLAT, Kinds.MINOR, Kinds.LINK].includes(kind) && "box-shadow: none;"}
```
tristanjasper, line 78: I think i'd rather see the `box-shadow` evaluated in its own function rather than overriding the previous line 77. Otherwise it these two could easily be ordered by accident down the line.

packages/Button/src/Button.styles.js
```diff
@@ -4,6 +4,28 @@ import { css, keyframes } from "styled-components";
...
 
+// Borders
+
+const borderColors = {
+  [Kinds.DEFAULT]: tokens.border.color,
+  [Kinds.PRIMARY]: tokens.color.green,
+  [Kinds.SECONDARY]: tokens.color.purple,
+  [Kinds.DESTRUCTIVE]: tokens.color.orange,
+  [Kinds.FLAT]: tokens.border.color,
+  [Kinds.MINOR]: "transparent",
```
tristanjasper, line 15: nit. use a variable for string "transparent"

packages/RawButton/src/RawButton.js
mikrotron: Anything that would interfere with the proper workings of the component if it were overridden should be after `moreProps`.  Or if it just doesn't make sense to be overridden (ex `isDisabled={isDisabled}` – this _can't_ be overridden since `isDisabled` is already excluded from `moreProps`).  But properties that won't break the component and may have a valid use case to be overridden (ex `data-pka-anchor` may be overridden in other components that want to identify this component more specifically, like `dropdown-menu.trigger-button`).

mikrotron: Short version: offer as much flexibility as possible, but restrict the ability to break it.

packages/Button/src/Button.styles.js
mikrotron: It's only used once in Button and once in RawButton now, and I think moving it to a common `helpers` helper in a third package is a bit harder to follow.  Prefer to follow the WET rule here vs the DRY one.  Or Martin Fowler's "Rule of Three", if you prefer:

https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

mikrotron:
> Perhaps you can add a link to the mock or screen shot of what the styling should be? I'm not familiar with the `whatinput` logic that you mention

Since the migration to Figma, I don't see all the states, this is the closest to the mockups I could find:
https://www.figma.com/file/DRhxHNWVzd94K8Tf0C3VTy/Starling-UI-Kit?node-id=68%3A305

But maybe @jonathanjakimon would know where more of the `Button` states are illustrated?

packages/Button/src/Button.styles.js
mikrotron: Actually decided that the "not disabled" selector would benefit from a helper. It's somewhat specific to a button though, so not sure it should get moved to the `helpers` package:

https://github.com/acl-services/paprika/blob/e9e1b3e76d3e0f848f547304c2e53347508555d8/packages/Button/src/Button.styles.js#L9-L13

PR #588, opened 2020-06-26
STAR-241 bug(DataGrid) fixing scroll horizontal misaligned  / 🦄.

nahumzs:
### Purpose 🚀
After introducing a dynamic scrollbar #566 , a bug was introduced preventing the header to synchronize with the body of the DataGrid, as well as a misaligned between the header and the content when scrolling full to the right side see images.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-241-scrolling-horizontal-fix

### Screenshots 📸
**before**
Screen Shot 2020-06-26 at 1 32 53 PM

**after**
Screen Shot 2020-06-26 at 1 32 41 PM

### References 🔗

### DIFF
--- a/packages/DataGrid/src/DataGrid.js
+++ b/packages/DataGrid/src/DataGrid.js
@@ -203,36 +203,33 @@ const DataGrid = React.forwardRef((props, ref) => {
     [i18n]
   );
 
-  const handleScroll = React.useCallback(
-    parameters => {
-      const { scrollLeft, scrollTop } = parameters;
+  const handleScroll = React.useCallback(parameters => {
+    const { scrollLeft, scrollTop } = parameters;
 
-      if (refScrollHeader.current && stickyColumnsIndexes.length) {
-        refScrollHeader.current.scrollTo({ left: scrollLeft, top: 0 });
-      }
+    if (refScrollHeader.current) {
+      refScrollHeader.current.scrollTo({ left: scrollLeft, top: 0 });
+    }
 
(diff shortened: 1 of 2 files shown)


mikrotron:
Now I see a gap at the right side of the last header cell (when the table doesn't need a horizontal scroll):
Screen Shot 2020-06-26 at 3 37 13 PM

I think this is a less impactful bug than what we had before, so I don't mind merging this one now... but if you know where to quickly fix it, that would be better 😁

nahumzs:
Sheesh lol, good catch, good news and interesting news:

`Good news` is fixed seems to work on Edge Microsoft (chrome version) as well but can't test on regular IE edge :/, but on chrome, firefox, opera, edge mac, 👍 

`Interesting news` the little square its actually the scrollbar lol, but it seems it is not possible to make it invisible with css but exists a deprecated obscure `overflow: overlay` props that seem to make the trick.

This will work at least for the foreseeable future I don't think will get remove from the browsers due to the backwards compatibility. 

        https://stackoverflow.com/questions/23200639/transparent-scrollbar-with-css
        https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#Values

nahumzs:
Screen Shot 2020-06-26 at 4 11 08 PM
Screen Shot 2020-06-26 at 4 14 03 PM
Screen Shot 2020-06-26 at 4 09 21 PM

Issue #562, opened 2020-06-12
DatePicker onMobile device closes calendar immediately on date selection but doesn't set date.


tristanjasper:
# Bug Report
## Prerequisites
**This is just a checklist, please delete this section after**
Please answer the following questions for yourself before submitting a bug report.
- [x] I checked to make sure that this bug has not already been filed
- [x] I reproduced the bug in the latest version of paprika
- [x] Please label as, low / medium / high priority

*Please fill in as much detail as possible below*

## Expected behavior
*Please describe the behavior you are expecting to happen.*

## Current behavior
*What is the current behavior?*

## Screenshots / Gifs / Codepens
![ios q](https://user-images.githubusercontent.com/4040102/84549557-18050500-acbd-11ea-9d0b-36ffa509d8d0.gif)

## Additional context
*Please provide any more additional details to illustrate the bug*

tristanjasper:
This issue is related to https://aclgrc.atlassian.net/browse/GRCPRD-3431

PR #595, opened 2020-07-09
UXD-30 Fix date picker on mobile

allison-c:
### Purpose 🚀
Fix date picker on mobile

### Notes ✏️

The issue was on mobile the blur event on `<DateInput />` from `hideCalendar` will be triggered after `onChange`, so `onChange` has been called twice. The second call is from `<DateInput />` and `<DateInput />` treat it as a `reset` event, since there's no value on the input element yet.

The solution is to make sure `onChange` from `handleSelect` always called after `onBlur`

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/date-picker`
### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-30-fix-date-picker-on-mobile

### Screenshots 📸

date-picker

### References 🔗

Resolve: #562


nahumzs:
clever
👍

PR #598, opened 2020-07-09
UXD-39 feature(DataGrid) Adding LinkAction component

nahumzs:
/ 🦄.

### Purpose 🚀
This PR adds a link component to the DataGrid, there plenty of occasions when a link needs to be added to the table either for opening a sidepanel or another kind of a modal, or for navigate to a new URL. 

There are few not obvious issues if we want to render a regular `<a />` element, therefore, this component hide those complications. 

The component's name is `<LinkAction />` and not `<Link />` because two reason one react router already has a component with that name and secondly the eslint has it register as a component as a regular `<a />` and complains if you don't provide an `href` prop.

### Links

To add link to the table we provided a component to achieve that specific scenario, the `linkAction` component can
behave as a button or as a regular link to use you can import it like:

```sh
import DataGrid, { LinkAction } from "@paprika/data-grid";
```

Once imported you can use it in the following way:

#### As a button

```js
<DataGrid.ColumnDefinition
  key={key}
  header="Link with onClick"
  cell={({ row }) => (
    <LinkAction
      onClick={() => {
        setIsOpen(true);
      }}
    >
      {row[key]}
    </LinkAction>
  )}
  onPressEnter={() => setIsOpen(true)}
/>
```

#### As a regular link

```js
<DataGrid.ColumnDefinition
  key={key}
  header="Link default"
  cell={({ row }) => <LinkAction href="http://design.wegalvanize.com">{row[key]}</LinkAction>}
  onPressEnter={() => {
    LinkAction.href("http://design.wegalvanize.com");
  }}
/>
```

#### LinkAction.href function

The `LinkAction` component also provides an `href` function that let you to implement the link funcionality in an onPressEnter function or any other
event supported by the `DataGrid.ColumnDefinition` component.

```js
// function signature
Link.href = (url, target) => {
  //... implementation
};

// use case:
<DataGrid.ColumnDefinition
  key={key}
  header="Link with _blank"
  cell={({ row }) => (
    <LinkAction href="http://design.wegalvanize.com" target="_blank">
      {row[key]}
    </LinkAction>
  )}
  onPressEnter={() => {
    LinkAction.href("http://design.wegalvanize.com", "_blank");
  }}
/>;
```

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-39-DataGrid-type-link

### Screenshots 📸
Screen Shot 2020-07-09 at 3 19 01 PM

### References 🔗


packages/DataGrid/src/components/Link/Link.js
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+import PropTypes from "prop-types";
+import * as sc from "./Link.styles";
+
+/**
```
oscarkwan, line 5: do you need these comments still?

Issue #520, opened 2020-05-12
Align ref API for Input, Textarea and all basic input forms.


nahumzs:
# Feature request

Currently, the textarea can't receive a direct ref, to be able to access to their ref you have to do a cumbersome approach

```js
          <Textarea
            inputRef={(ref) => {
              refTextAreaSerialized.current = ref;
            }}
            placeholder="paste serialized object"
          />

```

I would expect the textarea, input, checkbox etc, can expose their ref via a forwarded so we can hook the `useRef` into the components.

```js
const refTextarea = React.useRef(null);

<Textarea ref={refTextarea} />
```

I would expect the previous snippet to be the default way to pass refs. Also I'm proposing to remove support for callback ref, now all our ecosystem is on React 16 there it's not a strong reason anymore to support that kind of ref.

nahumzs:
cc @mikrotron and @tristanjasper what do you guys think about this?

PR #599, opened 2020-07-09
UXD-38 Using forwardRef for Select/Input/Textarea

allison-c:
### Purpose 🚀

Using forwardRef for Select/Input/Textarea

### Notes ✏️

Remove `inputRef` prop and forward `ref` to the `input` or `textarea`

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/input` `@paprika/select` `@paprika/textarea`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-38-Forwarding-ref

### Screenshots 📸

### References 🔗
Resolve: #520


nahumzs:
Can you add a changelog.md comments, please?

packages/Textarea/src/Textarea.js
```diff
@@ -85,7 +82,9 @@ function Textarea(props) {
 
   const setRef = node => {
```
nahumzs, line 83: I don't think we need the setRef at all any more right? this only is required if we have to support react 15 that we don't use in our platform.

packages/Input/src/Input.js
```diff
@@ -141,13 +137,13 @@ const Input = props => {
         data-pka-anchor="input"
         disabled={isDisabled}
         readOnly={isReadOnly}
-        ref={inputRef}
+        ref={ref}
```
nahumzs, line 140: this finally will let us have an uncontrolled input nicely and clean 👏 🙇

packages/Textarea/src/Textarea.js
allison-c: I think we still need it here, because the main objective is to assign `textareaRef` and also pass `ref` to consumer side and the `textareaRef` has been used here https://github.com/acl-services/paprika/pull/599/files#diff-a642316bf41d6a04a807281a1985e762R46

nahumzs:
can you referred the issue this is addressing? so we can close it as soon we merge them?

PR #600, opened 2020-07-09
UXD-37 Card style wrapped in a link

### LINKED ISSUE #559: Card - Wrapped in a link
I think it will be very common to wrap Card components in links. When this happens, the style of the Card is affected.  In this screenshot, the card on the left is wrapped in a link:

Screen Shot 2020-06-09 at 4 01 49 PM

It would be nice if we could handle this common use-case for the consumer.  Here is what I used in my app to handle it:

    @import '~@paprika/tokens/lib/tokens.scss';

    .asset-types__card-link {
      display: inline-block;
      text-decoration: none;

      [data-pka-anchor='card.footer'] {
        color: $color--black;
      }
    }

kcumlat:
### Purpose 🚀
To style the `<Card />` when wrapped in a link

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-37-card-style-wrap-link

### Screenshots 📸
Before -> After
link_beforeScreen Shot 2020-07-09 at 4 24 52 PM

### References 🔗
Resolves #559


nahumzs:
don't understand how removing the display type fixs the underline issue.

kcumlat:
> don't understand how removing the display type fixes the underline issue.

Originally `<Card>` uses flex, adding `text-decoration:none` only doesn't work. Tried using `inline-flex` works fine for Firefox but doesn't work in Chrome. I've read that since `<a>` is an inline element, the fix is to set the nested elements to `display: inline-block`, and text-decoration works fine.

nahumzs:
good 👍 , but this doesnt mean that we should encourage to wrap the card in a link right? I mean what if the wrapped link card has another link inside? cc @oscarkwan @mikrotron ther will be a11y concerns. 

But I think this PR is ok for now :)

mikrotron:
> good 👍 , but this doesnt mean that we should encourage to wrap the card in a link right? I mean what if the wrapped link card has another link inside? cc @oscarkwan @mikrotron ther will be a11y concerns.
> 
> But I think this PR is ok for now :)

Agree this is okay for now, but I think I'd prefer if this component accepted an `href` prop (or something along those lines, might not be quite that simple), but then it could wrap itself in an `<a>` element and the styling wouldn't be making assumptions about externalities.

PR #602, opened 2020-07-10
UXD-36 Focus Headings 🐐

mikrotron:
### Purpose 🚀
Implement a `focus()` method for the `<Heading>` component's imperative API + expose this via `<Modal.Header>`, `<SidePanel.Header>` and `<Takeover.Header>` for a11y purposes.

### Notes ✏️
- BREAKING CHANGE: removed the `domRef` prop from the `<Heading>` component.
- Pseudo breaking change? Updated a `pka-anchor` attribute for `<SidePanel>` – could break tests or CSS overrides.
- Modal, SidePanel, and Takeover will rely on the (unreleased) `v0.4.0` of Heading.  Will `lerna` update their dependencies automatically, or will a follow-up PR be required?
- Focusing on the heading element when a modal opens should be an accessibility best practice – especially now that the focus ring on the close button does not show (https://github.com/acl-services/paprika/pull/373)

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/heading`
- [x] MINOR (backward compatible) change to `@paprika/modal`, `@paprika/sidepanel`, `@paprika/takeover`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-36--heading-focus

### Screenshots 📸
heading
modal-header

Note: dotted focus ring is applied specifically to the Storybook stories – will not normally be shown in production.

### References 🔗


packages/Modal/stories/Modal.stories.js
```diff
@@ -106,7 +107,47 @@ storiesOf(`${storyName}/Examples`, module)
...
+            setOpen(false);
+          }}
+          onAfterOpen={() => {
+            if (refHeading.current) refHeading.current.focus();
+          }}
+          css={`
+            [data-pka-anchor="heading"]:focus {
+              ${stylers.focusRing.subtle()}
+            }
+          `}
```
mikrotron, line 136: Dotted focus ring just for the storybook examples.

packages/Takeover/package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "@babel/runtime-corejs2": "^7.3.1",
     "@paprika/button": "^0.3.2",
-    "@paprika/heading": "^0.2.2",
+    "@paprika/heading": "^0.3.20",
```
mikrotron, line 19: Surprising this was still depending on an old version of the Heading.

packages/SidePanel/src/components/Header/Header.js
```diff
@@ -23,57 +23,49 @@ const defaultProps = {
...
-            isSemantic={false}
-            onClick={onClose}
-            {...darkBackgroundProps(kind)}
-            size={isCompact ? ShirtSizes.SMALL : ShirtSizes.MEDIUM}
-          />
-        ) : null}
-      </div>
+      {hasCloseButton && (
+        <Button.Close
+          data-pka-anchor="sidepanel.header.close"
```
mikrotron, line 56: This was updated from `sidepanel-header-close` to align with our naming conventions.

packages/Heading/src/Heading.js
```diff
@@ -12,10 +12,6 @@ const propTypes = {
   /** Optional display level(1-6) affects styles only */
   displayLevel: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
 
-  /** Ref function that returns DOM node for heading */
-  // TODO: Replace with a focus() method via useImperativeHandle()
-  domRef: PropTypes.func,
```
mikrotron, line 17: This is a BREAKING CHANGE 💥 (but I don't think it was widely used, if at all).

packages/Modal/stories/Modal.stories.js
nahumzs: can we use dashed intead of dotted? dashed look better IMHO :P 🙈

mikrotron: This is a design decision, I don't really care. 
I'll change the `styler` if @sunminsarahlee gives it the 👍

PR #607, opened 2020-07-15
feature(DataGrid) OS-on-hightlighted-cell-support support for highlighted cell! :D

nahumzs:
#### Adding onHighlighted prop and isHighlighted prop.

This will unleash the datagrid to handle more complex interaction in their cells, 

#### Previous attempt
In my previous attempt, the `isHighlighted` prop was created by using forwardRef  inside of each`<Cell />` component. This proved to be a little too heavy to process, maybe because `forwardRef` its invoke per cell, so if you have a big table that means react has to call the function per cel and execute whatever that function internally does. (this approach was removed in my previous performance PR. 

#### New attempt
So now I'm trying to solve the issue  adding a global javascript object, `window.paprika.dataGridRef` this object is a pointer to the internal component function that can interact with the isHighlighted state. So far by my observation its working decently, and seems to be an easy way to expose internal Cell's methods to our consumers. 

#### What change
Nothing really change I just add a new feature and onHighlighted method that recieve the `prevCell and the nextCell` this helps the consumer turn on and off the highlighted in the previous cell and in the next cell. 

Also I'm providing a new prop to reset the default CSS style for the cell with the `cellPropsResetCSS` this lets the consumer  style their cell to however they want. this prop is part of `<ColumnDefinition />` component.

Finally now the `cell` prop on `<ColumnDefiniton />` receive two new props (**defaultCssCellStyle**, **isHighlighted**) that can be used to modify the style of one cell at the time like the one on the story

```js
<DataGrid.ColumnDefinition
  {...}
  cellProps={() => ({
    style: {
      background: tokens.color.yellowLighten20,
    },
  })}
  cell={({ row, defaultCssCellStyle, isHighlighted }) => {
    const cssStyle = isHighlighted
      ? {
          overflow: "hidden",
          background: "white",
          maxHeight: "88px",
          position: "absolute",
          top: "-2px",
          padding: "8px",
          lineHeight: "1.3",
          borderRadius: "4px",
          boxShadow: `0 0 0 2px ${tokens.color.blue}`,
        }
      : { ...defaultCssCellStyle };

    return <div style={cssStyle}>{row[key]}</div>;
  }}
/>
```

#### BONUS
I fixed a 🐛 where the highlighted border didn't show if using the `cellProps` prop has a background color.  

### GIF ✏️
What it's possible to do now, visually?

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/DataGrid/src/DataGrid.js
```diff
@@ -38,7 +41,10 @@ const propTypes = {
   onPressShiftSpaceBar: PropTypes.func,
   /** Callback when Spacebar is pressed */
   onPressSpaceBar: PropTypes.func,
+  /** Callback when user click the f key. Might change in the future */
```
oscarkwan, line 44: I didn't know we use the "f" key. What is this for?

packages/DataGrid/src/components/Cell/Cell.js
```diff
@@ -41,18 +48,47 @@ export default function Cell(props) {
...
+  }
+
+  React.useEffect(() => {
+    const key = getGridRefId({ gridId, rowIndex, columnIndex });
+    window.paprika.dataGridRef[key] = {
+      setHighlight,
+    };
+
+    return () => {
+      console.log("before:", Object.keys(window.paprika.dataGridRef).length);
```
oscarkwan, line 62: Do you need these console.logs?

packages/DataGrid/src/components/Cell/Cell.js
```diff
@@ -41,18 +48,47 @@ export default function Cell(props) {
...
+
+  const isColumnCellAFunction = typeof column.cell === "function";
   return (
     <sc.Cell
       borderType={borderType}
-      data-cell={`${gridId}.${columnIndex}.${rowIndex}`}
+      data-pka-cell-key-column-index={columnIndex}
+      data-pka-cell-key-row-index={rowIndex}
+      data-pka-cell-key={`${gridId}.${columnIndex}.${rowIndex}`}
+      data-pka="cell-container"
```
oscarkwan, line 75: This might be a breaking change. I think QA uses data-cell before for acceptance tests

oscarkwan: Are we using `data-pka` now? no more `data-pka-anchor`?

packages/DataGrid/src/DataGrid.js
nahumzs: sarah made the decision is to fire the keyboard event to select a row :)

oscarkwan:
Whats the rationale behind using `data-pka` now? We have a lot of `data-pka-anchors`, it might cause some confusion

packages/DataGrid/src/DataGrid.js
```diff
@@ -4,16 +4,19 @@ import { VariableSizeGrid as Grid } from "react-window";
...
-import useGridEventHandler from "./hooks/useGridEventHandler";
+import useGridEventHandler, { getGridRefId } from "./hooks/useGridEventHandler";
 import ColumnDefinition from "./components/ColumnDefinition";
 import * as sc from "./DataGrid.styles";
 import Basement, { End } from "./components/Basement";
 import InfiniteScroll from "./components/InfiniteScroll";
 import { Row, HeaderRow, StickyRow, StickyHeaderRow, createItemData } from "./DataGrid.Rows";
 import getScrollbarWidth from "./helpers/getScrollbarWidth";
 
+window.paprika = { dataGridRef: {} };
```
mikrotron, line 18: What if there are multiple `<DataGrids>` on a page that rely on these methods?

packages/DataGrid/src/components/Cell/Cell.js
mikrotron, line 75: Should be `data-pka-anchor='data-grid.cell-container'` unless there's a reason to deviate from the standard?

packages/DataGrid/src/components/Cell/Cell.js
```diff
@@ -41,18 +48,47 @@ export default function Cell(props) {
...
       hasZebraStripes={hasZebraStripes}
       rowIndex={rowIndex}
       style={style}
       tabIndex={-1}
     >
-      <sc.GridCell role="gridcell">{a11yText}</sc.GridCell>
-      <sc.InnerCell {...cellProps} aria-hidden="true" {...options.attrs}>
-        {typeof column.cell === "function" ? column.cell(options) : data[rowIndex][column.cell]}
+      <sc.GridCell data-pka="cell-a11y" role="gridcell">
```
mikrotron, line 81: Recommend `data-pka-anchor='data-grid.gridcell'` (or `grid-cell`) to align with the aria role, rather than just `a11y`.

packages/DataGrid/src/hooks/useGridEventHandler.js
```diff
@@ -38,30 +41,37 @@ export default function useGridEventHandler({
...
-  const setHighlight = React.useCallback(() => {
-    if (cell && cell.current && refGrid && refGrid.current) {
-      const $cell = $getCell();
+  const setHighlight = React.useCallback(
+    ({ rowIndex, columnIndex }) => {
+      if (cell && cell.current && refGrid && refGrid.current) {
+        const $cell = $getCell();
+
+        if ($cell) {
+          const classNameStr = "grid--is-active";
```
mikrotron, line 58: `data-grid__cell--is-active` would probably be a better BEM name, but since Paprika has managed to be very `className` agnostic, maybe we could use a `data-pka` attribute instead?

PR #614, opened 2020-07-22
PDTVT-337 add custom icon for clear

oscarkwan:
### Purpose 🚀
- adding an option to pass in a custom icon for the clear button

### Notes ✏️
**Why?**
- in mission control bulk update, clicking the clear button acts as a "reset" action, so we want to use the reset icon here.

### Updates 📦
- [x] MINOR (backward compatible) change to _Input_

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/PDTVT-337-update-input-icon

### Screenshots 📸
Screen Shot 2020-07-22 at 11 37 54 AM

### DIFF
--- a/packages/Input/src/Input.js
+++ b/packages/Input/src/Input.js
@@ -13,6 +13,8 @@ const propTypes = {
   /** Sets the class for the input. */
   className: PropTypes.string,
 
+  clearIcon: PropTypes.node,
+
   /** Sets the default input value  */
   defaultValue: PropTypes.string,
 
@@ -51,6 +53,7 @@ const propTypes = {
 
 const defaultProps = {
   a11yText: null,
+  clearIcon: <TimesCircleIcon />,
   className: null,
   defaultValue: "",
   hasClearButton: false,
@@ -77,6 +80,7 @@ const Input = props => {
     const { hasClearButton, isDisabled, isReadOnly, size, value } = props;
     if (!hasClearButton || isDisabled || isReadOnly || !value) return null;
     const iconSize = size === ShirtSizes.LARGE ? ShirtSizes.MEDIUM : ShirtSizes.SMALL;
+
     return (
       <Button.Icon
         a11yText="Clear Input" // TODO: add L10n
@@ -85,7 +89,7 @@ const Input = props => {
         size={iconSize}
         onClick={inputClearHandler}
       >
-        <TimesCircleIcon />
+        {props.clearIcon}
       </Button.Icon>
     );
   };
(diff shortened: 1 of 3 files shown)


packages/Input/src/Input.js
```diff
@@ -85,7 +89,7 @@ const Input = props => {
         size={iconSize}
         onClick={inputClearHandler}
       >
-        <TimesCircleIcon />
+        {props.clearIcon}
```
AndreyChernykh, line 92: Is adding jsx components in default props considered a good practice? 

I'd prefer:

```js
{props.clearIcon || <TimesCircleIcon />} 
```

instead

oscarkwan: Do we ignore the default prop then?

AndreyChernykh: Also,

```js
const { hasClearButton, isDisabled, isReadOnly, size, value, clearIcon } = props;

...
{clearIcon || <TimesCircleIcon />}
```
instesd of 

```
props.clearIcon
```

oscarkwan: Okok, ill just set it to null

AndreyChernykh: > Do we ignore the default prop then?

I guess we still need a default prop `null`. 

Imagine you had a more complex scenario rather than just `<TimesCircleIcon />` without any props? It could be not possible to put it into default props at all anyway. 

Maybe, if we provide a default icon and allow user to customize it, then `customClearIcon` would be a more descriptive name than just `clearIcon`

Also, if we want to allow the user to have no icon, then 

```
{clearIcon === undefined || <TimesCircleIcon />}
```

that way you can do `<Input clearIcon={null} />`. 

@nahumzs cc

oscarkwan: we also have `hasClearButton` that needs to be set to true, so probably don't need that above case to set clearIcon to null.

oscarkwan: I can also rename it to `customClearIcon`, thoughts? @nahumzs

nahumzs:
My two cents, clearIcon should have the default set to <TimesCircleIcon />, is totally fine having JSX as default, is just a function isn't? 

Then if you don't want to have the clear button, then use the proper already establish pattern of `hasClearButton` ex:

default
```js
<Input /> // no clear button
```

with icon
```js
<Input hasClearButton /> // renders the input with `x` clear button
```

with custom icon

```js
<Input hasClearButton clearButton={<MyThingy />} /> // renders MyThingy instead of <TimesCircleIcon />
```

with this, we should be able to have all scenarios I think so unless I'm missing something.

PR #619, opened 2020-07-23
Pdtvt 337 add listbox extra button

oscarkwan:
### Purpose 🚀
- add a prop to add extra buttons to listbox footer

### Updates 📦
- [x] MINOR (backward compatible) change to _ListBox_


nahumzs:
Just for context and future reference this diverge from the original meaning of this component but I guess a refactor of the ListBox.Footer is needed it.

Issue #603, opened 2020-07-10
Sortable inside SidePanel is not working


nahumzs:
# Bug Report

## Issue
![2020-07-10 11 07 55](https://user-images.githubusercontent.com/19418590/87186282-1456a500-c2a0-11ea-80e0-5eccbe641fc6.gif)

This occurs because of these lines: https://github.com/acl-services/paprika/blob/65f1bca8ed1db44451024f982a166570f301b681/packages/SidePanel/src/components/Dialog/Dialog.styles.js#L18-L38

once an element has applied a transformation the position fixed and absolute will change their context as explained on [this StackOverflow question](https://stackoverflow.com/a/36857029/196038)

This issue is not something we are not familiar we experience in our first version of the popover in our previous library, right now our current Popover is free of this issue. 

## Solution
Add a new prop to the Sidepanel can be `hasTransformAnimation` which can be turned on and off depending on the needs of the developers.

This will allow a better integration not only for the Sortable (which is a third party library) but other legacy cases.

PR #630, opened 2020-07-31
Uxd 66 sortable in sidepanel dragging bug - alternative

tristanjasper:
### Purpose 🚀
Allow Sortable to be used in a sidepanel without any dragging animation issues.

This is an alternative approach to adding an additional prop as suggested in the github issue (https://github.com/acl-services/paprika/issues/603) for which I have created a reference pr https://github.com/acl-services/paprika/pull/631

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to Sidepanel

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-66-sortable-in-sidepanel-dragging-bug-alternative

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-66

### DIFF
--- a/packages/SidePanel/src/components/Dialog/Dialog.js
+++ b/packages/SidePanel/src/components/Dialog/Dialog.js
@@ -9,7 +9,6 @@ const propTypes = {
   getPushContentElement: PropTypes.func,
   groupOffsetY: PropTypes.number,
   header: PropTypes.node,
-  hasTransformAnimation: PropTypes.bool.isRequired,
   kind: PropTypes.oneOf(["default", "child"]),
   isCompact: PropTypes.bool,
   isInline: PropTypes.bool,
@@ -38,6 +37,7 @@ const defaultProps = {
 };
 
 function Dialog(props) {
+  const [isAnimating, setIsAnimating] = React.useState(false);
   const refSidePanel = React.useRef(null);
 
   const {
@@ -47,7 +47,6 @@ function Dialog(props) {
     getPushContentElement,
     groupOffsetY,
     onAnimationEnd,
-    hasTransformAnimation,
     header,
     kind,
     isCompact,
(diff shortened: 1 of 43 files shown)


packages/SidePanel/src/components/Dialog/Dialog.js
```diff
@@ -80,8 +81,17 @@ function Dialog(props) {
     </React.Fragment>
   );
 
+  React.useEffect(() => {
+    setIsAnimating(true);
```
nahumzs, line 85: shouldn't this run only when `isOpen === true`? if not this effect will run each time isOpen prop switch no mattering if it's open or close.

tristanjasper: We want it to animate (run) when it is opening OR closing so I believe it should run for both situations

packages/SidePanel/src/components/Dialog/Dialog.styles.js
```diff
@@ -59,7 +37,35 @@ export const Dialog = styled.div`
 
   ${props => {
     const width = Number.isNaN(Number(props.width)) ? props.width : `${props.width}px`;
-    const animation = props.isOpen ? slideIn(props.isSlideFromLeft) : slideOut(props.isSlideFromLeft);
+    const slideOutTransform = props.isSlideFromLeft ? "translateX(-100%)" : `translateX(100%)`;
+    const slideInTransform = "translateX(0)";
+    function slideIn() {
```
nahumzs, line 42: can we extract this function outside will prevent of creating this function each time the stylecomponent function runs, you can do something like:

```js
function slideIn({ slideInTransform , slideOutTransform  l}) {
  return ...
}
```

also `props.isAnimating` should be desctructure at the top of the stylecomponent function as the style guide suggests

PR #633, opened 2020-08-04
STAR-265 bug(ListBox) working incorrectly with Enter keydown

nahumzs:
/ 🦄.

### Purpose 🚀
Before this PR  the onkeydown event for `↑`,`↓` , `SPACE`, `ESC` and `ENTER` where under the same function and under onKeyDown event, this creates an interesting behaviour for the Enter key where toggle the listbox opening and closing it when trying it to opening.

The reason why were bound into onKeyDown was because  ↑↓→←  have to have to ability to scroll up and down without lifting the key. 

With this PR this behaviour has been splitted between keydown and keyup, keydown will manage the arrow keys while the keyup will handle Space, Enter and Esc.

This results in a nicer behaviour.   

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-265-Listbox-on-enter

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/DataGrid/src/components/Cell/Cell.js
+++ b/packages/DataGrid/src/components/Cell/Cell.js
@@ -59,9 +59,7 @@ export default function Cell(props) {
     };
 
     return () => {
-      console.log("before:", Object.keys(window.paprika.dataGridRef).length);
       delete window.paprika.dataGridRef[key];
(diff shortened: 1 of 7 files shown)


nahumzs:
The only reason this is a breaking change is because QA in our consumer might be using onkeydown to hook their test and this will break those tests. cc @oscarkwan

PR #643, opened 2020-08-13
UXD-115 - Collapsible - Different default Icon depends on position

KDarcey:
### Purpose 🚀
To match current design guidelines for Collapsible. Adjust chevron icon to be an UpArrow when alligned right and expanded and to be a DownArrow when alligned right and collapsed. 

### Notes ✏️
N/A

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-115-Collapsible-default-chevron

### Screenshots 📸
image

### References 🔗
N/A

### DIFF
--- a/packages/Collapsible/src/Collapsible.js
+++ b/packages/Collapsible/src/Collapsible.js
@@ -2,6 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import RightArrowIcon from "@paprika/icon/lib/ArrowRight";
 import DownArrowIcon from "@paprika/icon/lib/ArrowDown";
+import UpArrowIcon from "@paprika/icon/lib/ArrowUp";
 import useI18n from "@paprika/l10n/lib/useI18n";
 import RawButton from "@paprika/raw-button";
 import collapsibleStyles from "./Collapsible.styles";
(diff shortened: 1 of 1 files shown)


packages/Collapsible/src/Collapsible.js
```diff
@@ -52,7 +53,16 @@ const defaultProps = {
...
+
+  const [collapsedIcon, setCollapsedIcon] = React.useState([props.iconExpand, props.iconCollapse]);
+
+  React.useEffect(() => {
+    if (props.iconAlign === "right") {
+      setCollapsedIcon([<UpArrowIcon />, <DownArrowIcon />]);
+    } else {
+      setCollapsedIcon([props.iconExpand, props.iconCollapse]);
+    }
+  }, []);
```
allison-c, line 65: I think `collapsedIcon` doesn't need to be a state, and using `useEffect` here. The default values of `iconExpand`, `iconCollapse` can be `null`, and then doing something like this:

```
const collapsedIcon = [props. iconExpand || props.iconAlign === "right" ? <UpArrowIcon /> : <RightArrowIcon /> , props. iconCollapse || props.iconAlign === "right" ? <DownArrowIcon /> : <RightArrowIcon />]
```

PR #644, opened 2020-08-13
Uxd 98 html table

nahumzs:
### Purpose 🚀
The Simple Table component provides the functionality of a regular HTML table with a similar API of the DataGrid, unlike the DataGrid the table component lacks certain futures and it's intended to be used when the number of rows is no overwhelming and dynamic height is required.

For cases where the table requires one or more of the following features, the use of the DataGrid can be a better alternative:
- Display hundreds or thousands of rows at the same time
- Fix columns and Fix headers
- Need for up, right, down, left arrow key navigation

Both components DataGrid and Table can be used with the ActionBar without a problem and both of them can have a custom header or custom cells. 

### Notes ✏️
Screen Shot 2020-08-13 at 1 14 35 PM
Screen Shot 2020-08-13 at 1 14 43 PM

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-98-HTML-Table

### Screenshots 📸

### References 🔗


midoritakeuchi:
Hi @nahumzs! Thanks for creating this component for us 👏🏼
One question - In the `DataGrid` component there's the count of rows and columns at the bottom. Will this new `Table` component have the option to hide that? No worries if it is too complicated.

nahumzs:
@midoritakeuchi yes this doesn't have it, actually, the DataGrid shouldn't have it either, but since the DataGrid is intended for displaying a vast amount of rows and columns sometimes make sense to know how many columns and rows are rendered. 

We have a ticket to make the DataGrid Footer Optional in the near future :).

packages/Table/src/Table.styles.js
```diff
@@ -0,0 +1,65 @@
...
+    `;
+  }
+  return css`
+    ${zebras};
+  `;
+});
+
+export const Thead = styled.thead`
+  background: ${tokens.table.header.backgroundColor};
+
```
tristanjasper, line 30: remove empty line

packages/Table/stories/Table.collections.stories.js
```diff
@@ -0,0 +1,125 @@
...
+};
+
+function Title(props) {
+  const { label, description, tag } = props;
+
+  return (
+    <div>
+      <titleStyles.Title>
+        <titleStyles.Label>
+          <Button.Link size="big">{label}</Button.Link>
```
tristanjasper, line 74: Shouldn't `size` be a `CustomPropTypes.ShirtSize` value? Do we need to create a new one?

packages/Table/stories/Table.collections.stories.js
```diff
@@ -0,0 +1,125 @@
...
+    width: 100%;
+  `,
+};
+
+function Item(props) {
+  const { value } = props;
+
+  return (
+    <itemStyles.Item>
+      <Button kind="flat">{value}</Button>
```
tristanjasper, line 99: use `Button.Kinds`? Perhaps not necessary in a story though

packages/Table/stories/Table.collections.stories.js
nahumzs: I'm not fan of using micro abstraction, which adds unnecessary complexity IMHO while reading the code. A lot of my thoughts about this are part of the comment that I added about our Constant package :)

tristanjasper: Ok. I figured we'd standardized the approach but maybe not. I'll update the types/enums pr to try and close it out and solidify an approach.

oscarkwan:
Semantic html checklist

- [ ] th are used for headers
- [ ] tr and td are used properly
- [] thead and tbody are used properly

Question
1. Will there be issues with global styling? Or should we just ignore that?
2. According to designs, for the zebra pattern I believe that it should start with `white/grey` and not `grey/white`

packages/Table/src/Table.js
```diff
@@ -0,0 +1,99 @@
...
+
+  let ColumnDefinitions = extractedColumnDefinitions;
+
+  if (!Array.isArray(extractedColumnDefinitions)) {
+    // when there is only one component extracted function return the element not an array of elements
+    ColumnDefinitions = [extractedColumnDefinitions];
+  }
+
+  return (
+    <sc.Table {...moreProps}>
```
oscarkwan, line 42: I feel like we should always provide an `aria-label`. Or else screen reader will just say "2 rows 11 columns" and give no context on what the table is

packages/Table/README.md
```diff
@@ -0,0 +1,49 @@
+<!-- autogenerated don't modify -->
+
+# @paprika/table - 0.1.0
+
+## Description
+
+The Table component is a simple wrap for the html table, for more advance feature please take a look into the DataGrid
```
oscarkwan, line 7:
```suggestion
The Table component is a simple wrap for the html table. For the use of a data table, please take a look into the DataGrid component.
```

packages/Table/src/Table.js
```diff
@@ -0,0 +1,99 @@
...
+import PropTypes from "prop-types";
+import ColumnDefinition from "./components/ColumnDefinition";
+import types from "./types";
+import * as sc from "./Table.styles";
+
+const propTypes = {
+  /** Define the look for borders in the table DataGrid.types.GRID, DataGrid.types.NONE, etc.  */
+  borderType: PropTypes.oneOf(["grid", "non", "horizontal", "vertical"]),
+  children: PropTypes.node.isRequired,
+  /** Add an alternate background on the DataGrid's rows */
```
oscarkwan, line 14:
```suggestion
  /** Add an alternating background on the table rows */
```

packages/Table/src/Table.js
```diff
@@ -0,0 +1,99 @@
...
+import React from "react";
+
+import extractChildren from "@paprika/helpers/lib/extractChildren";
+import PropTypes from "prop-types";
+import ColumnDefinition from "./components/ColumnDefinition";
+import types from "./types";
+import * as sc from "./Table.styles";
+
+const propTypes = {
+  /** Define the look for borders in the table DataGrid.types.GRID, DataGrid.types.NONE, etc.  */
```
oscarkwan, line 11:
```suggestion
  /** Define the look for borders in the table Table.types.GRID, Table.types.NONE, etc.  */
```

packages/Table/src/Table.js
```diff
@@ -0,0 +1,99 @@
...
+  borderType: PropTypes.oneOf(["grid", "non", "horizontal", "vertical"]),
+  children: PropTypes.node.isRequired,
+  /** Add an alternate background on the DataGrid's rows */
+  hasZebraStripes: PropTypes.bool,
+  /** Array of data to be stored in the DataGrid */
```
oscarkwan, line 16:
```suggestion
  /** Array of data to be stored in the Table */
```

packages/Table/src/Table.js
```diff
@@ -0,0 +1,99 @@
...
+  data: PropTypes.arrayOf(PropTypes.shape({})),
+  /** Sets the height of the DataGrid */
```
oscarkwan, line 18: Remove comment

oscarkwan: FYI, you will need to update readme after these changes

packages/Table/src/Table.js
nahumzs: I mean you can totally add it from the consumer side that's why moreProps are there, do you mean we should enforce something like `a11yText` so the consumer always has to provide a label?

nahumzs: > According to designs, for the zebra pattern I believe that it should start with white/grey and not grey/white

oh I see good catch will change it.

oscarkwan: `a11yText` would be nice i think. what do you think?

oscarkwan:
Reminder to add follow up for tests

nahumzs:
@oscarkwan I don't know what kind of test would we need? I added screener test already, what else do we need for this specific component?

oscarkwan:
> @oscarkwan I don't know what kind of test would we need? I added screener test already, what else do we need for this specific component?

Screener is good, and just some basic unit tests to make sure data passed in is rendered correctly

PR #646, opened 2020-08-14
STAR-264 Pagination Component

KDarcey:
### Purpose 🚀
To migrate the Pagination component to Paprika.

### Notes ✏️
- Refactored css with styled components.
- Added documentation (readme and mdx).
- Added spec and cypress tests.
- Added stories

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/STAR-264-Pagination

### Screenshots 📸
image
image


packages/Pagination/README.md
```diff
@@ -0,0 +1,30 @@
+<!-- autogenerated don't modify -->
+
+<!-- autogenerated don't modify -->
+<!-- content -->
+
+## Pagination
```
nahumzs, line 6: can you run `yarn readme` so can generate the prop table for you?

packages/Pagination/src/Pagination.js
```diff
@@ -0,0 +1,107 @@
...
+import ElipsisItem from "./components/EllipsisItem/ElipsisItem";
+import PageItem from "./components/PageItem/PageItem";
+import Divider from "./components/Divider";
+import * as sc from "./Pagination.styles";
+
+const propTypes = {
+  /** The number of current active page */
+  currentPage: PropTypes.number.isRequired,
+  /** Length of the pagination (the number of other pages will be displayed around active page). */
+  length: ShirtSizes,
```
nahumzs, line 15: 🤔 don't know if this prop is clear about what actually does... what does length do? and why has `shirtsize` attached to it? this is why I think micro abstraction like `shirtsize` doesn't escalate properly, doesn't have any meaning by itself, needs a prop to create one, and in this case, the prop doesn't help too much in providing a meaningful meaning. 

If we were using something like [types.small, types.big] then this would make more sense but `ShirtSize` is a term that is too abstract to be understood. 

cc @tristanjasper

nahumzs: For this case can we use a more meaningful prop name? 🙏

packages/Pagination/src/Pagination.js
```diff
@@ -0,0 +1,107 @@
...
+  /** Callback to be executed when current page is changed. */
+  onClickPage: PropTypes.func.isRequired,
```
nahumzs, line 17: we don't have another click event I would say its safe to just use `onClick` without the subfix.

packages/Pagination/src/components/Divider/Divider.js
```diff
@@ -0,0 +1,8 @@
+import React from "react";
```
nahumzs, line 1: you can remove Divider.styles.js and move all into `Divider.js` Right now Divider.js is not doing anything apart of re-exporting `sc.Divider`, after moving everything you can simple delete Divider.styles.js and everything should work correctly. 

you can address the same as:

like:

```js
// Divider.js

import styled from "styled-components";

const Divider = styled.span`
  background-color: ${tokens.border.color};
  display: inline-block;
  height: 24px;
  margin-left: ${tokens.space};
  margin-right: ${tokens.space};
  width: 1px;
`;

export default Divider;
```

packages/Pagination/stories/examples/InteractivePagination.js
```diff
@@ -0,0 +1,154 @@
+import React from "react";
+import { Story } from "storybook/assets/styles/common.styles";
+import Card from "@paprika/card";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import Heading from "@paprika/heading";
+import Avatar from "@paprika/avatar";
+import Pagination from "../../src/Pagination";
+
+const data1 = [
```
nahumzs, line 9: since this is a story and what you want is devs focusing in reading your code example, I would suggest to move all the data into a separate file and import it from the other file something like:
```
// mocks.js

export const data1 = [...]
export const data2 = [...]

```

and then in this file:

```js
import { data1, data2 } from "./ mocks"
```

And everything should keep working correctly but without data cluttering.

packages/Pagination/src/components/ArrowItem/ArrowItem.js
```diff
@@ -0,0 +1,44 @@
...
+  type: "right",
+};
+
+function ArrowItem(props) {
+  const { a11yText, isDisabled, onClick, type } = props;
+  const Icon = getArrowIcon(`Arrow-${type}`);
+  const I18n = useI18n();
+
+  return (
+    <sc.ArrowItem aria-label={a11yText || I18n.t(`pagination.${type}`)} isDisabled={isDisabled} onClick={onClick}>
```
nahumzs, line 34: This might fail accessibility test, it should be a RawButton or a Button, is not even showing the `cursor:pointer` when interacting with it. 

@allison-c can you confirm that this is the intended? I think we want to change it right? 

if we go with the RawButton you simple need to do in your `*.style.js` file the following:

```js
export const ArrowItem = styled(RawButton)`
  ...yourStuff
`
```

the previous should add all the accessibility to the number that should behave like a button

nahumzs:
One interaction feedback, I think that if we are on Page 1 we want to hide the left chevron or at least show it as disable, and do the same with the right chevron if there is no more pages we should show the chevron. Please sync with Sara to confirm this is the desire interaction. 🙏

packages/Pagination/src/components/ArrowItem/ArrowItem.js
allison-c: I agree we should use `RawButton` and `a11yText` prop, the cursor will be the correct one if you do it

packages/Pagination/src/Pagination.js
```diff
@@ -0,0 +1,102 @@
+import React from "react";
+import PropTypes from "prop-types";
+// import { ShirtSizes } from "@paprika/helpers/lib/customPropTypes";
```
allison-c, line 3: Can be removed

packages/Pagination/src/components/ArrowItem/ArrowItem.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import PropTypes from "prop-types";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import { getArrowIcon } from "./ArrowItem.styles";
+import * as sc from "./ArrowItem.styles";
+
+const propTypes = {
+  /* Description of the ArrowItem for assistive technology */
+  a11yText: PropTypes.string,
```
allison-c, line 9: How useful is this? I didn't see you pass in `a11yText ` from `<Pagination />`

packages/Pagination/src/components/ArrowItem/ArrowItem.js
```diff
@@ -0,0 +1,44 @@
...
+};
+
+function ArrowItem(props) {
+  const { a11yText, isDisabled, onClick, type } = props;
+  const Icon = getArrowIcon(`Arrow-${type}`);
+  const I18n = useI18n();
+
+  return (
+    <sc.ArrowItem aria-label={a11yText || I18n.t(`pagination.${type}`)} isDisabled={isDisabled} onClick={onClick}>
+      <Icon isDisabled={isDisabled} size={14} />
```
allison-c, line 35: Although icon components support `size` prop, but using styled-component to set the `font-size` should still be the first option
example: https://github.com/acl-services/paprika/blob/master/packages/Calendar/src/internal/calendar.styles.js#L9-L12

packages/Pagination/src/components/ArrowItem/ArrowItem.styles.js
```diff
@@ -0,0 +1,40 @@
+import styled, { css } from "styled-components";
+import RawButton from "@paprika/raw-button";
+import tokens from "@paprika/tokens";
```
allison-c, line 3: minor one, `import tokens` can go the last

packages/Pagination/src/components/ArrowItem/ArrowItem.js
```diff
@@ -0,0 +1,44 @@
...
+const defaultProps = {
+  a11yText: null,
+  isDisabled: false,
+  onClick: () => {},
+  type: "right",
+};
+
+function ArrowItem(props) {
+  const { a11yText, isDisabled, onClick, type } = props;
+  const Icon = getArrowIcon(`Arrow-${type}`);
```
allison-c, line 30: Since you use `type === "Arrow-Left"` in `getArrowIcon`,  can this line be simpler like `getIcon(type)` and use ``type === "left"` in `getIcon`

packages/Pagination/src/components/ArrowItem/ArrowItem.styles.js
```diff
@@ -0,0 +1,40 @@
...
+  `;
+};
+
+export const ArrowItem = styled(RawButton)`
+  ${({ isDisabled }) => css`
+    align-items: center;
+    border-radius: ${tokens.button.borderRadius};
+    box-sizing: border-box;
+    display: flex;
+    height: ${tokens.space * 3};
```
allison-c, line 28: I believe in paprika we usually use `stylers.spacer`

packages/Pagination/src/components/CurrentPageItem/CurrentPageItem.js
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import PropTypes from "prop-types";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import * as sc from "./CurrentPageItem.styles";
+
+const propTypes = {
+  /* Description of the CurrentPageItem for assistive technology */
+  a11yText: PropTypes.string,
```
allison-c, line 8: same here, I didn't see `a11yText ` passed in

packages/Pagination/src/components/CurrentPageItem/CurrentPageItem.js
```diff
@@ -0,0 +1,43 @@
...
+
+  /** The page number */
+  pageNumber: PropTypes.number.isRequired,
+
+  /** Value for role attribute to override the default */
+  role: PropTypes.string,
```
allison-c, line 14: wasn't passed in

packages/Pagination/src/components/Divider/Divider.js
```diff
@@ -0,0 +1,13 @@
+import styled from "styled-components";
+import tokens from "@paprika/tokens";
+
+const Divider = styled.span`
+  background-color: ${tokens.border.color};
+  display: inline-block;
+  height: 24px;
```
allison-c, line 7: spacer(3)

packages/Pagination/src/components/EllipsisItem/ElipsisItem.js
```diff
@@ -0,0 +1,33 @@
...
+import PropTypes from "prop-types";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import * as sc from "./ElipsisItem.styles";
+
+const propTypes = {
+  /* Description of the ElipsisItem for assistive technology */
+  a11yText: PropTypes.string,
+
+  /** Value for role attribute to override the default */
+  role: PropTypes.string,
```
allison-c, line 11: same here

packages/Pagination/src/components/EllipsisItem/ElipsisItem.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import PropTypes from "prop-types";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import * as sc from "./ElipsisItem.styles";
+
+const propTypes = {
+  /* Description of the ElipsisItem for assistive technology */
+  a11yText: PropTypes.string,
```
allison-c, line 8: same here

packages/Pagination/src/components/PageItem/PageItem.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import PropTypes from "prop-types";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import * as sc from "./PageItem.styles";
+
+const propTypes = {
+  /* Description of the PageItem for assistive technology */
+  a11yText: PropTypes.string,
```
allison-c, line 8: same here

packages/Pagination/src/components/PageItem/PageItem.js
```diff
@@ -0,0 +1,37 @@
...
+  /** Callback to be executed when the PageItem is clicked or activated by keyboard. */
+  onClick: PropTypes.func,
+
+  /** The page number */
+  pageNumber: PropTypes.number.isRequired,
+};
+
+const defaultProps = {
+  a11yText: null,
+  onClick: () => {},
```
allison-c, line 19: why this is optional but in `<Pagination />` `onClick` is required?

packages/Pagination/stories/examples/InteractivePagination.js
```diff
@@ -0,0 +1,66 @@
...
+                    <span>{props.footer1}</span>
+                    <span> {props.footer2}</span>
+                  </div>
+                </Card.Footer>
+              </Card>
+            </div>
+          );
+        })}
+      </div>
+      <Pagination onClick={ChangePage} style={{ marginTop: "100px" }} totalPages={3} currentPage={selectedPage} />
```
allison-c, line 63: should be `changePage` since it's just a function

packages/Pagination/tests/spec/Pagination.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    expect(getByText("2")).toBeVisible();
+    expect(getByText("3")).toBeVisible();
+    expect(getByText("4")).toBeVisible();
+    expect(getByText("12")).toBeVisible();
+    expect(getAllByRole("button")[0]).toBeVisible();
+    expect(getAllByRole("button")[1]).toBeVisible();
+  });
+  it("Should highlight current page", () => {
+    const { getByText } = renderComponent();
+    expect(getByText("2")).toHaveStyle("background-color:	#785cba");
```
allison-c, line 22: testing `aria-current` will be better

packages/Pagination/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "access": "public"
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/helpers": "^0.2.13",
+    "@paprika/icon": "^0.3.4",
+    "@paprika/l10n": "^0.4.6",
+    "@paprika/stylers": "^0.2.8",
+    "@paprika/tokens": "^0.1.14",
+    "@paprika/raw-button": "0.2.14",
```
allison-c, line 23: doesn't need to be a fixed version

packages/Pagination/src/components/PageItem/PageItem.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+function PageItem(props) {
+  const { a11yText, onClick, pageNumber } = props;
+  const I18n = useI18n();
+
+  return (
+    <sc.PageItem aria-label={(a11yText || I18n.t("pagination.page"), pageNumber)} onClick={onClick}>
```
allison-c, line 27: Should be `a11yText` since it's a RawButton now https://github.com/acl-services/paprika/blob/master/packages/RawButton/src/RawButton.js#L10

packages/Pagination/src/Pagination.js
```diff
@@ -0,0 +1,102 @@
...
+    const left = currentPage - delta;
+    const right = currentPage + delta;
+
+    const items = [];
+
+    let wasPreviousItemEllipsized = false;
+
+    for (let pageNumber = 1; pageNumber <= totalPages; pageNumber++) {
+      if (isCurrentPage(pageNumber)) {
+        items.push([<CurrentPageItem key={pageNumber} pageNumber={pageNumber} />, <Divider />]);
```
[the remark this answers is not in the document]
KDarcey: My thinking is that with the divider as a component it is easier to adjust the styling. Instead of having to add &::after css for all of the elements and have to adjust each one when we want to make changes, it is all in conveniently in one place. Additionally, it means we don't have to set up specific styling for the Arrow Icons when the arrow is left or right, instead it will be rendered in the appropriate place using the component. What do you think?

packages/Pagination/stories/examples/InteractivePagination.js
nahumzs: actually should be more in the terms of `handleClickChangePage`

packages/Pagination/src/Pagination.js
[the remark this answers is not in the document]
allison-c: Personally I'd prefer to simplify the number of DOM elements rendered, for example using `:after` when we can. But this is not a big deal if it's hard to do

packages/Pagination/src/Pagination.js
```diff
@@ -0,0 +1,102 @@
...
+import Divider from "./components/Divider";
+import * as sc from "./Pagination.styles";
+
+const propTypes = {
+  /** The number of current active page */
+  currentPage: PropTypes.number.isRequired,
+  /** The number of other pages that will be visible around the current/active page (not hidden by elipsis). */
+  pagesOnEachSide: PropTypes.number,
+  /** Callback to be executed when current page is changed. */
+  onClick: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
nahumzs: ohhh I thought its was a onClick, because was named onClickPage, but after reading the comment seems is an onChange and not a onClick, good call @allison-c can we rename it @KaanDarcey ?

packages/Pagination/src/Pagination.js
```diff
@@ -0,0 +1,106 @@
...
+
+  /** Callback to be executed when current page is changed. */
+  onChange: PropTypes.func.isRequired,
+
+  /** The number of total pages. */
+  totalPages: PropTypes.number.isRequired,
+};
+
+const defaultProps = {
+  pagesOnEachSide: "medium",
```
allison-c, line 24: So should this be a string or number? In the propTypes definition above you're having `pagesOnEachSide: PropTypes.number,`

packages/Pagination/src/components/PageItem/PageItem.styles.js
```diff
@@ -0,0 +1,27 @@
...
+    background: ${tokens.color.blackLighten70};
+  }
+`;
+
+export const Wrapper = styled.span`
+  border-right: 1px solid ${tokens.border.color};
+  margin: 0 ${tokens.spaceSm};
+  padding-right: ${stylers.spacer(1)};
+  position: relative;
+`;
```
allison-c, line 27: Since `Wrapper` is always same for all small components, we can extract it into a `Wrapper.styles.js` instead of repeating it many times

PR #648, opened 2020-08-17
UXD-131 Update lastKnownSelectedOptions when options are changed

allison-c:
### Purpose 🚀

`lastKnownSelectedOptions` wasn't updated when options are changed on the consumers' side. 
Even though `useChildrenChange` will trigger `updateOptions` action when options are updated, but in the reducer function, the action only update `options` and `selectedOptions` before. We should update `lastKnownSelectedOptions` too, otherwise the data are messed up in the internal state of listbox.

### Notes ✏️

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


nahumzs:
make sense thank you @allison-c

PR #655, opened 2020-08-18
UXD-110-Table-Spec Test

KDarcey:
### Purpose 🚀
To test the Table component.

### Notes ✏️
- spec file added.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-110-Table-spec

### Screenshots 📸
N/A

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-98


packages/Table/tests/spec/Table.spec.js
```diff
@@ -0,0 +1,56 @@
...
+      <Table.ColumnDefinition header="LastName" cell="lastName" />
+    </Table>
+  );
+}
+
+describe("Table", () => {
+  it("Should render data in the table component", () => {
+    const { getByText } = renderComponent();
+
+    expect(getByText("Name")).toBeVisible();
```
nahumzs, line 31: we can do the same than with cypress
```js
["name1", "name2", "..."].forEach(name => {
      expect(getByText(name)).toBeVisible();
})
```
we are duplicating testing here and with cypress, there is no harm in adding more test, but we should be aware if we are duplicating effort, it's ok for now :)

PR #659, opened 2020-08-20
UXD-155 Adding yarn readme into build process

allison-c:
### Purpose 🚀

Adding yarn readme into build process.

### Notes ✏️

Tested locally, `yarn readme` will be executed after lerna changing the verison in each `package.json` files. Also, the readme files should be committed together with the master push.

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


nahumzs:
do you know if there is a way to ignore `yarn readme` if for some reason failed to run the script, so the entire publish process doesn't stop just because the readme script failed?

allison-c:
> do you know if there is a way to ignore `yarn readme` if for some reason failed to run the script, so the entire publish process doesn't stop just because the readme script failed?

I think we can just simply add `"readme": "node scripts/readmeFiles.js || exit 0",`, but if that's the case, there's no point to add `yarn readme` into build process. If `yarn readme` failed and we still publish to npm, those readme files would be wrong or not updated.

nahumzs:
second thought, I think will be fine, because if fails will fail during the transpilation process and none step will be taken of the release process. we are good! :D

PR #663, opened 2020-08-21
Ux-157 FormElement api simplification

tristanjasper:
### Purpose 🚀
Proposal for FormElement api simplification so accessibility is handled by the component itself rather than the consumer.

Previously a render function was required as a child of `FormElement.Content` in order to expose accessibility tags `idForLabel, ariaDescribedBy` and `refLabel` which were not obviously available to the developer or how to use them, without looking into the details and storybook examples as each paprika component uses those tags differently.

This PR now takes care this by applying the accessibility tags automatically no matter what component you pass it as a child. This means it is less developer effort to work with. 

It is still expected that any other aria-labels such as `aria-required` should still be supplied by the consumer.

**When passing more than one `Input` or `input` element label focus**

It is expected that in a follow up PR a follow up Api change to allow overriding the behaviour of the default formElement labelId being applied to the first Input. 

Additional changes:
Added support for TimePicker with formElement

### Notes ✏️

### Updates 📦
- [x] MAJOR (breaking) change to formElement
- [X] PATCH (bug fix) change to TimePicker, Checkbox, DatePicker

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ux-157-form-element-simplification

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-157


tristanjasper:
> Did you explore to have a wrapper for transform something like this:
> 
> ```js
>       <FormElement label="Form Label">
>         <FormElement.Instructions>
>           <span>
>             Example text for extra panel for questionnaires. Example text for extra panel for questionnaires Example
>             text for extra panel for questionnaires. Example text for extra panel for questionnaires
>           </span>
>         </FormElement.Instructions>
>         <FormElement.Content>
>           <Input
>             onChange={handleChange}
>             value={value}
>             placeholder="Form placeholder"
>             hasError={Boolean(errorText.length)}
>           />
>         </FormElement.Content>
>         <FormElement.Description>
>           <span>This is description text</span>
>         </FormElement.Description>
>         <FormElement.Error>{errorText}</FormElement.Error>
>         <FormElement.Help>
>           Give me some help. <a href="wegalvanize.com">Learn more</a>.
>         </FormElement.Help>
>       </FormElement>
> ```
> 
> into:
> 
> ```js
> <FormElementShort 
>   label="Form Label"
>   instructions="..."
>   description="..."
>   error="..."
>   help="..."
> >
>   <Input
>      onChange={handleChange}
>      value={value}
>      placeholder="Form placeholder"
>      hasError={Boolean(errorText.length)}
>   />
> </FormElementShort>
> ```
> 
> ^ the previous wrapper could be an addition to our current API, is short and in the background will be rendering our current FormElement API

No I didn't look into that, This approach is new to me, I've never heard of adding additional shortcut components in any discussions. Perhaps elaborate on the reasons for it so we can have an open discussion? If we like it perhaps there are other components that could benefit from a similar shortcut wrapper approach? For me it appears to be adding non essential components though.

packages/FormElement/src/components/Content/Content.js
```diff
@@ -10,24 +11,132 @@ const propTypes = {
...
+  }
 
   if (!children) {
     return null;
   }
 
   return (
     <sc.ContentContainer data-pka-anchor="form-element.content" {...moreProps}>
-      {typeof children === "function" ? children({ idForLabel, refLabel, ariaDescribedBy }) : children}
+      {renderExtractedChildren()}
```
nahumzs, line 139: add a console.error or throw new Error if `typeof children === "function" `

packages/FormElement/src/components/Content/renderFunctions.js
```diff
@@ -0,0 +1,63 @@
...
+    id: idForLabel,
+  });
+
+const getAccessibleElement = (elements, cloneFn) =>
+  Array.isArray(elements) ? elements.map(cloneFn) : cloneFn(elements);
+
+export default {
+  Input: ({ extractedElements, idForLabel, ariaDescribedBy }) =>
+    getAccessibleElement(extractedElements, (element, index = 0) =>
+      cloneWithAriaAttributes({ element, idForLabel: index === 0 ? idForLabel : null, ariaDescribedBy })
```
tristanjasper, line 15: This behaviour of the first Input being provided the `id` (that is associated with the formElement label) will be able to be overridden in a follow up pr

Also on line https://github.com/acl-services/paprika/pull/663/files#diff-1a53d8778ef9b8807972d1aa4463c8c9R42

packages/FormElement/CHANGELOG.md
```diff
@@ -49,3 +49,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 -- Changed the spacing between label, content and instructions
 
 [@tristanjasper](https://github.com/tristanjasper).
+
+## [0.3.29] - 2020-08-20
```
nahumzs, line 53: this you should follow the new format that @allison-c  introduced, write this on the [unreleased] section read in our slack channel in paprika the announcement about it 🙏  :)

packages/FormElement/README.md
```diff
@@ -94,20 +95,16 @@ import Input from "@paprika/input";
...
-        placeholder="Form placeholder"
-        aria-describedby={ariaDescribedBy}
-        aria-required={hasRequiredLabel}
-        hasError={false}
-        isDisabled={isDisabled}
-        isReadOnly={isReadOnly}
-        size={size}
-      />
-    )}
+    <Input
```
nahumzs, line 98: is this a breaking change or our FormElements still support the renderFunction?

packages/FormElement/src/components/Content/Content.js
```diff
@@ -10,24 +12,54 @@ const propTypes = {
...
   ariaDescribedBy: PropTypes.string,
+  /** Used when form Elements are nested and is automatically appended to aria-describedby */
+  wrapperAriaDescribedBy: PropTypes.string,
 };
 
 const defaultProps = {
   idForLabel: null,
   ariaDescribedBy: null,
   refLabel: null,
+  wrapperAriaDescribedBy: null,
```
nahumzs, line 23: why do we have ariaDescribedBy and wrapperAriaDescribedBy, shouldn't this be part of a more universal prop named a11yText (this is what we use across our library)? 

Do you have an example where we are using `wrapperAriaDescribedBy`?

tristanjasper: You can see how we use it in this storybook example - https://github.com/acl-services/paprika/pull/663/files#diff-1a53d8778ef9b8807972d1aa4463c8c9R54

In the case of wrapping FormElements with other FormElements we need to make sure that the `aria-describedBy` strings coming from the wrapping FormElement are being passed down to the children 

Perhaps a better name for these props could be `a11yDescribedByIds` so I'll switch to that

alexzherdev:
Critiquing without offering an alternative is unfair but I'll try to offer at least something at the end 😄 I want to throw in a general idea: I think it's commendable to try to ease the burden for consumers, but this should always be weighed against complexity or potential for issues within the library. If the library's code becomes too convoluted or otherwise hard to maintain, we're ultimately doing a disservice to the consumers.

Particularly in this case, one thing I'm seeing is we're essentially making `Content` depend (in the npm/package sense) on all the different form components (input, checkbox, etc.). It's subtle, but the code that knows how to properly clone all the different kinds of components assumes a certain API (especially `children`) of those components. This is most obvious with `DatePicker`: the cloning code used by `Content` _knows_ that `DatePicker` has a single child which is the DatePicker's "input". If we change that API in a major bump for `DatePicker`, `Content` will break when used in conjunction with that new version.

This isn't a normal dependency though since `FormElement` doesn't import or compose the form components. It would be a peer dependency, because it's really saying hey, this version of me works in the presence of these versions of all those other components. It would actually be an [optional peer dependency](https://github.com/yarnpkg/rfcs/pull/105) since using `FormElement` doesn't _require_ each of the form components to be present. But honestly having any such sort of dependency seems weird.

I don't see a perfect way out, I feel like we may have painted ourselves into a corner introducing this generic `FormElement` component. The best thing I can see would be for us to rename the properties passed into the render prop so that they can be spread onto the child component without destructuring like so:
```jsx
<FormElement.Content>
  {a11yProps => (
    <Input
      {...a11yProps}
      {...otherProps}
    />
  )}
</FormElement.Content>
```
where `a11yProps` is `{ id: "someId", "aria-describedby": "something else" }`. Just makes it a bit less tedious.

tristanjasper:
> Critiquing without offering an alternative is unfair but I'll try to offer at least something at the end 😄 I want to throw in a general idea: I think it's commendable to try to ease the burden for consumers, but this should always be weighed against complexity or potential for issues within the library. If the library's code becomes too convoluted or otherwise hard to maintain, we're ultimately doing a disservice to the consumers.
> 
> Particularly in this case, one thing I'm seeing is we're essentially making `Content` depend (in the npm/package sense) on all the different form components (input, checkbox, etc.). It's subtle, but the code that knows how to properly clone all the different kinds of components assumes a certain API (especially `children`) of those components. This is most obvious with `DatePicker`: the cloning code used by `Content` _knows_ that `DatePicker` has a single child which is the DatePicker's "input". If we change that API in a major bump for `DatePicker`, `Content` will break when used in conjunction with that new version.
> 
> This isn't a normal dependency though since `FormElement` doesn't import or compose the form components. It would be a peer dependency, because it's really saying hey, this version of me works in the presence of these versions of all those other components. It would actually be an [optional peer dependency](https://github.com/yarnpkg/rfcs/pull/105) since using `FormElement` doesn't _require_ each of the form components to be present. But honestly having any such sort of dependency seems weird.
> 
> I don't see a perfect way out, I feel like we may have painted ourselves into a corner introducing this generic `FormElement` component. The best thing I can see would be for us to rename the properties passed into the render prop so that they can be spread onto the child component without destructuring like so:
> 
> ```js-jsx
> <FormElement.Content>
>   {a11yProps => (
>     <Input
>       {...a11yProps}
>       {...otherProps}
>     />
>   )}
> </FormElement.Content>
> ```
> 
> where `a11yProps` is `{ id: "someId", "aria-describedby": "something else" }`. Just makes it a bit less tedious.

Yeah, the reason the api is the way it is currently is exactly for that reason (we didn't want to tie the accessibility implementation too tightly to the components using it ) I see positives and negatives from both approaches.

Potentially that solution could be a good option as well as the proposal I have in this pr @alexzherdev. I think that could keep the separation of components, and make it a little bit easier for consumers to implement.  

There are some issues there though, in that not in all cases do be want to spread all of the `a11yProps` onto all children, but we could use that as a generic usage approach I think...

tristanjasper:
Closing this pr as want to reconsider the approach for api simplification.

packages/FormElement/src/components/Content/Content.js
```diff
@@ -9,25 +10,144 @@ const propTypes = {
...
+    if (!isLabelSet) {
+      const singleIdForLabel =
+        componentsRequiringIdForLabel.includes(key) || isOtherChildRequiringId(key, element) ? idForLabel : null;
+      isLabelSet = Boolean(singleIdForLabel);
+      return singleIdForLabel;
+    }
+  };
+
+  if (!extractedChildren?.children.length) {
+    delete extractedChildren.children;
```
tristanjasper, line 133: extractedChildrenHelper generates an empty children array when not needed, perhaps something we should prevent from happening

tristanjasper:
@alexzherdev Following on from your concerns in comment - https://github.com/acl-services/paprika/pull/663#issuecomment-681120673

I've modified certain paprika components (DatePicker, TimePicker) to expect a consistent application of `a11yDescribedByIds`, therefore if we do make changes to those components in future this shouldn't be too much of an issue of it being too tightly coupled to the formElements approach.

packages/FormElement/README.md
tristanjasper: This will be a breaking change. I don't think we should encourage using the renderFunction as an acceptable api moving forward

packages/TimePicker/src/TimePicker.js
```diff
@@ -149,20 +166,21 @@ function TimePicker(props) {
 
   const { t } = useI18n();
   return (
-    <sc.TimePicker onFocus={handleFocus} onBlur={handleBlur}>
+    <sc.TimePicker onFocus={handleFocus} onBlur={handleBlur} {...moreProps}>
```
tristanjasper, line 169: spreading `moreProps` here was the original intention so updated this

packages/FormElement/stories/examples/AlternateLayouts.js
```diff
@@ -331,20 +332,57 @@ const ExampleStory = () => {
...
+          <Input
+            onChange={handleChange}
+            value={value}
+            placeholder="Form placeholder"
+            aria-required={hasRequiredLabel}
+            hasError={Boolean(errorText.length)}
+            isDisabled={isDisabled}
+            isReadOnly={isReadOnly}
+            size={size}
+            overridelabelassociation="true" // Add to override default label association
```
tristanjasper, line 384: this is a proposal for how we can override which element is associated with the label, perhaps wrapping with a `FormElement.AssociateLabel` component could be another approach, or any better proposals?

nahumzs:
I think we have hit a point with this Component where is starting to be clear that our current approach doesn't escalate correctly neither play well for our consumer as being easier to use or an easy good pattern to compose our Form requirements. Currently is  verbose and keep going in this direction will add unnecessary complexity to the FormElement. 

In the previous attempt we discussed about this propose approach and agree wiring our custom inputs component to this FormElement will make it easy to break it and hard to maintain. 
And for the previous reason the render function was introduce, but that created a more "verbose" use of the component where you have to write multiple lines just to be able to use a `simple input`

Since the idea is to simplify this component, maybe would worth to take a step back approach and make a table of the requirements we need to support with this component and rethink what would be the best solutions for this. 

I think pushing this component to try to do what we want in their current form might not the best decision. 

Maybe is time to check https://github.com/adobe/react-spectrum/tree/main/packages/%40react-spectrum and see if we can adopt it or get inspiration how they solve this issue.

Another possible solution would be totally detach every piece of this component and provide helpers that generates ids to make easier to compose these elements.

like:

```js
import ListBox from "@paprika/listbox"
import Input from "@paprika/listbox"
import { Label, Description, Help, Error, etc, Elements, useFormElement } from "@paprika/formElement"

const { generateA11yProps, isVisuallyHiden } = useFormElement();

const inputA11yProps = generateA11yProps();
const listBoxA11yProps = listBoxA11yProps();

<Element.Group kind={Element.types.kind.horizontal}>
   <Element>
     <Label {...inputA11yProps.label}>Name</Label>
     <Input {....inputA11yProps.input} />
   </Element>
   <Element>
     <Label {...listBoxA11yProps} isVisuallyHidden>preference</Label>
     <ListBox {...stuff} />
     <Description>Some description</Description>
    <Element />
</Elements>
<Element.Group>
```

alexzherdev:
I generally agree with the above. What we're trying to do is the opposite of what's recommended here https://www.youtube.com/watch?v=AiJ8tRRH0f8 (which I agree with)

tristanjasper:
Ok I'll look into the above route

PR #664, opened 2020-08-21
UXD-153-constants-refactor

KDarcey:
### Purpose 🚀
Refactor all our components and unify the way we interact with constants in all our components

### Notes ✏️
- created constants package
- added `type.js` files for all relevant components.

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-153-constants-refactor

### Screenshots 📸
N/A

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-150
https://aclgrc.atlassian.net/browse/UXD-153


packages/Button/src/Button.js
```diff
@@ -194,7 +192,7 @@ const Button = React.forwardRef((props, ref) => {
 ButtonIcon.propTypes = buttonPropTypes;
 ButtonIcon.defaultProps = buttonDefaultProps;
 
-Button.Kinds = Kinds;
+Button.types = types;
```
nahumzs, line 195:
```js
Button.types = {
   kind: constant.kind,
   size: constant.size
}
```

packages/Button/src/Button.js
```diff
@@ -45,7 +43,7 @@ const propTypes = {
   isSubmit: PropTypes.bool,
 
   /** The visual style of the button. */
-  kind: PropTypes.oneOf(Kinds.ALL),
+  kind: PropTypes.oneOf(types.ALL),
```
nahumzs, line 46: does this render correctly on the readme file, types.ALL is a microabstractio maybe would worth to get ride off

packages/Constants/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@paprika/constants",
+  "version": "0.1.1",
```
allison-c, line 3: why the initial version is `0.1.1` ?

packages/Constants/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "directory": "packages/constants"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "peerDependencies": {
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0"
+  }
+}
```
allison-c, line 19: missing `@babel/runtime-corejs2` ?

packages/Avatar/README.md
```diff
@@ -22,12 +22,12 @@ npm install @paprika/avatar
...
-| backgroundColor | string             | false    | null     | Background color of the Avatar                        |
-| color           | string             | false    | null     | Color for the initial or icon                         |
-| size            | ShirtSizes.LIMITED | false    | "medium" | Size of Avatar                                        |
+| Prop            | Type                       | required | default      | Description                                           |
+| --------------- | -------------------------- | -------- | ------------ | ----------------------------------------------------- |
+| children        | node                       | false    | null         | Avatar content. It can be initial as a string or icon |
+| backgroundColor | string                     | false    | null         | Background color of the Avatar                        |
+| color           | string                     | false    | null         | Color for the initial or icon                         |
+| size            | [types.SMALL,types.MEDIUM] | false    | types.MEDIUM | Size of Avatar                                        |
```
allison-c, line 30: If we want to render `[Avatar.types.size.SMALL,Avatar.types.size.MEDIUM]` in the README.md, we'll need to use `size: PropTypes.oneOf([Avatar.types.size.SMALL, Avatar.types.size.MEDIUM]),` in propTypes definition. e.g. here: https://github.com/acl-services/paprika/pull/664/files#diff-1295811f8a1709758fc99b9871b2aba9R15
However since we usually put the propTypes definition at the beginning of files, if we make this change, there'll be some lint errors. Maybe need yours input @nahumzs 🙇‍♀️

nahumzs: yeah, I thought about that and then forgot about it, well I think with this PR and the one of Oscar and Karen we will have a way for avoiding to come to the source file to read the props. So theoretically we could move the props definition from the top to the bottom, or just relay on javascript and knowing that the compiler will move all those declaration to the top automatically and ignore the linter. (which I think is the easier one and really not harmful) 

What do you think?

allison-c: I'd prefer the second way too, combined with `/* eslint-disable no-use-before-define */` and `/* eslint-enable no-use-before-define */`

packages/ButtonGroup/stories/examples/Sandbox.js
```diff
@@ -1,12 +1,13 @@
 /* eslint-disable react/button-has-type */
 
 import React from "react";
+import * as constants from "@paprika/constants/lib/Constants";
```
alexzherdev, line 4: This here is a problem. It's fine to create an internally shared package to centralize those constants, but if the consumers need to install this package into their apps to access constants, we're signing up for version mismatches in the future as the package gets updates.
I _think_ the alternative would be to re-export so that they'd access e.g. `ButtonGroup.kinds.PRIMARY`

nahumzs: yeap that's the idea @alexzherdev , the constant package will hold only those constants that are "universal" and we are re-exporting all those via the component, so you could do something like:

```js
import Button from "@paprika/button"
function App() {
  const { kind, size } = Button.types
  return <Button kind={kinds.primary} size={size.medium} />
}

```

The constant package is just for internal/author use and we should write a disclaimer about that in the Readme file.

packages/Button/src/Button.styles.js
```diff
@@ -15,23 +14,23 @@ const enabled = content => css`
...
-  [Kinds.FLAT]: tokens.border.hoverColor,
-  [Kinds.MINOR]: "transparent",
-  [Kinds.LINK]: "transparent",
+  [types.DEFAULT]: tokens.border.hoverColor,
+  [types.PRIMARY]: tokens.color.greenDarken10,
+  [types.SECONDARY]: tokens.color.purpleDarken10,
+  [types.DESTRUCTIVE]: tokens.color.orangeDarken10,
+  [types.FLAT]: tokens.border.hoverColor,
+  [types.MINOR]: "transparent",
+  [types.LINK]: "transparent",
```
allison-c, line 33: Just wondering why not using `Button.types.kind.LINK` here? I saw inside side panel component you have `[Button.types.kind.PRIMARY]` in `styles.js` .  Maybe you don't need to change them right away but I think we should revisit and make them consistent.

allison-c:
I quickly looked through the changes, and checked all the stories in storybook and accepted changes in screener. Found some code styling issues, but just minor. What do you think @nahumzs

allison-c:
Sorry I saw many propType warnings when running this branch locally, could we solve them?

allison-c:
Found another issue 😅 , I think we have to export `types` again in `index.js` for example in `Button/src/index.js` we can do
```js
import * as constants from "@paprika/constants/lib/Constants";
import Button from "./Button";
import CloseButton from "./CloseButton";
import IconButton from "./IconButton";
import LinkButton from "./LinkButton";

Button.Close = CloseButton;
Button.Icon = IconButton;
Button.Link = LinkButton;
Button.types = {
  size: constants.defaultSize,
  kind: constants.kind,
};

export { Button as default };
```

Only by doing this, autocomplete can work.

packages/ActionBar/README.md
```diff
@@ -40,19 +40,19 @@ npm install @paprika/action-bar
...
+| columns          | arrayOf                                              | true     | -                               |             |
+| onAddFilter      | func                                                 | true     | -                               |             |
+| onApply          | func                                                 | true     | -                               |             |
+| onCancel         | func                                                 | false    | () => {}                        |             |
+| onChangeOperator | func                                                 | false    | null                            |             |
+| onClose          | func                                                 | false    | () => {}                        |             |
+| onOpen           | func                                                 | false    | () => {}                        |             |
+| operator         | [Filter.types.operator.AND,Filter.types.operator.OR] | false    | Filter.types.AND                |             |
```
nahumzs, line 54: why do these defaults are under Filter.types.AND instead of Filter.types.operator.AND its this expected?

packages/ActionBar/README.md
```diff
@@ -40,19 +40,19 @@ npm install @paprika/action-bar
...
+| rulesByType      | objectOf                                             | false    | Filter.types.defaultRulesByType |             |
```
nahumzs, line 55: same with this one `defaultRulesByType` is under types directly

packages/ActionBar/README.md
KDarcey: I think I must have missed it, thank you!

allison-c:
> Found another issue 😅 , I think we have to export `types` again in `index.js` for example in `Button/src/index.js` we can do
> 
> ```js
> import * as constants from "@paprika/constants/lib/Constants";
> import Button from "./Button";
> import CloseButton from "./CloseButton";
> import IconButton from "./IconButton";
> import LinkButton from "./LinkButton";
> 
> Button.Close = CloseButton;
> Button.Icon = IconButton;
> Button.Link = LinkButton;
> Button.types = {
>   size: constants.defaultSize,
>   kind: constants.kind,
> };
> 
> export { Button as default };
> ```
> 
> Only by doing this, autocomplete can work.

We decided do this in another PR

packages/CollapsibleText/src/CollapsibleText.js
```diff
@@ -81,7 +81,7 @@ function CollapsibleText(props) {
             aria-controls={contentId}
             aria-expanded={!isCollapsed}
             isSemantic={false}
```
nahumzs, line 83: This package lacks of changelog, don't remember the CollapsibleText component 🤔 why do we have a collapsible only for text? (rethorical question)

packages/Constants/README.md
```diff
@@ -0,0 +1,4 @@
+<!-- autogenerated don't modify -->
```
nahumzs, line 1: Please add content to the `CHANGELOG.md` file for the constant package 🙏

nahumzs, line 1: add explanation and disclaimer about the package is for internal use exclusively and to access the constants of a component they can do it over `Component.types`

packages/Constants/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@paprika/constants",
+  "version": "0.1.0",
+  "description": "Constants for components",
```
nahumzs, line 4: can you add in the description something like for library internal use only?

packages/Constants/src/Constants.js
```diff
@@ -0,0 +1,61 @@
...
+  MEDIUM: "medium",
+  LARGE: "large",
+};
+
+export const limitedSize = {
+  SMALL: "small",
+  MEDIUM: "medium",
+};
+
+export const gridTypes = {
```
nahumzs, line 56: ohh this are here because the DataGrid and the Table use them correct?

scripts/readmeFiles.js
```diff
@@ -5,7 +5,7 @@ const shell = require("shelljs");
 // eslint-disable-next-line import/no-extraneous-dependencies
 const reactDocs = require("react-docgen");
 
-const skipPackages = ["helpers", "Stylers", "Tokens", "Guard", "Icon", "Overlay", "Select", "Calendar"];
+const skipPackages = ["helpers", "Stylers", "Tokens", "Guard", "Icon", "Overlay", "Select", "Calendar", "Constants"];
```
nahumzs, line 8: @kcumlat this might get conflict with the rename of the script in your PR you might want to add it manually so don't have this issue.

packages/Constants/src/Constants.js
KDarcey: Correct!

PR #666, opened 2020-08-26
UXD-159 Create declaration files

kcumlat:
### Purpose 🚀
Add script to create a declaration `d.ts` file for every package

### Notes ✏️
- Declaration files for skipPackages. _Need to investigate more..._
- Check props values in `.d.ts` e.g. `ShirtSizes, KINDS`
_Some propTypes such are instanceOf, objectof, shape and custom uses `any` type_
### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


scripts/typesFiles.js
```diff
@@ -0,0 +1,153 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+const shell = require("shelljs");
+// eslint-disable-next-line import/no-extraneous-dependencies
+const reactDocs = require("react-docgen");
+
+const skipPackages = [
+  "Guard",
+  "Icon",
+  "Stylers",
+  "Tokens",
```
nahumzs, line 12: how Tokens has a declaration file but is skip it here? does the `tsc` script create for us? if so, then can we differentiated which packages are skip that we can't process by now, and which are those which are skip because `tsc` process via the `tscofing`?

package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "eslint packages --report-unused-disable-directives",
     "lint:jss": "stylelint './packages/**/src/**/*.styles.js'",
     "postpublish": "git add . && git commit -m 'Publish: automate readme and changelog update' --no-verify && git push origin master -f --no-verify",
-    "prepare": "yarn cleanup && lerna run pretranspile && yarn transpile && yarn readme",
+    "prepare": "yarn cleanup && lerna run pretranspile && yarn transpile && yarn readme && yarn types",
```
nahumzs, line 16: I gonna piggyback while you are here, can you unified the name convention for the reader, changelog and the types generation? 

maybe generateReadME, generateTypeDefinitions and generateChangeLOG? 🙏

scripts/typesFiles.js
kcumlat: Yes, `token.d.ts` was processed using `tsc` so it was added to the the skipPackages. Will look into a better way to differentiate skipPackages and files created via `tsc`. 👍

tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
```
nahumzs, line 1: are we using this declaration?

scripts/typesFiles.js
```diff
@@ -0,0 +1,153 @@
...
+const renderDeclarationTemplate = ({ displayName="", props = "" }) => { return `export default ${displayName};
+${props}`;};
+
+const createPropsList = ({ info }) => {
+  if (!info || !info.props) return "";
+  const subComponent = info.displayName.substring(info.displayName.indexOf(".") + 1, info.displayName.length);
+  const displayName = info.displayName.includes(".") ? subComponent : info.displayName;
+  /* prettier-ignore */
+  const list = [`
+declare function ${displayName}(props: any): JSX.Element;
```
nahumzs, line 42: if we already know which are the props we can declare the props instead of passing any:
using this:
```js
  Object.keys(info.props).map(key => {
```

the template would look like:
```js
interface IPRops {
    kind: "primary" | "secondary",
    size?: "medium" | "large",
}
```

we can even add the `?` if the prop is optional or without it if it's required. 

Maybe this is a stretch and the next step, but this could improve the developer experience.

tsconfig.json
kcumlat: was use to generate .d.ts file for `tokens`. I guess if we're only using `tsc` for tokens or any package, we can include only the specific folder to process, to be clear...

scripts/typesFiles.js
kcumlat: `?` for optional props has been added, but still need to check some of the  generated props value in the `d.t.s` file

packages/Button/index.d.ts
```diff
@@ -0,0 +1,37 @@
...
+  /** If the width of the button should span it's parent container (100%). */
+  isFullWidth?: boolean;
+  /** If the button should render in a pending state (with a spinner icon). */
+  isPending?: boolean;
+  /** If it will be rendered as a <button> element. If false, a <span> will be rendered via an accessible <RawButton>. */
+  isSemantic?: boolean;
+  /** If the type attribute should "submit", instead of the default "button". */
+  isSubmit?: boolean;
+  /** The visual style of the button. */
+  kind?: Kinds.ALL;
```
nahumzs, line 28: no for this PR but, after we merge the constant package we should be able to provide a better descriptive way for this kind of prop.

nahumzs:
this is great @karencumlat and @oscarkwan 💯 👏

nahumzs:
Please hold for #664 before merge this 🙏

nahumzs (approved):

PR #667, opened 2020-08-26
add default font styles on th

tristanjasper:
### Purpose 🚀
Add default font styling on th elements in Table Component

As discussed - https://github.com/acl-services/acl-exception/pull/7873/files#diff-ea428be058ad84e7b18651badfc2b143R7

### Notes ✏️
No visible changes to the component in storybook as the styles were already coming from other style sources. In a consuming application however that may not be the case

### Updates 📦
- [x] PATCH (bug fix) change to Table

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-162-Table-th-default-styling

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-162

### DIFF
--- a/packages/Table/src/Table.styles.js
+++ b/packages/Table/src/Table.styles.js
@@ -1,5 +1,6 @@
 import styled, { css } from "styled-components";
 import tokens from "@paprika/tokens";
+import { fontSize } from "@paprika/stylers/lib/helpers";
 import types from "./types";
 
 export const Table = styled.table`
@@ -58,9 +59,9 @@ export const TD = styled.td(({ borderType }) => {
 
 export const TH = styled.th(({ borderType }) => {
   return css`
+    ${fontSize()}
     ${borderType in borderTypesStyles ? borderTypesStyles[borderType] : ""}
-    font-size: 16px;
-    font-weight: 600;
+    font-weight: bold;
     padding: ${tokens.space};
     text-align: left;
   `;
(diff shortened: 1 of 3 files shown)


packages/Table/src/Table.styles.js
```diff
@@ -59,6 +59,8 @@ export const TD = styled.td(({ borderType }) => {
 export const TH = styled.th(({ borderType }) => {
   return css`
     ${borderType in borderTypesStyles ? borderTypesStyles[borderType] : ""}
+    font-size: 16px;
```
oscarkwan, line 62: we can use the font scale here

```
  ${stylers.fontSize(0)};
```

packages/Table/src/Table.styles.js
```diff
@@ -59,6 +59,8 @@ export const TD = styled.td(({ borderType }) => {
...
+    font-weight: 600;
```
oscarkwan, line 63: Do we use font-weight 600? I usually see 500/700, or just bold

packages/Table/src/Table.styles.js
tristanjasper: Actually would need to use `helpers.fontScale()` here as it's a Js file. In general I'm seeing quite a few components neglecting to use this so figured we were not mandating it but I think it's a good idea as well.

packages/Table/src/Table.styles.js
tristanjasper: Datagrid seems to use `600` so I was following suit, but `bold` makes sense to me as well.

PR #677, opened 2020-09-01
UXD-160 Style: refactor FormElement

KDarcey:
### Purpose 🚀
Refactor styling for FormElement

### Notes ✏️
N/A

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-160-refactor-styling-FormElement

### Screenshots 📸
N/A

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-160


packages/FormElement/src/FormElement.styles.js
```diff
@@ -7,19 +7,23 @@ const FontSizes = {
...
 };
 
-export const SectionsContainer = styled.div`
-  ${({ isInline }) => isInline && `flex-grow: 1;`}
-`;
+export const SectionsContainer = styled.div(
+  ({ isInline }) => css`
+    ${isInline && `flex-grow: 1;`}
+  `
+);
```
allison-c, line 14: There's no point adding `css` here. If you really want it you might want to add it like this
```
export const SectionsContainer = styled.div`
  ${({ isInline }) => isInline && css`flex-grow: 1;`}
`;
```

packages/FormElement/src/FormElement.styles.js
```diff
@@ -8,18 +8,24 @@ const FontSizes = {
 };
 
 export const SectionsContainer = styled.div`
```
nahumzs, line 10: why are yo not following the function format? I think this should be also changed isn't'?

nahumzs (approved):

PR #720, opened 2020-09-23
Tw 670 button link default styling resilient to global styles

tristanjasper:
### Purpose 🚀
Ensure `button.Links` with default button styling are not having their `hover` or `visited` states overridden by common global styles in consuming applications. 

Currently this covers any `a` tag hover, visited styles. 

As required to fix https://aclgrc.atlassian.net/browse/TW-670

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to Button

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/TW-670-Button-link-default-styling

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/TW-670

### DIFF
--- a/packages/Button/src/Button.styles.js
+++ b/packages/Button/src/Button.styles.js
@@ -182,13 +182,14 @@ export const sizeStyles = {
 export const kindStyles = ({ isDisabled }) => ({
   [types.DEFAULT]: css`
     ${skeuomorphicStyles}
-
     background-color: ${tokens.color.white};
     background-image: linear-gradient(${tokens.color.blackLighten90}, ${tokens.color.blackLighten70});
     color: ${tokens.color.black};
 
-    &:hover {
+    &:hover, &:visited {
       background: ${tokens.color.blackLighten70};
+      color: ${tokens.color.black};
+      text-decoration: none;
     }
 
     ${isDisabled && disabledStyles}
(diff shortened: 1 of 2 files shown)


packages/Button/src/Button.styles.js
```diff
@@ -182,13 +182,14 @@ export const sizeStyles = {
 export const kindStyles = ({ isDisabled }) => ({
   [types.DEFAULT]: css`
     ${skeuomorphicStyles}
-
     background-color: ${tokens.color.white};
     background-image: linear-gradient(${tokens.color.blackLighten90}, ${tokens.color.blackLighten70});
     color: ${tokens.color.black};
 
-    &:hover {
+    &:hover, &:visited {
```
mikrotron, line 189: I don't think you want `visited` styles to have the grey background... that one should be _only_ for `hover`.

`background-color`, `background-image`, `color`, and `text-decoration` could all be applied once in a block for `&, &:hover, &:visited, &:active` as long as there was another `&:hover` block _after_ that with a `background` override.

mikrotron:
For consistency, should all `kinds`, not just `DEFAULT` be given this same treatment?

tristanjasper:
> For consistency, should all `kinds`, not just `DEFAULT` be given this same treatment?

I tested with the other button types and there were no other issues with `a` colour or text decoration issues which is what this pr was specifically covering. 

We can go further and cover other styles that should be set for other scenarios though so I'll look at that as well.

mikrotron (approved):
llgtm 80

PR #722, opened 2020-09-24
UXD-75 Add notification card component

allison-c:
### Purpose 🚀

### Notes ✏️
Example: http://storybooks.highbond-s3.com/paprika/UXD-75-Notification-card-component/?path=/story/messaging-notificationcard--showcase

- [x] Component
- [x] Screener tests
- [x] Jest tests
- [x] mdx

Note: the generated doc shows `align` prop has default value `constants.align.TOP`. Ideally, it should be `Notification.Image.types.align.TOP`. I've looked into a fix, but not feasible to do it in this PR. Will create a paprika ticket for it.

image

### NotificationCard

| Prop     | Type   | required | default | Description                                 |
| -------- | ------ | -------- | ------- | ------------------------------------------- |
| children | node   | false    | null    |                                             |
| maxWidth | string | false    | null    | Dictates the maximum width of the component |

### NotificationCard.Image

| Prop     | Type                                           | required | default             | Description                      |
| -------- | ---------------------------------------------- | -------- | ------------------- | -------------------------------- |
| align    | [ constants.align.TOP, constants.align.CENTER] | false    | constants.align.TOP | Aligns the position of the image |
| children | node                                           | false    | null                |                                  |

### NotificationCard.Header

| Prop     | Type                | required | default | Description                     |
| -------- | ------------------- | -------- | ------- | ------------------------------- |
| children | node                | false    | null    |                                 |
| level    | [ 1, 2, 3, 4, 5, 6] | true     | -       | Heading level(1-6) is required. |

### NotificationCard.Body

| children | node | false    | null    |             |

### NotificationCard.Footer

| children | node | false    | null    |             |

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/notification-card`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-75-Notification-card-component

### Screenshots 📸

### References 🔗


packages/NotificationCard/src/components/Header/Header.js
```diff
@@ -0,0 +1,24 @@
...
+
+function NotificationHeader({ children, ...moreProps }) {
+  return <sc.NotificationHeader {...moreProps}>{children}</sc.NotificationHeader>;
+}
+
+const propTypes = {
+  children: PropTypes.node,
+
+  /** Heading level(1-6) is required. */
+  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]).isRequired,
```
nahumzs, line 13: why are required? can you not set one as default?

packages/NotificationCard/stories/NotificationCard.stories.js
```diff
@@ -0,0 +1,13 @@
+import { withKnobs } from "@storybook/addon-knobs";
+import { storiesOf } from "@storybook/react";
+import { getStoryName } from "storybook/storyTree";
+import Showcase from "./examples/Showcase";
+import Screener from "./examples/Screener";
+
+const storyName = getStoryName("NotificationCard");
+
+storiesOf(storyName, module)
```
mikrotron, line 9: I won't insist on it for this PR, but I think we should be writing stories for new components in the new CSF.

sunminsarahlee:
Allison, can we change the letter spacing for heading to 0.12em?
Screen Shot 2020-09-24 at 2 48 14 PM

packages/NotificationCard/src/components/Header/Header.js
allison-c: Maybe.. but it's required in heading component, so I'm not sure which number should I use for the default

allison-c: Discussed with @mikrotron, it's hard to tell what's the proper heading level for this component in different pages. If we set it as required, it forces the consumer to think about the proper heading level, accessibility will be improved. If someone complains about this, we can revisit for sure!

mikrotron: I'm undecided about this.
It could be just an `<h2>` by default (like SidePanel / Modal / Takeover), but that's not always going to be appropriate for this component's context. For those other components it's more likely that `<h2>` is appropriate.
If it's required though, then it forces the consumer to think about the proper heading level (for better accessibility).

nahumzs: ok, I don't know what would be worse not setting the header by default or setting to h2 or h3? but I'm not hard on this.

packages/NotificationCard/src/components/Header/Header.styles.js
```diff
@@ -5,8 +5,9 @@ import tokens from "@paprika/tokens";
 
 export const NotificationHeader = styled(Heading)`
   font-weight: bold;
-  letter-spacing: 0.05em;
+  letter-spacing: 0.12em;
```
nahumzs, line 8: we shouldn't use `em` those are based on the closer wrapper and this can change depending each app, if you can change it to px or lets just be aware this can create a small visual bug in certain scenarios.

nahumzs (approved):

PR #727, opened 2020-09-25
Uxd 257 field type component

nahumzs:
### Purpose 🚀

# @paprika/field-types - 0.1.0

- [x] FieldTypes for numeric
- [x] package ready
- [x] test in screener and small jest test
- [x] readme file
- [x] doc file
- [x] d.ts file
- [x] stories ready

Screen Shot 2020-09-25 at 2 24 59 PM

## Description

Field Types package is a set of components that help the Table and DataGrid to enhance the cell property to display different value in a ease manner

## Installation

```
yarn add @paprika/field-types
```

or with npm:

```
npm install @paprika/field-types
```

### FieldTypes

| children | node | true     | -       |             |

### Numeric

| Prop                | Type                                                       | required | default           | Description                                                                                                                                   |
| ------------------- | ---------------------------------------------------------- | -------- | ----------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| cell                | [string,number]                                            | true     | -                 | The cell value to be localize                                                                                                                 |
| align               | [ types.align.LEFT, types.align.RIGHT, types.align.CENTER] | false    | types.align.RIGHT | Text alignment for the number default is right                                                                                                |
| intl                | object                                                     | false    | {}                | The window.Intl.numberFormat option object https://mzl.la/3iW0ioQ                                                                             |
| currency            | string                                                     | false    | null              | When passing a currency string as 'EUR' or 'JPY' will display the correct currency symbol, is a short version instead of using the intl prop. |
| color               | string                                                     | false    | null              | Add a color to the number, accept any kind of html color #F60, rgba(100,100,100, 0.5), etc.                                                   |
| displayOnlyDecimals | bool                                                       | false    | true              | Controls if the number should be display with full delimiter or only the decimal separators                                                   |

#FieldTypes

FieldTypes component helps during your [Table](https://github.com/acl-services/paprika/tree/master/packages/Table) or [DataGrid](https://github.com/acl-services/paprika/tree/master/packages/DataGrid) implementation providing you with small components that encapsulate specific stylings for specific types that adapts to our Design System.

## Uses

Once the component its imported:

```js
import FieldTypes from "@paprika/field-types";
```

You can make use of different Fields in the following way:

```js
/**
 *  Using the Numeric field. The following applies either for the <Table /> or <DataGrid /> component
 */
import FieldTypes from "@paprika/field-types";
function App() {
  return <FieldTypes.Numeric align={FieldTypes.types.align.LEFT} currency="USD" number={1240} />;
}
```

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-257-FieldType-Component

### Screenshots 📸

### References 🔗


packages/FieldTypes/README.md
```diff
@@ -0,0 +1,76 @@
+<!-- autogenerated don't modify -->
+
+# @paprika/field-types - 0.1.0
+
+## Description
+
+Field Types package is a set of components that help the Table and DataGrid to enhance the cell property to display different value in a ease manner
```
KDarcey, line 7: Nitpick: adjust grammar slightly 'different values in an easy manner'

packages/FieldTypes/src/components/Container/Container.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import PropTypes from "prop-types";
+import * as sc from "./Container.styles";
```
KDarcey, line 3: Nitpick: I believe we usually want to put the styles import last

packages/FieldTypes/src/components/Container/Container.js
```diff
@@ -0,0 +1,25 @@
...
+import types from "../../types";
+
+export default function Container(props) {
+  const { children, align, color, ...moreProps } = props;
+
+  return (
+    <sc.Container $align={align} $color={color} {...moreProps}>
```
KDarcey, line 10: Question: Are the `$` intentional?

packages/FieldTypes/stories/FiledTypes.showcase.stories.js
```diff
@@ -0,0 +1,32 @@
...
+import FieldTypes from "../src";
+
+export default {
+  title: getStoryName("FieldTypes"),
+};
+
+const Container = styled.div(() => {
+  return css`
+    [data-pka-anchor="data-fields-numeric"] {
+      font-size: 32px;
```
KDarcey, line 15: Nitpick: Should we use stylers or tokens here instead?

packages/FieldTypes/src/components/Container/Container.js
mikrotron: I'm also unfamiliar with this convention, I though I did notice a prop with the `$` in the `DataGrid`.

mikrotron:
Seems like you would never need to use the `<FieldTypes>` component directly, correct?

I'm not so sure about the naming, I'd probably prefer something like `DataField`,  but since the top level is more like an abstract class, maybe it's okay.  Using `<DataField.Numeric>` seems a bit more descriptive though.

packages/FieldTypes/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  return (
+    <Container align={align} color={color} {...moreProps} data-pka-anchor="data-fields-numeric">
+      {typeof cell === "function" ? propNumber(i18nNumber) : i18nNumber}
+    </Container>
+  );
+}
+
+Numeric.propTypes = {
```
mikrotron, line 65: Aren't we still following the convention of declaring `propTypes` at the top, then just attaching at the bottom with `Numeric.propTypes = propTypes`?

packages/FieldTypes/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * The cell value to be localize
+   */
+  number: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
```
mikrotron, line 69: The PR description had this as `cell` not `number` and I came to complain, but `number` is better.  But... since there will be other "Field Type" components too, maybe we could use something like `value` which would allow a consistent API across other types.  This might be useful for dynamic type cells where we could use the `as` prop in a `map()` to set what `Field Type` component is used.

packages/FieldTypes/src/components/Numeric/Numeric.js
nahumzs: no we don't anymore, this was some changes introduce when we implemented the constant package, I remember discussing this with @allison-c but can't remember where is the PR discussion :)

packages/FieldTypes/src/components/Numeric/Numeric.js
nahumzs: sound good to me, yes I updated the readme and the code after reading my original PR, forgot to update the PR description.

packages/FieldTypes/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,98 @@
...
+   */
+  currency: PropTypes.string,
+  /**
+   * Add a color to the number, accept any kind of html color #F60, rgba(100,100,100, 0.5), etc.
+   */
+  color: PropTypes.string,
+  /**
+   * Controls if the number should be display with full delimiter or only the decimal separators
+   */
+  displayOnlyDecimals: PropTypes.bool,
```
mikrotron, line 89: How about `hasThousandsSeparators`?
Or even `hasOnlyRadixSeparator`?

packages/FieldTypes/src/components/Container/Container.js
nahumzs: if you pass color and align without the `$` will be spread on the final HTML dom, adding the `$` force the styledComponent component to drop it since is not a valid HTML attribute.

packages/FieldTypes/src/components/Container/Container.js
nahumzs: the only linter requirement that we have is to put imports from node_modules before any local import, I would think is Niptick but I can change it :)

packages/FieldTypes/README.md
nahumzs: ohhh there it's an option that you can make the change by yourself in github and propose the change and I can commit it from my side, try it. is the icon named "insert suggestion", use it :D

packages/FieldTypes/src/components/Numeric/Numeric.js
nahumzs: will do 👍

packages/FieldTypes/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,98 @@
+import React from "react";
+import PropTypes from "prop-types";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import Container from "../Container";
+import types from "../../types";
+
+// This let mimic the use of Field.types.align.LEFT in documentation and d.ts files
+// not other reason to have it
+const FieldTypes = {};
```
nahumzs, line 9: @karencumlat @allison-c this is how I was able to display FieldTypes.types.align.LEFT in documentation (reademe and d.ts) I'm not totally happy about it but I think works pretty well for the moment :/

packages/FieldTypes/README.md
KDarcey, line 7:
```suggestion
Field Types package is a set of components that help the Table and DataGrid to enhance the cell property to display different values in an easy manner
```

packages/FieldTypes/src/components/Container/Container.js
mikrotron: Ahh.  Is that a common convention, or a Nahum convention?

packages/FieldTypes/src/components/Numeric/Numeric.js
allison-c: Yes, totally! I also left some comments here https://github.com/acl-services/paprika/pull/727#discussion_r496134630

packages/FieldTypes/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,98 @@
...
+FieldTypes.types = types;
```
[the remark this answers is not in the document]
nahumzs: Correct, as long we follow the correct approach while using the types, that should be safe. We can do it in a new PR.

packages/FieldTypes/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,98 @@
...
+   */
+  intl: PropTypes.object, // eslint-disable-line react/forbid-prop-types
+  /**
+   * When passing a currency string as 'EUR' or 'JPY' will display the correct currency symbol, is a short version instead of using the intl prop.
+   */
+  currency: PropTypes.string,
+  /**
+   * Add a color to the number, accept any kind of html color #F60, rgba(100,100,100, 0.5), etc.
+   */
+  color: PropTypes.string,
```
mikrotron, line 85: Oh yeah, why does it take a `color` prop? Can't we just let it inherit the CSS from its parents?

nahumzs: I found useful to have it right there, cases exist when you want to mark one column as clickable (if it were a link) or marked a number in red color (if it's below `0`) and is quite useful to just to handle the color there instead of adding a className or styleComponent. 

For more complex cases totally you can use a className since all the props are spread.

oscarkwan:
@nahumzs There are a number of files with "FiledTypes" instead of "FieldTypes".

packages/FieldTypes/src/components/Container/Container.js
nahumzs: 🙈 nahum's

packages/FieldTypes/stories/FiledTypes.showcase.stories.js
nahumzs: is a story, its ok :)

nahumzs:
> @nahumzs There are a number of files with "FiledTypes" instead of "FieldTypes".
> 
> Screen Shot 2020-09-28 at 1 24 20 PM

solved

packages/DataField/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,99 @@
...
+      : propValue;
+
+  if (hasOnlyRadixSeparator) {
+    const numberWithDecimalsOnly = withDecimalSeparatorOnly({
+      number,
+      locale: i18n.locale,
+      options: { ...currency, ...decimalNumbers, ...intl },
+    });
+    return (
+      <Container align={align} color={color} {...moreProps} data-pka-anchor="data-fields-numeric">
```
mikrotron, line 53:
```suggestion
      <Container align={align} color={color} {...moreProps} data-pka-anchor="data-field.numeric">
```

packages/DataField/src/components/Numeric/Numeric.js
```diff
@@ -0,0 +1,99 @@
...
+        {typeof cell === "function" ? propValue(numberWithDecimalsOnly) : numberWithDecimalsOnly}
+      </Container>
+    );
+  }
+
+  return (
+    <Container align={align} color={color} {...moreProps} data-pka-anchor="data-fields-numeric">
```
mikrotron, line 60:
```suggestion
    <Container align={align} color={color} {...moreProps} data-pka-anchor="data-field.numeric">
```

packages/DataField/stories/DataField.examples.mdx.js
```diff
@@ -0,0 +1,74 @@
+import React from "react";
+import Table from "@paprika/table";
+import L10n from "@paprika/l10n";
+import styled, { css } from "styled-components";
+
+import DataField from "../src";
+
+const SimpleContainer = styled.div(() => {
+  return css`
+    [data-pka-anchor="data-fields-numeric"] {
```
mikrotron, line 10:
```suggestion
    [data-pka-anchor="data-field.numeric"] {
```

packages/DataField/stories/DataField.screener.stories.js
```diff
@@ -0,0 +1,162 @@
...
+
+const data = [
+  { name: "Abbeline Doe", income: 153234.87, taxes: 32300.34 },
+  { name: "Alivia Smith", income: 85720.92, taxes: 17300.43 },
+  { name: "Aniya Johanson", income: 45328.54, taxes: 14302.23 },
+];
+
+const Container = styled.div(() => {
+  return css`
+    [data-pka-anchor="data-fields-numeric"] {
```
mikrotron, line 24:
```suggestion
    [data-pka-anchor="data-field.numeric"] {
```

packages/DataField/stories/DataField.showcase.stories.js
```diff
@@ -0,0 +1,32 @@
...
+import styled, { css } from "styled-components";
+import DataField from "../src";
+
+export default {
+  title: getStoryName("DataField"),
+};
+
+const Container = styled.div(() => {
+  return css`
+    [data-pka-anchor="data-fields-numeric"] {
```
mikrotron, line 14:
```suggestion
    [data-pka-anchor="data-field.numeric"] {
```

mikrotron (approved):
llgtm 73

PR #736, opened 2020-10-01
UXD-305 Adding types handler for react-docgen

allison-c:
### Purpose 🚀
Adding types handler for react-docgen

### Notes ✏️

Added and extracted functions into `parseFileToReactDoc.js` .  Script will take the `types.propName.value` from `propTypes` and `defaultProps`, then convert them into `ComponentName.types.propName.value` by using `displayName`. By doing this, we don't need to create any fake wrapper for child component just for the constants types.

### Updates 📦
- [x] PATCH (bug fix) change to `DataField` `NotificationCard` `Radio`

### Storybook 📕

### Screenshots 📸

### References 🔗


scripts/parseFileToReactDoc.js
```diff
@@ -0,0 +1,40 @@
...
+        return { ...propValue };
+      });
+    }
+
+    if (value.defaultValue && new RegExp(/^types./).test(value.defaultValue.value)) {
+      documentation._props.set(key, {
+        ...value,
+        type: {
+          ...value.type,
+          value: newValue,
```
nahumzs, line 24: I'm not sure when the following could happen but:

wouldn't be better to newValue be set to `undefined` if you set it to null, the defaultValue will be `null` and be pass-down into another component, but if it's undefined the prop will not propagate.

scripts/readmeFiles.js
```diff
@@ -158,6 +155,11 @@ const processPropTables = ({ info, path, paprikaDocs = null }) => {
           arrayOfComponentsDefinitions,
         });
 
+        if (subComponent === "Image") {
```
nahumzs: I think we can delete this?

scripts/parseFileToReactDoc.js
```diff
@@ -0,0 +1,40 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+const reactDocs = require("react-docgen");
```
nahumzs, line 2: I didn't know you can extend react-docgen, awesome :) 👏

scripts/parseFileToReactDoc.js
```diff
@@ -0,0 +1,40 @@
...
+
+function typesHandler(documentation) {
+  const componentName = documentation._data.get("displayName").split(".")[0];
+
+  documentation._props.forEach((value, key) => {
+    let newValue = null;
+
+    if (value.type && value.type.name === "enum") {
+      newValue = value.type.value.map(propValue => {
+        if (new RegExp(/^types./).test(propValue.value)) {
```
alexzherdev, line 12:
```suggestion
        if (/^types\./.test(propValue.value)) {
```

nahumzs (approved):

PR #742, opened 2020-10-07
RAPRM-760  Propagate Table.ColumnDefinition moreProps to TH 🐐

mikrotron:
### Purpose 🚀
Propagate `<Table.ColumnDefinition>` `moreProps` attributes to the table header (`<th>`) elements.

### Notes ✏️
- `moreProps` were already being propagated to the cell elements (`<td>`), this applies them to the headers too, so that `classNames`, `test-ids`, etc can also be applied to the `th`, which is especially important for an empty table.
- Added a note to the `README` that explains how `<Table.ColumnDefinition>` `moreProps` are handled.
- Did not add to the MDX `Docs` page – the `README` will be added to that page in another upcoming task. 

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/table`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/RAPRM-760--table-moreprops

### References 🔗
https://aclgrc.atlassian.net/browse/RAPRM-760

This came as a request from @robyn3choi:
> I'm having a problem with `Table` right now, and maybe this is something that's missing in the docs, or just a missing feature. But it doesn't look like there's a way to control the column header widths when there are no cells in the table. So when cells get added, the column headers shift around. I saw in a storybook example that `Table.ColumnDefinition` accepts a `width` prop (it would be good to add this prop to the docs), but this doesn't kick in until cells are added.

### DIFF
--- a/packages/Table/src/Table.js
+++ b/packages/Table/src/Table.js
@@ -26,17 +26,17 @@ export default function Table(props) {
       <sc.Thead>
         <tr>
           {ColumnDefinitions.map((columnDefinition, columnIndex) => {
-            const { header } = columnDefinition.props;
+            const { cell, header, ...moreColumnProps } = columnDefinition.props;
 
             if (typeof header === "function")
               return (
-                <sc.TH borderType={borderType} key={columnIndex}>
+                <sc.TH borderType={borderType} key={columnIndex} {...moreColumnProps}>
                   {header({ header: columnDefinition.props, columnIndex })}
                 </sc.TH>
               );
             if (typeof header === "string")
               return (
-                <sc.TH borderType={borderType} key={columnIndex}>
+                <sc.TH borderType={borderType} key={columnIndex} {...moreColumnProps}>
                   {header}
                 </sc.TH>
               );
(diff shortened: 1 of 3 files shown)


packages/Table/src/Table.js
```diff
@@ -26,17 +26,17 @@ export default function Table(props) {
       <sc.Thead>
         <tr>
           {ColumnDefinitions.map((columnDefinition, columnIndex) => {
-            const { header } = columnDefinition.props;
+            const { cell, header, ...moreColumnProps } = columnDefinition.props;
 
             if (typeof header === "function")
               return (
-                <sc.TH borderType={borderType} key={columnIndex}>
+                <sc.TH borderType={borderType} key={columnIndex} {...moreColumnProps}>
```
alexzherdev, line 33: The fact that someone already assigning a `className` to the `td`s will suddenly see it being applied to `th`s and potentially messing up the presentation makes me think that this change is a breaking one.

Will also note that going this way prevents us from supplying different sets of props for `th`s vs `td`s which may or may not turn out to be a problem. (with `className`, can still work around by using `th.className { ... }` or `td.className { ... }`

mikrotron: I'm okay with introducing a breaking change. I think the use case here is most commonly going to be for selectors, so specify `th` or `td` will be acceptable for the convenience this will provide.  What would be the alternative? I can't think of anything that wouldn't be ugly, or a pretty significant change to the API.  

For more advanced use cases where different props are needed, it might be necessary to rely on passing render functions to `cell` and/or `header` which should provide a lot more flexibility (though it won't give you any control over the actual `th` or `td` elements).

mikrotron: @alexzherdev After discussing some alternative approaches with @nahumzs, we landed back on this same approach that will be a breaking change (but probably a low impact one).

nahumzs (approved):

PR #744, opened 2020-10-08
Feature(DataHeader): UX-336-Data-Header init dataheader component

nahumzs:
### Purpose 🚀
New component `DataHeader` please read the documentation and intention on this component on:  [story](http://storybooks.highbond-s3.com/paprika/UXD-336-Data-Header/?path=/story/table-dataheader-docs--page)

## NOTES

The `<DataHeader />` component helps consumer to deal with header with the `<Table />` and `<DataGrid />` components.

With this component you have access to three major features: render Label via a string or function, render an action button/menu at the right side of the header or render an icon `type`
before the label.

The component looks like:

```js
<DataHeader label={string|function} type={:string} renderActions={:function} />
```

## Basic example

A basic example using a `<Table />`

```js
<Table data={data}>
  <Table.ColumnDefinition header={() => <DataHeader label="name" />} cell="name" />
</Table>
```

## Render Actions Prop

The previous example is too simple, you can achieve the same using a single `header` prop on the `ColumnDefinition`
Let's add the `renderActions` prop to make it more interesting:

```js
import DropdownMenu from "@paprika/dropdown-menu";

function Menu() {
  return (
    <DropdownMenu>
      <DropdownMenu.Trigger buttonType="raw">
        <EllipsisVertical />
      </DropdownMenu.Trigger>
      <DropdownMenu.Item onClick={() => {}}>One</DropdownMenu.Item>
      <DropdownMenu.Item onClick={() => {}}>Two</DropdownMenu.Item>
      <DropdownMenu.Item onClick={() => {}}>Three</DropdownMenu.Item>
    </DropdownMenu>
  );
}

function App() {
  return (
    <Table data={data}>
      <Table.ColumnDefinition
        width="150px"
        header={() => <DataHeader label="name" renderActions={() => <Menu />} />}
        cell="name"
      />
    </Table>
  );
}
```
<Toast hasCloseButton={false}>
  You can render whatever you need into the `renderActions` space not only a menu, can be a button, icon, etc. It will always be align to the right.
</Toast>

## Using type prop for icons

The type props let you to assign an icon into the header to highlight the type of column, you can check the prop table on the readme to see which one are support it.

Worth noting the `DataHeader` component expose all their icons via an object declare on `DataHeader.icons` this is due licence issues with the use of some of our icons.
This object can be override in case you need to use the component outside of the [Highbond](https://www.wegalvanize.com/highbond/) platform.

```js
function App() {
  return (
    <Table data={data}>
      <Table.ColumnDefinition
        header={() => <DataHeader label="Value" type={DataHeader.type.types.TEXT} />}
        cell="name"
      />
      <Table.ColumnDefinition
        header={() => <DataHeader label="Value" type={DataHeader.type.types.NUMERIC} />}
        cell="name"
      />
      <Table.ColumnDefinition
        header={() => <DataHeader label="Value" type={DataHeader.type.types.DATA} />}
        cell="income"
      />
      <Table.ColumnDefinition
        header={() => <DataHeader label="Value" type={DataHeader.type.types.TIME} />}
        cell="taxes"
      />
      <Table.ColumnDefinition
        header={() => <DataHeader label="Value" type={DataHeader.type.types.DATE_TIME} />}
        cell="income"
      />
    </Table>
  );
}
```

## BackgroundColor and color

There are scenarios when you will want to highlight specific header column with a color, in this case you can make use of the `backgroundColor` and `color` props.

```js
function App() {
  return (
    <Table data={data}>
      <Table.ColumnDefinition
        header={() => <DataHeader backgroundColor="magenta" color="white" label="name" />}
        cell="name"
      />
      <Table.ColumnDefinition
        header={() => <DataHeader backgroundColor="cyan" label="Income" />}
        cell={({ row }) => row.income}
      />
      <Table.ColumnDefinition
        header={() => <DataHeader backgroundColor="black" color="white" label="Value" />}
        cell={({ row }) => row.taxes}
      />
    </Table>
  );
}
```

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-336-Data-Header

### Screenshots 📸

### References 🔗


packages/DataGrid/src/DataGrid.Rows.js
```diff
@@ -123,6 +123,7 @@ export const HeaderRow = React.memo(({ data: RowData, columnIndex, style }) => {
       role="columnheader"
       style={{ ...style, ...styleProps }}
       {...moreProps}
+      data-pka-header="true"
```
mikrotron, line 126: Why not `data-pka-anchor='data-grid.header'`?

packages/DataHeader/src/DataHeader.js
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Character from "@paprika/icon/lib/Character";
+import Numeric from "@paprika/icon/lib/Number";
+import DateTime from "@paprika/icon/lib/TimeAndDate";
+import ClockTime from "@paprika/icon/lib/ClockTime";
```
mikrotron, line 6: I think we renamed this to Clock?

packages/DataHeader/src/DataHeader.js
```diff
@@ -0,0 +1,58 @@
...
+  color: null,
+};
+
+DataHeader.icons = {
+  [types.type.TEXT]: <Character />,
+  [types.type.NUMERIC]: <Numeric />,
+  [types.type.DATE]: <Date />,
+  [types.type.DATE_TIME]: <DateTime />,
+  [types.type.TIME]: <ClockTime />,
+};
```
mikrotron, line 58: Do you still plan to extend the API for this component so that these icons can be customized?

packages/DataHeader/src/DataHeader.styles.js
```diff
@@ -0,0 +1,62 @@
...
+});
+
+export const Info = styled.div(() => {
+  return css`
+    align-items: center;
+    display: flex;
+    justify-content: flex-start;
+    width: 100%;
+    svg {
+      font-size: 1.3rem;
```
mikrotron, line 59: Why `rems`?

packages/DataHeader/src/DataHeader.styles.js
```diff
@@ -0,0 +1,62 @@
+import styled, { css } from "styled-components";
+import tokens from "@paprika/tokens";
+
+export const Header = styled.div(({ $color = tokens.color.black, $backgroundColor = "transparent", refHeader }) => {
```
mikrotron, line 4: These defaults for $color and $backgroundColor are kind of hidden, it seems they should be in the `defaultProps` instead. 

But having defaults makes it difficult to override with CSS.  If they can be `null` then we can just do nothing in `resetHeaderParentContainer()`.

packages/DataHeader/src/DataHeader.js
mikrotron: Also, what about Booleans? I know we made the Boolean icon ourselves, it should be in Wasabicons.

packages/DataHeader/stories/DataHeader.stories.mdx
```diff
@@ -0,0 +1,154 @@
+import { Meta } from "@storybook/addon-docs/blocks";
+import { getStoryName } from "storybook/storyTree";
+import Readme from "../README.md";
```
mikrotron, line 3: You import the README, but then don't render it.  Ideally we can save ourselves some work in the Docs page by reusing that content, which should demonstrate the basic use cases. Then the Docs page can illustrate more interesting/advanced use cases.

packages/DataHeader/test/DataHeader.spec.js
```diff
@@ -0,0 +1,5 @@
+describe("DataHeader", () => {
+  it("Testing data header", () => {
+    expect(true).toBe(true);
```
mikrotron, line 3: `// TODO` ?

packages/Icon/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Ellipsis vertical icon [@nahumzs](https://github.com/nahumzs).
```
mikrotron, line 11: Also Character, Number, DateTime? Will you also follow up by removing them all from Wasabicons?

packages/Icon/README.md
```diff
@@ -18,6 +18,10 @@ npm install @paprika/icon
 
 ## Usage
 
+```
+IMPORTANT: PAPRIKA MAKES USE OF SOME PROPRIETARY ICONS THAT REQUIRE A LICENCE IN ORDER TO BE CONSUMED. PLEASE REPLACE THOSE PROPRIETARY ICONS OUTSIDE OF THE HIGHBOND PLATFORM.
```
mikrotron, line 22: Should we say which ones? Or maybe just `PLEASE DO NOT CONSUME THESE ICONS DIRECTLY.` for the second sentence?

packages/Icon/CHANGELOG.md
nahumzs: I think we should keep the wasavicon icon we are the main users, but we are also allowing to overriding this, the component should work out of the box as all other ones.

packages/Icon/README.md
nahumzs: Yes let's add that phrase after it and we should be ok.

packages/DataHeader/stories/DataHeader.stories.mdx
```diff
@@ -0,0 +1,154 @@
...
+}
+```
+
+### Output
+
+<WithRenderActions />
+<br /><br />
+<Toast hasCloseButton={false}>
+  You can render whatever you need into the `renderActions` space not only a menu, can be a button, icon, etc. It will always be align to the right.
+</Toast>
```
mikrotron, line 85: Oh! I _like_ this 🤩 Nice easy way to make a callout.  
cc @oscarkwan @KaanDarcey 

However the `renderActions` just displays with backticks, you'd have to use a `<code>` block or something.

packages/DataHeader/stories/DataHeader.stories.mdx
```diff
@@ -0,0 +1,154 @@
...
+### Output
+
+<Basic />
+
+## Render Actions Prop
+
+The previous example is too simple, you can achieve the same using a single `header` prop on the `ColumnDefinition`
+Let's add the `renderActions` prop to make it more interesting:
+
+```js
```
mikrotron, line 50: `jsx` produces nicer syntax highlighting, in my opinion.

packages/DataHeader/stories/DataHeaderWithBackground.datagrid.stories.js
```diff
@@ -0,0 +1,51 @@
+import React from "react";
```
mikrotron, line 1: This seems too similar to the `DataHeader.datagrid.stories.js` story.  We can probably eliminate one of them.

packages/DataHeader/stories/DataHeader.screener.stories.js
```diff
@@ -0,0 +1,180 @@
...
+const icons = {
+  ...DataHeader.icons,
+  [DataHeader.types.type.DATE_TIME]: <CaretUp />,
+};
+
+export const TypesScreener = () => {
+  return (
+    <Story>
+      <h3>TEXT</h3>
+      <DataHeader icons={DataHeader.icons} label="name" type={DataHeader.type.types.TEXT} />
```
mikrotron, line 169:
```suggestion
      <DataHeader icons={DataHeader.icons} label="name" type={DataHeader.types.type.TEXT} />
```

mikrotron: Not sure if that's the right fix, but the story is currently broken

packages/Icon/CHANGELOG.md
mikrotron: If an icon is in Paprika though, it will automatically get pulled into Wasabicons. So we can remove it from the wasabicons repo and just let it get imported from here, rather than have two copies that might get out of sync.

packages/DataHeader/src/DataHeader.styles.js
nahumzs: no entirely sure what would be better, I can't be sure all our apps have a `16px` base font-size: so better to use a root definition than setup this at 16px? I don't think is a big issue if we see is not working correctly we can setup in pixels?

packages/Icon/CHANGELOG.md
nahumzs: ohh I see, I understand incorrectly what you mean, so you are saying that I have to go and remove these icons from wasabicons?

mikrotron: Yes, exactly. No rush, just a good neighbourly thing to do 🙂

packages/DataHeader/stories/DataHeader.screener.stories.js
nahumzs: yes its, I already fixed it thank you :)

packages/Icon/CHANGELOG.md
nahumzs: Will add a ticket for it.

packages/DataHeader/stories/DataHeader.stories.mdx
nahumzs: what's the difference? adding those ticks is quite simply no need to learn anything neither remember :)

mikrotron: Just a suggestion from what I've found.  Using `'''jsx` instead of `'''js` makes it prettier.

Screen Shot 2020-10-15 at 10 24 53 AM

vs

Screen Shot 2020-10-15 at 10 25 13 AM

packages/DataHeader/stories/DataHeader.stories.mdx
```diff
@@ -0,0 +1,158 @@
+import { Meta } from "@storybook/addon-docs/blocks";
```
mikrotron, line 1:
```suggestion
import { Meta, Description } from "@storybook/addon-docs/blocks";
```

packages/DataHeader/stories/DataHeader.stories.mdx
```diff
@@ -0,0 +1,158 @@
...
+    isToolshown: true,
+    showPanel: false,
+  },
+};
+
+<Meta title={`${getStoryName("DataHeader")}/Docs`} parameters={params} />
+
+# README
+
+<Description markdown={Readme} />
```
mikrotron, line 18: Without the import above, this doesn't render (but also doesn't produce an error, I guess because it's MDX).

packages/DataHeader/stories/DataHeader.stories.mdx
```diff
@@ -0,0 +1,158 @@
...
+The `<DataHeader />` component helps consumer to deal with header with the `<Table />` and `<DataGrid />` components.
+
+With this component you have access to three major features: render Label via a string or function, render an action button/menu at the right side of the header or render an icon `type`
+before the label.
+
+The component looks like:
+
+```js
+<DataHeader label={string|function} type={:string} renderActions={:function} />
+```
```
mikrotron, line 33: This is all in the README, so if that get's inserted then this can be removed.

packages/DataHeader/stories/DataHeaderWithBackground.datagrid.stories.js
mikrotron: I still see both... did you mean to leave them both still?

packages/DataHeader/stories/DataHeader.stories.mdx
nahumzs: do you mean in the linter? we are not lingering mdx as you commenting

mikrotron (approved):
llgtm 100

PR #751, opened 2020-10-20
UXD-390 Upgrade screener-storybook

jamiek-acl:
Upgrade `screener-storybook`, which will upgrade one of its dependencies (`http-proxy`) which has a high-severity security vulnerability.


nahumzs (approved):
🦖 👍

PR #753, opened 2020-10-22
Uxd 374 list box tags

nahumzs:
#### Uncontrolled example:

```jsx
import React from "react";
import ListBoxWithTags, { useListBoxWithTags } from "@paprika/list-box-with-tags";
import animals from "../mocks";

const defaultFilteredData = animals.slice(0, 20);
const defaultData = animals;

  // All props are provided by useListBoxWithTags
  const { isSelected, filteredData, getSelectedOptions, ...moreUseListBoxWithTagsProps } = useListBoxWithTags("label", {
    defaultData,
    defaultFilteredData,
  });

  return (
    <div style={{ padding: "32px" }}>
      <ListBoxWithTags selectedOptions={getSelectedOptions()} {...moreUseListBoxWithTagsProps}>
        {filteredData.map(option => {
          if (typeof option.isCustom !== "undefined") {
            return null;
          }

          return !isSelected(option.label) ? (
            <ListBoxWithTags.Option value={option.label} key={option.label} label={option.label}>
              {option.label}
            </ListBoxWithTags.Option>
          ) : null;
        })}
      </ListBoxWithTags>
    </div>
  );
}

// Note
// At any point you can execute `getSelectedOptions()` which returns an array of the current selected options.
```

Will produce:

#### useListBoxWithTags hook

To make the component uncontrolled the ListBoxWithTags provides a hook

`useListBoxWithTags` receive two parameters the first one is the `key` property from your defaultData items, which let the hook to know what's the unique key to use for storing selected items and other internal features.

The second parameter received and object with the following properties: 

```
defaultData = []

it is used to store you default Data and is an object array, it's expected that at least those object included a "label:string" property.
```

```
defaultFilteredData = []

A subset of `defaultData` which is used to store the current filtered items once the user type on the filter.
```

```
defaultSelectedKeys = [] 

An array with keys that can be used to prefill the component with selected items.
```

```
filterAttribute = "label"

in case you want to make use of another property to filter your results. 
```

It will return the props you need to fill your ListBoxWithTags, ex:

```js
  const { isSelected, filteredData, getSelectedOptions, ...moreUseListBoxWithTagsProps } = useListBoxWithTags("label", {
    defaultData,
    defaultFilteredData,
  });
```

#### Custom Pill

You can take full control when rendering the Pills to adapt it to your needs:

```jsx
  function renderPill({ option, Pill, onRemove }) {
    const color = getAvatarColors(option.label);
    return (
      <Pill onRemove={onRemove} key={option.label}>
        <div style={{ display: "flex", justifyContent: "center", alignItems: "center" }}>
          <div style={styleForPill(color)}>{option.label.substring(0, 2)}</div>
          {option.label}
        </div>
      </Pill>
    );
  }
```

#### Prop Table 

| Prop              | Type       | required | default        | Description                                                                                                       |
| ----------------- | ---------- | -------- | -------------- | ----------------------------------------------------------------------------------------------------------------- |
| children          | node       | true     | -              | Expect <ListBoxWithTags.Option />                                                                                 |
| filter            | func       | false    | undefined      | filter function for the ListBoxWithTags can be pair with ListBoxWithTags.filter                                   |
| noResultsMessage  | node       | false    | null           | String message to be display when there are not results                                                           |
| onChange          | func       | false    | () => {}       | Callback whenever the user change a selection on the ListBoxWithTags                                              |
| onCustomOption    | func       | false    | null           | Callback whenever the user input a new custom option like some@email.com, pass undefined to ignore this behaviour |
| onRemove          | func       | false    | () => {}       | Callback once a pill is remove from the Trigger                                                                   |
| regexCustomOption | instanceOf | false    | /^.+@.+\..+\$/ | Regex that match the input of the user and reports to onCustomOption. The default is a basic email regex          |
| renderPill        | func       | false    | null           | Render prop to override the default Pill style, see example for it's uses.                                        |
| selectedOptions   | arrayOf    | false    | null           | An array of id that helps the ListBoxWithTags to known what elements are selected                                 |

#### Controlled component
This is a more contrived example but shows you how to do a Fully controlled component.

```jsx
import React from "react";
import ListBoxWithTags from "../../src";
import animals from "../mocks";

// please use a proper debounce function in a your application
const debounce = (func, wait) => {
  let idTimeout = null;
  return (...args) => {
    clearTimeout(idTimeout);
    idTimeout = setTimeout(() => {
      func(...args);
    }, wait);
  };
};

const animalSubset = animals.slice(0, 20);

export default function Controlled() {
  const [data] = React.useState(animalSubset);

  const [selectedKeys, setSelectedKeys] = React.useState(["Alpaca"]);
  const [dataFiltered, setDataFiltered] = React.useState(data);

  function checkIfIsSelected(id) {
    return selectedKeys.includes(id);
  }

  function handleChange(option, options, selectedOption) {
    setSelectedKeys(prev => {
      const prevClone = prev.slice(0);
      const id = options[selectedOption].value;
      if (prev.includes(id)) {
        prev.splice(prevClone.indexOf(id), 1);
        return prev;
      }

      return [...new Set(prevClone.concat([id]))];
    });

    setDataFiltered(data);
  }

  function handleRemove(option) {
    const index = selectedKeys.indexOf(option.label);
    if (index >= 0) {
      setSelectedKeys(prev => {
        const clone = prev.slice(0);
        clone.splice(index, 1);
        return clone;
      });
    }
  }

  function handleAddedOption(label) {
    const option = { label, isCustom: true };
    // this is just an example
    // but in real life you could have a list/dictionary with the
    // value of the selected items where you could push the new value
    animals.push(option);

    setDataFiltered(() => data);

    setSelectedKeys(prev => {
      return prev.concat(option.label);
    });
  }

  const handleFilterDebounce = React.useMemo(
    () =>
      debounce(({ search }) => {
        if (search === "") {
          setDataFiltered(data);
          return;
        }

        const result = ListBoxWithTags.filter(search, animals);
        setDataFiltered(result);
      }, 250),
    [data]
  );

  return (
    <div style={{ padding: "32px" }}>
      <ListBoxWithTags
        noResultsMessage="No results found, but you can add an email and then press enter..."
        onChange={handleChange}
        onCustomOption={handleAddedOption}
        onRemove={handleRemove}
        selectedOptions={selectedKeys.length ? animals.filter(item => selectedKeys.includes(item.label)) : []}
        filter={handleFilterDebounce}
      >
        {dataFiltered.map(option => {
          if (typeof option.isCustom !== "undefined") {
            return null;
          }

          return !checkIfIsSelected(option.label) ? (
            <ListBoxWithTags.Option value={option.label} key={option.label} label={option.label}>
              {option.label}
            </ListBoxWithTags.Option>
          ) : null;
        })}
      </ListBoxWithTags>
    </div>
  );
}

```

### Updates 📦
`new`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-374-ListBoxTags

### Screenshots 📸

### References 🔗


nahumzs:
Note: Written documentation and split for Pill in their own component in the following PR, Don't want this one to grow too much.

packages/ListBoxWithTags/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@paprika/list-box-with-tags",
+  "version": "0.0.1",
+  "description": "ListBoxWithTags behave as a ListBox with the difference that the selected options stacked on pills next to each other",
```
jamiek-acl, line 4: Since there is no `Tag` component, maybe this should be `ListBoxWithPills`?
Maybe they dont know about the `ListBox`, so we shouldn't reference it in the description?

```
The ListBoxWithTags component allows users to select one or more options from a list contained within the component, showing a pill next to each.
```

As discussed, maybe we don't even need a separate component... it's still just a `ListBox`

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -0,0 +1,156 @@
...
+import PropTypes from "prop-types";
+import ListBox from "@paprika/list-box";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import Pill, { Pills } from "./components/Pill";
+import { filter } from "./helpers";
+import * as sc from "./ListBoxWithTags.styles";
+
+const propTypes = {
+  /** Expect <ListBoxWithTags.Option /> */
+  children: PropTypes.node.isRequired,
```
jamiek-acl, line 11: `PropTypes.instanceOf(ListBoxWithTags.Option)`

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -0,0 +1,156 @@
...
+  /** filter function for the ListBoxWithTags can be pair with ListBoxWithTags.filter  */
+  filter: PropTypes.func,
+  /** String message to be display when there are not results  */
+  noResultsMessage: PropTypes.node,
+  /** Callback whenever the user change a selection on the ListBoxWithTags  */
+  onChange: PropTypes.func,
+  /** Callback whenever the user input a new custom option like some@email.com, pass undefined to ignore this behaviour */
+  onCustomOption: PropTypes.func,
```
jamiek-acl, line 19: I think `onAddCustomOption` is more clear

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -0,0 +1,156 @@
...
+  /** Callback once a pill is remove from the Trigger */
+  onRemove: PropTypes.func,
+  /** Regex that match the input of the user and reports to onCustomOption. The default is a basic email regex */
+  regexCustomOption: PropTypes.instanceOf(RegExp),
```
jamiek-acl, line 23: I think it should be `customOptionRegex` or `customOptionRegexPattern`

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -0,0 +1,156 @@
...
+  /** Render prop to override the default Pill style, see example for it's uses.  */
+  renderPill: PropTypes.func,
+  /** An array of id that helps the ListBoxWithTags to known what elements are selected  */
+  selectedOptions: PropTypes.arrayOf(PropTypes.shape({})),
```
jamiek-acl, line 27: `arrayOf(PropTypes.object)`?

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -0,0 +1,156 @@
...
+};
+
+const defaultProps = {
+  filter: undefined,
```
jamiek-acl, line 31: Why `undefined` and not `null`?

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -0,0 +1,156 @@
...
+    renderPill,
+    selectedOptions,
+    ...moreProps
+  } = props;
+  const { t } = useI18n();
+
+  const refDivRoot = React.useRef(null);
+  /* eslint-disable react/prop-types */
+  const size =
+    typeof props.size !== "undefined" &&
```
jamiek-acl, line 109: Why not just put it in `defaultProps`?

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -0,0 +1,156 @@
...
+    [ListBox.types.size.MEDIUM, ListBox.types.size.SMALL, ListBox.types.size.LARGE].includes(props.size)
+      ? props.size
+      : ListBox.types.size.MEDIUM;
+  /* eslint-enable react/prop-types */
+
+  const refFilter = React.useRef(null);
+
+  function handleKeyDown(event) {
+    const label = event.target.value;
+    const regexEmail = regexCustomOption;
```
jamiek-acl, line 119: dont need `regexEmail` variable

packages/ListBoxWithTags/src/ListBoxWithTags.styles.js
```diff
@@ -0,0 +1,56 @@
...
+  `;
+});
+
+export const PlaceHolderText = styled.div(() => {
+  return css`
+    border-radius: ${tokens.border.radius};
+    line-height: 1;
+    margin-bottom: ${tokens.spaceSm};
+    margin-right: ${tokens.spaceSm};
+    padding: ${tokens.spaceSm[0] / 2}px ${tokens.spaceSm};
```
jamiek-acl, line 54: why is `spaceSm` an array...?

packages/ListBoxWithTags/src/components/Pill/Pill.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Close from "@paprika/icon/lib/Times";
+import * as sc from "./Pill.styles";
+
+const propTypes = {
+  onRemove: PropTypes.func.isRequired,
+  children: PropTypes.node.isRequired,
+  size: PropTypes.oneOf(["small", "medium", "large"]),
```
jamiek-acl, line 9: Constants. On lines 13 and 39, too.

packages/ListBoxWithTags/src/helpers.js
```diff
@@ -0,0 +1,11 @@
...
+  const regex = new RegExp(`${search}`, "gi");
+  const results = [];
+  for (let i = 0, len = data.length; i < len; i++) {
+    if (data[i][id].match(regex) !== null) {
+      results.push(data[i]);
+    }
+  }
+
+  return results;
+};
```
jamiek-acl, line 11: I think this is just:

```
export const filter = (pattern, data, id = "label") => {
  const regex = new RegExp(pattern, "gi");  
  return data.filter(datum => datum[id].match(regex) !== null)
};
```

packages/ListBoxWithTags/src/useListBoxWithTags.js
```diff
@@ -0,0 +1,110 @@
...
+
+      return [...new Set(prevClone.concat([id]))];
+    });
+
+    setFilteredData(defaultData);
+  }
+
+  function getSelectedOptions() {
+    if (selectedKeys.length) {
+      const options = [];
```
jamiek-acl, line 48: Put `options` outside of the `if` statement, and at the end of this function `return options;`

packages/ListBoxWithTags/src/useListBoxWithTags.js
```diff
@@ -0,0 +1,110 @@
...
+        return clone;
+      });
+    }
+  }
+
+  function isSelected(id) {
+    return selectedKeys.includes(id);
+  }
+
+  const handleCustomOption = (func = label => ({ label, isCustom: true })) => label => {
```
jamiek-acl, line 74: 😨

packages/ListBoxWithTags/stories/examples/UncontrolledUser.js
```diff
@@ -0,0 +1,166 @@
+import React from "react";
+import Popover from "@paprika/popover";
+import { getAvatarColors } from "@paprika/avatar/lib/helpers";
+import Avatar from "@paprika/avatar";
+import ListBoxWithTags, { useListBoxWithTags } from "../../src";
+
+// prettier-ignore
```
jamiek-acl, line 7: This should be extracted to a separate file, and prettified

packages/ListBoxWithTags/src/useListBoxWithTags.js
```diff
@@ -0,0 +1,110 @@
...
+
+  data.forEach(d => {
+    dictionary[d[key]] = d;
+  });
+
+  return dictionary;
+}
+
+export default function useListBoxWithTags(
+  key,
```
jamiek-acl, line 16: What do you think about putting `key` in the object, too?

packages/ListBoxWithTags/test/ListBoxWithTags.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      { id: 2, label: "The Fox" },
+      { id: 3, label: "The Raccoon" },
+      { id: 4, label: "Dinosaur" },
+    ];
+
+    expect(ListBoxTags.filter("dog", data)).toMatchSnapshot();
+    expect(ListBoxTags.filter("fox", data)).toMatchSnapshot();
+    expect(ListBoxTags.filter("Raccoon", data)).toMatchSnapshot();
+    expect(ListBoxTags.filter("the", data)).toMatchSnapshot();
+    expect(ListBoxTags.filter("saur", data)).toMatchSnapshot();
```
jamiek-acl, line 45: I thought we weren't using snapshots anymore

packages/ListBox/src/hooks/hooks.js
```diff
@@ -111,43 +111,40 @@ export function useOnScrolled() {
...
       const rectOption = $option.getBoundingClientRect();
 
-      // if options its between the box don't do anything
-      if (rectOption.top >= rectBox.top && rectOption.bottom <= rectBox.bottom) {
+      const nextTopPositionForOption =
+        rectOption.height * state.activeOption + state.activeOption * (marginBottomNumber + marginTopNumber);
+
+      // this occurs when the first element is a Divider so we want to scroll to the top
+      // so the user can see the Divider text instead to see only the option
+      if (state.activeOption - 1 === 0 && state.options[state.activeOption - 1].isDisabled) {
```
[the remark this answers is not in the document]
nahumzs: 🤔 it might be that all of them are disabled but at that point, you couldn't use the up/down arrow, is common to have the first element to be a Divider (group) and unlikely you have two separators at the beginning. Could happen just doesn't make sense from UI/UX perspective.

packages/ListBoxWithTags/stories/mocks.js
```diff
@@ -0,0 +1,2 @@
+// prettier-ignore
```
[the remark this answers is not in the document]
nahumzs: so it doesn't count as lines and make my PR looks bigger haha when is not :P

packages/ListBoxWithTags/package.json
nahumzs: Remove ListBowWithTags as a package and put it inside of ListBox, the dev experience has increased I think.

packages/ListBoxWithTags/src/ListBoxWithTags.js
nahumzs: yeap the user can have any kind of object data on it. Can't validate that :P

packages/ListBoxWithTags/src/ListBoxWithTags.js
nahumzs: if you set null the filter prop will be spread if you put undefined the filter prop will be removed by react and will not be spread.

packages/ListBoxWithTags/src/ListBoxWithTags.js
nahumzs: Ohh this is so I don't expose this prop because is already part of the ListBox.

packages/ListBoxWithTags/stories/mocks.js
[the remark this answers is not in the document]
jamiek-acl: I think that ship has sailed

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -0,0 +1,157 @@
...
+  /** Callback whenever the user input a new custom option like some@email.com, pass undefined to ignore this behaviour */
+  onCustomOption: PropTypes.func,
+  /** Callback once a pill is remove from the Trigger */
+  onRemove: PropTypes.func,
+  /** Regex that match the input of the user and reports to onCustomOption. The default is a basic email regex */
+  regexCustomOption: PropTypes.instanceOf(RegExp),
+  /** Render prop to override the default Pill style, see example for it's uses.  */
+  renderPill: PropTypes.func,
+  /** An array of id that helps the ListBoxWithTags to known what elements are selected  */
+  selectedOptions: PropTypes.arrayOf(PropTypes.shape({})),
```
jamiek-acl, line 28:
```suggestion
  selectedOptions: PropTypes.arrayOf(PropTypes.object),
```

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -0,0 +1,157 @@
...
+    selectedOptions,
+    ...moreProps
+  } = props;
+  const { t } = useI18n();
+
+  const refDivRoot = React.useRef(null);
+  /* eslint-disable react/prop-types */
+  const size =
+    typeof props.size !== "undefined" &&
+    [ListBox.types.size.MEDIUM, ListBox.types.size.SMALL, ListBox.types.size.LARGE].includes(props.size)
```
jamiek-acl, line 111:
```suggestion
    [Object.keys(ListBox.types.size)].includes(props.size)
```

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -0,0 +1,157 @@
...
+  selectedOptions: null,
+};
+
+const renderTrigger = ({ t, size, selectedOptions, onRemove, renderPill }) => (...args) => {
+  const [, , , attributes] = args;
+  const { propsForTrigger, refTrigger, dispatch, types, handleKeyDown, handleKeyUp } = attributes;
+
+  function handleClick(event) {
+    event.stopPropagation();
+    if (event.target.dataset.pkaAnchor === "listbox-tags-pill-delete") {
```
jamiek-acl, line 48:
```suggestion
    if (event.target.dataset.pkaAnchor === "list-box-tags.pill.delete") {
```

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -0,0 +1,157 @@
...
+    [ListBox.types.size.MEDIUM, ListBox.types.size.SMALL, ListBox.types.size.LARGE].includes(props.size)
+      ? props.size
+      : ListBox.types.size.MEDIUM;
+  /* eslint-enable react/prop-types */
+
+  const refFilter = React.useRef(null);
+
+  function handleKeyDown(event) {
+    const label = event.target.value;
+    const regexEmail = regexCustomOption;
```
jamiek-acl, line 120: You don't need the variable `regexEmail`, and who says it will always be an `email` regular expression?

packages/ListBox/src/helpers/filter.js
```diff
@@ -0,0 +1,11 @@
...
+  const regex = new RegExp(`${search}`, "gi");
+  const results = [];
+  for (let i = 0, len = data.length; i < len; i++) {
+    if (data[i][id].match(regex) !== null) {
+      results.push(data[i]);
+    }
+  }
+
+  return results;
+};
```
jamiek-acl, line 11:
```suggestion
export const filter = (pattern, data, id = "label") => {
  const regex = new RegExp(pattern, "gi");  
  return data.filter(datum => datum[id].match(regex) !== null)
};
```

packages/ListBox/src/hooks/hooks.js
jamiek-acl, line 125:
```suggestion
      if (state.activeOption === 1 && state.options[0].isDisabled) {
```

packages/ListBox/src/components/WithTags/useWithTags.js
```diff
@@ -0,0 +1,109 @@
...
+      const options = [];
+      selectedKeys.forEach(key => {
+        options.push(dataDictionary[key]);
+      });
+
+      return options;
+    }
+
+    return [];
+  }
```
jamiek-acl, line 56:
```suggestion
  function getSelectedOptions() {
    return selectedKeys.map(key => dataDictionary[key]);
  }
```

packages/ListBox/src/components/WithTags/useWithTags.js
```diff
@@ -0,0 +1,109 @@
...
+    const index = selectedKeys.indexOf(option[key]);
+    if (index >= 0) {
+      setSelectedKeys(prev => {
+        const clone = prev.slice(0);
+        clone.splice(index, 1);
+        return clone;
+      });
+    }
+  }
+
```
jamiek-acl, line 68: Can `.filter` be used here?

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -0,0 +1,157 @@
...
+      {...propsForTrigger()}
+      onClick={handleClick}
+      onKeyUp={handleKeyUp}
+      onKeyDown={handleKeyDown}
+      size={size}
+    >
+      <Pills>
+        {selectedOptions.map(item => {
+          if (typeof renderPill === "function") {
+            return renderPill({ option: item, Pill, onRemove: handleRemove(item) });
```
alexzherdev, line 73: Any benefit to passing `Pill` in parameters vs exporting it from the package? Technically there's no need to pass it in since it's the same every time

packages/ListBox/src/components/WithTags/useWithTags.js
```diff
@@ -0,0 +1,109 @@
...
+    const option = func(label);
+    setData(prev => {
+      return prev.concat(option);
+    });
+
+    setSelectedKeys(prev => {
+      return prev.concat(option[key]);
+    });
+
+    setData(prevData => {
```
alexzherdev, line 85: Why is this necessary in addition to line 77?

packages/ListBox/stories/WithTags/examples/CustomTrigger.js
```diff
@@ -0,0 +1,57 @@
...
+import animals from "../mocks";
+
+const defaultFilteredData = animals.slice(0, 20);
+const defaultData = animals;
+
+// prettier-ignore
+const styleForPill = (color) => ({ alignItems: "center", backgroundColor: color.backgroundColor, borderRadius: "50%", boxSizing: "border-box", color: color.fontColor, display: "flex", height: "24px", justifyContent: "center", marginRight: "8px", padding: "3px", width: "24px", fontSize: ".8rem", lineHeight: 1})
+
+export default function CustomTrigger() {
+  const { isSelected, filteredData, getSelectedOptions, ...moreUseListBoxWithTagsProps } = useWithTags("label", {
```
alexzherdev, line 14: `useListBoxWithTags`/`Pills` seems a better name, otherwise no idea what exactly I'm using with tags?

packages/ListBox/stories/WithTags/examples/Controlled.js
```diff
@@ -0,0 +1,106 @@
...
+      <ListBox
+        noResultsMessage="No results found, but you can add an email and then press enter..."
+        onChange={handleChange}
+        onCustomOption={handleAddedOption}
+        onRemove={handleRemove}
+        selectedOptions={selectedKeys.length ? animals.filter(item => selectedKeys.includes(item.label)) : []}
+        filter={handleFilterDebounce}
+      >
+        {dataFiltered.map(option => {
+          if (typeof option.isCustom !== "undefined") {
```
alexzherdev, line 93: This boilerplate is present in all of the stories but it's not clear why it's necessary

packages/ListBoxWithTags/src/ListBoxWithTags.styles.js
nahumzs: there is a conversation why having tokens with px and not units creates complexity in some scenarios, I propose to refactor styles so it's possible to do this without "alternatives" / "hacks" / "importing stylers".

For this specific case `tokens.spaceSm` is equal to "4px" which is a string, so I'm just taking the first element which is the 4. https://app.slack.com/client/T02SWPEUM/CB3QRPDJ4/thread/CB3QRPDJ4-1605751958.213700

nahumzs: all strings are arrays...

packages/ListBoxWithTags/src/components/Pill/Pill.js
nahumzs: yeah I'm fine with this duplication, reads correctly wherever you are and doesn't provide any extra effort that would be adding it to a constant. You know me I'm against micro abstractions.

nahumzs: Also the Pill will be their own component so this will get refactor :)

packages/ListBoxWithTags/src/helpers.js
nahumzs: yeah in this specific case I used a `for` instead of the `filter` method because theoretically filter it should be slower than a regular `for`, and while 99% of the cases I would be used the `Array.filter` method, in this case, might get call multiple cases in a long list, so didn't add too much effort to me to add this "small" boost of performance even if I haven't been measure it.

packages/ListBoxWithTags/src/useListBoxWithTags.js
nahumzs: in case you want to override the way, you create your custom function. 

maybe the following is easier to read?:
```
  const factoryCustomOption = label => ({ label, isCustom: true });
  const handleCustomOption = (func = factoryCustomOption) => label => {
```

packages/ListBoxWithTags/test/ListBoxWithTags.spec.js
nahumzs: these are not html snapshot these are javascript objects snapshots and are really handy when you want to assert that a specific object is returned whenever running a function. 

Instead of manually written the expected return object just assert that match the snapshot :)

packages/ListBox/src/components/WithTags/WithTags.js
nahumzs: would make any difference? the linter will complain if I use an object isn't it?

packages/ListBox/src/helpers/filter.js
nahumzs: https://github.com/acl-services/paprika/pull/753#discussion_r529857074

packages/ListBox/src/components/WithTags/WithTags.js
jamiek-acl: I think both behave the same, just one is more concise

packages/ListBox/stories/WithTags/examples/Controlled.js
nahumzs: yes, this is for the case when the user is allowed to add custom options via the filter input. 

The `onAddCustomOption()` has a default factory function that look like:

```
(label) => {
  return { label, isCustom: true};
}
```

But can be override as:

```js
onAddCustomOption(label => {
 return { id: nanoid, label, ...yourStuff }
})
```

And delegate the control about how to create this option to the consumer.

I'm planning to add all the full documentation in the next PR with proper `md` and `mdx` documentation.

packages/ListBox/src/components/WithTags/useWithTags.js
nahumzs:
```js
    const option = func(label);
    
    setData(prev => {
      return prev.concat(option);
    });

    setSelectedKeys(prev => {
      return prev.concat(option[key]);
    });

    setData(prevData => {
      return prevData.concat(option);
    });
```

Seem I applying two times the same line, brain fart.


But this is the full explanation:
```js
// this create the custom options with the filter's input 
    const option = func(label);
// this add the option to the entire data set and force useMemo function to recreate the dictionary line 26
    setData(prev => {
      return prev.concat(option);
    });
// this add the "key" to the selected item
    setSelectedKeys(prev => {
      return prev.concat(option[key]);
    });

// this restart the options filtered to the original defaultFilterData, the initial state.
    setFilteredData(defaultData);
```

packages/ListBox/src/components/WithTags/WithTags.js
nahumzs: Mmm, for this pass it's easier to expose it like this, I will extract the Pill to its own package soon, and then can be imported, don't want to export it as part of the ListBox package.

Meanwhile, this allowed us to expose it easily and also can be fixed quickly whenever we do a breaking change or we can even just import the `Pill` inside of ListBox and still be backward compatible.

packages/ListBoxWithTags/src/useListBoxWithTags.js
nahumzs: I thought it for the past two hours and agree makes sense, will be more descriptive having the word key on the object.

packages/ListBox/stories/WithTags/examples/Controlled.js
nahumzs: I have removed it from the examples because it no longer needs it in most cases.

packages/ListBox/src/components/WithTags/WithTags.js
nahumzs: ohh miss this one, fixing it.

packages/ListBox/src/index.d.ts
```diff
@@ -170,6 +170,28 @@ declare namespace ListBox {
...
+  onChange?: (...args: any[]) => any;
+  /** Callback whenever the user input a new custom option like some@email.com, pass undefined to ignore this behaviour */
+  onAddCustomOption?: (...args: any[]) => any;
+  /** Callback once a pill is remove from the Trigger */
+  onRemove?: (...args: any[]) => any;
+  /** Regex that match the input of the user and reports to onAddCustomOption. The default is a basic email regex */
+  customOptionRegex?: instanceOf;
+  /** Render prop to override the default Pill style, see example for it's uses. */
+  renderPill?: (...args: any[]) => any;
+  /** An array of id that helps the ListBoxWithTags to known what elements are selected */
```
jamiek-acl, line 192: A few outdated references to `ListBoxWithTags` in this block

packages/ListBox/src/components/Pill/Pill.styles.js
```diff
@@ -0,0 +1,80 @@
...
+    padding: 0;
+  `;
+});
+
+export const Pill = styled.li(() => {
+  return css`
+    align-items: center;
+    background: ${tokens.backgroundColor.level0};
+    border: 1px solid ${tokens.border.color};
+    border-radius: ${tokens.space[0] * 2}px;
```
jamiek-acl, line 21: These `space[0]` will work... until the token changes to be `10px`

packages/ListBox/src/components/Pill/Pill.js
```diff
@@ -0,0 +1,40 @@
...
+  onRemove: PropTypes.func.isRequired,
+  children: PropTypes.node.isRequired,
+  size: PropTypes.oneOf(["small", "medium", "large"]),
+};
+
+const defaultProps = {
+  size: "medium",
+};
+
+export function Delete({ size, onRemove = () => {} }) {
```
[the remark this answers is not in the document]
nahumzs: will be replaced in their own package I think is ok for now.

jamiek-acl (approved):
tenor

PR #759, opened 2020-10-22
UXD-427  Add README and Docs pages for Tokens + Stylers 🐐

mikrotron:
### Purpose 🚀
Add information about how to use the `tokens` + `stylers` in their `README` files and on Storybook.

### Notes ✏️
- README content is re-used in Storybook Docs pages.
- Stylers stories also converted to CSF.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-427--tokens-stylers-docs

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-427


nahumzs:
👏 this is awesome, sometimes I'm reluctant to use stylers because I don't know which one exists!! now will be easier!

alexzherdev:
This one would really benefit from having code samples in the docs IMO. Especially since I think some of the stylers are full declarations while others are just values? (e.g. I never remember whether it's `line-height: ${stylers.lineHeight(1)}` or just `${stylers.lineHeight(1)}`? Also being able to just copy-paste into the app from storybook would be nice.

packages/Stylers/stories/Stylers.sass.stories.js
```diff
@@ -0,0 +1,200 @@
...
+export default {
+  title: `${storyName}/Sass`,
+};
+
+export const mixins = () => (
+  <Story className="story--stylers">
+    <h1>Mixin Examples</h1>
+
+    <h4>
+      <code>@truncate-text</code>
```
alexzherdev, line 21: These seem misleading since the `@` is on `@include`, not the mixin name (`@include truncate-text;`). Again, for people less familiar with the syntax, having an actual copy-pasteable bit would be nice. I know it's mentioned in `Docs` but just looking at the story you're not seeing that.

oscarkwan (approved):

PR #763, opened 2020-10-23
chore(Paprika): UXD-409 Adding a commit format validator

jamiek-acl:
### Purpose 🚀
Implement Husky commit hook that forces commit messages to follow `Conventional commits` pattern and include JIRA ticket number.  

There are existing tools to do something similar:
[commitlint](https://github.com/conventional-changelog/commitlint#getting-started) - forces commits to use conventional commits pattern, but not start with JIRA ticket number
[commitlint-jira](https://github.com/Gherciu/commitlint-jira) -forces commits to start with JIRA ticket number, but not use conventional commits pattern
but we wanted a tool that did both.

### Usage

Commit messages will now have to be like:

```
chore(TimePicker): add isReadOnly prop
feat(ListBoxTags): created new component
refactor(MultiSelect): now uses hooks
```
The valid types are: ["feat", "fix", "style", "test", "docs", "build", "chore", "ci", "perf", "refactor", "revert"]

If a commit message does not follow this pattern, the commit will fail and an error message is displayed:
Screen Shot 2020-10-23 at 10 23 57 AM

The script will also grab the JIRA ticket number from the branch name and inject it into the commit message, so it also forces branch names to be structured like: `abc-123--blahblah-blah`

After grabbing the branch name, the script will automatically convert the above commit messages to:

```
chore(TimePicker): ABC-123 add isReadOnly prop
feat(ListBoxTags): ABC-123 created new component
refactor(MultiSelect): ABC-123 now uses hooks
```

If a branch name does not contain a JIRA ticket number, the commit will fail and an error message is displayed:
Screen Shot 2020-10-23 at 10 31 44 AM

### NOTE TO ALL DEVS

If you previously configured git to use a ruby script to add the JIRA ticket number (instead of Husky), you will need to undo that by running this command: 
```
$ cp ~/src/paprika/.git/hooks/applypatch-msg ~/src/paprika/.git/hooks/commit-msg
```


nahumzs:
maybe worth mentioning why we didn't use commitlint or the jira plugin to do this?

nahumzs (approved):
Let's merge it and have tacos! 

Screen Shot 2020-10-23 at 1 48 36 PM

PR #768, opened 2020-10-28
Uxd 464 expose onclose and card

jamiek-acl:
### Purpose 🚀
I added two things to `<DropdownMenu />`:

1. You can now pass in an `onClose` prop that is fired after the component is closed.  I need this so I can change the active state of the trigger that opened it.
2. You can now add a `<DropdownMenu.Content />` as a child of `<DropdownMenu>` to add props (e.g. a "className") to the internal `Popover.Card`:

Screen Shot 2020-10-28 at 10 18 45 AM

 I need this so I can target the Popover for style changes, i.e. remove this gap:
Screen Shot 2020-10-27 at 4 11 59 PM

### DIFF
--- a/packages/DropdownMenu/src/components/Content/Content.js
+++ b/packages/DropdownMenu/src/components/Content/Content.js
@@ -1,5 +1,3 @@
-// import React from "react";
-
 export default function Content() {
   return null;
 }
--- a/packages/DropdownMenu/src/components/Content/Content.js
+++ b/packages/DropdownMenu/src/components/Content/Content.js
@@ -1,0 +1,7 @@
+// import React from "react";
+
+export default function Content() {
+  return null;
+}
+
+Content.displayName = "DropdownMenu.Content";
--- a/packages/DropdownMenu/src/components/Content/Content.js
+++ b/packages/DropdownMenu/src/components/Content/Content.js
@@ -1,5 +1,7 @@
+import React from "react";
+
 export default function Content() {
-  return null;
+  return <React.Fragment />;
 }
 
 Content.displayName = "DropdownMenu.Content";
--- a/packages/DropdownMenu/README.md
+++ b/packages/DropdownMenu/README.md
@@ -72,6 +72,7 @@ import Confirmation from "@paprika/confirmation";
 
 <DropdownMenu>
   <DropdownMenu.Trigger>Trigger</DropdownMenu.Trigger>
+  <DropdownMenu.Content className="my-popover-classname" />
   <DropdownMenu.Item onClick={() => {}}>Edit</DropdownMenu.Item>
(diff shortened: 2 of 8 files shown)


nahumzs:
~The way the className is passed down will work but is no aligned with the way our API works, the correct solution is to expose a Popover shell component for the DropDownMenu Component and extract the props in the `DropDownMenu.Popover`with extractChildrenComponent function and spread them over the popover inside of the DropDownMenu implementation something like this:

Add [in this line](https://github.com/acl-services/paprika/blob/uxd-464-expose-onclose-and-card/packages/DropdownMenu/src/DropdownMenu.js#L84) something like:
```js
  const extractedChildren = extractChildren({"DropDownMenu.Trigger": popoverExtracted, children}, ["DropdownMenu.Trigger", "DropDownMenu.Popover"]);
```

Then because this line is already spreading (in this line)[]

```js
<DropDownMenu>
 <DropDownMenu.Popover className="myClassName" />
 <DropDownMenu.Option>option</DropDownMenu.Option>
</DropDownMenu>
```~

---

Ok the problem is that we want to target the `Popover.Content` :/ we already spreading the props on the popover so maybe you can do something like:

```js
<DropDownMenu>
 <DropDownMenu.Content className="myClassName" />
 <DropDownMenu.Option>option</DropDownMenu.Option>
</DropDownMenu>
```
and then:
```js
 const { "DropdownMenu.Content": Content, children: extractedChildren } = extractChildren(children, ["DropdownMenu.Trigger", "DropdownMenu.Conent"]);
```

finally [in this line](https://github.com/acl-services/paprika/blob/master/packages/DropdownMenu/src/DropdownMenu.js#L175):

```js
<Popover.Content {...Content.props}>
```

packages/DropdownMenu/src/components/Content/Content.js
```diff
@@ -0,0 +1,5 @@
+export default function Content() {
```
jamiek-acl, line 1: This "component" was required because it needs a `displayName` for `extractChildren` to recognize it.

packages/DropdownMenu/src/components/Content/Content.js
```diff
@@ -0,0 +1,5 @@
...
+  return null;
```
nahumzs, line 2: Can you return <React.Fragment></React.Fragment> this helps React DocGen to recognize this as a ReactComponent and include it into the documentation, returning null will not be recognized as a react component.

nahumzs:
also, can you run these two https://github.com/acl-services/paprika/blob/master/package.json#L11-L12 scripts so the readme and types get updated it?

nahumzs (approved):
Screen Shot 2020-10-28 at 1 58 05 PM

hungry!

PR #770, opened 2020-10-28
Uxd-478 form element margins

tristanjasper:
### Purpose 🚀

1. Ensure formElement has margins applied so that application css won't override. (eg in the case of a fieldset)
2. Cleanup margins on subcomponents

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to FormElement

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-478-formElement-margins

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/FormElement/src/FormElement.styles.js
+++ b/packages/FormElement/src/FormElement.styles.js
@@ -1,6 +1,5 @@
 import styled, { css } from "styled-components";
 import stylers from "@paprika/stylers";
-import tokens from "@paprika/tokens";
 
 const FontSizes = {
   small: stylers.fontSize(-2),
@@ -25,7 +24,7 @@ export const FormElement = styled.div(
     ${isInline && inlineFormElementStyles};
     ${isDisabled && `opacity: 0.5;`}
     border: none;
-    margin: 0 0 ${tokens.space} 0;
+    margin: 0;
     padding: 0;
   `
 );
--- a/packages/FormElement/src/FormElement.styles.js
+++ b/packages/FormElement/src/FormElement.styles.js
@@ -1,5 +1,6 @@
 import styled, { css } from "styled-components";
 import stylers from "@paprika/stylers";
+import tokens from "@paprika/tokens";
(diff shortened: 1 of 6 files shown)


packages/FormElement/src/FormElement.styles.js
```diff
@@ -24,6 +25,7 @@ export const FormElement = styled.div(
     ${isInline && inlineFormElementStyles};
     ${isDisabled && `opacity: 0.5;`}
     border: none;
+    margin: 0 0 ${tokens.space} 0;
```
tristanjasper, line 28: now ensuring a bottom margin is specified only on the root element. This is important to ensure it won't be overridden by application css eg when it is rendered as a `fieldset`

packages/FormElement/src/components/Content/Content.js
```diff
@@ -26,9 +25,9 @@ function Content(props) {
   }
 
   return (
-    <sc.ContentContainer data-pka-anchor="form-element.content" {...moreProps}>
+    <div data-pka-anchor="form-element.content" {...moreProps}>
       {typeof children === "function" ? children({ idForLabel, refLabel, ariaDescribedBy }) : children}
-    </sc.ContentContainer>
+    </div>
```
tristanjasper, line 30: There was no need for the margin-bottom on the content so removed it. This is important if there is no description  or error components below the content then we don't get unwanted bottom margins still existing

packages/FormElement/src/FormElement.styles.js
mikrotron: Could we alternatively set `margin: 0` and rely on the app to apply any bottom margin it needs?  

I think it's ideal for the root element of a component to have no margin.

tristanjasper: Sure, yeah that probably makes more sense. Will change.

mikrotron (approved):
llgtm 90

PR #771, opened 2020-10-29
UXD-458  Side bar nav

allison-c:
### Purpose 🚀

Using `SideNavigation` wrap `SideNavigation.Item`. Sometimes there'll be a top navigation on the page. We don't want the side navigation overlapping with the top one, so we can add `offsetY`, which will be passed to the internal `SidePanel` component.

```jsx
<div style={{ display: "flex", flexDirection: "column", height: "100vh" }}>
  <div
    style={{
      backgroundColor: "#f0f0f0",
      fontSize: "24px",
      height: "40px",
      paddingLeft: "8px",
    }}
  >
    Top navigation
  </div>
  <div style={{ display: "flex", flexDirection: "row", flexGrow: 1 }}>
    <SideNavigation header="Assets" offsetY={40}>
      <SideNavigation.Item href={mockURL}>First item (0)</SideNavigation.Item>
      <SideNavigation.Item isCurrent href={mockURL}>
        Second item (1000)
      </SideNavigation.Item>
      <SideNavigation.Item href={mockURL}>Long text item with (100000000) items inside</SideNavigation.Item>
    </SideNavigation>
    <div>Page content</div>
  </div>
</div>
```

### Result

http://storybooks.highbond-s3.com/paprika/UXD-458--side-bar-nav/iframe.html?id=navigation-sidenavigation--showcase&viewMode=story

Jietu20201030-094838-HD

### Notes ✏️

Will ask @sunminsarahlee to review the design.

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/side-navigation`
- [x] PATCH (bug fix) change to `@paprika/sidepanel`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-458--side-bar-nav

### Screenshots 📸

### References 🔗


packages/SidePanel/src/components/Content/Content.js
```diff
@@ -11,15 +11,17 @@ const defaultProps = {
   children: null,
 };
 
-export default function Content(props) {
+const Content = React.forwardRef((props, ref) => {
```
allison-c, line 14: It turns out I don't really need the ref on side panel content for the new component, but this change might be useful later on, so I didn't revert it.

packages/SidePanel/src/components/Dialog/Dialog.js
```diff
@@ -34,7 +34,7 @@ function Dialog(props) {
   const dialogMain = (
     <React.Fragment>
       <sc.DialogContent
-        data-pka-anchor="sidepanel.content"
+        data-pka-anchor="sidepanel.dialogContent"
```
allison-c, line 37: There were two elements having `sidepanel.content` as pka-anchor

packages/SideNavigation/src/components/Item/Item.styles.js
```diff
@@ -0,0 +1,45 @@
+import styled, { css } from "styled-components";
+import Button from "@paprika/button";
+import tokens from "@paprika/tokens";
+import { spacer, fontSize } from "@paprika/stylers/lib/helpers";
+
+export const Item = styled(Button.Link)`
```
nahumzs, line 6: 🤔 so all items on the side menu need to be a button.link? what about those cases where we want to render a div container, or just a regular button that doesn't jump into a new page?

Wouldn't make more sense to set this a `styled.div`

And then you could do `<SideNavigation.Item as={Button.Link} href="">You link?</SideNavigation.Item>`

I guess you can do the opposite like `<SideNavigation.Item as={div} data-type="this is a regular div">My item</SideNavigation.Item>`

If you think `Button.Link` should be the default, then can you add a story pointing out how to render regular items on it?

allison-c: From the current design `Button.Link` looks like the default, sure I'll add a story for rendering custom items!

nahumzs (approved):

Issue #735, opened 2020-10-01
DropdownMenu LinkItem appearance and refactor


tristanjasper:
After more discussion with @krismckinnonacl and other Ux Designer input, it has been concluded that the `DropdownMenu.LinkItem` appearance should not be any different that a regular `DropdownMenu.Item`. 

We should change the LinkItem to only be used for ExternalLinks (and it should use the paprika ExternalLink component which includes the open in new tab symbol)

There is also a bug in that if the first item in the dropdown is a Dropdown.LinkItem, it will not show focus style correctly when navigated to with keyboard

# Bug Report
## Prerequisites
**This is just a checklist, please delete this section after**
Please answer the following questions for yourself before submitting a bug report.
- [ ] I checked to make sure that this bug has not already been filed
- [ ] I reproduced the bug in the latest version of paprika
- [ ] Please label as, low / medium / high priority

*Please fill in as much detail as possible below*

## Expected behavior
*Please describe the behavior you are expecting to happen.*

## Current behavior
*What is the current behavior?*

## Screenshots / Gifs / Codepens
*Include media to help illustrate the bug. Gifs help a lot!*

## Additional context
*Please provide any more additional details to illustrate the bug*

tristanjasper:
Closing this as is now no longer an issue.
Links are now styled to look like regular dropdown items. 
There doesn't appear to be a bug with the first item being a linkItem either

PR #772, opened 2020-10-29
chore(DropdownMenu): Make links black. Add external link icon onto ex…

jamiek-acl:
Before:
Screen Shot 2020-10-29 at 10 39 26 AM

After:
Screen Shot 2020-10-29 at 11 06 48 AM


packages/DropdownMenu/src/components/LinkItem/LinkItem.js
```diff
@@ -33,15 +34,18 @@ const LinkItem = props => {
     ...moreProps,
   };
 
+  let externalLinkIcon = null;
   if (isExternal) {
     linkItemProps["aria-label"] = I18n.t("dropdownMenu.isExternal", { link: children });
     linkItemProps.target = "_blank";
     linkItemProps.rel = "noopener noreferrer";
+    externalLinkIcon = <NewTabIcon size="10px" />;
```
tristanjasper, line 42: Perhaps we could leverage the `Button.Link` component 
```
if (isExternal) {
    linkItemProps.a11yText = I18n.t("dropdownMenu.isExternal", { link: children });
    linkItemProps.shouldOpenNewTab = true;
    linkItemProps.suffixIcon = null;
  }

  return (
    <Button.Link css={linkItemStyles} {...linkItemProps}>
      {children}
    </Button.Link>
  );
```

tristanjasper:
Looks like this is digging into the issue I created recently - https://github.com/acl-services/paprika/issues/735
that myself and @krismckinnonacl were discussing.

packages/DropdownMenu/src/Item.styles.js
```diff
@@ -5,12 +5,17 @@ import stylers from "@paprika/stylers";
 const itemStyles = css`
   ${stylers.fontSize()};
   ${stylers.lineHeight()}
+  color: ${tokens.color.black};
```
tristanjasper, line 8: I know when I was consuming the dropdownMenu component in results and used the non external links I had application styles causing styling issues, so I was experimenting with adding 

```
border-radius: 0;
border-width: inherit;
min-height: inherit;
text-align: left;
``` 
 to fix some of those application styles from affecting this. But can revisit that later.

packages/DropdownMenu/src/components/LinkItem/LinkItem.js
jamiek-acl: Hmm I don't think that's necessary, I'm not experiencing any focus ring style issues as-is.

tristanjasper: Ok np I will address the keyboard navigation focus bug (when first item is a link) in a follow up pr.

tristanjasper (approved):

PR #774, opened 2020-10-29
UXD-157 formElement component api refactor

tristanjasper:
### Purpose 🚀

This PR changes the formElement

- Use a context based approach to sub component rendering
- Content receives a `a11yProps` single render prop which can be spread onto children creating a simpler api for consumers

- Updates to DatePicker, Radio, Checkbox to be able to take this `a11yProps` prop and use it correctly.

This is a followup Pr to a previous approach that was not pursued - https://github.com/acl-services/paprika/pull/663

It is still expected that any other aria-labels such as aria-required should still be supplied by the consumer.

As an simple example api change comparison is as follows:

Before:
```
import FormElement from "@paprika/form-element";

<FormElement label="Name">
  <FormElement.Content>
    {({ idForLabel, ariaDescribedBy }) => (
      <Input
        id={idForLabel}
        onChange={handleChange}
        value={value}
        aria-describedby={ariaDescribedBy}
        aria-required={hasRequiredLabel}
        hasError={false}
        isDisabled={isDisabled}
        isReadOnly={isReadOnly}
        size={size}
      />
    )}
  </FormElement.Content>
</FormElement>;
```

New api:

```
import FormElement, { useFormElement, Label, Content } from "@paprika/form-element";

<FormElement>
  <Label>Name</Label>
  <Content>
{a11yProps => (
    <Input
      onChange={handleChange}
      value={value}
      aria-required={hasRequiredLabel}
      hasError={false}
      isDisabled={isDisabled}
      isReadOnly={isReadOnly}
      size={size}
      {...a11yProps}
    />
    )}
  </Content>
</FormElement>;
```

### Notes ✏️

### Updates 📦
- [x] MAJOR (breaking) change to FormElement, DatePicker, Checkbox, Radio

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-157-formElement-api-refactor-2

### Screenshots 📸

### References 🔗


packages/FormElement/src/components/Label/Label.js
```diff
@@ -4,56 +4,70 @@ import PropTypes from "prop-types";
...
+
+  const a11yProps = formElementA11yProps
+    ? { htmlFor: formElementA11yProps.labelA11yProps.id, ref: formElementA11yProps.labelA11yProps.ref }
+    : {};
+
+  const renderQuestionRequirement = () => {
+    if (hasRequiredLabel) {
+      return I18n.t("formElement.required");
+    }
+    if (hasOptionalLabel) {
```
tristanjasper, line 51: For now i'm leaving in support for the "Optional" tag to be shown with the formElement, although i'm not sure if we really need or want it anymore. Perhaps designers can confirm @sunminsarahlee

packages/DatePicker/src/DatePicker.js
```diff
@@ -26,9 +26,6 @@ const propTypes = {
   /** Date format used while displaying date. It should be human-friendly and spelled out, default is MMMM DD,YYYY */
   humanFormat: PropTypes.string,
 
-  /** ID for the `<input>`. */
-  id: PropTypes.string,
```
tristanjasper, line 30: Unrelated change here: I suggested we remove the `id` from the datePicker component (which is spread onto the `Input` subcomponent, so that we encourage consumers to use the `<DatePicker.Input id={id}>` instead and can add the id there.

Apart from being easier to understand and be more explicit, this avoids the situation where an id maybe applied to root component, but is overridden by `<DatePicker.Input {...datePickerA11yProps}>`

packages/FormElement/src/hooks/useFormElement.js
```diff
@@ -0,0 +1,56 @@
...
+    radioA11yProps: {
+      "aria-describedby": ariaDescribedBy,
+    },
+    datePickerA11yProps: {
+      id: idForLabel,
+      "aria-describedby": ariaDescribedBy,
+    },
+    listBoxA11yProps: {
+      refLabel,
+    },
```
mikrotron, line 36: I don't like that this component would need to know about its consumer.  Can we just offer up all the props (`id`, `ariaDescribedBy`, `refLabel`) and let the consumer destructure the ones it needs?

alexzherdev:
to go on record with some of the previous discussions - I don't know of a substantially better way to implement this particular approach compared to what's in this PR. But overall I think this approach is flawed and creates the wrong dependency between the form element and its fields ([see from before](https://github.com/acl-services/paprika/pull/663#issuecomment-681120673)).
I would really love if those not satisfied with the render prop approach voiced their concerns. I personally think this is an inversion of control situation that calls for a render prop. (don't care for my opinion? read KCD here https://kentcdodds.com/blog/react-hooks-whats-going-to-happen-to-render-props#the-remaining-use-case-for-render-props)
I don't see what using hooks buys us here. A potential advantage of hooks would be better composability but I just don't see consumers creating custom hooks that would compose `useFormElement`.
¯\\\_(ツ)_/¯

packages/FormElement/src/hooks/useFormElement.js
tristanjasper: I can see that it is easier for a consumer to just get the spread the props they need without having to know which props they should be spreading, but I agree it does create a strange binding based on component name....
Personally I am more in favour of your suggested approach Mike but will wait to hear what others think.

tristanjasper:
> to go on record with some of the previous discussions - I don't know of a substantially better way to implement this particular approach compared to what's in this PR. But overall I think this approach is flawed and creates the wrong dependency between the form element and its fields ([see from before](https://github.com/acl-services/paprika/pull/663#issuecomment-681120673)).
> I would really love if those not satisfied with the render prop approach voiced their concerns. I personally think this is an inversion of control situation that calls for a render prop. (don't care for my opinion? read KCD here https://kentcdodds.com/blog/react-hooks-whats-going-to-happen-to-render-props#the-remaining-use-case-for-render-props)
> I don't see what using hooks buys us here. A potential advantage of hooks would be better composability but I just don't see consumers creating custom hooks that would compose `useFormElement`.
> ¯\_(ツ)_/¯

I believe the primary motivation for the change is to go with an approach more consistent with other components, but other than that I don't have a strong opinion on either approach so interested to hear others views on this.

alexzherdev:
> I believe the primary motivation for the change is to go with an approach more consistent with other components, but other than that I don't have a strong opinion on either approach so interested to hear others views on this.

the argument for consistency only stands when you're solving the same or a similar problem. You don't use a screwdriver on nails just because you've used screwdrivers on screws before. I'm not aware of other components using hooks that would be solving a similar problem to form element.

packages/FormElement/src/components/Content/Content.styles.js
```diff
@@ -0,0 +1,12 @@
...
+
+const inlineContentStyles = css`
+  align-items: baseline;
+  display: flex;
+`;
+
+export const Content = styled.div(
+  ({ isInline }) => css`
+    ${isInline && inlineContentStyles};
+  `
```
mikrotron, line 11: I think the `Content` component could be agnostic of this `isIinline` prop (which should be set only on the parent `FormElement`).  The styling for FormElement could then apply any transformation CSS necessary to both the Label and the Content (but maybe that could even be avoided if it simply wraps them both in a flex layout).

packages/FormElement/src/hooks/useFormElement.js
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import nanoid from "nanoid";
+import isNil from "lodash/isNil";
+
+export default function useFormElement(props = { id: "", wrapperA11yProps: {} }) {
+  const { id, wrapperA11yProps } = props;
+  const wrapperAriaDescribedBy = wrapperA11yProps ? wrapperA11yProps["aria-describedby"] : "";
+  const ariaDescriptionId = React.useRef(nanoid()).current;
```
nahumzs, line 8: state is better for nanoid https://github.com/ai/nanoid#react

packages/FormElement/src/hooks/useFormElement.js
nahumzs: I second mike, we should strive for making the form element fully agnostic.

packages/FormElement/src/hooks/useFormElement.js
mikrotron: Why? It doesn't change or need to invoke a rerender, so why not just use a static ref?

packages/FormElement/src/FormElement.js
```diff
@@ -1,93 +1,18 @@
...
-  const idForLabel = generateLabelId(id);
-  const refLabel = React.useRef(null);
+  const { children, isDisabled, isInline, size, hasFieldSet, formElementA11yProps, ...moreProps } = props;
 
-  const getClonedElement = (displayName, extraProps = {}) => {
-    if (!extractedChildren[displayName]) return null;
-    return React.cloneElement(extractedChildren[displayName], extraProps);
+  const value = {
+    formElementA11yProps,
+    hasFieldSet,
```
mikrotron, line 13: Wouldn't it make sense to put the rest of the props – `isInline`, `size`, `isDisabled` on context too for any of the sub components that may rely on those for styling adjustments?

packages/FormElement/src/hooks/useFormElement.js
alexzherdev: pretty sure it's a sub-millisecond matter but with a ref there's a `nanoid()` call happening on every invocation of the outer function even though its result is immediately discarded. not so with state

alexzherdev:
Take note of the last remark in this section, it applies to this PR https://github.com/ai/nanoid#react

packages/FormElement/src/FormElement.js
tristanjasper: Currently I don't see any other sub components making use of those props so don't need to add them to context right now

packages/FormElement/src/components/Label/Label.js
```diff
@@ -4,56 +4,68 @@ import PropTypes from "prop-types";
 import useI18n from "@paprika/l10n/lib/useI18n";
 
 import labelStyles, { ruleStyles } from "./Label.styles";
+import Help from "../Help/Help";
```
tristanjasper, line 7: Still passing the help as a prop instead of just passing it as regular children to this component, purely because we have a specific order that we want the children to render in, ie the help should be to the right of the "optional" or "required" tag (which is rendered via prop control)

packages/Checkbox/src/Checkbox.js
```diff
@@ -41,8 +31,11 @@ const Checkbox = props => {
     size,
   };
 
+  if (extendedInputProps.refLabel) {
+    delete extendedInputProps.refLabel;
```
mikrotron, line 35: What harm is there in leaving this?

packages/Checkbox/src/Checkbox.js
```diff
@@ -41,8 +31,11 @@ const Checkbox = props => {
...
+  }
+
   const inputProps = {
-    "aria-describedby": ariaDescribedBy,
```
mikrotron, line 45: This isn't needed anymore? It's a breaking change – should be added to the `CHANGELOG`

packages/DatePicker/src/DatePicker.js
mikrotron: `datePickerA11yProps` → `a11yProps` (from the FormElement.Content render prop argument).  Still a valid argument, but maybe @allison-c or @DanilAgafonov would have some concerns with this change to the API since they are more familiar with the `<DatePicker>`?

packages/FormElement/README.md
```diff
@@ -140,19 +131,20 @@ Using with DOM input
...
-    {({ idForLabel, ariaDescribedBy }) => (
+    {a11yProps => (
       <input
         aria-required={hasRequiredLabel}
-        aria-describedby={ariaDescribedBy}
+        aria-describedby={a11yProps[aria-describedby]}
         aria-invalid={Boolean(errorText.length)}
         disabled={isDisabled}
-        id={idForLabel}
+        id={a11yProps.id}
```
mikrotron, line 143: Can you not spread them on a native `<input>`?

packages/FormElement/src/components/Description/Description.js
```diff
@@ -12,10 +14,16 @@ const defaultProps = {
 };
 
 function Description(props) {
-  const { children, ariaDescriptionId, ...moreProps } = props;
+  const [ariaDescriptionId] = React.useState(nanoid);
```
mikrotron, line 17: Interesting change. This used to be generated in the FormElement and passed _down_. Now it's generated here and passed _up_.  I don't disagree with this approach, but curious if it was needed for some reason, or if you think it just made more sense this way?

packages/FormElement/src/components/Error/Error.js
```diff
@@ -13,22 +13,28 @@ const defaultProps = {
...
 
+  React.useEffect(() => {
+    addIdToAriaDescribedBy({ ariaErrorId });
+  }, []);
+
   if (!children) return null;
 
   return (
-    <div css={errorMessageStyles} data-pka-anchor="form-element.error" {...moreProps}>
+    <div role="alert" css={errorStyles} data-pka-anchor="form-element.error" {...moreProps} id={ariaErrorId}>
```
mikrotron, line 28: Oh, this didn't have `role=alert` before? Hmm. I noticed in Safari it no longer reads it out, but that was in a static example, not dynamic.  But still, I am a bit worried about JAWS, since this makes it a "live region" and I think with the `<Toast>` we needed to implement a timeout to get it to work properly.  Let's omit it for now but maybe open a GitHub issue, because I think it's semantically correct to make this an `alert`, just think we should carefully test it before making the change.  This is an important component.

packages/FormElement/src/FormElement.js
```diff
@@ -122,59 +45,29 @@ FormElement.types = {
...
 FormElement.displayName = "FormElement";
 FormElement.propTypes = propTypes;
 FormElement.defaultProps = defaultProps;
 
-FormElement.subComponentDisplayNames = subComponentDisplayNames;
-FormElement.Content = Content;
-FormElement.Instructions = Instructions;
-FormElement.Description = Description;
-FormElement.Error = ErrorMessage;
-FormElement.Help = Help;
```
mikrotron, line 178: I prefer this style of packaging up the component.  I see it is how you are packaging the `Fieldset` and I think they should be consistent.  It means you can't import the subcomponents separately on the same line, but if you really want to, you could destructure the `FormElement` (or `Fieldset`) to clean up the JSX.  

If this is confusing, let's discuss.  I tried it out locally, and I think it's a bit clearer.

packages/FormElement/src/components/Label/Label.js
mikrotron: Last I saw, both are still supported in the Forms patterns pitch:
https://aclgrc.atlassian.net/wiki/spaces/UX/pages/676398749/Forms+pattern+v2+in+review#Element-instruction%3A-Concise

packages/FormElement/src/components/Label/Label.js
```diff
@@ -4,56 +4,68 @@ import PropTypes from "prop-types";
...
-          {hasRequiredLabel ? I18n.t("formElement.required") : null}
-          {hasOptionalLabel ? I18n.t("formElement.optional") : null}
-        </span>
-      ) : null}
-
-      {help}
+    <div data-pka-anchor="form-element.label" css={labelStyles} {...labelProps} {...moreProps} {...a11yProps}>
+      {children}
+      {hasOptionalLabel || hasRequiredLabel ? <span css={ruleStyles}> {renderQuestionRequirement()}</span> : null}
+      {renderHelp()}
```
mikrotron, line 65: This is minor, and can be addressed later, but since the `<label>` `<legend>` wraps the `<Help>` trigger, the "Information" label for that trigger gets appended to label that the screen reader announces, which can be a bit confusing.  

Wrapping the whole label in a `<div>` and then just the `children` and "required/optional" bit in the `<label>` needs to be thought through though, because then were would the `moreProps` go? Probably have to go on the `div`, but then you have no way of putting more props on the `label` without a `propsCollector`.  That _might_ be okay, but not sure.  So yeah, okay for now.

packages/FormElement/src/components/Layout/Layout.js
```diff
@@ -0,0 +1,17 @@
...
+export const LayoutFlexParent = styled.div`
+  display: flex;
+`;
+
+export const LayoutLeftCol = styled.div(
+  ({ width }) => css`
+    width: 180px;
+    ${width && `width: ${width}px;`}
+  `
+);
```
mikrotron, line 12: Exporting just the styled component like this means the `width` prop isn't documented anywhere (automatically anyway).  I think we should wrap this one in a component with PropTypes.

Also, what about a percentage for `width`? That would allow nice, fluid layouts. Maybe something like `if typeof width === number` add `"px"`?

packages/FormElement/src/components/Layout/Layout.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const LayoutRightCol = styled.div`
+  flex: 1;
```
mikrotron, line 15:
```suggestion
  flex-grow: 1;
```

packages/FormElement/src/components/Layout/index.js
```diff
@@ -0,0 +1 @@
+export { LayoutRightCol, LayoutLeftCol, LayoutFlexParent } from "./Layout";
```
mikrotron, line 1: Instead of exporting each, what if we just export `Layout` as default?

Something like:

```js
const Layout = styled.div...
export const LeftCol = styled.div...
export const RightCol = styled.div...

Layout.LeftCol = LeftCol;
Layout.RightCol = RightCol;

export default Layout;
```

Then to use, it's like this:
```js
import FormElement from "../../src";
const { Layout } = FormElement;

<Layout>
  <Layout.LeftCol width="120">...</Layout.LeftCol>
  <Layout.RightCol>...</Layout.RightCol>
</Layout>
```

Does that seem cleaner to you?

packages/FormElement/stories/FormElement.examples.stories.js
```diff
@@ -10,10 +9,11 @@ import DatePicker from "./examples/DatePicker";
 import ListBox from "./examples/ListBox";
 import ButtonGroup from "./examples/ButtonGroup";
 import NativeInput from "./examples/NativeInput";
-import Nested from "./examples/Nested";
+import Nested from "./examples/NestedExample";
```
mikrotron, line 12: I've been trying to move away from the unnecessary `Example` suffix on these story components.

packages/FormElement/stories/FormElement.examples.stories.js
```diff
@@ -102,6 +102,13 @@ nestedExample.story = {
   parameters: exampleStoryParameters,
 };
 
+export const inlineContentExample = () => (
+  <FormElementStory storyName="Inline Content" fileName="examples/InlineContent.js">
```
mikrotron, line 106: It will need the `component="FormElement"` prop for the `source` link to appear (it needs it to reconstruct the GitHub path/URL).

packages/DatePicker/CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+2020-10-29
```
allison-c, line 9: You don't need to put the date here, it's auto-generated when publish

packages/FormElement/stories/examples/NestedExample.js
```diff
@@ -9,63 +12,80 @@ export default function NestedExample() {
...
+                        readOnly={isReadOnly}
+                        id={a11yProps.id}
+                        aria-describedby={a11yProps["aria-describedby"]}
+                      />
+                      <input
+                        aria-required={hasRequiredLabel}
+                        disabled={isDisabled}
+                        readOnly={isReadOnly}
+                        aria-describedby={a11yProps["aria-describedby"]}
+                      />
```
mikrotron, line 49: Is there something special you're trying to illustrate with these native `<inputs>` as opposed to the Paprika `<Inputs>` below?

packages/DatePicker/src/DatePicker.js
allison-c: 👍 I'm ok with it

mikrotron:
It's getting close! I messed with it locally to try some out some of my suggestions.  Maybe the branch will help you understand what I was getting at (I definitely broke at least the "Nested" story, maybe others on this branch.  But the "Accessibility" and "Inline" stories are working):

https://github.com/acl-services/paprika/compare/uxd-157-formElement-api-refactor-2...uxd-157-formElement-api-refactor-2--mike2

packages/Checkbox/src/Checkbox.js
tristanjasper: We can avoid the console warnings if we delete the prop, I don't see why we wouldn't do preemptively do that

packages/Checkbox/src/Checkbox.js
tristanjasper: Yes, not necessary when we are spreading the `a11yProps` on the `Checkbox.Input` itself (https://github.com/acl-services/paprika/pull/774/files/d3d5d79b37a2579b3b51be592fca03140a7395d0#diff-412375913c3db82b1954e7b42b97f36557cef23c5dbb7c5bcae4fe3d00066177R16)

packages/FormElement/README.md
tristanjasper: Yes but then we get warnings about `refLabel` not being an supported attribute of an input element. Which is perhaps fine 🤷‍♂️  But i would have thought we would want to avoid. I would think most of the time developers would be using the paprika `Input` element instead which deletes the `refLabel` prop for them...

packages/FormElement/src/components/Description/Description.js
tristanjasper: I think it makes more sense to only generate these variables if they are needed, (Dictated by what gets added via the api). I don't think we need to create ids for components if they are not used.

packages/FormElement/src/components/Layout/index.js
tristanjasper: Yes but I was trying to keep consistent with the approach of the api, eg not having to use the `.` syntax. But I perhaps it's fine as they are all related subcomponents so perhaps in this case it does make it more clearer associating them in this way, will go with that

packages/FormElement/src/FormElement.js
tristanjasper: Ok i've switched it

packages/FormElement/README.md
mikrotron: I see 🤔 

Well, I think the README should be focused exclusively on the most common use cases, we can use the Docs page or the Examples stories to demonstrate these edgier cases.  Maybe just explain this in the `NativeInput` story, and cut this example from README.

packages/Checkbox/src/Checkbox.js
mikrotron: Ah... I have mixed feelings about that.  On the one hand, you're right, we shouldn't promote a technique that produces console warnings.  

On the other hand, they're _just warnings_ and I'm not keen on the `<Checkbox>` knowing details about the `<FormElement>` implementation.  What if the `FormElement` changes again, will anyone know to come looking in the `Checkbox` to remove or change this code?

Maybe @nahumzs or @alexzherdev have some opinions or suggestions about this detail?

packages/FormElement/src/components/Error/Error.js
tristanjasper: 👌

packages/FormElement/src/components/Label/Label.js
tristanjasper: 👌

packages/FormElement/stories/examples/NestedExample.js
tristanjasper: Not really, just showing the difference when using each type, native or Paprika Input

packages/Checkbox/src/Checkbox.js
tristanjasper: Good point, Ok sure I can remove but yep we shall get warnings and unwanted attributes being added to elements but perhaps that is acceptable 🤷‍♂️

packages/Checkbox/src/Checkbox.js
```diff
@@ -41,16 +31,19 @@ const Checkbox = props => {
...
     checked: checkedState === types.checkboxStates.CHECKED,
     disabled: isDisabled,
-    id: checkboxId,
     onChange,
     ref: inputRef,
     tabIndex,
     type: "checkbox",
     ...extendedInputProps,
+    id: checkboxId,
```
tristanjasper, line 46: we need to make sure the id that the `Checkbox` or `Radio` component generates is used on the input

packages/Checkbox/src/Checkbox.js
alexzherdev: no strong feelings but I'm leaning towards what @mikrotron said

packages/FormElement/stories/examples/InlineContent.js
```diff
@@ -0,0 +1,84 @@
...
+                    placeholder="Form placeholder"
+                    hasError={Boolean(errorText.length)}
+                    {...a11yProps}
+                  />
+                  <Input
+                    onChange={handleChange}
+                    value={value}
+                    placeholder="Form placeholder"
+                    hasError={Boolean(errorText.length)}
+                    aria-describedby={a11yProps["aria-describedby"]} // Only aria-describedby applied to subsequent Content components
```
nahumzs, line 63: 💯 👏 🏆

packages/FormElement/src/FormElement.js
```diff
@@ -1,118 +1,47 @@
...
-  const ariaErrorId = React.useRef(uuidv4()).current;
-  const ariaInstructionsId = React.useRef(uuidv4()).current;
-  const uniqueInputId = React.useRef(uuidv4()).current;
-  const hasError =
-    !!extractedChildren[subComponentDisplayNames.Error] &&
-    !!extractedChildren[subComponentDisplayNames.Error].props.children;
-
+  const { id, children, isDisabled, size, hasFieldSet, ...moreProps } = props;
+  const [ariaDescribedBy, setAriaDescribedBy] = React.useState({});
+  const uniqueInputId = React.useRef(nanoid()).current;
```
nahumzs, line 20: This should be `React.useState(nanoid)` is how their library suggested to use it.

nahumzs (dismissed):

packages/Checkbox/src/Checkbox.js
tristanjasper: Ok, I think for now i've removed those prop deletions, but in future if it becomes more of an issue we can always patch it using the propsCollector to prevent `refLabel` being added to the element.

nahumzs (changes requested):
hold while releasing

nahumzs (dismissed):

packages/FormElement/stories/examples/Nested.js
```diff
@@ -1,59 +1,91 @@
...
-import FormElement from "../../src/FormElement";
-
-const InputSet = styled.div`
-  display: flex;
-  > div {
-    margin-left: 8px;
-  }
-`;
+import FormElement from "../../src";
+import Fieldset from "../../src/components/Fieldset";
```
mikrotron, line 6: To demonstrate how we would use this in the wild, we should collapse this to a single `import` like:
```js
import FormElement, { Fieldset } from "../../src";
```

mikrotron:
For posterity (and a follow-up documentation cleanup task) can you also add in the PR description how the `isInline` prop was replaced by the `FormElement.Layout` component, and the `hasFieldset` prop has been retained, but the `<Fieldset>` component introduced for clearer JSX markup?

packages/FormElement/stories/examples/Nested.js
tristanjasper: Didn't work for me, I think i was getting unit test issues with that import pattern

mikrotron (approved):
llgtm 75

packages/FormElement/stories/examples/Nested.js
mikrotron: Really? The story works fine that way... 🤔 
Anyway, this is a minor detail we can adjust in a follow up PR to improve docs.

PR #786, opened 2020-11-05
feat(Panels): UXD-499 Deprecate SidePanel and move to Panels

jamiek-acl:
### Purpose 🚀
Deprecate SidePanel and move to Panels

### Updates 📦


jamiek-acl:
Gonna call it `Panel` instead of `Panels`

Issue #794, opened 2020-11-11
OverflowMenu focus


jamiek-acl:
# Bug Report

## Current behavior
1. When you open the `OverflowMenu`, it automatically puts focus on the first item (https://github.com/acl-services/paprika/blob/master/packages/OverflowMenu/src/OverflowMenu.js#L72). Maybe that's fine when using the keyboard, but when using the mouse it is a distraction.  We prevent a similar behaviour using `what-input` on the `@paprika/Button` component and should do the same here: https://paprika.highbond.com/?path=/story/buttons-button-examples--focus).
![focus-mouse](https://user-images.githubusercontent.com/23224777/98750911-ed0a0480-2373-11eb-8afe-629abcde23be.gif)

2. When you try and tab out of an expanded `OverflowMenu` , the focus goes to the next element, but then jumps back to the collapsed `OverflowMenu`. Not cool.
![stealing-focus](https://user-images.githubusercontent.com/23224777/98750511-06f71780-2373-11eb-9b2c-f6269b1f99df.gif)

nahumzs:
> When you try and tab out of an expanded OverflowMenu , the focus goes to the next element, but then jumps back to the collapsed OverflowMenu. Not cool.

Should we check on the `blur` event `event.key === "Tab"` and then just close the menu and ignore the focus part?

jamiek-acl:
https://github.com/acl-services/paprika/pull/804

PR #804, opened 2020-11-16
Issue 794  overflow focus

jamiek-acl:
### Purpose 🚀
1. Dont focus on first element when expand OverflowMenu using mouse
2. Dont focus on OverflowMenu trigger when exit OverflowMenu by pressing the Tab key.

### Notes ✏️
This is related to this ticket: https://aclgrc.atlassian.net/browse/UXD-464
The two issues can be seen here: https://github.com/acl-services/paprika/issues/794

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/issue-794--overflow-focus

### Screenshots 📸
Before can be seen here: https://github.com/acl-services/paprika/issues/794

After:
after4

### DIFF
--- a/packages/OverflowMenu/package.json
+++ b/packages/OverflowMenu/package.json
@@ -26,7 +26,8 @@
     "@paprika/stylers": "^1.0.0",
     "@paprika/tokens": "^1.0.0",
     "prop-types": "^15.7.2",
-    "uuid": "^3.3.2"
+    "uuid": "^3.3.2",
+    "what-input": "^5.2.10"
   },
   "peerDependencies": {
     "@paprika/l10n": "^0.2.5",
--- a/packages/OverflowMenu/src/OverflowMenu.js
+++ b/packages/OverflowMenu/src/OverflowMenu.js
@@ -55,7 +55,7 @@ function OverflowMenu(props) {
     setFocusIndex(index);
   }
 
-  const handleCloseMenu = (key) => {
+  const handleCloseMenu = key => {
     setIsOpen(false);
 
     if (isConfirming) {
@@ -92,7 +92,7 @@ function OverflowMenu(props) {
   const overflowLastItemIndex =
     React.Children.toArray(
       extractedChildren.filter(
-        (child) =>
+        child =>
           child.type &&
           (child.type.displayName === "OverflowMenu.Item" || child.type.displayName === "OverflowMenu.LinkItem")
       )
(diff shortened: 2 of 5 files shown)


packages/OverflowMenu/package.json
```diff
@@ -26,7 +26,8 @@
     "@paprika/stylers": "^1.0.0",
     "@paprika/tokens": "^1.0.0",
     "prop-types": "^15.7.2",
-    "uuid": "^3.3.2"
+    "uuid": "^3.3.2",
```
nahumzs, line 29: can we use nanoid? is a smaller lib.

packages/OverflowMenu/src/OverflowMenu.js
```diff
@@ -71,11 +74,12 @@ function OverflowMenu(props) {
 
   const handleOpenMenu = () => {
     setIsOpen(true);
-    // https://github.com/acl-services/paprika/issues/316
-    // Todo Should focus the first item via an onAfterOpen event callback in popover
-    setTimeout(() => {
-      focusAndSetIndex(0);
-    }, 250);
+
+    if (document.querySelector("html").getAttribute("data-whatinput") === "keyboard") {
```
nahumzs, line 78: I think you can get the same with `document.querySelector("[data-whatinput]").dataset.whatIput`

packages/OverflowMenu/src/OverflowMenu.js
```diff
@@ -71,11 +74,12 @@ function OverflowMenu(props) {
...
+      setTimeout(() => {
```
nahumzs, line 79: it might safet if you do `window.requestAnimationFrame(() => {
  // or does this get triggers when the overflow menu is still visible?
})`

packages/OverflowMenu/src/OverflowMenu.js
```diff
@@ -108,7 +112,7 @@ function OverflowMenu(props) {
     } else if (event.key === "Enter" || event.key === " ") {
       // do nothing
```
nahumzs, line 113: is weird to have an if statement with an empty body 🤔 isn't?

packages/OverflowMenu/src/OverflowMenu.js
jamiek-acl: oh yeah, good idea

packages/OverflowMenu/src/OverflowMenu.js
jamiek-acl: @nahumzs are you suggesting that I change it to this?

```
  const handleOpenMenu = () => {
    setIsOpen(true);

    if (document.querySelector("[data-whatinput]").dataset.whatinput === "keyboard") {
      setTimeout(() => {
        window.requestAnimationFrame(() => {
          focusAndSetIndex(0);
        });
      }, 250);
    }
  };
```

packages/OverflowMenu/src/OverflowMenu.js
jamiek-acl: I think it's ok because it's purpose is to not go into the `else` block.

tristanjasper:
LGTM

packages/OverflowMenu/src/OverflowMenu.js
nahumzs: ohh you don't need the settimeout if you are using the requestAnimationFrame, but I'm suspecting that the real problem is that we don't have a callback for the popover which has animation to be called once the animation finished :/.

In that case requestAnimationFrame will be not sufficient.

jamiek-acl: Yeah replacing `setTimeout` with `requestAnimationFrame` does not work.

nahumzs (approved):

PR #808, opened 2020-11-17
RAPRM-926  Fix Radio.Group size prop 🐐

mikrotron:
### Purpose 🚀
The `<Radio.Group>` has a `size` prop that should be passed along to its `<Radio>` `children`.

### Notes ✏️
- `<Radio.Group>` was not including `size` when it did `React.cloneElement()`.
- `A11yText` prop was removed from `<Radio.Group>` – it was being implemented incorrectly as `aria-labelledby`, and had no effect.  The radio group needs to get a label from a parent `<FormElement hasFieldSet>`. 
- This _would_ be a `MAJOR` breaking change normally, but because it was already "broken", this can remain a `PATCH` update.
- Fixed the `Showcase` story, converted all stories to CSF, and split the examples in  the `Controlled` story into `Controlled` and `Uncontrolled` example stories.

### Updates 📦
- [x] PATCH (bug fix) change to `Radio`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/RAPRM-926--fix-radio-button/?path=/story/forms-radio--showcase

### References 🔗
https://aclgrc.atlassian.net/browse/RAPRM-926

### DIFF
--- a/packages/Radio/stories/examples/Controlled.js
+++ b/packages/Radio/stories/examples/Controlled.js
@@ -1,49 +1,32 @@
 import React from "react";
 import { action } from "@storybook/addon-actions";
-import { Rule } from "storybook/assets/styles/common.styles";
-import Heading from "@paprika/heading";
-import { RadioStory } from "../Radio.stories.styles";
+import { Gap } from "storybook/assets/styles/common.styles";
+import StoryHeading from "storybook/components/StoryHeading";
 import Radio from "../../src/Radio";
 
-export const radioItems = ["Radio 1", "Radio 2 label", "Radio 3 option"];
+const radioItems = ["Radio 1", "Radio 2", "Radio 3"];
 
-const ExampleStory = props => {
+const ExampleStory = () => {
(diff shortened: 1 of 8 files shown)


packages/Radio/stories/examples/Controlled.js
```diff
@@ -1,49 +1,32 @@
...
-    setCheckedIndex(Number(e.target.value));
-  };
+
+  function handleIndexChange(event) {
+    setCheckedIndex(Number(event.target.value));
+  }
+
+  function handleRadioChange(activeIndex) {
+    setCheckedIndex(activeIndex);
+    action("Radio selection changed to index ")(activeIndex);
```
oscarkwan, line 18: is the space on purpose?

mikrotron: Yeah, it will display the index after that string, so it is easier to read this way

oscarkwan (dismissed):

oscarkwan (approved):

PR #809, opened 2020-11-18
UXD-531 CopyInput component

robyn3choi:
### Purpose 🚀
Added a new CopyInput component, along with stories and jest test.

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-531--copyinput

### Screenshots 📸
On hover:
image

On Click:
image

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-531


packages/CopyInput/src/CopyInputInputPropsCollector.js
```diff
@@ -0,0 +1,6 @@
+// Shell component for enhancing DX (developer experience)
+export default function CopyInputInputPropsCollector() {
```
nahumzs, line 2: this one should return `<></>` a fragment so ReatDocGen can pick it up and report it back in the ReadMe File otherwise DocGen will interpret the null as a regular function and not a React Component. 

After doing that you can go to `package.json` and add it as a subcomponent so the generateReadMe file can find it while visiting the package like here: https://github.com/acl-services/paprika/blob/master/packages/DatePicker/package.json#L37

nahumzs, line 2: `nit` I would rename this one to Input there is not really a reason to describe it as a PropCollector. Just a few other components are named like that, for example, the DatePicker was renamed by Allison a couple of days ago. From PropCollectors to just Input, cc @mikrotron

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,69 @@
...
+function CopyInput(props) {
+  const { defaultValue, children, ...moreProps } = props;
+  const I18n = useI18n();
+  const extendedInputProps = extractChildrenProps(children, CopyInputInputPropsCollector);
+
+  const [value, setValue] = React.useState(defaultValue);
+  const [isClickedTooltipOpen, setIsClickedTooltipOpen] = React.useState(false);
+
+  function handleButtonClick() {
+    navigator.clipboard.writeText(value);
```
nahumzs, line 21: can you add a note on the ReadMe file that this doesn't support `ie11`? https://caniuse.com/mdn-api_navigator_clipboard 

This might be fine for now since this is primarily for next-gen AN which has end support for ie11 but if this would be used in results or another part of our platform this might not be enough. 

you would need to do something like this to make it work in IE11 as well:

```js
function copyToClipboard(element) {
    var $temp = $("<input>");
    $("body").append($temp);
    $temp.val($(element).text()).select();
    document.execCommand("copy");
    $temp.remove();
}
```

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,69 @@
...
+    setIsClickedTooltipOpen(true);
+    setTimeout(() => setIsClickedTooltipOpen(false), 5000);
+  }
+
+  return (
+    <sc.CopyInput data-pka-anchor="copy-input" {...moreProps}>
+      <Input defaultValue={defaultValue} onChange={e => setValue(e.target.value)} {...extendedInputProps} />
```
AndreyChernykh, line 28: What if a user of this component wants to use the `value`. E.g.

```js
const [controlledValue, setControlledValue] = React.useState(defaultValue);

<CopyInput defaultValue="Copy me">
  <CopyInput.Input isReadOnly value={controlledValue} onChange={e => setControlledValue(e.target.value)} />
</CopyInput.Input>
```

In that case user would have to override the default `onChange`, hence, the `value` inside of the `CopyInput` will not be updated. And that will defeat the purpose of the component.

packages/CopyInput/README.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+Providing more props to the internal `Input`
+
+```js
+import CopyInput from "@paprika/copy-input";
+
+<CopyInput defaultValue="Copy me">
+  <CopyInput.Input isReadOnly />
+</CopyInput.Input>;
```
AndreyChernykh, line 48: As a user, I'd love to see a real-world usage example, including controlling the Input.

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,69 @@
...
+      <Popover
+        align="bottom"
+        isDark={!isClickedTooltipOpen}
+        isEager={!isClickedTooltipOpen}
+        isOpen={isClickedTooltipOpen || null}
```
AndreyChernykh, line 33: why `|| null` instead of `|| false`? Shouldn't `isOpen` be a boolean?

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,69 @@
...
+  const I18n = useI18n();
+  const extendedInputProps = extractChildrenProps(children, CopyInputInputPropsCollector);
+
+  const [value, setValue] = React.useState(defaultValue);
+  const [isClickedTooltipOpen, setIsClickedTooltipOpen] = React.useState(false);
+
+  function handleButtonClick() {
+    navigator.clipboard.writeText(value);
+    setIsClickedTooltipOpen(true);
+    setTimeout(() => setIsClickedTooltipOpen(false), 5000);
```
nahumzs, line 23: Maybe you would want to move this into a sideEffect? so can be clean after the component is removed?

Let's suppose that this component is used in a sidepanel, the user clicks the button to copy but then close the panel.

What will happens is that the setTimeout will execute once the side panel is not there and will try to `setIsClickedTooltipOpen(false)` in a component that doesn't exist creating a minor memory leak. 


you could do something like this:

```js
// inside your component
const refSetTimeout = React.useRef(null);

React.useEffect(() => {
  if(isClickedTooltipOpen) {
     refSetTimeout.current = setTimeout(()=>{
     refSetTimeout.current = null
     ... // more stuff
     }, 5000)
  }

   return () => {
	if (refSetTimeout.current) {
       clearTimeOut(refSetTimeout.current)
    }
   }
}, [isClickedTooltipOpen])
```

nahumzs: ^ the return function in the effect is the cleaning function that will be executed before the component will be unmounted.

packages/CopyInput/src/CopyInput.js
robyn3choi: We're going to use this component in Robots as well, but I'm not sure if Robots will need to support ie11 either since it's going to be pretty tied with the flow builder.

packages/CopyInput/src/CopyInput.js
nahumzs: I was going to add this as my feedback as well, this component seems that should be a controlled component.

robyn3choi: thanks for catching that, i'll fix it

packages/CopyInput/src/CopyInput.js
nahumzs: Ohh I see, so maybe two options:
1. Testing in ie11 is not that simple with our current gear because the storybook doesn't render in ie11 :/, so a possibility is let it like this add a note in the readme pointing to this PR and if we later need to support it, we can let an ux-dev add the support and properly test it in ie11.

2. We replaced your current code with a similar that I'm suggesting and then we just asked a ux-dev later to test it in ie11 whenever they need to implemented in that env to check if it's working correctly.

What do you think?

packages/CopyInput/src/CopyInput.js
robyn3choi: It needs to be `null` or else the popover won't appear on hover (from `isEager`). This is a bit of a weird popover because it needs to render one way on hover, and another way on click.

nahumzs:
I just tested it by clicking the button but doesn't seem to be copying the text in the input for me.

packages/CopyInput/src/CopyInput.js
mikrotron: I think it makes the most sense to use code compatible with ie11 now.  Some customers are going to continue to use ie11, and will suffer degraded capabilities in some places, but if this works with vanilla js, why not?

I did implement a vanilla js version, with in important a11y detail (can you see it?) that you may want to copy over:
https://github.com/acl-services/paprika/blob/002d64e1d4dea17e1530c3ba2c25a3279cc47645/.storybook/components/CodeViewer/CodeViewer.helpers.js#L31

mikrotron:
> I just tested it by clicking the button but doesn't seem to be copying the text in the input for me.

Me too.  And I don't see the confirmation Popover either?
http://storybooks.highbond-s3.com/paprika/UXD-531--copyinput/?path=/story/forms-copyinput--showcase

robyn3choi:
> I just tested it by clicking the button but doesn't seem to be copying the text in the input for me.

@nahumzs That might be because of https://github.com/acl-services/paprika/pull/809#discussion_r526423452

robyn3choi:
> > I just tested it by clicking the button but doesn't seem to be copying the text in the input for me.
> 
> Me too. And I don't see the confirmation Popover either?
> http://storybooks.highbond-s3.com/paprika/UXD-531--copyinput/?path=/story/forms-copyinput--showcase

Wait it's very broken on storybook but it works for me locally

packages/CopyInput/src/CopyInput.js
mikrotron: It _is_ a tricky one! Currently the `isEager` version of the tooltip is wrapping the `<Button>` below with a `<RawButton>` (which isn't ideal), because the `<Popover.Trigger>` doesn't have a function for a children.  Probably you'll want to use that render prop pattern.  

I recently revamped the Popover Storybook stories – I hope they now do a better job of explaining how the compoent works, and with clearer code examples. 

https://paprika.highbond.com/?path=/story/messaging-popover-docs--page

packages/CopyInput/src/CopyInput.styles.js
```diff
@@ -0,0 +1,19 @@
+import styled, { css } from "styled-components";
+import tokens from "@paprika/tokens";
+
+export const CopyInput = styled.div(
+  () => css`
```
mikrotron, line 5: If you don't need any of the `props` for styling, you don't need the function here.  You can just do:
```js
export const CopyInput = styled.div`
```
Followed by your css.

packages/CopyInput/stories/CopyInput.stories.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import { withKnobs } from "@storybook/addon-knobs";
+import { getStoryName } from "storybook/storyTree";
+import Showcase from "./examples/Showcase";
+import WithInputProps from "./examples/WithInputProps";
+
+const storyName = getStoryName("CopyInput");
+
+storiesOf(storyName, module)
```
mikrotron, line 10: Can we go with the Storybook compact story format (CSF) instead of using `storiesOf`?  

I actually don't mind if you skip the Showcase story entirely though, and put a few examples all in one story.  Whatever makes it easier for you to build the component.

robyn3choi: Sure, is there an example of a component with CSF stories? Doing a "find all" for CSF doesn't return anything really

mikrotron: Sure, I just refactored the `<Radio>` stories, you can check it out.  
https://github.com/acl-services/paprika/tree/master/packages/Radio/stories

You'll need a separate `stories.js` file for each "folder" in the Storybook navigation menu though, because the default export defines that folder.  So if you maintain both of your stories you'll need a `CopyInput.stories.js` and a `CopyInput.examples.stories.js`

packages/CopyInput/src/CopyInput.js
nahumzs: sound good if @robyn3choi can add that would be 💯 🙇

nahumzs:
> > > I just tested it by clicking the button but doesn't seem to be copying the text in the input for me.
> > 
> > 
> > Me too. And I don't see the confirmation Popover either?
> > http://storybooks.highbond-s3.com/paprika/UXD-531--copyinput/?path=/story/forms-copyinput--showcase
> 
> Wait it's very broken on storybook but it works for me locally

:/ yeah sometimes this happens, the Jedi team hasn't had time to check this out. If you can add an images of the component working would be enough :)

nahumzs:
:100: :party-dino: one minor thing, I think you need to run again generateReadMe script so can pick up the subcomponent @robyn3choi

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,99 @@
...
+    document.execCommand("copy");
+    document.body.removeChild(textbox);
+
+    setIsClickedTooltipOpen(true);
+    setIsHoverTooltipOpen(false);
+  }
+
+  return (
+    <sc.CopyInput data-pka-anchor="copy-input" {...moreProps}>
+      <Input ref={inputRef} defaultValue={defaultValue} {...extendedInputProps} />
```
nahumzs, line 49: ha! https://github.com/acl-services/paprika/blob/master/packages/Input/src/Input.js#L54 I was wondering how is that react it is not complaining about having a defaultValue and a value pass at the same time for and uncontrolled and controlled component, seems like the Input component removes the defaultValue if a value is passed down. :magic:

robyn3choi:
I just found out that the Input is supposed to always be read-only, so I'll make those changes shortly

nahumzs:
@robyn3choi @mikrotron shouldn't be by default read-only but with the ability to make it optional?

robyn3choi:
> @robyn3choi @mikrotron shouldn't be by default read-only but with the ability to make it optional?

@nahumzs  I'd be alright with that

mikrotron:
> @robyn3choi @mikrotron shouldn't be by default read-only but with the ability to make it optional?

From @jonathanjakimon:
> For the moment we always used it in as `read only`. My recommendation is to follow this way and one day if we have an exception, we will think about it (it will probably never happen)

But the difference in the coding might be trivial to allow it.

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,99 @@
...
+      </Popover>
+    </sc.CopyInput>
+  );
+}
+
+const propTypes = {
+  /** Used for CopyInput.Input */
+  children: PropTypes.node,
+  /** Default value for the input. */
+  defaultValue: PropTypes.string,
```
mikrotron, line 85: I think we should just rename this to `value`.

packages/CopyInput/src/CopyInput.js
mikrotron: If you add a `isReadOnly` and `value` _after_ the `extendedInputProps`, then there's no way to make this into an uncontrolled, editable component.  But if you add them _before_ then you can probably still achieve it.  
I didn't try, but I _think_ if the consumer put `<CopyInput.Input value={null} defaultValue={something} isReadOnly={false} />` it would work.

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,99 @@
...
+    document.body.removeChild(textbox);
+
+    setIsClickedTooltipOpen(true);
+    setIsHoverTooltipOpen(false);
+  }
+
+  return (
+    <sc.CopyInput data-pka-anchor="copy-input" {...moreProps}>
+      <Input ref={inputRef} defaultValue={defaultValue} {...extendedInputProps} />
+      <div ref={buttonRef}>
```
mikrotron, line 50: Let's add a `data-pka-anchor='copy-input.button'` here.

We _could_ add a `data-pka-anchor='copy-input.input'` to the `<Input>` as well maybe, or for symmetry, wrap it in a `div` with that anchor as well. But this wouldn't be strictly necessary, because the `Input` is still selectable with its own default anchor.

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,99 @@
...
+        <Button.Icon
+          kind="primary"
+          onClick={handleButtonClick}
+          onMouseEnter={() => setIsHoverTooltipOpen(true)}
+          onMouseLeave={() => setIsHoverTooltipOpen(false)}
```
mikrotron, line 55: Why not `onMouseOver` / `onMouseOut`?

mikrotron: This should also have:
```js
onFocus={() => setIsHoverTooltipOpen(true)}
onBlur={() => setIsHoverTooltipOpen(false)}
```
For keyboard only operation.

mikrotron: Also, for screen readers, we should use the `a11yText` prop here.  When the `<Popover.Trigger>` render prop is used, the popover content can be associated with the supplied `a11yAttributes` value, but since you don't use the Trigger, this should work just fine:

```js
a11yText={isClickedTooltipOpen ? I18n.t("copyInput.clicked_tooltip") : I18n.t("copyInput.hover_tooltip")}
```

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,99 @@
...
+        >
+          <CopyIcon />
+        </Button.Icon>
+      </div>
+      <Popover
+        align="bottom"
+        isDark
+        isOpen={isHoverTooltipOpen && !isClickedTooltipOpen}
```
mikrotron, line 63: This should have `isEager` to get the keyboard focus working right.

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,99 @@
...
+        getPositioningElement={() => buttonRef.current}
+      >
+        <Popover.Content>
+          <Popover.Card>{I18n.t("copyInput.hover_tooltip")}</Popover.Card>
+        </Popover.Content>
+        <Popover.Tip />
+      </Popover>
+      <Popover align="bottom" isOpen={isClickedTooltipOpen} getPositioningElement={() => buttonRef.current}>
```
mikrotron, line 71: This should have `shouldKeepFocus`, also for proper keyboard focus.

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,99 @@
...
+        <Popover.Content>
+          <Popover.Card>{I18n.t("copyInput.clicked_tooltip")}</Popover.Card>
+        </Popover.Content>
+        <Popover.Tip />
+      </Popover>
+    </sc.CopyInput>
```
mikrotron, line 77: And finally, one missing bit of accessibility, the "copied" message isn't getting read by the screen reader 😞.

But, I have a one line solution for you:
```js
{isClickedTooltipOpen && <Toast kind={Toast.types.kind.VISUALLY_HIDDEN} canAutoClose aria-hidden>Copied</Toast>}
```

packages/CopyInput/src/CopyInput.styles.js
```diff
@@ -0,0 +1,17 @@
+import styled from "styled-components";
+import tokens from "@paprika/tokens";
+
+export const CopyInput = styled.div`
+  display: flex;
+  [data-pka-anchor="heading"] {
+    margin-bottom: 0;
+  }
```
mikrotron, line 8: Confused by this...? Was it an experiment?

packages/CopyInput/src/CopyInput.styles.js
```diff
@@ -0,0 +1,17 @@
...
+
+  input.form-input__input {
```
mikrotron, line 10: That `className` is an artifact left over from the migration from `acl-ui`.  We will likely remove it at some point.  You could use a `[data-pka-anchor="input"]` selector, but actually I think just `input` is fine too.

mikrotron:
And... one more thing. Can you please add a `CHANGELOG.md` with this template:
```
# Changelog

All notable changes to this project will be documented in this file.

The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).

## [Unreleased]
```

packages/CopyInput/src/CopyInput.styles.js
robyn3choi: hmm mustve been leftover from a copy pasta

packages/CopyInput/src/CopyInput.styles.js
robyn3choi: I actually needed 2 classnames here or else the border-radius would get overridden by `Input` component styling. Maybe I could use `input[data-pka-anchor="input"]`?

packages/CopyInput/src/CopyInput.js
```diff
@@ -46,48 +47,65 @@ function CopyInput(props) {
...
+          onMouseOver={() => setIsHoverTooltipOpen(true)}
+          onMouseOut={() => setIsHoverTooltipOpen(false)}
+          onFocus={() => setIsHoverTooltipOpen(true)}
+          onBlur={() => setIsHoverTooltipOpen(false)}
         >
           <CopyIcon />
         </Button.Icon>
       </div>
       <Popover
+        a11yText={isClickedTooltipOpen ? I18n.t("copyInput.clicked_tooltip") : I18n.t("copyInput.hover_tooltip")}
```
mikrotron, line 64: This needs to go on the `<Button.Icon>` actually.

packages/CopyInput/src/CopyInput.js
```diff
@@ -46,48 +47,65 @@ function CopyInput(props) {
...
+      >
         <Popover.Content>
           <Popover.Card>{I18n.t("copyInput.clicked_tooltip")}</Popover.Card>
         </Popover.Content>
         <Popover.Tip />
       </Popover>
+      {isClickedTooltipOpen && (
+        <Toast kind={Toast.types.kind.VISUALLY_HIDDEN} aria-hidden>
+          Copied
+        </Toast>
```
mikrotron, line 90: My bad, I thought I checked this, but the `aria-hidden` is counter productive it seems.  Let's get rid of that and instead use `canAutoClose autoCloseDelay={1000}` to make it disappear from the DOM after it gets read by the screen reader.  

If this seems really weird, I can explain what's going on offline.

mikrotron: Oh, and it needs the actual `I18n.t("copyInput.clicked_tooltip")` too.

packages/CopyInput/src/CopyInput.js
```diff
@@ -46,48 +47,65 @@ function CopyInput(props) {
...
   );
 }
 
 const propTypes = {
   /** Used for CopyInput.Input */
   children: PropTypes.node,
-  /** Default value for the input. */
-  defaultValue: PropTypes.string,
+  /** Is the input read-only. */
+  isReadOnly: PropTypes.bool,
```
mikrotron, line 100: This _could_ simply be added to the `<CopyInput.Input>` without being an explicit part of the `<CopyInput>` API.  Or did you have a reason you thought it would be good to have here?

packages/CopyInput/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@paprika/copy-input",
+  "version": "1.0.0",
```
mikrotron, line 3: @nahumzs should this be set to a lower version so we can "promote" it when we publish?

packages/CopyInput/src/CopyInput.js
```diff
@@ -0,0 +1,117 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Input from "@paprika/input";
+import Popover from "@paprika/popover";
+import Button from "@paprika/button";
+import Toast from "@paprika/toast";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import extractChildrenProps from "@paprika/helpers/lib/extractChildrenProps";
+import CopyIcon from "@paprika/icon/lib/ArrowDown"; // TODO: Replace with actual icon when ready
```
mikrotron, line 9: If you want, I can help you do this in this PR (though a follow-up PR will be required in the `wasabicons` repo). 

Otherwise it can be a separate sub-task / PR.

packages/CopyInput/src/CopyInput.js
robyn3choi: Hmm i guess it might as well be just be a prop for `CopyInput.Input`, since the user probably wants it to be a controlled component if the input is editable. But since `isReadOnly` should default `false`, it's a bit weird to have it as a `CopyInput.Input` prop, since the user would expecting the normal `Input` default props. If we have `isReadOnly` props is at the root level component, the API is more clear.

packages/CopyInput/src/CopyInput.js
robyn3choi: But it feels weird Button's a11yText would change to "Copied" after you click it

packages/CopyInput/src/CopyInput.js
robyn3choi: Let's do it in a different subtask

packages/CopyInput/src/CopyInput.js
mikrotron: Oh, good point! And we have the toast already to announce copied. 
It should just be the hover tooltip instead! 🤦

packages/CopyInput/src/CopyInput.js
mikrotron: Fair point. I'll let you make the decision – it's your component 😄

robyn3choi: Alright, i'll leave it as it is. Users who just want an uncontrolled editable CopyInput can rejoice since they won't need to give the component any children

packages/CopyInput/package.json
nahumzs: I think is fine it will get published with 1.0.0-alpha and when promoted will be 1.0.0 👍

mikrotron (approved):
llgtm 90

nahumzs:
👏 💯 😄

PR #815, opened 2020-11-19
Uxd 546 panel from below

jamiek-acl:
### Purpose 🚀
Add ability for the Panel component to slide up from the bottom.
The Panel can leave a gap on the left or on the right.

### Notes ✏️
Made some breaking changes to prop names.

### Updates 📦
- [x] MAJOR (breaking) change to Panel

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-546-panel-from-below

### Screenshots 📸

with-sidenav


packages/Panel/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/panel",
-  "version": "1.0.4-alpha.2",
+  "version": "2.0.0",
```
nahumzs, line 3: 🤔 @allison-c its ok to update the version manually?

packages/Panel/src/Panel.js
```diff
@@ -25,23 +26,32 @@ export default function Panel(props) {
...
     onAfterClose,
     onAfterOpen,
     onClose,
+    slideFrom,
     width,
     zIndex,
     ...moreProps
   } = props;
 
+  const offsetTop = Object.prototype.hasOwnProperty.call(offset, "top") ? offset.top : 0;
```
nahumzs, line 44: you can do `if ("top" in offset)` will do the same but it's shorter.

packages/Panel/src/Panel.js
```diff
@@ -25,23 +26,32 @@ export default function Panel(props) {
...
+
   // Hooks
   const [isVisible, setIsVisible] = React.useState(isOpen);
-  const offsetScroll = useOffsetScroll(offsetY);
+  const offsetScroll = useOffsetScroll(offsetTop);
+
+  const _offset = {
+    top: offsetScroll,
+    left: Object.prototype.hasOwnProperty.call(offset, "left") ? offset.left : 0,
```
nahumzs, line 52: same for this use `in` is easier to read IMO.

packages/Panel/src/Panel.js
```diff
@@ -192,52 +209,60 @@ const propTypes = {
...
-  /** Control if the side panel slides from the left */
-  isSlideFromLeft: PropTypes.bool,
-
   /** Modify the look of the Panel */
   kind: PropTypes.oneOf([Panel.types.kind.DEFAULT, Panel.types.kind.CHILD, Panel.types.kind.PRIMARY]),
 
-  /** Control y offset of the panel */
-  offsetY: PropTypes.number,
+  /** Control offset of the Panel. Only use 'top' when sliding in from the left or right. Only use 'left' or 'right' when sliding in from the bottom. */
+  offset: PropTypes.shape({ top: PropTypes.number, left: PropTypes.number, right: PropTypes.number }),
```
nahumzs, line 228: can you have offset from both left and right when sliding from the bottom?

packages/Panel/package.json
allison-c: If you want to update the version manually better to run `yarn lerna publish --force-publish=@paprika/panel`, we shouldn't need to manually update the verison in package.json

packages/Panel/src/Panel.js
```diff
@@ -192,52 +209,60 @@ const propTypes = {
...
   onAfterOpen: PropTypes.func,
 
-  /** Callback triggered when the side panel needs to be close */
+  /** Callback triggered when the Panel needs to be close */
   onClose: PropTypes.func,
 
-  /** The width of the open panel. */
+  /** Control where the Panel slides in from */
+  slideFrom: PropTypes.oneOf([
+    Panel.slideFromDirections.RIGHT,
```
nahumzs, line 241: This should follow the convention of `Panel.types.slide.RIGHT`

just add a declaration inside of Panel.js and do

```js	
Panel.types = {
  slide: {
    RIGHT: "right",
    LEFT: "left",
    etc
  }
}
```

packages/Panel/src/helpers.js
```diff
@@ -25,3 +27,29 @@ export function extractChildren(children, types) {
 
   throw new Error("extractChildren types parameter must be an Array");
 }
+
+export function warnOfPropErrors(props) {
```
nahumzs, line 31: 💯 extra Dev Experience!!

packages/Panel/src/Panel.js
jamiek-acl: Yes that would work

packages/Panel/src/Panel.js
```diff
@@ -177,6 +191,7 @@ export default function Panel(props) {
 
 Panel.types = {
   kind: types.sidePanelKinds,
+  slide: types.slide,
```
nahumzs, line 194: this should be `slideFrom: types.slide` it should be the same name of the prop that it's representing.

packages/Panel/src/Panel.js
```diff
@@ -25,23 +25,32 @@ export default function Panel(props) {
...
   } = props;
 
+  const offsetTop = "top" in offset ? offset.top : 0;
+
   // Hooks
   const [isVisible, setIsVisible] = React.useState(isOpen);
-  const offsetScroll = useOffsetScroll(offsetY);
+  const offsetScroll = useOffsetScroll(offsetTop);
+
+  const _offset = {
```
mikrotron, line 49: `currentOffset`? `calculatedOffset`? 
What would best distinguish this value from the prop value?

jamiek-acl: The underscore?  😄

nahumzs (approved):

nahumzs:
Looklike the types are not being exported correctly @jamiek-acl and is breaking the story at least on paprika.highbond

packages/Panel/README.md
```diff
@@ -22,23 +22,24 @@ npm install @paprika/panel
...
+| onAfterOpen           | func                                                                                                 | false    | () => {}                        | Callback once the Panel has been opened event                                                                                                   |
+| onClose               | func                                                                                                 | false    | null                            | Callback triggered when the Panel needs to be close                                                                                             |
+| slideFrom             | [ Panel.slideFromDirections.RIGHT, Panel.slideFromDirections.LEFT, Panel.slideFromDirections.BOTTOM] | false    | Panel.slideFromDirections.RIGHT | Control where the Panel slides in from                                                                                                          |
```
nahumzs, line 40: I think you need to re-run generatReadme, also it should be:

`types.slideFroms`=> `types.slideFrom` because you match the prop name

nahumzs (approved):

PR #817, opened 2020-11-24
UXD-539 locked items in column arrangement

allison-c:
### Purpose 🚀

image

### Notes ✏️

- Added `isDragDisabled` and `handle` prop on `<Sortable.item />` to be able to disable the drag and customize the icon (or any other element consumer want to put there)
- Added some logic in `useColumnsArrangement ` hook to support `disabledColumnIds`
- Removed the tip and changed the offset of the popovers in the action bar

### Updates 📦
- [x] MINOR (backward compatible) change to `@paprika/sortable` `@paprika/action-bar`

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/ActionBar/src/hooks/useColumnsArrangement/useColumnsArrangement.js
```diff
@@ -1,18 +1,29 @@
...
+    const movedChild = newOrder.splice(source, 1);
+    newOrder.splice(destination, 0, ...movedChild);
+
+    return disabledColumnIds.every(
+      disabledColumnId => defaultOrder.indexOf(disabledColumnId) === newOrder.indexOf(disabledColumnId)
+    );
+  }
+
+  function isColumnHidden(columnId) {
+    return hiddenColumnIds.indexOf(columnId) > -1;
```
nahumzs, line 21: `nit` 
```js
hiddenColumnIds.includes(columnId)
```

nahumzs: I don't know why we don't do it, but I think we should start using more Javascript Sets, this and remove would be super simple if we were using sets instead of arrays. 

`SomeSet.has(columnId) // :P`

nahumzs:
The UI looks great 🐱.

packages/ActionBar/stories/ShowcaseApp/ShowcaseApp.js
```diff
@@ -69,16 +69,10 @@ export default function App() {
...
-    "name",
-    "status",
-    "country",
-    "joined",
-    "shareable",
-    "level",
-    "position",
-  ]);
+  const { orderedColumnIds, isColumnHidden, ...handlers } = useColumnsArrangement(
+    ["goals", "name", "status", "country", "joined", "shareable", "level", "position"],
```
nahumzs, line 73: have you considered to use an object instead of parameters for useColumnArragment?

const order = ["goals", "name", "status", "country", "joined", "shareable", "level", "position"];
const disable = ["name"]

```js
const { orderedColumnIds, isColumnHidden, ...handlers } = useColumnsArrangement({ order, disable })
```

I think would be more flexible and future resilience in case we want to add more parameters. 

but I'm not super hard on this just a suggestion.

allison-c: I like it! Maybe a starling day for grouping the handlers/props returned in another object(`const {filteredData, filterProps} = useFilter()`), so they can be released together as a breaking change

packages/ActionBar/src/hooks/useColumnsArrangement/useColumnsArrangement.js
allison-c: I think usually the main reason for using an array is the order matters.I can convert `hiddenColumnIds ` to a set for sure!

packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
```diff
@@ -4,8 +4,10 @@ import Button from "@paprika/button";
 import Input from "@paprika/input";
 import Popover from "@paprika/popover";
 import Sortable from "@paprika/sortable";
+import HandleIcon from "@paprika/icon/lib/DragHandle";
```
jamiek-acl, line 7: You don't need this, since it is the default

packages/Sortable/src/components/SortableItem/SortableItem.js
```diff
@@ -10,27 +10,33 @@ import { itemStyles, itemIndexStyles, itemHandleStyles, itemBodyStyles, itemClos
...
+const SortableItem = ({ children, index, handle, hasNumbers, isDragDisabled, onRemove, ...moreProps }) => {
   const I18n = useI18n();
 
   const handleRemove = () => {
     onRemove(index);
   };
 
   return (
-    <Draggable draggableId={`draggable-${children.props.sortId}`} index={index}>
+    <Draggable draggableId={`draggable-${children.props.sortId}`} index={index} isDragDisabled={isDragDisabled}>
```
jamiek-acl, line 39: Can you please add a story to the `Sortable` component that demonstrates a locked Sortable.Item

jamiek-acl:
So I cannot move the `Name` column in your example, which is great, but that also means I cannot move the `Goals` column (as that would move the `Name` column).  Thoughts on that?

allison-c:
> So I cannot move the `Name` column in your example, which is great, but that also means I cannot move the `Goals` column (as that would move the `Name` column). Thoughts on that?

Because if you move the `Goals` column to the bottom, `Name` will be the first, which turns out `Name` has been moved even though you didn't drag it. I add this logic in the `useColumnsArrangement` hook. In the most common cases, disabled/fixed column will be always on the top (cause they will be sticky columns in the data grid), so the only problem will be you cannot drag a column to the position before a sticky column

jamiek-acl:
Right. So the order is:
```
Goals
Name (locked)
Country
...
```

What if I dragged `Country` to be the first one, and `Goals` automatically moved to the third spot (keeping `Name` in the second spot)?  May not be worth the hassle as you mentioned, if most often it is the first column that consumers are locking.

allison-c:
> Right. So the order is:
> 
> ```
> Goals
> Name (locked)
> Country
> ...
> ```
> 
> What if I dragged `Country` to be the first one, and `Goals` automatically moved to the third spot (keeping `Name` in the second spot)? May not be worth the hassle as you mentioned, if most often it is the first column that consumers are locking.

That works too, but will it be a little bit confusing if I didn't want to move `Goals` ? btw this behavior is customizable if they use their own hook.

jamiek-acl:
> That works too, but will it be a little bit confusing if I didn't want to move `Goals` ? btw this behavior is customizable if they use their own hook.

If `Goals` was locked, it wouldn't move either.

I just wanted to have the conversation; it's probably only worth addressing if consumers start requesting it/providing their own hook.

allison-c:
> > That works too, but will it be a little bit confusing if I didn't want to move `Goals` ? btw this behavior is customizable if they use their own hook.
> 
> If `Goals` was locked, it wouldn't move either.
> 
> I just wanted to have the conversation; it's probably only worth addressing if consumers start requesting it/providing their own hook.

I'll definitely check with Sarah once this is demonstrable on asset inventory!

jamiek-acl (approved):
925683124s

PR #818, opened 2020-11-24
UXD-543 Avoid double escaped strings

allison-c:
### Purpose 🚀

In the consumer apps, there might be some string like ` Product/Service` `<Product><Service>` since React already escaped the characters when we render all the components/elements. i18next doesn't need to escape them again. 

### Notes ✏️

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/l10n`

### Storybook 📕

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/L10n/src/i18n.js
+++ b/packages/L10n/src/i18n.js
@@ -9,6 +9,9 @@ export const i18n = locales => {
     fallbackLng: defaultLocale,
     lowerCaseLng: true,
     resources: locales || paprikaLocales,
+    interpolation: {
+      escapeValue: false,
+    },
   };
 
   const newInstance = i18next.createInstance(i18nConfig, err => {
--- a/packages/CopyInput/README.md
+++ b/packages/CopyInput/README.md
@@ -57,7 +57,6 @@ return (
 <!-- eoContent -->
 
 ## Links
-
 - [Storybook showcase](https://paprika.highbond.com/?path=/story/copy-input--showcase)
 - [Github source code](https://github.com/acl-services/paprika/tree/master/packages/CopyInput/src)
 - [Github create issue](https://github.com/acl-services/paprika/issues/new?label=[]&title=@paprika/copy-input%20[help]:%20your%20short%20description&body=%0A%23%20Help%20wanted%0A%0A%23%23%20Please%20write%20your%20question.%0A*A%20clear%20and%20concise%20description%20of%20what%20the%20question%20is*%0A%0A%23%23%20Additional%20context%0A*Add%20any%20other%20context%20or%20screenshots%20about%20your%20question%20here.*%0A)
--- a/packages/L10n/CHANGELOG.md
+++ b/packages/L10n/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Avoid double escaped strings.
+
 ## [0.4.1] - 2020-07-09
 
 ### Removed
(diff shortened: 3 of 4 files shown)


nahumzs:
Why would be this a breaking change?

packages/L10n/src/i18n.js
```diff
@@ -9,6 +9,9 @@ export const i18n = locales => {
     fallbackLng: defaultLocale,
     lowerCaseLng: true,
     resources: locales || paprikaLocales,
+    interpolation: {
+      escapeValue: false,
```
nahumzs, line 13: I don't think we want this @allison-c , why would we want to be false by default? 

```
escapeValue
(default) true
escapes passed in values to avoid XSS injection
```

doesn't `i18next.t('keyEscaped', { myVar: '', interpolation: { escapeValue: false } });` do what you want to?

[interpolation i18next](https://www.i18next.com/translation-function/interpolation)

allison-c: Yes, but there're too many places I need to add `interpolation: { escapeValue: false }` in asset inventory, maybe more in the future. e.g. {{assetName}} details,add new {{assetName}}.... we can also create a helper function for I18n.t with the configuration but I think fix it in paprika side it's easier for developers. 
BTW react will escape strings anyway even if we don't use i18next. As long as we are cautious when using dangerouslyInsert
What do you think if I add another prop on L10n component like `options` and pass it here? So this won't be a breaking change. I put it as a breaking change because some consumer app might already did something to address this issue?

jamiek-acl: But is it necessary, @nahumzs ?

This code:
```
export default function Greeting() {
  const i18n = useI18n();
  return (
    <p>Custom locale greeting: {i18n.t("customlocalegreeting", { name: "<script>alert('what up');</script>" })}</p>
  );
}
```
outputs this:
Screen Shot 2020-11-24 at 10 56 06 AM

allison-c: @jamiek-acl suggested an easier way to solve the issue by adding `{{-name}}` I'll close the PR 😅

nahumzs: @jamiek-acl because by default i18n escape it and react also do it. is a double layer of security as Allison commented. 

I don't think we should disable it by default or even allowed developers to disabled it globally even in their pillars. 


So this is my suggestion, expose a new method in i18n to ease the process for using it, like: 

```
// i18n paprika file
const dangerouslyT = (key, t, options = {})=> {
  return t(key, {interpolation: { escapeValue: false }, ...moreProps})
}
export dangerouslySetText

// your app

import {dangerouslyT, i18n} from "@paprika/l10n"

dangerouslyT("mykey", i18n.t)

```

nahumzs: even in this line https://github.com/acl-services/paprika/blob/master/packages/L10n/src/i18n.js#L18

you could do:

```js
export const i18n = locales => {
  const i18nConfig = {
    lng: defaultLocale,
    fallbackLng: defaultLocale,
    lowerCaseLng: true,
    resources: locales || paprikaLocales,
  };

  const newInstance = i18next.createInstance(i18nConfig, err => {
    if (err) return console.log("something went wrong loading i18next", err);
    return null;
  });

  newInstance.dangerously_t = (key, moreProps = {}) => {
     const {t} = newInstance
     return t(key, {interpolation: { escapeValue: false }, ...moreProps})
  }

  return newInstance;
};

```

nahumzs: yeah ok, but we should disable escaping at any level just in specific scenarios.

jamiek-acl: > is it a double layer of security

boxer

PR #819, opened 2020-11-25
feat(Tabs): UXD-474 add shouldUnmount prop

allison-c:
### Purpose 🚀

In asset inventory, users will be able to switch between 2 tab panels. Both of them contains collapsible card and closed by default. This PR avoids lots of implementation required on the app side to preserve the opening state and avoid re-fetch data.

### Notes ✏️

### Updates 📦
- [x] MINOR (backward compatible) change to `@paprika/tabs`

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/Tabs/src/components/Panel/Panel.js
```diff
@@ -5,25 +5,35 @@ import * as sc from "./Panel.styles";
 
 const propTypes = {
   children: PropTypes.node,
-  /** Controls if the option is selected or not */
 
+  /** Controls if the option is selected or not */
   isSelected: PropTypes.bool,
+
+  /** Should unmount or not when isSelected is false */
+  shouldUnmount: PropTypes.bool,
```
nahumzs, line 13: what about `preventUnmount` so it's aligned with preventDefault `javascript land`

packages/Tabs/src/components/Panel/Panel.js
```diff
@@ -5,25 +5,35 @@ import * as sc from "./Panel.styles";
...
 
-  if (!isSelected) {
+  if (!isSelected && shouldUnmount) {
     return null;
   }
 
   return (
-    <sc.Panel {...moreProps} role="tabpanel" data-pka-anchor="tabs.panel">
+    <sc.Panel
+      isSelected={isSelected}
```
nahumzs, line 31: I guess you should add aria-hidden="true" depending on shouldUnmount too? cc @mikrotron

packages/Tabs/src/components/Panel/Panel.js
allison-c: preventDefault is a function though, I thought we prefer to use `should` `has` for boolean value?

packages/Tabs/src/components/Panel/Panel.js
allison-c: There're a `display:none` in styles.js, so aria-hidden will be unnecessary

jamiek-acl (approved):

PR #826, opened 2020-11-30
feat(StatusTracker): UXD-470 add component

allison-c:
### Purpose 🚀

http://storybooks.highbond-s3.com/paprika/UXD-470-status-tracker/?path=/story/buttons-statustracker--showcase

Please note some colors are TBD, will adjust them while addressing the review comments

### Notes ✏️

# @paprika/status-tracker - 0.0.1

## Description

StatusTracker component.

## Installation

```
yarn add @paprika/status-tracker
```

or with npm:

```
npm install @paprika/status-tracker
```

### StatusTracker

| children | node | false    | null    |             |

### StatusTracker.Point

| Prop        | Type                                       | required | default      | Description                                                           |
| ----------- | ------------------------------------------ | -------- | ------------ | --------------------------------------------------------------------- |
| children    | node                                       | false    | null         |                                                                       |
| description | node                                       | false    | null         | The description of current status point will be shown in the tooltip. |
| kind        | [ kinds.CURRENT, kinds.FUTURE, kinds.PAST] | false    | kinds.FUTURE | Kind of current status point.                                         |
| name        | node                                       | false    | null         | The name of current status point will be shown in the tooltip.        |

```jsx
import StatusTracker from "@paprika/status-tracker";
import OverflowMenu from "@paprika/overflow-menu";

<StatusTracker>
  <StatusTracker.Point name="Draft" kind={StatusTracker.types.kind.PAST} description={getDesc()} />
  <StatusTracker.Point
    name="In review"
    description={
      <>
        <span>10 mins ago</span>
        <span>by someone</span>
      </>
    }
    kind={StatusTracker.types.kind.CURRENT}
    ref={currentPointRef}
  >
    <OverflowMenu>
      <OverflowMenu.Trigger isPending={isPending} />
      <OverflowMenu.Item onClick={handleClick}>First Item</OverflowMenu.Item>
      <OverflowMenu.Item onClick={handleClick}>Second Item</OverflowMenu.Item>
      <OverflowMenu.Divider />
      <OverflowMenu.Item onClick={handleClick}>View workflow</OverflowMenu.Item>
    </OverflowMenu>
  </StatusTracker.Point>
  <StatusTracker.Point name="Approve" kind={StatusTracker.types.kind.FUTURE} />
  <StatusTracker.Point name="Done" kind={StatusTracker.types.kind.FUTURE} />
</StatusTracker>;
```

image

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/status-tracker`

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/StatusTracker/src/components/Point/Point.js
```diff
@@ -0,0 +1,172 @@
...
+      </sc.Point>
+    );
+  }
+
+  return (
+    <sc.PointWrapper aria-current={kind === kinds.CURRENT} kind={kind} {...moreProps}>
+      <sc.Popover isEager isDark isOpen={isTooltipOpen}>
+        <Popover.Trigger>{getTrigger}</Popover.Trigger>
+        <Popover.Content>
+          <sc.PopoverCard hasOnlyName={!!name && !!description}>
```
nahumzs, line 121: nit: Boolean(name) && Boolean(description) is slightler cleaner than coercion?

packages/StatusTracker/src/components/Point/Point.js
```diff
@@ -0,0 +1,172 @@
...
+            {name && <sc.NameInTooltip data-pka-anchor="status-tracker.point.name">{name}</sc.NameInTooltip>}
```
nahumzs, line 122: this will fail on `"" && <sc.NameInTooltip data-pka-anchor="status-tracker.point.name">{name}</sc.NameInTooltip>` it will print the `""`. I know is a little more typing but going with the ternary is slighter safer. or change it to `Boolean(name)`

packages/StatusTracker/src/components/Point/Point.styles.js
```diff
@@ -0,0 +1,108 @@
...
+  &:last-child {
+    flex-grow: 0;
+
+    ::after {
+      content: none;
+    }
+  }
+`;
+
+const basicPoint = css`
```
nahumzs, line 41: `nit` currentPoint?

packages/StatusTracker/src/components/Point/Point.styles.js
```diff
@@ -0,0 +1,108 @@
...
+const pastPoint = css`
+  background-color: ${tokens.color.greenLighten10};
+`;
+
+const futurePoint = css`
+  background-color: ${tokens.color.blackLighten70};
+  border: 2px solid ${tokens.color.blackLighten20};
+`;
+
+export const Point = styled.div`
```
nahumzs, line 56: as standard this should be:

styled.div(({ kind }) => {
    switch (kind) {
      case types.kinds.PAST:
        return css`
          ${basicPoint}
          ${pastPoint}
        `;
      case types.kinds.CURRENT:
        return css`
          > [data-pka-anchor="popover"] > [data-pka-anchor="button"] {
            margin-top: -${tokens.spaceSm};
          }
        `;
      case types.kinds.FUTURE:
        return css`
          ${basicPoint}
          ${futurePoint}
        `;
      default:
        return null;
    }
  }} 
})

packages/OverflowMenu/src/OverflowMenu.js
```diff
@@ -49,14 +53,23 @@ function OverflowMenu(props) {
   const triggerId = React.useRef(uuid());
   const overflowListRef = React.useRef(null);
 
+  function getIsOpenValue() {
+    return controlledIsOpen === null ? isOpen : controlledIsOpen;
```
jamiek-acl, line 57: Can you please add a spec for these scenarios

packages/StatusTracker/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@paprika/status-tracker",
+  "version": "0.0.1",
+  "description": "StatusTracker component.",
```
jamiek-acl, line 4: Please elaborate a bit

packages/StatusTracker/src/StatusTracker.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Point from "./components/Point";
+import { kinds } from "./types";
+import * as sc from "./StatusTracker.styles";
+
+function StatusTracker(props) {
+  const { children } = props;
```
jamiek-acl, line 8: why not destructure one line up?

nahumzs:
Side question: 

If you click on active status and the overflow appears seems like the focus doesn't go to the overflow and then the up and down arrows doesn't work for the overflow.

1. I guess the consumer should deal with this?
2. Does the drop-down menu lack up and down arrows, implementation 😮?

packages/helpers/src/extractChildren.js
```diff
@@ -3,8 +3,13 @@ import React from "react";
 export default function extractChildren(children, types) {
   const _children = [];
   const components = {};
+
   if (Array.isArray(types)) {
-    React.Children.toArray(children).forEach(child => {
+    for (const child of React.Children.toArray(children)) {
```
jamiek-acl, line 8: What is the benefit of `for (child of ...` vs `.forEach(...`?

packages/StatusTracker/src/components/Point/Point.styles.js
allison-c: This has been shared in both past point and future point

packages/StatusTracker/src/StatusTracker.js
allison-c: It just in paprika we seem like always destructure inside the function

packages/helpers/src/extractChildren.js
allison-c: I had to return the final results early if there's React.Fragment

packages/StatusTracker/src/StatusTracker.js
nahumzs: ^ yeap is a paprika thing 🙏

nahumzs:
Seem a good alpha version for me, let's use it and see if we need to refine it. 

I still a little off about how the overflow works with the focus, that's something that we might want to fix in a next PR? cc @mikrotron

packages/helpers/src/extractChildren.js
alexzherdev: [this transpiled code tho](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=GYewTgBAFAxiB2BnALhAHhEwIE8CUEA3gL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.12.9&externalPlugins=)
image

nahumzs: wao, that's a lot of code for just the `const x of y` didn't know that, any way to just ship that for browsers that don't support it natively? @alexzherdev ?

nahumzs: seem like all that code is just for IE11 correct? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

allison-c: I can use `some` or `every` then!

allison-c:
> Side question:
> 
> If you click on active status and the overflow appears seems like the focus doesn't go to the overflow and then the up and down arrows doesn't work for the overflow.
> 
> 1. I guess the consumer should deal with this?
> 2. Does the drop-down menu lack up and down arrows, implementation 😮?

This issue has been fixed, it was because status tracker overrides the onClick handler on Overflow.Trigger.

packages/helpers/src/extractChildren.js
alexzherdev: > seem like all that code is just for IE11 correct? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

pretty much yeah

> wao, that's a lot of code for just the `const x of y` didn't know that, any way to just ship that for browsers that don't support it natively? @alexzherdev ?

paprika is transpiled and some level of browser support has to be set for that transpilation. I'm not aware of an easy way to publish two sets of artifacts (e.g. modern vs IE) that could be seamlessly consumed and further used in an app. It's possible to transpile paprika just enough to be evergreen-browser-compatible, and then configure apps to produce two bundles (where IE bundle would transpile paprika's code further). But that'd be a pretty significant change

nahumzs (approved):

PR #829, opened 2020-12-02
UXD-595 feature(ListBox): withTags adding ability to use an alternative renderTriggerPillLabel method

nahumzs:
### Purpose 🚀
Add ability for the ListBox (withTags) to use an alternative property apart from label. 

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-595-listbox-with-tags-small-fix

### Screenshots 📸

### References 🔗


jamiek-acl:
Can you please address these issues as well:

1. when there are lots of selections and the select grows in height, the expand/collapse arrow should remain on the first line -- not be centred
Screen Shot 2020-12-15 at 1 50 18 PM


2. when all the options have been selected, i need the ability to pass in a message that can be displayed (similar to how the "No search results found" message appears)
Screen Shot 2020-12-15 at 1 51 08 PM


3. pass through extra props.  I need to apply a class to `<ListBoxWithTags>` so I can set its height (my pills are tall, so the div is a different height when there are/are not options selected).  WithTags passes the props onto ListBox (https://github.com/acl-services/paprika/blob/master/packages/ListBox/src/components/WithTags/WithTags.js#L149), but ListBox doesnt have a `moreProps`.

nahumzs:
btw I will move this component into their own package, will be hard to keep with version control if we have it under the listbox

nahumzs:
> 1. when there are lots of selections and the select grows in height, the expand/collapse arrow should remain on the first line -- not be centred

Mmmm I don't know if this approach is incorrect or correct, if it's incorrect will be annoying to have to calculate the original height of the Trigger so we can calculate the position depending on the size of the Trigger isn't it?

jamiek-acl:
> > 1. when there are lots of selections and the select grows in height, the expand/collapse arrow should remain on the first line -- not be centred
> 
> Mmmm I don't know if this approach is incorrect or correct, if it's incorrect will be annoying to have to calculate the original height of the Trigger so we can calculate the position depending on the size of the Trigger isn't it?

It is what Sarah requested :).  I dont think you have to calculate the height -- why cant you just do a static `padding-top`?

packages/ListBox/src/ListBox.js
```diff
@@ -54,10 +54,10 @@ export function ListBox(props) {
   return (
     <React.Fragment>
       {trigger}
-      <Content onCancelFooter={footer ? footer.props.onClickCancel : null}>
+      <Content onCancelFooter={footer ? footer.props.onClickCancel : null} hasOptions={React.Children.count(children)}>
         <Box {...box.props}>
           {filter}
-          <List height={height}>
+          <List height={height} hasOptions={React.Children.count(children)}>
```
jamiek-acl, line 60: `hasOptions` is boolean, but you are passing in a number?  `!!React.Children.count(children)`?

packages/ListBox/src/components/List/List.styles.js
```diff
@@ -16,6 +16,6 @@ export const List = styled.ul(
       box-sizing: border-box;
     }
     padding: 8px;
-    ${noResultsFound ? "padding: 0" : ""}
+    ${noResultsFound || !hasOptions ? "padding: 0" : ""}
```
jamiek-acl, line 19: You could combine this with the line above it, but no big deal:

```
${noResultsFound || !hasOptions ? "padding: 0" : "padding: 8px;"}
```

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -73,18 +82,21 @@ const renderTrigger = ({ t, size, selectedOptions, onRemove, renderPill, renderT
...
 
-          const label = renderTriggerPillLabel === null ? item.label : renderTriggerPillLabel(item);
+          const label = pillLabelKey === null ? item.label : item[pillLabelKey];
 
           if (typeof label !== "string") {
             throw Error(
-              "Label property in your object should be a string, are you returning a string from renderTriggerPillLabel?"
+              `Either your item ${JSON.stringify(
+                item
+              )} lacks of an attribute "label", or you are not passing down the pillLabelKey prop on the component to indicate which attribute from your data should be use to render the pill label`
```
jamiek-acl, line 99:
```suggestion
              `Your item ${JSON.stringify(item)} must include the attribute "label", or you must indicate which attribute is should be rendered as the label via the "pillLabelKey" prop.`
```

packages/ListBox/stories/WithTags/examples/AllOptionsAreSelected.js
```diff
@@ -0,0 +1,35 @@
...
+    key: "label",
+    defaultData,
+    defaultFilteredData,
+    defaultSelectedKeys: ["Alpaca", "Ant", "Ape", "Armadillo"],
+  });
+
+  return (
+    <div style={{ padding: "32px" }}>
+      <ListBox
+        allOptionsAreSelected
```
jamiek-acl, line 20: So if all options are selected, it shows the message even when I haven't expanded the ListBox?  That's kinda weird.

Screen Shot 2020-12-18 at 10 26 37 AM

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -31,20 +32,31 @@ const propTypes = {
   renderPill: PropTypes.func,
   /** An array of id that helps the ListBoxWithTags to known what elements are selected  */
   selectedOptions: PropTypes.arrayOf(PropTypes.shape({})),
+  /** Provides an alternative for rendering the Pill label instead of using the default [{label:value}] coming from the og data */
+  pillLabelKey: PropTypes.string,
+  /** When this is true, it will display a message indicating all options are selected without showing the popover nor the search input */
+  allOptionsAreSelected: PropTypes.bool,
```
jamiek-acl, line 38: Instead of passing in another prop, could you loop over the children and count how many are displayType `ListBox.Option`, and see if that count equals `selectedOptions.length`?

packages/ListBox/src/components/WithTags/WithTags.js
```diff
@@ -31,20 +32,31 @@ const propTypes = {
...
+  /** Message to display when all options have been selected */
+  allOptionsAreSelectedMessage: PropTypes.string,
```
jamiek-acl, line 40: A default for this would be nice

packages/ListBox/stories/WithTags/examples/AllOptionsAreSelected.js
nahumzs: isn't this the example that @sunminsarahlee shared with you? if you don't use allOptionAreSelected option, still render the filter and shows you the legend that there are not results.

packages/ListBox/src/components/WithTags/WithTags.js
nahumzs: No, I can't, I thought about it for some time, and couldn't figure out a way for knowing it, the children that are pass down are only the ones that are rendering either because a filter has been applying on top of them, or the initial list. I don't have control to know how big the list is or know what has been selected to be able to compared it. 

look at the story so you can see it.

packages/ListBox/stories/WithTags/examples/AllOptionsAreSelected.js
nahumzs: also if this is now what she wants we can modify it once she is back from 🌴 @sunminsarahlee 😄

jamiek-acl (approved):

PR #832, opened 2020-12-03
feat(CollapsibleCard): UXD-470 support render label function

allison-c:
### Purpose 🚀

In asset inventory, we have to have the ability to render a custom label to align all the items in the collapsible card header
image

### Notes ✏️

image

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/collapsible`

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/CollapsibleCard/src/index.d.ts
```diff
@@ -7,8 +7,8 @@ interface CollapsibleCardProps {
   children?: React.ReactNode;
   /** If has a divider between collapsible header and content. */
   hasDivider?: boolean;
-  /** Label text as the card title. */
-  label?: React.ReactNode;
+  /** Label text as the card title, can be a render function. */
+  label?: node | func;
```
nahumzs, line 11: this is something that we are starting to do different across our components, in some component whenever you provide a render function we prefix it with it, but in this case, is "hidden" inside the label.

What do you think that if you don't put the prefix `render` the API should assume that it's either text or node. 

And for render function, we do it explicitly `renderLabel` this way is transparent to the user that there is two alternatives.

**or**

we just use labels for function, node and text, but we should always support it. :/ which is not something that we have been doing. 


cc @mikrotron

allison-c: 1. `label` prop + `renderLabel` prop: Since the IDE will tell the developer what type of data the prop can take, I don't think adding a separated prop is very necessary
2. `<Collapsible.Label>Label</Collapsible.Label>` + `<Collapsible.Label>{() => {}}</Collapsible.Label>`: I prefer this way better, the only thing is `CollapsibleCard` is a component from `Collapsible`, and the `label` is a prop on `Collapsible`, can I only change the API of `CollapsibleCard` first? And when we have time we can change the one on `Collapsible` and maybe other collapsible series

mikrotron: @nahumzs Can you clarify what you mean by "support it" when you say _"we just use labels for function, node and text, but we should always support it."_?

I think it's okay to omit the `render` prefix on a prop if it supports multiple types (`string`, `node` AND `function`). I think keeping a `render` prefix for props that ONLY support a render `function` is not a bad idea though.  

I also agree with @allison-c that in this specific case, option `2.` where there's a `<Label>` subcomponent instead of a `label` prop would be ideal.  I think we should introduce that breaking change sooner than later!

nahumzs: Yeah ok, lets keeps with this direction,

@mikrotron I was meaning that if we support Node and text we should support function as default but that might be an overkilled, let's see how our library keep evolving and once we see more cases maybe we can take a decision a refactor whatever its needs to be refactor.

nahumzs (approved):

PR #833, opened 2020-12-04
docs(Paprika): UXD-409 update readme on enforcing conventional commits

jamiek-acl:
### Purpose 🚀
update readme on enforcing conventional commits


mikrotron (approved):
llgtm 31

Issue #831, opened 2020-12-03
FormElement - small defects


mikrotron:
After testing new API changes introduces in the `<FormElement>` `v2.0` in Robots, a few small defects were uncovered, including:

- Styling of `help` icon in `<FormElement.Label>`, when `<Layout.LeftCol>` is given a custom width.
- `<FormElement.Error>` sub component cannot be used without a `<FormElement>` wrapper.
- `<Fieldset.Label>` should not include a `for` attribute.

PR #835, opened 2020-12-04
UXD-601 FormElement fixes and documentation updates 🐐

mikrotron:
### Purpose 🚀
1. Address some small issues with new `<FormElement>` API uncovered during QA (fixes https://github.com/acl-services/paprika/issues/831).
2. Update documentation for new API and to improve clarity.

### Notes ✏️
- README, Docs MDX page, Showcase, Variations and Examples all revisited.
- Fixes and updates to `<FormElement>` and it's subcomponents detailed in [CHANGELOG](https://github.com/acl-services/paprika/pull/835/files#diff-6a4790fc18dd5152be31aefb69397d5fcf22653845d87d6a5834b13994069496)
- BREAKING CHANGES are fairly minor.

### Updates 📦
- [x] MAJOR (breaking) change to `FormElement`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-601--formelement-docs-n-fixes/?path=/story/forms-formelement-docs--page

### Screenshots 📸
New story listing for FormElement
Screen Shot 2020-12-04 at 9 58 32 AM

New Everything Bagel Example Story
Screen Shot 2020-12-04 at 9 58 56 AM

New version of Variations Story
Screen Shot 2020-12-04 at 10 01 18 AM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-601
https://github.com/acl-services/paprika/issues/831


packages/FormElement/stories/examples/Fieldset.js
```diff
@@ -0,0 +1,90 @@
...
+              width: 100%;
+            `}
+          >
+            <div
+              css={`
+                width: 20%;
+              `}
+            >
+              <FormElement>
+                <Label isVisuallyHidden>Salutation</Label>
```
tristanjasper, line 34: Wonder if the `isVisuallyHidden` functionality should be handled by the component if it is wrapped in a `FieldSet` rather than leaving the consumer to add it. How do they know if it should or shouldnt' be only showing for screen readers? Perhaps we should encourage them to always add `Labels` by default. But the component can decided whether it should be showing or not to the user.

packages/FormElement/stories/examples/ButtonGroup.js
```diff
@@ -1,27 +1,38 @@
...
 export default function ButtonGroupExample() {
   const refButtonGroup = React.useRef(null);
-  const buttonOptionsArray = ["Black Panther", "Wonder Woman", "Spiderman"];
-  const buttonGroupOptions = buttonOptionsArray.map(hero => (
-    <ButtonGroup.Item value={hero} key={hero}>
-      {hero}
-    </ButtonGroup.Item>
-  ));
+  const labelId = "label-buttongroup";
+  const buttonOptions = ["Batman", "Wolverine", "Daredevil"];
```
tristanjasper, line 11: What do you have against Black Panther, Wonder woman and Spiderman. They are all very capable super heroes.

packages/FormElement/stories/examples/ButtonGroup.js
```diff
@@ -1,27 +1,38 @@
...
-    </FormElement>
+    <L10n>
+      <FormElement>
+        <Label onClick={() => refButtonGroup.current.focus()} id={labelId}>
+          Form Label
+        </Label>
+        <Content>
+          {a11yProps => (
+            <ButtonGroup ref={refButtonGroup}>
+              {renderGroupOptions(`${labelId} ${a11yProps["aria-describedby"]}`)}
```
tristanjasper, line 30: how come you don't use the `id` coming from `a11yProps`?

packages/FormElement/stories/examples/Checkbox.js
```diff
@@ -1,34 +1,33 @@
...
 import { Fieldset } from "../../src";
 
-const { Label, Description, Content } = Fieldset;
+const { Label, Instructions, Content } = Fieldset;
 
 export default function CheckboxExample() {
-  const isDisabled = false;
-  const size = constants.size.MEDIUM;
-  const optionsArray = ["Black Panther", "Spiderman", "Wonder Woman"];
+  const checkboxOptions = ["Black Panther", "Spiderman", "Wonder Woman"];
```
tristanjasper, line 10: Ah... You do still like them 🥇

packages/FormElement/stories/examples/Checkbox.js
```diff
@@ -1,34 +1,33 @@
...
+  const handleChange = activeIndex => () => {
+    action(`Checkbox ${activeIndex + 1} was clicked.`)();
+  };
+
+  const renderCheckboxes = a11yProps =>
+    checkboxOptions.map((item, index) => (
+      <Checkbox key={item} onChange={handleChange(index)}>
         <Checkbox.Input {...a11yProps} />
-        {hero}
+        {item}
```
tristanjasper, line 20: I would still consider them a `hero` 🤷‍♂️

packages/FormElement/stories/examples/Checkbox.js
```diff
@@ -1,34 +1,33 @@
...
-    <Fieldset>
-      <Label>Form Label</Label>
-      <Description>A description</Description>
-      <Content>{a11yProps => <CheckboxGroup>{getCheckboxOptions(a11yProps)}</CheckboxGroup>}</Content>
-    </Fieldset>
+    <L10n>
+      <Fieldset>
+        <Label>Checkbox group label</Label>
+        <Instructions>Some instructions for these checkboxes.</Instructions>
+        <Content>{a11yProps => renderCheckboxes(a11yProps)}</Content>
```
tristanjasper, line 29: Why no CheckboxGroup? I think previously it was rendering too close vertically

packages/FormElement/stories/examples/DatePicker.js
```diff
@@ -1,21 +1,30 @@
...
-      </Content>
-    </FormElement>
+    <L10n>
+      <FormElement>
+        <Label>Start Date</Label>
+        <Instructions>Enter date in MM/DD/YYYY format.</Instructions>
+        <Content>
+          {a11yProps => (
+            <DatePicker date={date} onChange={handleChange}>
+              <DatePicker.Input placeholder="MM/DD/YYYY" {...a11yProps} />
```
tristanjasper, line 23: this placeholder string should also be the `dateFormat` on the DatePicker itself. (Although it is `MM/DD/YYYY` by default

packages/FormElement/stories/examples/Everything.js
```diff
@@ -0,0 +1,124 @@
...
+            </Content>
+            {!errorText ? <Description>This is a description of the fieldset.</Description> : null}
+            <Error>{errorText}</Error>
+          </RightCol>
+        </Layout>
+      </Fieldset>
+      <Rule />
+      <Button onClick={() => setErrorText("There was an error.")} size={Button.types.size.SMALL}>
+        Show Error
+      </Button>{" "}
```
tristanjasper, line 118: whats this " " ?

packages/FormElement/stories/examples/Inline.js
```diff
@@ -0,0 +1,48 @@
...
+      <FormElement>
+        <Layout>
+          <LeftCol width={180}>
+            <Label>Form Label</Label>
+          </LeftCol>
+          <RightCol>
+            <Instructions>These are some field instructions.</Instructions>
+            <Content>
+              {a11yProps => (
+                <Input
```
tristanjasper, line 26: I still think worth showing multiple inputs inline here

packages/FormElement/stories/examples/ListBox.js
```diff
@@ -1,17 +1,28 @@
...
 import ListBox from "@paprika/list-box";
 import FormElement from "../../src";
 
-const { Label, Content } = FormElement;
+const { Label, Content, Description } = FormElement;
 
 export default function ListBoxExample() {
-  const optionsArray = ["Black Panther", "Wonder Woman", "Spiderman", "The Incredibles", "Thor", <span>test</span>];
-  const listboxOptions = optionsArray.map(hero => <ListBox.Option key={hero}>{hero}</ListBox.Option>);
+  const optionsArray = ["Catwoman", "Doctor Octopus", "Thanos"];
```
tristanjasper, line 9: Seriously. Catwoman.

packages/FormElement/stories/examples/NativeInput.js
```diff
@@ -1,30 +0,0 @@
-import React from "react";
```
tristanjasper, line 1: Why remove a native element example? The way you apply certain attribute is different to using an `Input` component. Or are we now ok with them just spreading `a11yProps` on the `input` in general even though there are warnings

mikrotron: Same warnings appear with `<Input>` since the `moreProps` go right onto the `<input>`.  I didn't see any real value in adding a native example since the usage is the same, and it's a very rare use case in our applications.  Certainly not one we would encourage as design system enforcers.

packages/FormElement/stories/examples/ListBox.js
mikrotron: Batman has all the best villians!

packages/FormElement/README.md
```diff
@@ -22,107 +22,172 @@ npm install @paprika/form-element
...
-  <Label>Name</Label>
+  <Label help="Help text">Field label</Label>
+  <Instructions>Instruction text</Instructions>
+  <Content>
+    {a11yProps => (
+      <Input hasError={Boolean(errorMsg)} {...a11yProps} />
+    )}
+  </Content>
+  {errorMsg ? <Description>Description text</Description> : null}
+  <Error>{errorMsg}</Error>
```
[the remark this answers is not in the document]
mikrotron: Yes, should be `!errorMsg` 🤦 

My thinking was that since `<Error>` already [returns `null` when its `children` are falsey](https://github.com/acl-services/paprika/pull/835/files#diff-f984d98b1cdde18b6eac74859149f5629445adb2a89cded3a57de9c28c3c0197R16), it's not necessary to conditionally render it... but your suggestion doesn't hurt anything and I agree it _IS_ clearer.  I'll change it.

packages/FormElement/README.md
```diff
@@ -22,107 +22,172 @@ npm install @paprika/form-element
...
+</FormElement>
+```
+
+### Using `<Fieldset />` with `<Checkbox />`
+
+Since the `<Checkbox />` and `<Radio />` components already render with a `<label>` element, the label for the checkbox or radio _group_ needs to be a `<legend>` element, instead of another redundant `<label>`, which needs to be wrapped in a `<fieldset>` element. For this purpose, the `<FormElement />` has a `hasFieldSet` prop, but for clearer JSX markup, you can use the `<Fieldset>` component, provided as a named import from the same `@paprika/form-element` component.
```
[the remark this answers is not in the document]
mikrotron: Yeah, probably a good suggestion.  Easier to write this stuff in `md` than `html`, but it would be good content to have in that story.  I guess I don't mind having it in both places, but maybe I can trim the detail from the README, but keep it all in the story.

packages/FormElement/src/components/Content/Content.js
```diff
@@ -26,14 +24,14 @@ function Content(props) {
   const ariaDescribedByIdsString = ariaDescribedByIdsArray.join(" ");
 
   const a11yProps = {
-    refLabel,
+    "aria-describedby": ariaDescribedByIdsString.length ? ariaDescribedByIdsString : null,
+    "aria-required": isRequired || null,
+    "aria-disabled": isDisabled || null,
+    disabled: isDisabled || null,
```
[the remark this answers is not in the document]
mikrotron: It seems to work like this for `Input`, `Select` and `Textarea`.  If you add `isDisabled` to the `<Input>` _too_ you get a _very_ faint looking input, since the whole `<FormElement>` gets `opacity: 0.5` when it has `isDisabled`.  But that's 🤷‍♂️ 

I guess you might get weirdness if you set `isDisabled` on `FormElement` and `isDisabled={false}` on the `<Input>` but why would you ever do that?

If you _only_ set `isDisabled` on the `<Input>` that's fine too, the rest of the `<FormElement>` pieces just won't be dimmed.

packages/FormElement/src/components/Content/Content.js
```diff
@@ -51,6 +49,11 @@ function Content(props) {
   );
 }
 
+const propTypes = {
+  /** Input field and layout elemements. May be a render funtion with ({ id, refLabel, ariaDescribedBy }) as argument. */
```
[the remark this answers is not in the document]
mikrotron: I was debating _where_ to document what's actually in `a11yProps` though... I feel like that render function signature should be included _somewhere_.

packages/FormElement/src/components/Label/Label.js
```diff
@@ -1,74 +1,68 @@
...
-      {renderHelp()}
-    </div>
+      {isOptional || isRequired ? (
+        <>
+          &nbsp;<sc.Requirement>{renderQuestionRequirement()}</sc.Requirement>
+        </>
+      ) : null}
+      {help ? (
+        <>
+          &nbsp;
```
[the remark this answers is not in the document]
mikrotron: Oh, I was trying my best to avoid an indent when the line breaks here.  It mostly works with `&nbsp;`.  The line breaks _after_ `the spacing and the part that breaks to a new line renders flush with the above line.

packages/FormElement/stories/examples/ButtonGroup.js
mikrotron: Variety is the Paprika of life.

packages/FormElement/stories/examples/ButtonGroup.js
mikrotron: The `id` from `a11yProps` is destined for the input element (which we don't have in this case).  This is a workaround for that – I can append the content of the `<label>` to the ARIA description this way (without overriding the accessible name, which should remain the `children` content of the `<ButtonGroup.Item>`.

packages/FormElement/stories/examples/Checkbox.js
mikrotron: I think it just adds noise to the example.  The primary purpose is to convey how to use the component, not make it look beautiful.

packages/FormElement/stories/examples/Everything.js
mikrotron: Just a cheap + dirty way to keep the buttons from sticking together. Not how I would do it in production (I'd use CSS) but seems like the least distracting way to accomplish it.

packages/FormElement/stories/examples/Fieldset.js
mikrotron: Oh I think it has to depend on the design.  Sometimes the labels should show, sometimes only _one_ should show, sometimes they can be hidden if the `<Fieldset.Label>` is sufficient.  I think this is still the best way to provide necessary flexibility.

packages/FormElement/stories/examples/Inline.js
mikrotron: I saved that for the `Everything Bagel` story.  I wanted to keep this one streamlined though, so the focus is on the `<Layout>`.

packages/FormElement/src/components/Help/Help.js
```diff
@@ -1,42 +1,49 @@
...
+import * as sc from "./Help.styles";
+
+function Help(props) {
+  const { children, isDisabled, a11yText } = props;
+  const I18n = useI18n();
 
-import helpStyles, { StyledTrigger, iconStyles } from "./Help.styles";
+  return (
+    <sc.Help align="bottom" data-pka-anchor="form-element.help">
+      {isDisabled ? (
```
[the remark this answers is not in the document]
mikrotron: Oh, I remember now that I want to be able to use the `<Label>` without the context provider (just for styling) sometimes.  Since the `<Help>` isn't used on its own, and `<Label>` gets `isDisabled` from context already, I think it's okay that this component just has the prop.

But you make a good point, and the `<Label>` should have an `isDisabled` prop, in case it's being used without context and should be disabled. The prop will take precedence if both exist.

tristanjasper:
LGTM

packages/FormElement/stories/examples/DatePicker.js
```diff
@@ -1,21 +1,32 @@
...
-      </Content>
-    </FormElement>
+    <L10n>
+      <FormElement>
+        <Label>Start Date</Label>
+        <Instructions>Enter date in {dateFormat} format.</Instructions>
+        <Content>
+          {a11yProps => (
+            <DatePicker date={date} onChange={handleChange}>
+              <DatePicker.Input placeholder={dateFormat} dateFormat={dateFormat} {...a11yProps} />
```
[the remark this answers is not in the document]
tristanjasper: Oh no it doesn't actually but perhaps we should encourage it to just be added to the root component? Or should it be just an inputProp?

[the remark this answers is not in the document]
mikrotron: Interesting 🤔 

It works on either, but the one on the `Input` takes precedence.  I think since it's the `Input` that _needs_ it, it makes more sense to remove the prop (and `humanFormat`, etc) from the `<DatePicker>`.  But that's for a different issue/PR.

cc @DanilAgafonov @allison-c

tristanjasper (approved):

PR #838, opened 2020-12-04
feat(Avatar): UXD-588 Added round ones.

jamiek-acl:
### Purpose 🚀

- Breaking change: the old "MEDIUM" square Avatar is now "LARGE".
- Breaking change: the old "SMALL" square Avatar is now "MEDIUM". A "small" square Avatar no longer exists, so I show a console warning if they chose that combo.  
- Breaking change: the Avatar will display whatever you pass into it as-is (including capitalization). Created the `getInitialsFromText` helper function so the consumer can get the text they want to pass in.
- Added new "isRound" prop so you can have round Avatars.

### Notes ✏️
Screen Shot 2020-12-07 at 9 08 03 AM

### Updates 📦
- [x] MAJOR breaking changes to Avatar


packages/Avatar/src/Avatar.styles.js
```diff
@@ -4,21 +4,39 @@ import stylers from "@paprika/stylers";
...
+  `,
+};
 
-export const avatarSizeStyles = {
+const squareStyles = {
   small: `
-    border-radius: ${tokens.button.borderRadius};
-    height: ${smallSize};
-    width: ${smallSize};
+    border-radius: ${tokens.card.borderRadius};
```
jamiek-acl, line 30: I don't know why it's saying I changed this line, cuz i didnt.  But the new one is the correct one... 🤔

mikrotron:
Would you ever want a round avatar with a one character abbreviation? Or a square one with two characters?  
Seems a bit weird to me that the shape also determines the abbreviation style. 
More of a UI Kit question, maybe?

jamiek-acl:
> Would you ever want a round avatar with a one character abbreviation? Or a square one with two characters?
> Seems a bit weird to me that the shape also determines the abbreviation style.
> More of a UI Kit question, maybe?

The existing square ones would not show two characters, so that works the same. If they wanted a single character in the round one they'd have to be sure to pass in a single word.

The component matches what they have in Figma.

packages/Avatar/src/Avatar.styles.js
allison-c: `card.borderRadius` is 6px, but here I believe we need 4px, that's why it was `tokens.button.borderRadius` before
image

jamiek-acl: What? no, where you lookin'?

https://www.figma.com/file/DRhxHNWVzd94K8Tf0C3VTy/%F0%9F%90%A6Starling-UI-Kit?node-id=1698%3A33820

Screen Shot 2020-12-04 at 4 13 59 PM

And if you inspect on Paprika storybook, it is 6px

jamiek-acl: oh you were looking at the medium one and i'm looking at the large one

sunminsarahlee:
@mikrotron Currently when we add users, their first name and last name are mandatory fields. But it is kinda weird to me the shape also determines the abbreviation style. Also, we will have an icon in the circular avatar for group. 
Screen Shot 2020-12-04 at 4 19 56 PM

packages/Avatar/src/Avatar.styles.js
allison-c: 😂 Yeah I'm looking at the small one

sunminsarahlee: We are using 6px rounded corner for 40px avatar and 4px for 32px one.

packages/Avatar/src/helpers.js
```diff
@@ -25,3 +25,8 @@ export function getAvatarColors(text) {
     backgroundColor: COLORS[index].background,
   };
 }
+
+export function getInitialsFromText(text, length = 1) {
```
mikrotron, line 29: Oh, I was thinking it would go in the `@paprika/helpers` package, but this works too.  But maybe it would be nice to export it in the `/src.index.js` file as well, so consumers can get it like: 

```js 
import Avatar, { getInitialsFromText } from '@paprika/avatar';
```

instead of 

```js
import Avatar from '@paprika/avatar';
import { getInitialsFromText } from '@paprika/avatar/lib/helpers';
```

jamiek-acl: I've updated the example story to grab the helper function as you describe

mikrotron: Oh! I didn't realize it was already being exported from `index.js` 🤦 
https://github.com/acl-services/paprika/blob/master/packages/Avatar/src/index.js#L2

mikrotron (approved):
xmas llgtm 1

PR #844, opened 2020-12-10
fix(ListBox): UXD-588 Provide proper props to Avatar component stories

jamiek-acl:
### Purpose 🚀
Fix ListBoxWithTags stories

### Notes ✏️
Two stories on prod were broken because of bad props (caused by recent changes):
https://paprika.highbond.com/?path=/story/forms-listbox--list-box-with-tags-uncontrolled-with-custom-trigger
https://paprika.highbond.com/?path=/story/forms-listbox--list-box-with-tags-uncontrolled-users

### Updates 📦
- [x] PATCH (bug fix) change to some stories

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-588--fix-uncontrolled-list-box


nahumzs (approved):

nahumzs:
thank you for cleaning some of my mess :D

PR #848, opened 2020-12-11
fix(OverflowMenu): FS-564 pass dynamic props to Confirmation on update

arkadiybutermanov:
### Purpose 🚀
Fix passing dynamic props to Confirmation on update

### Notes ✏️
`renderConfirmation` state gets updated only when you click on the Item: https://github.com/acl-services/paprika/blob/4d84b80efcf86d70066fe4dd113a1f6f2a5b5dc0/packages/OverflowMenu/src/OverflowMenu.js#L183 This makes it impossible to use dynamic props in `Confirmation` component (e.g. `isPending`), since they never get updated in state until the next click.

This PR replaces `renderConfirmation` state with reference and adds explicit updates when required.
See discussion in https://wegalvanize.slack.com/archives/CB3QRPDJ4/p1607629290311200

### Updates 📦
- [x] PATCH (bug fix) change to `OverflowMenu` component

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/FS-564-fix-OverflowMenu-rerendering

### References 🔗
https://aclgrc.atlassian.net/browse/FS-564

### DIFF
--- a/packages/OverflowMenu/src/OverflowMenu.js
+++ b/packages/OverflowMenu/src/OverflowMenu.js
@@ -46,12 +46,12 @@ const OverflowMenu = React.forwardRef((props, ref) => {
 
   const [isOpen, setIsOpen] = React.useState(false);
   const [isConfirming, setIsConfirming] = React.useState(false);
-  const [renderConfirmation, setRenderConfirmation] = React.useState(null);
   const [currentFocusIndex, setFocusIndex] = React.useState(0);
   const triggerRef = React.useRef(null);
   const menuId = React.useRef(uuid());
   const triggerId = React.useRef(uuid());
   const overflowListRef = React.useRef(null);
+  const renderConfirmationRef = React.useRef(null);
 
   function focusAndSetIndex(index) {
     if (overflowListRef && overflowListRef.current && index !== undefined)
@@ -77,7 +77,7 @@ const OverflowMenu = React.forwardRef((props, ref) => {
 
     if (isConfirming) {
       setIsConfirming(false);
-      setRenderConfirmation(null);
+      renderConfirmationRef.current = null;
     }
 
     if (triggerRef.current && key === "Tab") {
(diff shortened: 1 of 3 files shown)


packages/OverflowMenu/src/OverflowMenu.js
```diff
@@ -151,6 +154,14 @@ const OverflowMenu = React.forwardRef((props, ref) => {
       });
   };
 
+  const extractRenderConfirmation = () => {
```
arkadiybutermanov, line 157: Not very elegant solution, however it should unblock us. I think, we should revisit this component and refactor the way we render confirmation modal. Extracting it from children - doesn't seem to be the best option? Perhaps we should render it from inside `OverflowMenu.Item` ? Also, the today's implementation doesn't support multiple items with Confirmation dialogs.

packages/OverflowMenu/test/overflowMenu.spec.js
```diff
@@ -101,5 +104,26 @@ describe("OverflowMenu", () => {
         expect(queryByText(/edit/i)).not.toBeInTheDocument();
       });
     });
+
+    it("should pass dynamic props to Confirmation on update", async () => {
+      expect(getByText(/confirm delete/i)).toHaveAttribute("aria-disabled", "false");
+
+      const newConfirmation = (
+        <Confirmation
+          isPending
```
arkadiybutermanov, line 113: The exact case I've discovered. It's often needed to set confirm button state to `pending` (e.g. when making delete request to backend)

AlinaMoskieva (approved):

nahumzs:
Thank you @arkadiybutermanov I agree that the confirm scenario is too tight to this component, maybe we should revisit as you suggest, also I wonder while doing that would worth actually use the ListBox as a base, I see many similarities with this component. 

cc @mikrotron @tristanjasper

mikrotron:
> Thank you @arkadiybutermanov I agree that the confirm scenario is too tight to this component, maybe we should revisit as you suggest, also I wonder while doing that would worth actually use the ListBox as a base, I see many similarities with this component.
> 
> cc @mikrotron @tristanjasper

I'm not so sure it should be a `<ListBox>` the two components look similar but are semantically quite different and adding more complexity to an already complex ListBox may not be ideal for maintainability.  

I think @arkadiybutermanov's point that this component does not support multiple confirmable items is a good problem for us to focus on solving, and rethinking the way this component is composed in the process.

Issue #782, opened 2020-11-05
DataGrid `lazy` example


jamiek-acl:
# Bug Report
This story: https://paprika.highbond.com/?path=/story/table-datagrid-examples--lazy hits an external API.  Some of the Cypress tests use it.  I ran into a problem where the API was down, so cypress tests were failing and I couldn't merge my PR.  Update this story to not use an external API.

jamiek-acl:
The `Search` component now also hits an external API (the Marvel API), so it is susceptible to this and should be updated as well.

allison-c:
We had to disable the cypress tests using DataGrid lazy example to unblock the pre-release build. https://github.com/acl-services/paprika/pull/872/commits/2af4abc1ba461c4d2b579021051c3206f62b6939

Definitely better to use a fake server.

jamiek-acl:
Done here: https://github.com/acl-services/paprika/pull/883

PR #849, opened 2020-12-11
feature(Search): init UXD-596

nahumzs:
### Purpose 🚀
Implementing a simple autocomplete search on top of the ListBox, this component renders an Input on the trigger and filter the options, still let you navigate the option while focusing on the Input.

An implementation looks like:

```js
import React from "react";
import Search from "../../src";

const data = [
  { label: "vendor (assets type)" },
  { label: "Vendor Controls Critically (questionnaire)" },
  { label: "Vendor Control full (questionnaire)" },
  { label: "Types and more (assets)" },
  { label: "Investment and controls" },
  { label: "Objectives for audit" },
  { label: "Assets for third party vendor" },
  { label: "Finding automatic vendor requirements" },
];

export default function Controlled() {
  const [filterData, setFilterData] = React.useState(data);
  const [valueSelected, setValueSelected] = React.useState("");
  function handleChangeSearch(term) {
    setFilterData(ListBox.filter(term, data));
  }

  function handleSelected(value, actions) {
    /**
     * after receiving the selected value there are some alternatives paths you might want to take
     * 1. Close the popover
     * 2. Set the selected value on the input
     * 3. Clear the input
     *
     * These actions might change depending on what are you implementing.
     * all these options are possible you can invoke any of the actions to do that.
     * if there is something else specific you might want to do you either can add
     * expose it on the Search.js File
     * or request help from the paprika team.
     * */
    console.log(actions);
    const { setInput, close } = actions;

    setFilterData(ListBox.filter(value.label, data));
    setInput(value.label);
    close();
    console.log(value);
    setValueSelected(`→${value.label}`);
  }

  return (
    <div style={{ padding: "32px" }}>
      {valueSelected}
      <Search onChangeSearch={handleChangeSearch} onSelected={handleSelected}>
        {filterData.map(option => {
          return (
            <Search.Option value={option.label} key={option.label} label={option.label}>
              {option.label}
            </Search.Option>
          );
        })}
      </Search>
    </div>
  );
}

```

### Notes ✏️

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -239,6 +254,7 @@ export function handleEnterOrSpace({ event, state, dispatch, onChangeContext })
         activeOptionIndex: state.activeOption,
         dispatch,
         onChangeContext,
+        eventType: "keypress",
```
allison-c, line 257: all these event types strings can be in the types.js ?

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -88,7 +88,7 @@ export default function Trigger(props) {
     refLabel,
     size,
   } = state;
-  const triggerButtonId = React.useRef(nanoid());
+  const triggerButtonId = React.useState(nanoid);
```
allison-c, line 91: const [triggerButtonId] = React.useState(nanoid);

packages/Search/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@paprika/search",
+  "version": "1.0.0",
```
allison-c, line 3: the initial version is better to < 1.0.0, otherwise the pre-release will publish `1.0.1-alpha0`, and the actual stable version will be at least 1.0.1

packages/Search/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/button": "^1.0.3",
+    "@paprika/helpers": "^1.0.2",
+    "@paprika/icon": "^1.0.1",
+    "@paprika/input": "^1.0.3",
+    "@paprika/l10n": "^1.1.1",
+    "@paprika/list-box": "1.2.1-alpha.2",
+    "@paprika/overflow-menu": "^1.0.5",
```
allison-c, line 25: Some unused dependencies

packages/Search/src/Search.js
```diff
@@ -0,0 +1,138 @@
...
+      },
+    };
+    refSelected.current = null; // selected ref value
+    onSelected(value, actions);
+  }
+
+  function handleChange(index, options, attributes) {
+    const { eventType } = attributes;
+
+    if (eventType.includes("click")) {
```
allison-c, line 79: Only `click` event and trigger a select? keyboard `enter` is not allowed?

allison-c:
image
This second magnifier icon is interesting, I thought if I select this option the list will be refreshed, but it actually just takes my input value

packages/Search/stories/examples/Basic.js
```diff
@@ -0,0 +1,59 @@
...
+    setInput(value.label);
+    close();
+    console.log(value);
+    setValueSelected(`→${value.label}`);
+  }
+
+  return (
+    <div style={{ padding: "32px" }}>
+      {valueSelected}
+      <ListBox onChangeSearch={handleChangeSearch} onSelected={handleSelected}>
```
allison-c, line 48: Wait.. is this a new component or just an extended ListBox

packages/Search/src/renderTrigger.js
```diff
@@ -0,0 +1,130 @@
...
+          icon={<SearchIcon />}
+          onBlur={onBlurInput({ dispatch, types })}
+          onChange={handleChange}
+          onClick={handleClickInput}
+          onKeyDown={handleKeyDownInput}
+          onKeyUp={handleKeyUpInput}
+          ref={refInput}
+          type="text"
+          value={inputValue}
+          placeholder={t("listBox.filter.placeholder")}
```
allison-c, line 123: Are we missing `onClear` here? The clear button in storybook example clears the input but doesn't clear the value in the example.

allison-c:
Not sure if this is a bug with the storybook example only or related to the component itself

- Go to http://storybooks.highbond-s3.com/paprika/UXD-596-Search/?path=/story/forms-search--search-basic
- Click on the input trigger and type something
- Remove the value just typed and wait for the popover to be closed
- Hit `Enter` on the keyboard
You can see a value will be selected

submitbug

packages/Search/stories/examples/LazyLoading.js
```diff
@@ -0,0 +1,83 @@
...
+    setInput(value.label);
+    setSelection(value.label);
+    close();
+    console.log(`😎 `, value);
+  }
+
+  return (
+    <div style={{ padding: "32px" }}>
+      <ListBox onChangeSearch={handleChangeSearchDebounce} onSelected={handleSelected}>
+        {filterData.map(option => {
```
jamiek-acl, line 72: So I guess a `loading` spinner would be an implementation detail

packages/Search/stories/examples/helpers.js
```diff
@@ -0,0 +1,9 @@
+export async function fetchAPI(term, offset = null, limit = 20) {
+  const url = "https://gateway.marvel.com:443/v1/public/characters?nameStartsWith=";
```
jamiek-acl, line 2: Didn't we find an alternative before so we didn't have to hit an external API?  (If I recall, some tests were failing because of it... not that this has any tests)

packages/ListBox/src/components/Trigger/Trigger.js
nahumzs: yeap, I think we can replace all nanoid-related tasks in another PR.

packages/Search/src/Search.js
nahumzs: yes is allowed under the submit event on the form element.

packages/Search/stories/examples/Basic.js
nahumzs: is a wrapper on top of the listbox, let me rename the default import to Search to avoid confusion :P

packages/ListBox/src/components/List/List.js
```diff
@@ -12,7 +12,7 @@ const propTypes = {
   children: PropTypes.node.isRequired,
 
   /** Body content of the list. */
-  hasOptions: PropTypes.bool.isRequired,
+  hasOptions: PropTypes.oneOfType([PropTypes.bool, PropTypes.number]).isRequired,
```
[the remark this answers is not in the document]
nahumzs: Because sometimes people pass a number or boolean and is really annoying to see the console log in the test full of this kind of errors :/

packages/Search/stories/examples/LazyLoading.js
nahumzs: yes indeed, is hard to have a search component fit them all. 

The search can have pagination, no pagination, :/ better to hand that control to the consumer.

packages/Search/stories/examples/helpers.js
nahumzs: Do you have an alternative API that we could use? this is quite simple and mostly I'm using it to demonstrated a case. 

I wonder if would be beneficial to create our own testing API for our component :P

nahumzs:
> Not sure if this is a bug with the storybook example only or related to the component itself
> 
> * Go to http://storybooks.highbond-s3.com/paprika/UXD-596-Search/?path=/story/forms-search--search-basic
> * Click on the input trigger and type something
> * Remove the value just typed and wait for the popover to be closed
> * Hit `Enter` on the keyboard
>   You can see a value will be selected
> 
> submitbug

Indeed!! thank you for finding out, I fix it 🙏

nahumzs:
> image
> This second magnifier icon is interesting, I thought if I select this option the list will be refreshed, but it actually just takes my input value

🤔 I guess this is between a limitation of the list box and a case for when you start browsing the element and making selections and typing something.

You can end into a state when there is an option selected and you have text on the input, so if you click enter what are you trying to fire? the selected option? or the input text?

I think having this extra option is clear your intention that if you click enter will fire the one that is selected.

But to be honest we can modify this in the future if we find that our internal users don't like this approach.

packages/ListBox/src/components/List/List.js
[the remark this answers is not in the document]
jamiek-acl: I don't think we should support people passing in the wrong data types, but i'm not going to die on this hill.

packages/Search/stories/examples/helpers.js
jamiek-acl: Oh here is the `Issue` i made before to replace an external API: https://github.com/acl-services/paprika/issues/782.
I'll just add the `Search` component to that `Issue`, so when it is done, it is fixed here too.

packages/Search/src/renderTrigger.js
nahumzs: I moved the {...moreProps} to be spread on the Input instead of the ListBox. So people can spread props in the Input, I don't know if that you mean about the onClear Button?.

I tested and onClear seems to be working correctly

packages/Search/src/renderTrigger.js
```diff
@@ -0,0 +1,130 @@
...
+  refListBoxReducer.current.dispatch = dispatch;
+  // eslint-disable-next-line
+  refListBoxReducer.current.types = types;
+
+  function handleClickInput(event) {
+    event.stopPropagation();
+    if (!isOpen) dispatch({ type: types.openPopover });
+    if (refInput.current.value === "" && countOptions === 0) {
+      dispatch({ type: types.closePopover });
+    }
```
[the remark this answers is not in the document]
nahumzs: Yeah I changed the code so now the popover will appears only if the user input a value and have children

packages/Search/src/renderTrigger.js
allison-c: 👍 all good now, just checked the example

allison-c (approved):

PR #850, opened 2020-12-15
feature(Uploader): UXD-636 add custom error handler and remover retry

nahumzs:
### Purpose 🚀
- Remove the ability to retry to upload the file since is not current possible to do it and is unstable
- Replace `onError` for `onEnd` in the parameters whenever you use the `onRequest` function prop.
- Now there is an onError prop in the `<Uploader onError />` which allow the consumer to customize the message per file error. 

### Notes ✏️
Screen Shot 2020-12-16 at 5 21 05 PM

Screen Shot 2020-12-16 at 4 25 59 PM

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-636-uploader-improves

### Screenshots 📸

### References 🔗


packages/Uploader/src/components/File/File.js
```diff
@@ -47,14 +52,8 @@ function File({ error, fileKey, name, progress, size, status }) {
...
-              <Button.Icon
-                kind={Button.Icon.types.kind.MINOR}
-                onClick={() => {
-                  restartFileUpload(fileKey);
-                }}
-                size={Button.Icon.types.size.SMALL}
-              >
-                <RetryIcon />
+              <Button.Icon kind={Button.Icon.types.kind.MINOR} onClick={() => {}}>
+                <InfoCircle />
```
jamiek-acl, line 56: Maybe this should be the caution icon and be red like the error message?  The black is quite eye catching.

packages/L10n/src/locales/en.yml
```diff
@@ -19,7 +19,7 @@ en:
       choose_from_computer: choose from your computer
       choose_from_computer_a11y: Choose files from your computer
       cancel_upload: Cancel upload
-      restart_upload: Restart upload
+      restart_upload: Restart upload by selecting the file again
```
jamiek-acl, line 22: Minor, but maybe `Select the file again to retry`

packages/Uploader/src/components/File/File.js
jamiek-acl: Screen Shot 2020-12-15 at 4 20 25 PM

jamiek-acl (dismissed):
Approving so not a blocker for Kuldeep, but I think the comments I added should be addressed

packages/Uploader/src/components/File/File.js
mikrotron: It should at least be `tokens.textColor.icon`, as that black is quite stark.

packages/Uploader/CHANGELOG.md
```diff
@@ -6,6 +6,32 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+```js
+        <Uploader
+          {...}
+          onRequest={({ file, onProgress, onEnd, formData }) => {
+            file.request
+              .get("https://api.example.com:4001/")
+              .withCredentials()
+              .on("progress", onProgress)
+              .end(onEnd);
+          }}
```
mikrotron, line 30: This would be helpful to have in the documentation too, at least in the README.

mikrotron: Oh, I see now it's in one of the stories.  That should be sufficient actually, you can disregard the above recommendation.

packages/Uploader/src/components/File/File.js
nahumzs: actually if you cancel the file why would you want to retry? now second thoughts we shouldn't even have an icon, and the infoIcon is actually correct isn't it?

cc @sunminsarahlee

jamiek-acl: > actually if you cancel the file why would you want to retry?

yeah that's true

nahumzs: Screen Shot 2020-12-16 at 11 36 29 AM

like this

sunminsarahlee: I am okay with removing `i` icon for cancelling action.

jamiek-acl:
Can you update this story too, please:
https://paprika.highbond.com/?path=/story/forms-uploader-examples--file-statuses

package.json
```diff
@@ -17,7 +17,7 @@
     "postpublish": "git add packages/*.md packages/*/index.d.ts && git commit -m 'Publish: automate readme, type definitions and changelog update' --no-verify && git push origin master -f --no-verify",
     "prepare": "yarn cleanup && lerna run pretranspile && yarn transpile && yarn generateReadMe && yarn generateTypeDefinitions",
     "prepublishOnly": "yarn generateChangelog",
-    "prettier": "prettier \"**/*.+(js|json|yml|css|scss)\"",
+    "prettier": "prettier \"packages/**/*.+(js|json|yml|css|scss)\"",
```
allison-c, line 20: a small comment, we already have prettierignore set up, do we really need this to be run in the packages folder?

packages/Uploader/CHANGELOG.md
```diff
@@ -17,3 +17,31 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Updated: Created and placed all constants into type.js file. [@kaan.darcey](https://github.com/KDarcey).
+
+## [1.1.1-alpha.0] - 2020-12-16
```
allison-c, line 21: another small comment, you don't need to put the version and date here, you can just put everything under the unreleased section. The version actually should be  `1.1.1-alpha.1`. but even you put it here, the build won't fail, so it's ok

packages/Uploader/src/components/File/File.js
```diff
@@ -23,45 +24,46 @@ const propTypes = {
...
             <Popover.Trigger>
-              <Button.Icon
-                kind={Button.Icon.types.kind.MINOR}
-                onClick={() => {
-                  restartFileUpload(fileKey);
-                }}
-                size={Button.Icon.types.size.SMALL}
-              >
-                <RetryIcon />
+              <Button.Icon kind={Button.Icon.types.kind.MINOR} onClick={() => {}}>
```
allison-c, line 56: Will this render two nested buttons? one for trigger, one for button.icon, I usually use `<Popover.Trigger>{() => <Button />}</Popover.Trigger>` for a case like this. But I can see this is not an issue introduced by this PR, so I guess it's ok to ignore it?

allison-c (approved):

packages/Uploader/CHANGELOG.md
jamiek-acl: @allison-c this was added automatically by running `yarn generateChangelog`... am i not supposed to run that?  I don't know about anyone else, but I'm getting confused by what is done manually, and what is done automatically (especially now that I'm working in sriracha, too, which has its own scripts) in the changelog and package.json version.

Package.json versions get updated automatically when deployment happens, right?  Maybe we can have the changelog get auto-generated at that point?

allison-c: Yes `yarn generateChangelog` will be run at that point. So we never need to do that manually

package.json
nahumzs: I'm not entirely sure what other files/libraries are being scanned apart from those ignore, but in my case read files that are not relevant plus the one in packages, making it slower, I can remove it, but is speeding the process by targeting exactly the files that we care.

packages/Uploader/CHANGELOG.md
nahumzs: ohhh not you shouldn't run it, changeLog runs automatically after the release :)

packages/Uploader/src/components/File/File.js
nahumzs: Yeah I think might render two buttons didn't notice before, I still have to work on this component after some decisions, will fix it in the next PR.

PR #865, opened 2021-01-07
Pdtvt 685  dynamic hyperlinks

jamiek-acl:
```
<DynamicHyperlink onFetch={() => {...}} />
```
Screen Shot 2021-01-07 at 9 47 02 AM

### Purpose 🚀
This component doesn't render anything.  Add it to your bundle and it will run a function(1) on each "dynamic-hyperlink"(2) that exists in the DOM. If the DOM changes, it will look for them again.

  (1) The function that runs is passed into the component by the consumer, and it must return a Promise. When the Promise resolves, it must return these properties:

- name (string)
- term (string)
- error (string)
 
If there is an error, neither the `name` nor `term` will appear. Valid errors are `access_denied` and `invalid_url`.

  (2) A "dynamic-hyperlink" is an anchor tag with the `data-dynamic-hyperlink` attribute.

Screen Shot 2021-01-12 at 2 20 30 PM

### Updates 📦
- [x] MAJOR (breaking) change to DynamicHyperlink (new)

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/pdtvt-685--dynamic-hyperlinks


packages/DynamicHyperlink/src/DynamicHyperlink.js
```diff
@@ -0,0 +1,47 @@
...
+  const I18n = useI18n();
+
+  React.useEffect(() => {
+    function updateDynamicHyperlinks() {
+      const dynamicHyperlinks = Array.from(document.querySelectorAll("[data-dynamic-hyperlink]"));
+
+      dynamicHyperlinks
+        .filter(dynamicHyperlink => !dynamicHyperlink.hasAttribute("data-dynamic-hyperlink--processed"))
+        .forEach(dynamicHyperlink => {
+          const originalLinkUrl = dynamicHyperlink.innerHTML;
```
raman-gill-acl, line 16: i think we should use the href value

packages/DynamicHyperlink/src/DynamicHyperlink.js
```diff
@@ -0,0 +1,47 @@
...
+          dynamicHyperlink.innerHTML = I18n.t("dynamicHyperlink.loading"); // eslint-disable-line no-param-reassign
+          dynamicHyperlink.setAttribute("data-dynamic-hyperlink--processed", "true");
+
+          // the consumer does the fetching of data
+          onFetch(originalLinkUrl).then(({ error, name, term }) => {
```
raman-gill-acl, line 21: onFetch should take in the value of the data-dynamic-hyperlink value.  Let's chat about the interface today

packages/DynamicHyperlink/src/DynamicHyperlink.js
```diff
@@ -0,0 +1,47 @@
...
+import "./DynamicHyperlink.scss";
+
+export default function DynamicHyperlink({ onFetch }) {
+  const I18n = useI18n();
+
+  React.useEffect(() => {
+    function updateDynamicHyperlinks() {
+      const dynamicHyperlinks = Array.from(document.querySelectorAll("[data-dynamic-hyperlink]"));
+
+      dynamicHyperlinks
```
raman-gill-acl, line 13: will this be able to handle multiple dynamic links in parallel?  lets say that there are 50 links on the page...will they all get spinners and wait for API response to come in?

packages/DynamicHyperlink/src/DynamicHyperlink.js
jamiek-acl: it's the same value, is it not?

raman-gill-acl: It is right now.  But it is possible that in the future we store transformed links in the db and the frontend will refetch.  

Another reason to use the Href attribute is that is what the dyn-hyperlinking package in Projects is using - so a change here will make the this component consistent.  Maybe we can chat about this during the meeting today.

raman-gill-acl (commented):
this is awesome Jaime!

packages/DynamicHyperlink/src/DynamicHyperlink.js
nahumzs, line 21: you are expecting a promise but haven't check if you received one.

const valueResponse = onFetch(originalLinkUrl);

if ("then" in valueRespues) {
 /// your code
 return
}

throw new Error("onFetch must return a promise");

nahumzs:
Is the consumer able to create their own link styled? seems like the style is predefined is this something expected?

jamiek-acl:
> Is the consumer able to create their own link styled? seems like the style is predefined is this something expected?

No they are not able to create their own style, this is a Paprika component!

packages/DynamicHyperlink/src/DynamicHyperlink.js
jamiek-acl: yes

nahumzs:
I don't see any test added, maybe add a screener test, I think it should be sufficient for this component.

packages/DynamicHyperlink/src/DynamicHyperlink.js
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from "prop-types";
+import useI18n from "@paprika/l10n/lib/useI18n";
+import "./DynamicHyperlink.scss";
+
+export default function DynamicHyperlink({ onFetch }) {
+  const I18n = useI18n();
+
+  React.useEffect(() => {
+    function updateDynamicHyperlinks() {
+      const dynamicHyperlinks = Array.from(document.querySelectorAll("[data-dynamic-hyperlink]"));
```
alexzherdev, line 11: just pointing out that this assumes [the component is a singleton](https://overreacted.io/writing-resilient-components/#principle-3-no-component-is-a-singleton). Would it make sense for it to take in a parent element and then do `parent.querySelectorAll`?

nahumzs: The other thing that I was thinking is what would happen if the link resides under a react tree, and this component which is at the top goes and modify the DOM. 

React would not know about this change, and then what would happen later if that component updated its tree, I guess will ignore the DOM and will put back an unstyled link?

This component is an interesting one.

packages/DynamicHyperlink/src/DynamicHyperlink.js
raman-gill-acl: onFetch(originalLinkUrl, LinkType)

packages/DynamicHyperlink/src/DynamicHyperlink.js
jamiek-acl: @alexzherdev Hmm I think we might learn more after it is implemented for testing.

@nahumzs The script includes a `MutationObserver`, so if the DOM changes, the script re-runs.

jamiek-acl:
> I don't see any test added, maybe add a screener test, I think it should be sufficient for this component.

I've added one now:

Screen Shot 2021-01-12 at 3 18 01 PM

nahumzs (approved):

PR #868, opened 2021-01-08
Uxd 683 advanced filter

allison-c:
### Purpose 🚀

New package for advanced filters component, pitch is here: https://aclgrc.atlassian.net/wiki/spaces/UX/pages/1188069912/Filter+ver.+1+in+ActionBar+Approved

Note this component is only for the advanced filters case described in the design pitch.
This component will replace the `Filter` component in `ActionBar`.

Tests and cleanup/deprecation for `Filter` in `ActionBar` will be done in different tickets.

- Clean up `Filter` in `ActionBar` https://github.com/acl-services/paprika/pull/869

storybook: 
http://storybooks.highbond-s3.com/paprika/UXD-683-advanced-filter/?path=/story/table-filters--showcase

### Notes ✏️

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


jamiek-acl:
In the story when i choose to filter by `position`, it dies.

jamiek-acl:
Maybe an implementation detail, but the `Showcase` story should list the players by default when I start it (as no filters are applied). I wasn't sure I'd be able to test the filters.

packages/Filters/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "dependencies": {
+    "@babel/runtime-corejs2": "^7.3.1",
+    "@paprika/button": "^1.0.4-alpha.0",
+    "@paprika/constants": "^1.0.0",
+    "@paprika/date-picker": "^1.0.5-alpha.0",
+    "@paprika/icon": "^1.0.1",
+    "@paprika/input": "^1.0.4-alpha.0",
+    "@paprika/panel": "^2.0.3-alpha.0",
+    "@paprika/radio": "^1.0.3-alpha.0",
+    "@paprika/select": "^1.0.3-alpha.0",
```
jamiek-acl, line 25: Do you need `alpha` versions?

packages/Filters/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "@paprika/stylers": "^1.0.0",
+    "@paprika/tokens": "^1.0.0",
+    "immer": "^6.0.2",
+    "nanoid": "^2.1.7",
```
jamiek-acl, line 29: since starting from scratch, `uuid`?

packages/Filters/src/Filters.js
```diff
@@ -0,0 +1,148 @@
...
+import * as sc from "./Filters.styles";
+
+function getLabelText(numberOfFilters, I18n) {
+  switch (numberOfFilters) {
+    case 0:
+      return I18n.t("filters.label");
+    case 1:
+      return I18n.t("filters.singular_label");
+    default:
+      return I18n.t("filters.plural_label", { numberOfFilters });
```
jamiek-acl, line 22: This used to be done automatically, but i haven't done it in at well over a year.  I think all you have to do is:
```yaml
// en.yml
filters:
  label: filter
  label_plural: {count} filters
```
```javascript
//Filters.js
I18n.t("filters.label", { count: numberOfFilters });
```
and it will choose the proper one automatically.
Have to do something specific for the "0" case, though.

https://www.i18next.com/translation-function/plurals

allison-c:
> Maybe an implementation detail, but the `Showcase` story should list the players by default when I start it (as no filters are applied). I wasn't sure I'd be able to test the filters.

Yes, I've added a default filter in the showcase example, and also fixed the bug you mentioned above, should be able to see it soon on the link! thank you!

packages/Filters/src/Filters.js
```diff
@@ -0,0 +1,148 @@
...
+          ) : (
+            children
+          )}
+          <Button data-pka-anchor="filters.addFilterButton" icon={<AddIcon />} onClick={onAddFilter}>
+            {I18n.t("filters.actions.add")}
+          </Button>
+        </sc.FiltersPanel>
+        <Panel.Footer isSticky>
+          <sc.Footer>
+            <Button data-pka-anchor="filters.applyButton" onClick={handleApply} kind="primary">
```
jamiek-acl, line 93: what are these `data-pka-achor`s for, a future test?

packages/Filters/src/Filters.js
```diff
@@ -0,0 +1,148 @@
...
+  onCancel: PropTypes.func,
+  onChangeOperator: PropTypes.func,
+  onClear: PropTypes.func,
+  operator: PropTypes.oneOf([Filters.types.operator.AND, Filters.types.operator.OR]),
+
+  rulesByType: PropTypes.objectOf(PropTypes.arrayOf(PropTypes.objectOf(Filters.types.rule))),
+};
+
+const defaultProps = {
+  appliedNumber: 0,
```
jamiek-acl, line 131: Small nit that i'm not adamant about, the proper english would be `numberApplied`

packages/Filters/package.json
allison-c: Yes, I'll need to keep them in sync with the latest packages, so that lerna can take care of the rest

packages/Filters/package.json
allison-c: good idea!

packages/Filters/src/Filters.js
allison-c: I didn't know i18n can handle that! I'll try it out thank you!

packages/Filters/src/Filters.js
allison-c: Yes!

packages/Filters/src/components/DatePicker/DatePicker.js
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export default function DatePickerWrapper(props) {
+  const { initialDate, onChange, parsingFormat } = props;
+
+  function getMomentDate(dateStr) {
+    return dateStr === "" ? null : moment(dateStr, parsingFormat);
+  }
+
+  const memorizedGetMomentDate = React.useCallback(getMomentDate, [parsingFormat]);
```
jamiek-acl, line 19: The goal if this `React.useCallback()` is to... what?  Did you mean to use `React.useMemo()`?

nit: Also you called it `memorize` instead of `memoize`... even though I think the term _should_ be `memorize`

packages/Filters/src/components/Item/Item.js
```diff
@@ -0,0 +1,188 @@
...
+  } = props;
+  const { columns, data, filtersRef, onChangeOperator, operator, rulesByType } = React.useContext(FilterContext);
+  const I18n = useI18n();
+
+  const selectedColumnType = columns.find(({ id }) => id === selectedColumnId).type;
+
+  const selectOptions = React.useMemo(() => {
+    return selectedColumnType === types.columnTypes.SINGLE_SELECT
+      ? data.filter((obj, index, arr) => {
+          return arr.map(mapObj => mapObj[selectedColumnId]).indexOf(obj[selectedColumnId]) === index;
```
jamiek-acl, line 48: Can you come up with a better name than `mapObj`?

`obj` isn't my favourite either, i usually go with `datum` (the singular of `data`), but that's not as big a deal

packages/Filters/src/components/Item/Item.js
```diff
@@ -0,0 +1,188 @@
...
+      selectedRule === types.rules.IS_NOT_EMPTY;
+
+    if (shouldNotShowValueField) return null;
+
+    if (renderCustomValueField) return renderCustomValueField();
+
+    switch (selectedColumnType) {
+      case types.columnTypes.BOOLEAN:
+        return (
+          <sc.ValueInput as={Select} value={`${value}`} onChange={handleChangeBooleanFilterValue}>
```
jamiek-acl, line 125:
```suggestion
          <sc.ValueInput as={Select} value={value} onChange={handleChangeBooleanFilterValue}>
```

packages/Filters/src/types.js
```diff
@@ -0,0 +1,87 @@
+export const rules = {
```
jamiek-acl, line 1: `rules`, `defaultRulesByType` and `localeKeysByRule ` are duplicated in `types.js` and `rules.js`... think its better to extract?

packages/Filters/stories/MockServerFilter/api.js
```diff
@@ -0,0 +1,17 @@
...
+  });
+};
+
+export const filterBy = () => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve([data[1]]);
+    }, 2000);
+  });
+};
```
jamiek-acl, line 17: LOL
```suggestion
import data from "../ShowcaseApp/data";

export const fetchAll = () => {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve(data);
    }, 500);
  });
};

export const filterBy = () => {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve([data[1]]);
    }, 500);
  });
};
```

packages/Filters/src/components/DatePicker/DatePicker.js
allison-c: Because the function `getMomentDate` is a dependency in useEffect below, so it has to be a memorized function

packages/Filters/src/components/Item/Item.js
allison-c: This is for a reason, had to convert boolean value to string, I guess use `toString` will be better

packages/Filters/src/types.js
allison-c: This is a mistake for sure!

packages/Filters/stories/MockServerFilter/api.js
allison-c: I'd prefer 800 to have the experience of the spinner 😅

packages/Filters/src/components/DatePicker/DatePicker.js
jamiek-acl: Ah right -- like `use case #2` here: https://github.com/alpo22/react-notes/blob/master/hooks/useCallback.md#use-case-2

jamiek-acl:
In the future please try to have more descriptive commit messages; if you ever need to track something down in the future, you'll be glad you did!

packages/Filter/src/Filter.js
```diff
@@ -0,0 +1,143 @@
...
+    onCancel();
+  }
+
+  function handleClear() {
+    onClear();
+  }
+
+  return (
+    <FilterContext.Provider value={{ filterRef, columns, data, operator, onChangeOperator, rulesByType }}>
+      <Panel
```
nahumzs, line 58: 🤔 debating with myself about if including the panel inside of the filter component or not, did you considered an option where you split this component in two, so the user assemble those by themselves but you provide them ready for use?

Something like

```js
import Filter, { Panel } from "@paprika/filter"

<Panel>
  <Filter />
</Panel>
```

An approach like this would decouple the panel from the filter, so the filter could actually render wherever it needs to render. 

Just a thought.

allison-c: I'm checking with Sarah if we have a case of using filters but not the panel. If not, I think the only benefit would be bringing the flexibility to the consumer app to control the props on `Panel` `Panel.Trigger`, that's a good point, because I forgot support `zIndex` in this PR. 

Other than that I don't think there're many values to ask them to write `<Filter.Item>` inside `<Filter>` and `<Filter>` inside `<Panel>`. It's only for the decoupling purpose like you said.

allison-c: Oh just got Sarah's comment, she said we don't have any case about using advanced filters outside the panel

jamiek-acl (approved):
looks good to me

PR #878, opened 2021-01-14
feat(Filter): UXD-698 change to listbox

allison-c:
### Purpose 🚀

### Notes ✏️

Replaced `<Select />` by `<ListBox />` in `<Filter />` component, only the column dropdown and single-select value dropdown have the search input, because they might be very long, the rule dropdown is usually limited length

### Updates 📦
- [x] MINOR (backward compatible) change to `@paprika/filter`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-698-replace-select-by-listbox

### Screenshots 📸

### References 🔗


nahumzs:
This might be a small bug?

packages/Filter/src/components/Item/Item.js
```diff
@@ -80,31 +81,29 @@ function Item(props) {
...
   }
 
-  function handleChangeSingleSelectFilterValue(event) {
-    onChangeFilter(types.changeTypes.FILTER_VALUE, { id, value: event.target.value });
+  function handleChangeSingleSelectFilterValue(index, options) {
+    onChangeFilter(types.changeTypes.FILTER_VALUE, { id, value: options[index].value });
   }
 
   function renderRuleField() {
     switch (selectedColumnType) {
```
nahumzs, line 93: curious why to use a switch for two options?

packages/Filter/src/components/Item/Item.js
```diff
@@ -123,9 +122,16 @@ function Item(props) {
...
       case types.columnTypes.BOOLEAN:
         return (
-          <sc.ValueInput as={Select} value={value.toString()} onChange={handleChangeBooleanFilterValue}>
-            <option value="true">{I18n.t("filter.rules.true")}</option>
-            <option value="false">{I18n.t("filter.rules.false")}</option>
+          <sc.ValueInput data-pka-anchor="filter.item.valueInput" value={value.toString()}>
```
nahumzs, line 125: Each time I review this switch, I want to comment about turning it into an object, but then I realize that each of these statements would end having a different signature if they were turned into a function. I think this is one of the few cases where I find that the switch statement performs better than using an object pattern.

:)

nahumzs (approved):

packages/Filter/src/components/Item/Item.js
allison-c: Maybe when I worked on it I thought there might be many column types? To be honest I don't recall, or maybe just a simple mistake 😅

PR #879, opened 2021-01-14
feature(Table): UXD-720 adding keyboard arrow navigability

nahumzs:
### Purpose 🚀

### Notes ✏️

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-720-arrow-keys-for-table/?path=/story/table-table--arrow-navigation

### Screenshots 📸

**`NOTE`** the following gif where snapshots before the integration of the `enableArrowKeyNavigation` prop, now there is a proper story where you can test it 

[Story with navigation](http://storybooks.highbond-s3.com/paprika/UXD-720-arrow-keys-for-table/?path=/story/table-table--arrow-navigation)

### References 🔗


packages/Table/src/Table.js
```diff
@@ -21,8 +25,23 @@ export default function Table(props) {
...
+      onFocus: handleFocus({ refFocus, tableId }),
+      onBlur: handleBlur({ refFocus, tableId }),
+      onKeyDown: handleKeyDown({ refFocus, tableId, ...qty }),
+      onClick: handleClick({ refFocus, tableId }),
+    };
+  }
+
   return (
-    <sc.Table aria-label={a11yText} {...moreProps}>
+    <sc.Table aria-label={a11yText} id={tableId} tabIndex={0} {...moreProps} {...arrowKeyNavigationProps}>
```
allison-c, line 44: Should the tabIndex value be set on the `td` element dynamically to let the screen reader read out the cell content? or maybe there's a better way to support screen reader?

nahumzs: yes!!! I saw that approach in the w3c, but let me find out how will be work within the inline editing, because we might have edition on popover, takeover, side panels, so don't really know yet who should control the tabindex yet or what should hold the focus. the editing doesn't happened directly on the td.

nahumzs: you know let me change it to the `td` I can go back and change the official way of doing it.

allison-c: Thank you! It would be super nice, hopefully not too much refactoring, I guess later on if there's input in a popover we can just move the focus to that input, arrow key doesn't work in that case, but it's totally fine I think.  It's a different workflow

nahumzs: ok fixed it!

nahumzs: was quite simple to change it actually good think you catch it.

allison-c (approved):

nahumzs:
I'm thinking that eventually we will need a `cellA11yText` like in the DataGrid so we can pass correct content to the screen reader with those cells with complex content/components.

PR #881, opened 2021-01-14
feat(MockEndpoints): UXD-714 Add an easy way to mock endpoints

jamiek-acl:
### Purpose 🚀
Create a new component (that is really just a helper file) that makes mocking API endpoints easy.

mocked


packages/MockEndpoints/src/MockEndpoints.js
```diff
@@ -0,0 +1,14 @@
+import fetchMock from "fetch-mock";
```
nahumzs, line 1: what about this change, doing this give us flexibility to have a easy to use fetch mockup and also gave the ability to and advance user to use fetchMock at their will?

```js
export default fetchMock;

export handlers => {
  // validate that the handlers have a certain shape (url, responseData)
  handlers.forEach(handler => {
    fetchMock.get(handler.url, handler.responseData);
  });

  fetchMock.catch(url => {
    const error = `MockEndpoints.js: You did not mock the endpoint "${url}", but did request it.`;
    console.error(error);
    return { error };
  });
};
```

jamiek-acl: Good idea. Done!

nahumzs (approved):

PR #891, opened 2021-01-19
feat(DynamicHyperlink): PDTVT-753 Make it run on dynamic hyperlinks w…

jamiek-acl:
Make it run on dynamic hyperlinks within ckEditor.
Note that within ckEditor, the style of the dynamic hyperlinks isn't applied - the consumer has to add those styles to the ckEditor.
Renamed the component

### Updates 📦
- [x] MAJOR (breaking) change to DynamicHyperlink

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/pdtvt-753--iframe

### Screenshots 📸
Screen Shot 2021-01-19 at 9 29 02 AM


packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.js
```diff
@@ -1,25 +1,36 @@
...
+      const dynamicHyperlinkSelector = "[data-dynamic-hyperlink]";
+      const dynamicHyperlinks = Array.from(document.querySelectorAll(dynamicHyperlinkSelector));
+
+      document.querySelectorAll("iframe.cke_wysiwyg_frame").forEach(iframe => {
+        if (iframe.contentWindow.document.body) {
+          iframe.contentWindow.document.body.querySelectorAll(dynamicHyperlinkSelector).forEach(dynamicHyperlink => {
+            dynamicHyperlinks.push(dynamicHyperlink);
+          });
+        } else {
+          setTimeout(updateDynamicHyperlinks, 100);
```
frankliu81, line 24: is this timeout just to ensure that the iframe content is loaded before we load the links, is there a more reliable way than using timeouts?

frankliu81 (approved):

packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.js
jamiek-acl: Yeah i don't like it either, but without it, it will not work on links inside ckEditor.
I don't think there is a better way to do it since the ckEditor is not React.  @nahumzs do you have any thoughts?

jamiek-acl:
Inside the ckEditor, `updateDynamicHyperlinks` does not get called on every key press, but it does on some of them (like enter) because i think ckEditor changes the classes on its toolbar items.

MutationObserver doesnt have a way to ignore certain children (the ckEditors), so i'm not sure of a way to prevent it.

packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.js
yukikywong: Wouldn't the first iframe be skipped since it'll just go to the else case and never back to the if case? Perhaps need a while loop with condition `!iframe.contentWindow.document.body` to wait for the content to be loaded.

nahumzs: I answered a similar question here https://stackoverflow.com/a/44979019/196038

I think that should be enough to just wait until `iframe.contentWindow.document.body.querySelectorAll(dynamicHyperlinkSelector)` return something. 

By the way, in that example, I received a string as a DOM parameter, but it should be a function instead, so the query always gets executed with the current context after each `requestNextAnimationFrame()`

Issue #393, opened 2020-01-15
Tab component doesn't work correctly with `null` elements


AndreyChernykh:
# Bug Report

## Expected behavior
 
```js
<Tabs.List>
          {content1 ? (
            <Tabs.Tab>{content1}</Tabs.Tab>
          ) : null}
          {content2 ? (
            <Tabs.Tab>{content2}</Tabs.Tab>
          ) : null}
        </Tabs.List>
```

If `content1 === true` and `content2 === false`, the only first tab should be rendered.

## Current behavior

```
 Uncaught TypeError: Cannot read property 'type' of null
    at Object.React$$1.cloneElement 
```

## Screenshots / Gifs / Codepens
N/A

## Additional context

It would be nice, actually, to have a `Tab` component NON to be separated from the `Tabs.Panels` component. 

E.g.: 

```js
<Tabs>
  <Tabs.Tab>
    <Tabs.Header>Hello</Tabs.Tab>
    <Tabs.Content>World</Tabs.Tab>
  </Tabs.Tab>
  {content ? (<Tabs.Tab>
    <Tabs.Header>Optional content</Tabs.Tab>
    <Tabs.Content>{content}</Tabs.Tab>
  </Tabs.Tab>) : null}
</Tabs>;
```

PR #892, opened 2021-01-19
UXD-716  Vertical Tabs 🐐

mikrotron:
### Purpose 🚀
Add an `isVertical` prop to `<Tabs.List>` that results in a vertically stacked list of tabs.

#### Added
- `isVertical` prop to render tabs in a stacked list.
- `hasInsetFocusStyle` prop for inset focus ring on Tab.Link children.
- `hasTruncation` prop to truncate tab labels when vertical. 
- 💥 BREAKING CHANGE: `size` prop to control font size – results in default font size of 14px instead of 16px. Original 16px style now requires `size={Tabs.types.size.LARGE}` prop on `<Tabs>`. 

#### Fixed
- Restored broken styling for link `<Tab>`.
- Moved `a11yText` prop from `<Tabs.List>` to `<Tab>`.
- `<Tabs.List>` and `<Tabs.Panels>` will now support `null` children (resolves https://github.com/acl-services/paprika/issues/393).

#### Changed
- 💥 BREAKING CHANGE: Moved `height` prop from `<Tabs.List>` to `<Tabs>` and renamed to `tabHeight`.

### Updates 📦
- [x] MAJOR (BREAKING) change to `Tabs`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-716--vertical-tabs/?path=/story/navigation-tabs-examples--vertically-stacked-tabs-with-icons

### Screenshots 📸
Screen Shot 2021-01-19 at 5 12 34 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-716
Issue https://github.com/acl-services/paprika/issues/393


packages/Tabs/src/Tabs.js
```diff
@@ -29,26 +42,24 @@ const Tabs = props => {
...
 
+  // TODO: Disabled tab items should also get focus on keyboard interaction
   // https://github.com/acl-services/paprika/issues/310
-  // Todo Disabled tab items should also get focus on keyboard interaction
   const onKeyDown = (event, currentIndex) => {
-    const tabList = React.Children.toArray(props.children)[0];
-    const enabledIndexes = tabList.props.children
-      .map((tab, index) => (tab.props.isDisabled === true ? null : index))
+    const { "Tabs.List": TabsList } = extractChildren(children, ["Tabs.List"]);
+    const tabs = TabsList.props.children.filter(child => child);
```
jamiek-acl, line 49: is this `.filter` needed? It's not filtering on anything

packages/Tabs/stories/Tabs.stories.js
```diff
@@ -67,15 +76,77 @@ storiesOf(`${storyName}/Examples`, module)
...
-  .add("Compact Tabs with custom height", () => (
+  .add("Tabs with custom height", () => (
     <Story>
-      <Tabs>
-        <Tabs.List height={80}>
-          <Tabs.Tab hasInsetFocusStyle>These tabs</Tabs.Tab>
-          <Tabs.Tab hasInsetFocusStyle>Are tall</Tabs.Tab>
-          <Tabs.Tab hasInsetFocusStyle>And narrow</Tabs.Tab>
-          <Tabs.Tab hasInsetFocusStyle>With inset focus</Tabs.Tab>
+      <Tabs tabHeight={80} hasInsetFocusStyle>
```
jamiek-acl, line 81: What is the argument for adding the new `hasInsetFocusStyle` prop? Shouldn't vertical tabs just always have inset focus style, and horizontal tabs dont even get the option?

jonathanjakimon:
Small detail:
For the hover state, I suggest having the whole line in grey instead of just the small rectangle on the far left.
It is slightly different from horizontal tabs but we need to increase the visibility of the vertical ones.

image

packages/Tabs/src/Tabs.js
robyn3choi: if it's to handle nulls, you can do the same thing on line 52 to keep things consistent

packages/Tabs/src/components/List/List.js
```diff
@@ -1,48 +1,56 @@
...
+    isVertical,
+    onKeyDown,
+    onClickTab,
+    setTabListRef,
+    size,
+  } = context;
 
   const childrenWithProps = React.Children.map(children, (tab, index) => {
-    const isSelected = activeIndex === index;
+    if (!tab) return;
```
robyn3choi, line 26: I think this could be removed since we're already filtering out `null` TabsList children in `Tabs.js`

robyn3choi: Whatever we decide for this, we should apply to `Panels` as well to keep things consistent. Right now, looks like the `null` check is done in line 15 in `Panels.js`

packages/Tabs/src/Tabs.js
mikrotron: Yeah, it's to filter out `nulls` and I agree `child => child != null` would be more clear 👍

packages/Tabs/stories/Tabs.stories.js
mikrotron: Sometimes they're needed for horizontal tabs too, if their container has `overflow:hidden`.  It wasn't explicitly part of the API before, you could add it to the `<Tab>` and since it was a `<RawButton>` that would get any `moreProps` it would work... but it wouldn't work when that `<Tab>` was a link (`<a>`).  

So this approach is a bit easier, you just need to add the prop once on the root `<Tabs>` and it gets propagated, and it works for links too.

packages/Tabs/src/Tabs.js
```diff
@@ -61,43 +72,64 @@ const Tabs = props => {
...
   isDisabled: PropTypes.bool,
+
+  /** If the tabs are stacked vertically. */
+  isVertical: PropTypes.bool,
+
+  /** Size of the tab label text.  */
+  size: PropTypes.oneOf([Tabs.types.size.MEDIUM, Tabs.types.size.LARGE]),
+
+  /** Height, in pixels, of the tabs (ignored when isVertical is true). */
+  tabHeight: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
mikrotron, line 117: Oh, this is another change, the `tabHeight` can be a number or a string (CSS value).  But I need to update the comment description.

packages/Tabs/src/components/List/List.js
mikrotron: The null check in `Tabs.js` is just for handling keyboard navigation, not for rendering purposes.  

This line here (and the corresponding one in `Panels.js`) is to allow `null` children by avoiding the attempt to `cloneElement()` on them, which throws an error.

robyn3choi:
Now that the selected tab has a grey background, should the text remain black like in @jonathanjakimon 's screenshot?

mikrotron:
> Now that the selected tab has a grey background, should the text remain black like in @jonathanjakimon 's screenshot?

Jonathan's last screenshot was just to illustrate the hover state, which does maintain black text.  When selected though, the text should be blue – consistent with the horizontal tabs.

robyn3choi (approved):

PR #898, opened 2021-01-21
fix(Modal): UXD-773 fixed mapping error in footer

### LINKED ISSUE #387: <Modal.Footer> assumes children are React elements
# Bug Report
## Expected behavior
I'm expecting to be able to use React idioms like the following in `<Modal.Footer>`:
```jsx
<Modal.Footer>
  {hasDoneStuff && (
    <span>Success</span>
  )}
  <Button onClick={() => { setHasDoneStuff(true); }}>
    Do stuff
  </Button>
</Modal.Footer>
```
## Current behavior
Component breaks.

## Screenshots / Gifs / Codepens
```
Footer.js?2928:34 Uncaught TypeError: Cannot read property 'key' of null
```
https://github.com/acl-services/paprika/blob/c4f2593292a1709638a8b45d84ca0243d488f44d/packages/Modal/src/components/Footer.js#L18-L20

## Additional context
I'm not sure mapping over children is necessary here. (The code also seems to assume every child has a key, which I'm guessing is very unlikely in this case)
Would we be able to achieve the same styling by removing `ChildWrapper` div and instead adding margin via a `& > *` selector?

allison-c:
### Purpose 🚀

### Notes ✏️

Resolves #387

Avoid using React.Children.map and read the keys of each child. Also checked other places using similar method and fixed them

### Updates 📦
- [x] PATCH (bug fix) change to`@paprika/modal`

### Storybook 📕

### Screenshots 📸

### References 🔗


nahumzs (approved):

nahumzs:
🆒 😄

PR #899, opened 2021-01-21
L10N-315 2021-01-21-260 - translations batch

agherreras:


jamiek-acl (approved):

cristinajimenez (approved):

packages/L10n/src/locales/de.yml
```diff
@@ -189,7 +189,37 @@ de:
...
       access_denied: "Zugriff verweigert"
       invalid_url: "Ungültige URL"
       loading: Wird geladen...
+    filter:
+      actions:
+        add: Filter hinzufügen
+        clear: Löschen
+        delete: Löschen
+      no_filters_applied: Auf diese Ansicht wurden keine Filter angewendet
+      label: Filtern
```
alexandermccate, line 202: Filter

packages/L10n/src/locales/es.yml
```diff
@@ -189,7 +189,37 @@ es:
...
       access_denied: "Acceso denegado"
       invalid_url: "URL no válida"
       loading: Cargando...
+    filter:
+      actions:
+        add: Agregar filtro
+        clear: Borrar
+        delete: Eliminar
+      no_filters_applied: No se aplicaron filtros a esta vista
+      label: Filtrar
```
alexandermccate, line 202: Filtro

packages/L10n/src/locales/es.yml
```diff
@@ -189,7 +189,37 @@ es:
...
+      or: O
+      rules:
+        is: es
+        is_not: no es
+        contains: contiene
+        does_not_contain: no contiene
+        is_blank: está en blanco
+        is_not_blank: no está en blanco
+        equals: es igual a
+        not_equal_to: es distinto de
```
alexandermccate, line 215: `no igual a` - seems better for consistency. thoughts?

packages/L10n/src/locales/fr.yml
```diff
@@ -189,7 +189,37 @@ fr:
...
       access_denied: "Accès refusé"
       invalid_url: "URL non valide"
       loading: Chargement...
+    filter:
+      actions:
+        add: Ajouter un filtre
+        clear: Effacer
+        delete: Supprimer
+      no_filters_applied: Aucun filtre appliqué à cette vue
+      label: Filtrer
```
alexandermccate, line 202: Filtre

packages/L10n/src/locales/pt.yml
```diff
@@ -189,7 +189,37 @@ pt:
...
       access_denied: "Acesso negado"
       invalid_url: "URL inválido"
       loading: Carregando...
+    filter:
+      actions:
+        add: Adicionar filtro
+        clear: Limpar
+        delete: Excluir
+      no_filters_applied: Nenhum filtro aplicado a esta exibição
+      label: Filtrar
```
alexandermccate, line 202: Filtro

packages/L10n/src/locales/pt.yml
```diff
@@ -189,7 +189,37 @@ pt:
...
+        does_not_contain: não contém
+        is_blank: está em branco
+        is_not_blank: não está em branco
+        equals: igual a
+        not_equal_to: diferente de
+        greater_than: ">"
+        less_than: "<"
+        greater_than_or_equal_to: "≥"
+        less_than_or_equal_to: "≤"
+        is_empty: está em branco
```
alexandermccate, line 220: `está vazio` could be better. please fix and double check with translator in the meantime

packages/L10n/src/locales/zh.yml
```diff
@@ -189,7 +189,37 @@ zh:
...
       access_denied: "拒绝访问"
       invalid_url: "无效的 URL"
       loading: 正在加载…
+    filter:
+      actions:
+        add: 添加过滤器
+        clear: 清除
+        delete: 删除
+      no_filters_applied: 没有过滤器应用于此视图
+      label: 筛选
```
alexandermccate, line 202: 过滤器

alexandermccate (commented):
@agherreras I've added my comments for a number of fixes, please apply all as needed. Also, share a screenshot of all those new strings with translators when working on QA step and get their feedback. We might have to further edit some translations, but we have a solid first draft.

packages/L10n/src/locales/es.yml
agherreras: We could use "no es igual a" for consistency, but I still prefer "es distinto de".
What do you think, @cristinajimenez?

cristinajimenez: Agree, if we can choose, I'd rather use `es distinto de`

alexandermccate: ...but note that it will be quite different than what we have in docs, Academy, AN, etc. That's why I meant to use `no es igual a`. ..and we don't want to edit legacy content in that case.

alexandermccate: image

alexandermccate: image

agherreras: Sure. We can use `no es igual a` for consistency with other TMs...

PR #900, opened 2021-01-22
Uxd 756 title prop should be optional

allison-c:
### Purpose 🚀

### Notes ✏️

I have to hardcode the props parameter in a template to make the default value to `null`. SVGR doesn't have a configuration for it.

Also I noticed there's no js files in src folder so I deleted the ts files too

Will do the same thing in wasabicons 😺 

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/icon`

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/Icon/svgr.config.js
```diff
@@ -0,0 +1,21 @@
...
+  svgProps: {
+    // We need to render string template
+    // eslint-disable-next-line no-template-curly-in-string
+    css: "{`color: ${props.color};width: ${props.size};height: ${props.size};vertical-align: text-top;`}",
+    "data-pka-anchor": "icon",
+    focusable: false,
+  },
+  titleProp: true,
+  template,
+};
```
allison-c, line 21: `.svgrrc` cannot handle template correctly, so uses `svgr.config.js` instead

packages/Icon/helper/template.js
```diff
@@ -0,0 +1,12 @@
+module.exports = ({ template }, opts, { imports, interfaces, componentName, jsx, exports }) => {
```
nahumzs, line 1: maybe would worth putting a link or reference where to read more about it? so the next developer can get an idea of how this works?

nahumzs (approved):

packages/Icon/helper/template.js
```diff
@@ -0,0 +1,12 @@
...
+  const plugins = ["jsx"];
+  const typeScriptTpl = template.smart({ plugins });
+  return typeScriptTpl.ast`
+  ${imports}
+  ${interfaces}
+  function ${componentName}({title=null, ...props}) {
```
nahumzs, line 7: I guess this is trigger by typescript and modified the `ast` while executing the code? or how does this works?

allison-c: This triggered by svgr plugin https://github.com/gregberge/svgr/blob/master/packages/babel-plugin-transform-svg-component/src/index.js
You can find the default template on that page as well

PR #904, opened 2021-01-25
chore(DynamicHyperlinkTranslator): PDTVT-784 Tweak link styles.

jamiek-acl:
### Purpose 🚀
Don't underline the right-hand-side of the cards on hover. Note: they are still clickable links, as everything is under a link..
hover

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/pdtvt-784--no-underline

### DIFF
--- a/packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.js
+++ b/packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.js
@@ -36,15 +36,24 @@ export default function DynamicHyperlinkTransformer({ onFetch }) {
               .then(response => {
                 const { error, name, term } = response;
 
-                const className = error ? "dynamic-hyperlink--invalid" : "dynamic-hyperlink--valid";
-                const linkText = error ? originalLinkUrl : name;
-                const errorText = error ? I18n.t(`dynamicHyperlinkTransformer.${error}`) : "";
+                dynamicHyperlink.innerHTML = ""; // eslint-disable-line no-param-reassign
 
-                dynamicHyperlink.innerHTML = linkText; // eslint-disable-line no-param-reassign
-                const typeOrErrorSpan = document.createElement("span");
-                typeOrErrorSpan.innerHTML = error ? `- ${errorText}` : term;
-                typeOrErrorSpan.className = className;
-                dynamicHyperlink.appendChild(typeOrErrorSpan);
+                const labelSpan = document.createElement("span");
+                labelSpan.className = "dynamic-hyperlink--label";
+                labelSpan.innerHTML = error ? originalLinkUrl : name;
+                dynamicHyperlink.appendChild(labelSpan);
+
+                if (error) {
+                  const errorSpan = document.createElement("span");
+                  errorSpan.className = "dynamic-hyperlink--invalid";
+                  errorSpan.innerHTML = `- ${I18n.t(`dynamicHyperlinkTransformer.${error}`)}`;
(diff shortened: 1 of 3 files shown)


packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.scss
```diff
@@ -10,30 +10,32 @@ a[data-dynamic-hyperlink] {
   border-radius: $button--border-radius;
   box-shadow: 0 0 2px $color--black;
   color: $color--blue-darken-10;
+  cursor: default;
```
allison-c, line 13: Since you said this is still under the clickable link, should the cursor be a pointer as before?

packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.js
```diff
@@ -36,15 +36,24 @@ export default function DynamicHyperlinkTransformer({ onFetch }) {
...
-                dynamicHyperlink.innerHTML = linkText; // eslint-disable-line no-param-reassign
-                const typeOrErrorSpan = document.createElement("span");
-                typeOrErrorSpan.innerHTML = error ? `- ${errorText}` : term;
-                typeOrErrorSpan.className = className;
-                dynamicHyperlink.appendChild(typeOrErrorSpan);
+                dynamicHyperlink.innerHTML = ""; // eslint-disable-line no-param-reassign
+
+                const labelSpan = document.createElement("span");
+                labelSpan.className = "dynamic-hyperlink--label";
+                labelSpan.innerHTML = error ? originalLinkUrl : name;
```
nahumzs, line 43: can be a case where the content on the tag requires sanitization? would be a way to inject malicious code via this link?

packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.scss
jamiek-acl: That is up for debate.
I didn't like the hand cursor appearing when hovering over the error message/type on the right hand side of the card.

packages/DynamicHyperlinkTransformer/src/DynamicHyperlinkTransformer.js
jamiek-acl: This code only gets run on URLs with the 'data-dynamic-hyperlink' attribute, and links only get that attribute if they match a certain regex pattern defined in the back-end. So i think the back-end is essentially doing the sanitization with their regex.

nahumzs: what about adding this, I think wouldn't hurt adding more security? https://github.com/apostrophecms/sanitize-html#readme

jamiek-acl: `... It is well suited for cleaning up HTML fragments such as those created by CKEditor and...`.  Projects is already doing some stripping of fragments added by CKEditor in the back-end onSave.

Also, I am just putting the original content back into the link... I don't see how not sanitizing that can be dangerous?

jamiek-acl: i'll follow up with r&d in another ticket to see if this is necessary. for now please just review the styling changes here.

nahumzs (approved):

PR #907, opened 2021-01-27
Uxd 707 upgrade cypress

allison-c:
### Purpose 🚀

Upgrade cypress to 6

### Notes ✏️

- `get*` should be replaced to `find*`
-  Chained command will be applied to the parent element only
-  Using `find` and `findAll` explicitly
-  If `type()` doesn't work, check if the focus has been moved away, or reset the focus to the input
- Consider replacing `cy.contains(xx).should('not.be.visible')` to `xx.should('not.contain', xx)` or `.should('not. exist')`

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


nahumzs:
that was fast, was just find + replace? 😮

allison-c:
> that was fast, was just find + replace? 😮

Most of them yes, some of them did some investigation but still easy to fix 😆

nahumzs (approved):
did you test that the change also throws an error if they were checking to the opposite? just to check that the new introduction of commands always yields true?

allison-c:
> did you test that the change also throws an error if they were checking to the opposite? just to check that the new introduction of commands always yields true?

Yes, `find*` will throw an error if they didn't find anything, and in our tests, we usually chain `find` with `should` or `click/type` together

Issue #483, opened 2020-04-14
Tabs [feature] - Adding capability to programatically set the active tab


nahumzs:
## Describe the solution you'd like
The Tabs components should provide a way to select a tab programmatically.

nahumzs:
Maybe we can do this via https://reactjs.org/docs/hooks-reference.html#useimperativehandle

jamiek-acl:
Jira ticket: https://aclgrc.atlassian.net/browse/UXD-786

Issue #888, opened 2021-01-18
Tab onClick doesnt fire


jamiek-acl:
# Bug Report
The documentation says I should be able to add `onClick` to a `Tab`: https://github.com/acl-services/paprika/tree/master/packages/Tabs#tabstab

But if you pass in a function, it does not fire because `onClick` is overridden by the parent: https://github.com/acl-services/paprika/blob/master/packages/Tabs/src/Tabs.js#L28

So either make it work (preferable), or remove the prop.

jamiek-acl:
Looking into this more closely, making `Tabs.onClick` work isn't necessary as there is a `Tab.onClickTab` prop that does the same thing.

The `Tabs.onClick` prop was automatically added to the documentation (along with others), so in my PR I've updated the docs so it is clear that it is for internal use only.

PR #910, opened 2021-01-27
feat(Tabs): Controlled version

jamiek-acl:
### Purpose 🚀
Previously the `<Tabs>` component could only be run "uncontrolled" (active tab stored in internal state).  Now it can be also run "controlled" (active tab stored in external state).  This allows the consumer to change the active tabs if they want.

Issue: https://github.com/acl-services/paprika/issues/483

### Updates 📦
- [x] MINOR (backward compatible) change to Tabs

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-786--set-active-tab

Screen Shot 2021-01-27 at 3 50 12 PM


nahumzs:
Normally defaultValue, defaultIndex, etc are "standards" for uncontrolled components, I think would worth adding a property name `index` and used that one when it should behave as a controlled component, similarly to value and defaultValue in the Input component.

This also adds an extra benefit you now can test to check if the user intends to use the Tabs as controlled or uncontrolled, if both defaultIndex and index have an index then you can throw an error. 

Side comment, `onClick` would not be sufficient instead of `onClickTab`?

jamiek-acl:
I don't love `index` since it is so generic.  Umm...  `activeIndex`?

re: `onClickTab`, i just like verbose names to reduce potential confusion.

nahumzs:
sure `activeIndex` but that will break unification with defaultIndex? maybe we should deprecate defaultIndex and call it `defaultActiveIndex`,

We can make it as an alias, and console something like `console.warn("defaultActiveIndex is preferable than defaultIndex, please try to change the prop name?")` in case the user use it?


That path would not create a breaking change and alternative is just simple to make a breaking change and rename the `defaultIndex`

mikrotron:
> Side comment, `onClick` would not be sufficient instead of `onClickTab`?

`onClick` would have to be renamed when cloned or it could conflict with a custom `onClick` passed to the individual `Tab`.  

This rename should also resolve this issue: https://github.com/acl-services/paprika/issues/888

jamiek-acl:
I have had a classic change of heart and gone with `index`, so we can leave `defaultIndex` alone and maintain the `xxx/defaultXxx` pattern.

nahumzs (approved):

PR #918, opened 2021-02-01
UXD-803 docs(ListBox): added Showcase

oscarkwan:
### Purpose 🚀
- Make a better Showcase story for Listbox that has knobs
- Clean up some code / typos
- Deleting some stories that aren't really that relevant because we have more complete Showcase story

### Updates 📦
- No updates made to source code, strictly documentation

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-803-listbox-showcase

### Screenshots 📸

### References 🔗


packages/ListBox/stories/examples/Showcase.js
```diff
@@ -0,0 +1,57 @@
...
+import { Rule } from "storybook/assets/styles/common.styles";
+import * as characters from "../fixtures/characters";
+import ListBox from "../../src";
+
+const getKnobs = () => ({
+  hasImplicitAll: boolean("hasImplicitAll", false),
+  isDisabled: boolean("isDisabled", false),
+  placeholder: text("placeholder", "Select..."),
+  showFilter: boolean("Show Filter", false),
+  showFooter: boolean("Show Footer", false),
```
mikrotron, line 13: Should we add `height`, `isInline`, and `size`?

packages/ListBox/stories/examples/Showcase.js
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import { action } from "@storybook/addon-actions";
+import { boolean, text } from "@storybook/addon-knobs";
+import { Rule } from "storybook/assets/styles/common.styles";
+import * as characters from "../fixtures/characters";
+import ListBox from "../../src";
+
+const getKnobs = () => ({
+  hasImplicitAll: boolean("hasImplicitAll", false),
```
mikrotron, line 9: Does this have an effect in the Showcase? If we can't see the difference (even in the `actions` panel) then we should just omit it and explain it in the README, and/or MDX and/or Examples stories.

packages/ListBox/stories/examples/Showcase.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default function Showcase() {
+  const { showFilter, showFooter, ...listBoxProps } = getKnobs();
+
+  const ListBoxChildren = (
+    <>
+      {showFilter && <ListBox.Filter />}
+      {characters.villians}
+    </>
+  );
```
mikrotron, line 24: I'm not sure it matters so much for the Showcase source code (more important for Examples, for sure), but it's still a bit more readable to copy this inline.

packages/ListBox/stories/examples/Showcase.js
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <>
+      <h3>Single</h3>
```
mikrotron, line 28: `<StoryHeader>` please!
Also, maybe we can be a bit more descriptive? "Single-select – default" / "Multi-select – `isMulti`" or something like that? 

I think it would be good to indicate that the `isMulti` prop is the difference here.  Maybe it even works if `isMulti` is a knob...?

packages/ListBox/stories/examples/Showcase.js
oscarkwan: There were issues with `isInline, and size`. I talked to Nahum about `size` and it's because its not a controlled prop. It only works as like an "initial value". If i changed the prop it won't re-render with the new size

packages/ListBox/stories/examples/Showcase.js
oscarkwan: isMulti won't work as a knob unfortunately as the component doesn't re-render

packages/ListBox/stories/examples/Showcase.js
oscarkwan: I'll remove it for now and include it somewhere else in a future PR

mikrotron:
Interesting that the `clear` button doesn't fire the `onChange` callback (🐞?)

mikrotron (approved):
llgtm 22

nahumzs:
http://storybooks.highbond-s3.com/paprika/UXD-803-listbox-showcase/?path=/story/forms-listbox-docs--page#adding-custom-children-to-the-listbox 

has big images can you set a image size for it?

nahumzs:
http://storybooks.highbond-s3.com/paprika/UXD-803-listbox-showcase/?path=/story/forms-listbox-docs--page#with-custom-checkbox is missing a closing that the example should close `</ListBox.Option>`

nahumzs:
The `doc` page looks really good, do you think it might worth putting a link to the lazy example code? That's a common requirement.

oscarkwan:
@nahumzs this ticket is just primarily for the showcase, i can fix your comments on my upcoming PRs

PR #919, opened 2021-02-02
feat(StatusTracker): SNP-1125 support status without actions

allison-c:
### Purpose 🚀

### Notes ✏️
If the current status doesn't have actions available (maybe because of permission issues, or maybe it's already the last step) we will show a tabbable div there

Screen Shot 2021-02-02 at 1 19 21 PM

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/status-tracker`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/SNP-1125-support-status-without-actions

### Screenshots 📸

### References 🔗


packages/StatusTracker/src/components/Point/Point.js
```diff
@@ -106,13 +107,33 @@ const Point = React.forwardRef((props, ref) => {
...
+          kind={kind}
+          tabIndex={0}
+          data-pka-anchor="status-tracker.point"
+          {...a11yAttributes}
+        >
+          {name}
+        </sc.Point>
+      );
+    }
+
```
jamiek-acl, line 128: Hmmm, instead of having two `return` statements, do you think it's better to have just one?

```
isCurrent = kind === kinds.CURRENT
name = isCurrent ? name : " ";
as = isCurrent ? null : RawButton;
tabIndex = isCurrent ? 0 : null;

return (
  <sc.Point
    ...
  />
);

File: packages/StatusTracker/src/components/Point/Point.styles.js [hunk 2]
```diff
@@ -53,7 +53,40 @@ const futurePoint = css`
...
+  justify-content: center;
+  min-height: ${spacer(4)};
+  padding: ${tokens.spaceSm} ${tokens.spaceLg};
+  text-align: center;
+  transition-duration: 0.2s;
+  transition-property: border;
+  vertical-align: middle;
+
+  &:focus {
+    border-color: #276cf5;
```
jamiek-acl, line 77: $color--blue

packages/StatusTracker/src/components/Point/Point.styles.js
```diff
@@ -53,7 +53,40 @@ const futurePoint = css`
...
+    box-shadow: 0 0 0 2px #99cbfc;
```
jamiek-acl, line 78: $color--blue-lighten-30

packages/StatusTracker/src/components/Point/Point.styles.js
```diff
@@ -53,7 +53,40 @@ const futurePoint = css`
   border: 2px solid ${tokens.color.blackLighten20};
 `;
 
-export const Point = styled.div(({ kind }) => {
+const currentPointWithoutActions = css`
+  ${fontSize(-1)};
+  ${lineHeight(1)};
+  align-items: center;
+  background-color: ${tokens.color.white};
+  border: 1px solid rgb(215, 215, 215);
```
jamiek-acl, line 61: $border--color

packages/StatusTracker/src/components/Point/Point.styles.js
```diff
@@ -53,7 +53,40 @@ const futurePoint = css`
...
+  &:focus {
+    border-color: #276cf5;
+    box-shadow: 0 0 0 2px #99cbfc;
+    outline: none;
+  }
+
+  &:hover {
+    background: #f0f0f0;
+    border-color: #a4a4a4;
+    color: #3f3d3c;
```
jamiek-acl, line 85: these ones too, ha

packages/StatusTracker/src/components/Point/Point.js
allison-c: There's are 3 in total 😄 personally I prefer to show all the props within the returned element to make it easier to understand

packages/StatusTracker/src/components/Point/Point.js
```diff
@@ -60,6 +60,7 @@ const Point = React.forwardRef((props, ref) => {
           onFocus={handleTooltipOpen}
           onBlur={handleTooltipClose}
           kind={kind}
+          isDropdown
```
AndreyChernykh, line 63: `isDropdown === false` by default? 

If so, then this would be considered a braking-change (major version bump). Just a heads-up.

allison-c: This is an internal prop for css, apis are still same, only if you didn't put an Overflow menu inside a Point, we will show this white button-like trigger

jamiek-acl (approved):

Issue #909, opened 2021-01-27
Add `data-qa-anchor` on deeply nested components for e2e tests


tristanjasper:
.... For instance the close button in modal header needs to be simulated to have click events on it, so we should add a `data-qa-anchor` prop on this component.

Also look for any other components with a similar situation

# Bug Report
## Prerequisites
**This is just a checklist, please delete this section after**
Please answer the following questions for yourself before submitting a bug report.
- [ ] I checked to make sure that this bug has not already been filed
- [ ] I reproduced the bug in the latest version of paprika
- [ ] Please label as, low / medium / high priority

*Please fill in as much detail as possible below*

## Expected behavior
*Please describe the behavior you are expecting to happen.*

## Current behavior
*What is the current behavior?*

## Screenshots / Gifs / Codepens
*Include media to help illustrate the bug. Gifs help a lot!*

## Additional context
*Please provide any more additional details to illustrate the bug*

nahumzs:
To scope the report I think what we mmight need is to add props collectors for those components like `<OverflowMenu.Popover />` etc. instead of physically add an anchor. 

So people would be able to add whatever they want to the specific component.

PR #921, opened 2021-02-02
refactor(Modal): use context, header cleanup, add qa-anchor on close …

tristanjasper:
…button

### Purpose 🚀
Pr to address issue https://github.com/acl-services/paprika/issues/369

Primary Change 
- Refactored Modal to use Context over extractChildren for Modal components

Additional changes
- Added data-qa-anchor to modal header close button (issue https://github.com/acl-services/paprika/issues/909)
- Added isSingleLine prop to modal header to support text truncating long strings
- Added storybook examples for heading background colour change, and form as child example

### Notes ✏️
Note: Changing Modal to use a context approach is not going to solve the issue of Paprika `Button` components with `isSemantic="false"` attributes. We still need to dynamically add a new `button` or `input` element to trigger the onSubmit action. (we still need the logic at https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.js#L46)

### Updates 📦
- [x] MAJOR (breaking) change to Modal: there will be more top padding

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-812-modal-use-context

### Screenshots 📸
image
With heading string truncation

### References 🔗


packages/Modal/src/Modal.styles.js
```diff
@@ -80,25 +77,3 @@ export const Modal = styled.div(
     ${stylers.z(1)};
   `
 );
-
```
tristanjasper, line 83: moving these styles to be with the subcomponents files themselves

packages/Modal/src/Modal.styles.js
```diff
@@ -80,25 +77,3 @@ export const Modal = styled.div(
...
-  overflow-y: auto;
-
-  &:focus {
-    ${stylers.focusRing.subtle(true)};
-  }
-`;
-
-export const Content = styled(OriginalContent)`
-  ${Header} + ${ContentWrapper} > & {
-    margin-top: -${OriginalContent.SPACE};
```
tristanjasper, line 98: I removed this style. I'm not sure it is necessary, and doesn't look good if you change the background colour of the header for example. The content that is passed into the `Modal.Content` subcomponent should have the correct margins applied (eg margin-top: 0;) http://storybooks.highbond-s3.com/paprika/uxd-812-modal-use-context/?path=/story/messaging-modal-examples--with-header-coloured

packages/Modal/src/components/Header/Header.js
```diff
@@ -2,34 +2,54 @@ import React from "react";
...
+        isSingleLine={isSingleLine}
+        hasCloseButton={hasCloseButton}
+      >
         {children}
-      </Heading>
+      </sc.HeaderHeading>
       {hasCloseButton && (
-        <Button.Close data-pka-anchor="modal.header.close-button" isSemantic={false} onClick={onClose} size="medium" />
+        <Button.Close
+          data-qa-anchor="modal.header.close-button"
```
tristanjasper, line 47: Added this attribute for consumers to use (e2e and system tests). Addresses issue https://github.com/acl-services/paprika/issues/909

packages/Modal/src/components/Header/Header.styles.js
```diff
@@ -1,9 +1,21 @@
...
   padding: 0 ${spacer(2)};
 `;
+
+export const HeaderHeading = styled(Heading)(
+  ({ isSingleLine, hasCloseButton }) => css`
+    margin-right: ${spacer(2)};
+    ${isSingleLine && stylers.truncateText};
+    ${isSingleLine && `padding: ${spacer(2)} 0;`};
+    ${isSingleLine && hasCloseButton && `margin: 0 ${spacer(2)} 0 0; width: 95%;`};
+    ${isSingleLine && !hasCloseButton && `margin: 0;`};
```
tristanjasper, line 19: Anyone know a better way to write this, eg 
```
if(isSingleLine){
stylers.truncateText
padding: ${spacer(2)} 0;
closeButton ? margin: 0 ${spacer(2)} 0 0; width: 95% : margin: 0;
}
```
I couldn't get the syntax to work

packages/Modal/stories/Modal.stories.js
```diff
@@ -140,40 +144,72 @@ storiesOf(`${storyName}/Examples`, module)
...
         </Modal>
       </Story>
     );
-  });
-
-storiesOf(`${storyName}/Backyard/Sandbox`, module)
-  .add("with form render", () => (
-    <Modal
-      isOpen
-      as="form"
```
tristanjasper, line 161: Removed this story as I don't think we still want to render `as="form"` but prefer to now pass `form` element as a child. Added that story below

packages/Modal/src/Modal.js
```diff
@@ -49,13 +54,8 @@ const Modal = props => {
...
         <sc.Modal size={size} data-pka-anchor="modal.wrapper" {...moreProps}>
-          <sc.Dialog state={state} role="dialog" aria-modal="true" aria-label={ariaLabel} data-pka-anchor="modal">
-            {headerExtracted && <sc.Header {...headerExtracted.props} onClose={onClose} />}
-            <sc.ContentWrapper role="region" tabIndex="0">
-              {contentExtracted && <sc.Content {...contentExtracted.props} />}
-              {children}
-            </sc.ContentWrapper>
-            {footerExtracted && <sc.Footer {...footerExtracted.props} />}
+          <sc.Dialog state={state} role="dialog" aria-modal aria-label={ariaLabel} data-pka-anchor="modal">
+            <ModalContext.Provider value={contextValue}>{children}</ModalContext.Provider>
```
nahumzs, line 58: super `nip:` could you move the const `contextValue` closer to the return statement? so is easier to spot the values that will be share across the context 🙏

packages/Modal/src/components/Content/Content.styles.js
```diff
@@ -0,0 +1,17 @@
+import styled from "styled-components";
+import stylers from "@paprika/stylers";
+import { spacer } from "@paprika/stylers/lib/helpers";
+
+const SPACE = spacer(2);
```
jamiek-acl, line 5: this variable isnt needed

packages/Modal/src/Modal.styles.js
jamiek-acl: I think it makes sense to remove this and leave the content styling up to the consumer; but does this make it a breaking change?  Existing modals will look slightly different (the content will be 16px lower).

packages/Modal/src/components/Footer/Footer.styles.js
```diff
@@ -5,6 +5,7 @@ export const Footer = styled.div`
   align-items: center;
   background-color: ${tokens.modal.footer.backgroundColor};
   display: flex;
+  flex: none;
```
jamiek-acl, line 8:
```suggestion
  flex-grow: 0;
  flex-shrink: 0;
  flex-basis: auto;
```

packages/Modal/src/components/Header/Header.js
```diff
@@ -2,34 +2,54 @@ import React from "react";
...
+          data-pka-anchor="modal.header.close-button"
+          isSemantic={false}
+          onClick={onClose}
+          size="medium"
```
jamiek-acl, line 51: nit, `constants.size.MEDIUM`

packages/Modal/src/components/Header/Header.styles.js
jamiek-acl:
```javascript
export const HeaderHeading = styled(Heading)(
  ({ isSingleLine, hasCloseButton }) => css`
    margin-right: ${spacer(2)};

    ${isSingleLine &&
      `
      ${stylers.truncateText};
      padding: ${spacer(2)} 0;
      
      ${hasCloseButton ? `margin: 0 ${spacer(2)} 0 0; width: 95%;` : `margin: 0;`}
    `}
  `
);
```

tristanjasper: thanks @jamiek-acl

packages/Modal/src/Modal.styles.js
tristanjasper: I'm not entirely clear what constitutes a breaking change as this will be a minor visual change, but perhaps? 🤷‍♂️

jamiek-acl: @nahumzs what do you think, is this a breaking change?

jamiek-acl: or @mikrotron, do you think this is style change makes this a breaking change?

mikrotron: Yeah, even I find [this topic](https://github.com/semver/semver/issues/161) ambiguous. I don't think semver was created with CSS (or even UI) in mind.   

I guess I would frame it like this: _"Is this detail a part of the API contract that Paprika has with its consumers?"_

If we break the contract, it should be a breaking change.  If the margin of a subcomponent is part of the "contract", then that's a pretty specific and detailed contract! So I'd be inclined to say it is NOT a breaking changed, but should be clearly documented in the CHANGELOG.  

cc @alexzherdev

alexzherdev: yeah this is tricky and more of a judgement call. You could consider this a bug fix for the component in which case it's fair not to announce this as a breaking change (bug fixes often _are_ breaking changes but you don't make a major bump for every fix).
Since this is a pretty high-traffic component we may want to play it safe and notify consumers they may need a code change via a major bump.

packages/Modal/stories/Modal.stories.js
```diff
@@ -12,6 +13,9 @@ import * as types from "../src/types";
 
 import Modal from "../src";
 
+// eslint-disable-next-line import/no-named-as-default-member
+const { Header, Content, Footer, FocusLock } = Modal;
```
jamiek-acl, line 17: Can you not do this?  Consumers will use Paprika stories as their guide and I think it makes it harder to find, for example, any `Modal.Header`s in the code, as there are lots of `Header`s out there.

packages/Modal/src/Modal.styles.js
tristanjasper: Ok In my opinion also this should be considered a breaking change so lets go with that.

packages/Modal/src/components/Content/Content.styles.js
tristanjasper: Oh my change couldn't have been pushed..

jamiek-acl: i think it is resolved, i needed to refresh

packages/Modal/stories/Modal.stories.js
tristanjasper: Perhaps for a story it is ok? I tend to prefer the shorthand syntax, but don't mind. Have we established a pattern as in FormElement stories we use this approach

packages/Modal/src/components/Content/Content.styles.js
tristanjasper: Oh i already updated this https://github.com/acl-services/paprika/pull/921/files#diff-3371c2ca8904e71ec6ae31a706a9566fdcd72674217750415c25a93d261f1ea6R1

jamiek-acl (approved):

PR #925, opened 2021-02-04
Uxd 805 listbox stories csf refactor

oscarkwan:
### Purpose 🚀
- Refactor all stories to correct format for listbox
- Refactor stories to use CSF
- Refactor many irrelevant stories, replaced it with (subcomponents, and other stories)
- Clean up random UI

### Updates 📦
- Documentation only

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-805-listbox-stories-csf-refactor

### Screenshots 📸

### References 🔗


mikrotron:
`ListBox.storiesblah.js` – looks like you need at least one more commit 😁

packages/ListBox/stories/Trigger.stories.js
```diff
@@ -1,94 +0,0 @@
...
-
-function renderOptions() {
-  return options.map(option => <ListBox.Option key={option}>{option}</ListBox.Option>);
-}
-
-const togglePopover = (dispatch, types) => () => {
-  dispatch({ type: types.togglePopover });
-};
-
-storiesOf(`${storyName}/Subcomponents/Trigger`, module)
```
mikrotron, line 19: There's still a simple example in `Example/Trigger.js` and another example in the Docs page, but are you sure there's nothing good in this file that we're removing?

packages/ListBox/stories/Trigger.stories.js
```diff
@@ -1,94 +0,0 @@
...
-import { storiesOf } from "@storybook/react";
-import { getStoryName } from "storybook/storyTree";
-import Button from "@paprika/button";
-import RefreshIcon from "@paprika/icon/lib/Refresh";
-import ListBox from "../src";
-
-const storyName = getStoryName("ListBox");
-const options = ["Punisher", "Catwoman", "Venom", "Thunderbolts", "Deadpool", "Spawn", "Wolverine"];
-
-function renderOptions() {
```
mikrotron, line 11: Just a note: While usually I'd say this would be more readable to have this function inline, in this specific case where the focus is on the Trigger element, I'd argue this abstraction makes it _easier_ to focus on the important details.

packages/ListBox/stories/examples/multi.js
```diff
@@ -1,257 +0,0 @@
...
-  <ListBox isMulti>
-    {characters.villians.map(Option => (
-      <ListBox.Option label={Option.props.children}>
-        {({ isSelected }) => <OptionCustomCheckbox isSelected={isSelected} character={Option.props.children} />}
-      </ListBox.Option>
-    ))}
-  </ListBox>
-);
-
-export const WithGroups = () => (
```
mikrotron, line 35: We need to preserve this example – but in the Docs page is probably a better place for it.  There's not currently an example of groups there though, so I'm worried we'll forget if we don't keep anything in this PR.

packages/ListBox/stories/examples/multi.js
```diff
@@ -1,257 +0,0 @@
...
-export const TriggerIsHidden = () => {
-  return (
-    <ListBox
-      isInline
-      isMulti
-      onChange={selectedOptions => {
-        console.log(selectedOptions);
-      }}
-    >
-      <ListBox.Trigger isHidden />
```
mikrotron, line 186: We should make sure this little nugget gets kept somewhere.  Maybe just reinstated in a later PR, but let's make sure we don't forget.

packages/ListBox/stories/Trigger.stories.js
oscarkwan: That simple example is being used in the Docs page.

oscarkwan: I will expand on this more when i do the Docs page and explain more about the Trigger

mikrotron (approved):
llgtm 92

PR #926, opened 2021-02-04
UXD-822  Fix clear button for uncontrolled Input 🐐

mikrotron:
### Purpose 🚀
Refactor `<Input>` to all use of `hasClearButton` prop when component is uncontrolled.

### Notes ✏️
- These are `BREAKING CHANGES` – the previous implementation was flawed, and the new approach of firing `onChange(null)` makes the `onClear` callback prop redundant, so it was removed (see [inline comments](https://github.com/acl-services/paprika/pull/926/files#r570626922) for details).
- Also localization for the `aria-label` of the Clear Button.
- Added a `callAll()` helper function to `helpers` package.

### Updates 📦
- [x] MAJOR (breaking) change to `@paprika/input`
- [x] MINOR (backward compatible) change to `@paprika/helpers`
- [x] PATCH (bug fix) change to `@paprika/l10n`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-822--input-clear-fix

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-822


packages/Input/src/Input.js
```diff
@@ -4,25 +4,76 @@ import classNames from "classnames";
 import * as constants from "@paprika/constants/lib/Constants";
 import TimesCircleIcon from "@paprika/icon/lib/TimesCircle";
 import Button from "@paprika/button";
+import useI18n from "@paprika/l10n/lib/useI18n";
 import * as types from "./types";
 import * as sc from "./Input.styles";
 
+function callAll(...fns) {
```
mikrotron, line 11: Maybe this should be added to `@paprika/helpers` to be shared?

packages/Input/src/Input.js
```diff
@@ -38,33 +89,11 @@ const Input = React.forwardRef((props, ref) => {
...
-    size,
-    ...moreProps
-  } = props;
-
-  if (moreProps.value || moreProps.value === "") {
-    delete moreProps.defaultValue;
-  } else {
-    delete moreProps.value;
-  }
-
```
mikrotron, line 60: New approach is to use `isControlled`, defined above to more explicitly branch control.

packages/Input/src/Input.js
```diff
@@ -77,15 +106,19 @@ const Input = React.forwardRef((props, ref) => {
...
         data-pka-anchor="input"
         disabled={isDisabled}
         readOnly={isReadOnly}
-        ref={ref}
         {...moreProps}
+        className="form-input__input"
+        value={isControlled ? value : undefined}
+        defaultValue={!isControlled ? defaultValue : undefined}
+        onChange={callAll(handleChange, onChange)}
+        ref={ref || inputRef}
```
mikrotron, line 121: These should be protected from being overridden by `moreProps`

packages/Input/src/Input.js
```diff
@@ -153,7 +186,7 @@ const defaultProps = {
   a11yText: null,
   clearIcon: null,
   className: null,
-  defaultValue: "",
+  defaultValue: null,
```
mikrotron, line 189: Slight change to the API, but I don't think this constitutes a breaking change.

packages/Input/src/Input.js
```diff
@@ -4,25 +4,76 @@ import classNames from "classnames";
...
+    for (const fn of fns) {
+      if (fn) fn(...args);
+    }
+  };
+}
+
 const Input = React.forwardRef((props, ref) => {
-  const inputClearHandler = e => {
-    e.target.value = "";
-    props.onChange(e);
```
mikrotron, line 13: This is potentially a breaking change, but it only worked in some situations by accident.  The `e` in this case is the `click` event on the Clear Button, not the `change` event on the `<input>` so if the consuming app sets the state based on the `e` passed to `onChange` then the `target.value` is undefined and the input will be cleared.  

This is flawed.

Going forward the consumer should also connect the `onClear` callback to a handler that will clear the value (if the component is controlled).

packages/Input/stories/examples/InputExample.js
```diff
@@ -4,7 +4,19 @@ import Input from "../../src";
...
-  return <Input {...props} onChange={e => setValue(e.target.value)} ref={ref} value={value} />;
+  return (
+    <Input
+      {...props}
+      onChange={e => {
+        setValue(e.target.value);
+      }}
+      onClear={() => {
+        setValue("");
+      }}
```
mikrotron, line 15: This illustrates how a controlled component _should_ be used if it has a clear button.  It should not rely on the `onChange` prop alone.

packages/Input/src/Input.js
nahumzs: wouldn't hurt, up to you :)

packages/Input/src/Input.js
nahumzs: 🤔 I don't think this is completely wrong, this has good abstraction, the end-user doesn't need to know that the onClear Button even exist, Possible solution is to manually dispatch the event.

```
const event = new Event('change');

// Dispatch it.
refInput.current.dispatchEvent(event);
```

I think that would be a nicer solution than forcing a developer to wire the onClear callback.

packages/Input/src/Input.js
```diff
@@ -4,25 +4,76 @@ import classNames from "classnames";
...
+    hasClearButton,
+    hasError,
+    onChange,
+    onClear,
+    size,
+    value,
+    ...moreProps
+  } = props;
+
+  const isControlled = value !== null;
```
nahumzs, line 37: `null` is a value for any prop you should test for undefined.

React automatically filter those props with the value `undefined` so I think is closer of what we can to compare?

packages/Input/src/Input.js
```diff
@@ -4,25 +4,76 @@ import classNames from "classnames";
...
+
+  const [shouldShowClearButton, setShouldShowClearButton] = React.useState(
+    hasClearButton && !isDisabled && !isReadOnly && (value || defaultValue)
+  );
+
+  React.useEffect(() => {
+    setShouldShowClearButton(hasClearButton && !isDisabled && !isReadOnly && (value || defaultValue));
+  }, [hasClearButton, isDisabled, isReadOnly, value, defaultValue]);
+
+  const inputRef = React.useRef();
```
nahumzs, line 49: `nip` in mostly all of our components the variable using `useRef()` start with the prefix `ref`

packages/Input/src/Input.js
mikrotron: I actually tried this. It did not actually trigger the `handleChange` or `onChange` callbacks though 😒 

Maybe I was doing something wrong, but yeah, I tried `bestRef.current.dispatchEvent(new Event('change');` and nothing would actually happen.

packages/Input/src/Input.js
```diff
@@ -4,25 +4,76 @@ import classNames from "classnames";
...
+  const bestRef = ref || inputRef;
```
nahumzs, line 50: is bestRef trying to decide if you should consume the ref pass down from the developer or to use an internal `useRef` ?

packages/Input/src/Input.js
mikrotron: The `defaultProps` assign `null` as a default value for `value`.  I can't think of a time I've seen `undefined` as a value in `defaultProps`.

packages/Input/src/Input.js
mikrotron: Yes, exactly.

packages/Input/src/Input.js
nahumzs: 🤔 but `null` value yields that inputs need a string value, wouldn't be better `undefined` ?

packages/Input/stories/examples/InputExample.js
nahumzs: Would be awesome if we can avoid onClear and just dispatch an onChange event with `""` as value.

packages/Input/src/Input.js
mikrotron: Okay, I see we do sometimes use `undefined` in `defaultProps`, so I can change this.

packages/Input/stories/examples/InputExample.js
mikrotron: Agreed. But was unable to get that to work.

packages/Input/src/Input.js
nahumzs: `onChange({ target: { value: "" }})` 🙈

mikrotron: That will work in a very specific use, when the consumer only cares about `event.target.value`... do you think that's acceptable?

nahumzs: jokes aside if you can make it work with creating the event would be a way fancier and better experience, what you can do is to create an `id={`input_uuidV4`}` and attach it to the input, then with javascript find the real dom element, and use tha one to fire the Event, I'm suspecting that react doesn't work because we are trying to mount on top of the synthetic event.

mikrotron: I also tried:

```js
const event = document.createEvent("HTMLEvents");
event.initEvent("change", true, false);
document.getElementById("tester").dispatchEvent(event);
```

From https://stackoverflow.com/questions/39065010/why-react-event-handler-is-not-called-on-dispatchevent

But still no luck.  Looked at this package: https://github.com/vitalyq/react-trigger-change

But it says clearly in the README
> This module is a hack and is tightly coupled with React's implementation details. Not intended for production use. Useful for end-to-end testing and debugging.

It seems like this kind of thing is deliberately not supported by React.

nahumzs: what would happens if we trigger the onchange with a null no with the event, would that be too bad?
```
handleChange(event) {
  if (event) set(event.target.value) : set("")
}
```
this will work as well in ts signature because you could say is an event type | null

mikrotron: Yeah, I can dig that. But then I think we should remove the `onClear` prop, which is definitely a breaking change.

mikrotron: Also, we should match this behaviour in the `<ListBox>`, but that's a separate task 😉

nahumzs: unless some bug was introduced the onclear should report an onChange as well in the listbox...

nahumzs: yeah I think is a breaking change...

nahumzs (approved):

PR #928, opened 2021-02-05
UXD-826 conventional commits with release

allison-c:
### Purpose 🚀

- Updated pre-release.yml
  - Checkout from `master` branch and do the publishing work on a temporary clean branch
  - Using `--conventional-commits` option versioning the packages
- Added a similar semaphore yml for official release

Explanations are on the files changed page https://github.com/acl-services/paprika/pull/928/files please check the details there.

### Notes ✏️

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


.semaphore/pre-release.yml
```diff
@@ -22,8 +22,18 @@ blocks:
       jobs:
         - name: "Publish to NPM"
           commands:
+            - git checkout -b pre-release-$SEMAPHORE_PIPELINE_ID
+            - git push origin pre-release-$SEMAPHORE_PIPELINE_ID --no-verify
```
allison-c, line 26: Checkout and do the publishing work on a clean branch, to avoid the error happening because of a PR just merged in during the publishing task.
I have to push this branch, otherwise lerna will complain about this https://github.com/lerna/lerna/blob/main/commands/version/lib/is-behind-upstream.js

.semaphore/pre-release.yml
```diff
@@ -22,8 +22,18 @@ blocks:
...
             - nvm use
             - npx lerna bootstrap
             - yarn
-            - yarn lerna version prerelease --yes
+            - yarn lerna version --conventional-commits --conventional-prerelease --no-changelog --no-push --yes
```
allison-c, line 30: Using `--conventional-commits --conventional-prerelease` will automatically assign the correct alpha version to packages.
They're using this preset by default: https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular I didn't find any problem with it

allison-c: Adding `--no-changelog` because we currently have our own changelog generator, they might conflict with each other. We can revisit this issue once we resolve auto versioning.

allison-c: Adding `--no-push` just for the convenience of the CI, and we will push everything (including tags) at line https://github.com/acl-services/paprika/pull/928/files#diff-c9c651f685e154b266bdbb551f05c643a0a41991154dac0413c2b8f63b19f7a0R37

.semaphore/pre-release.yml
```diff
@@ -22,8 +22,18 @@ blocks:
...
             - yarn lerna publish from-package --yes --dist-tag next
+            - git add packages/*.md packages/*/index.d.ts
+            - git commit -m 'chore(Publish): automate readme, type definitions and changelog update' --no-verify
+            - git push origin pre-release-$SEMAPHORE_PIPELINE_ID --follow-tags --no-verify
```
allison-c, line 34: Commit the changes for readme and declaration files. These tasks were done by the `postpublish` script before. But since now I want to do the publishing work on a clean branch, so I have to explicitly say push to xx branch rather than just master branch like before.
Also I removed `postpublish` script in `package.json`

.semaphore/pre-release.yml
```diff
@@ -22,8 +22,18 @@ blocks:
...
+            - git checkout master
+            - git merge origin/pre-release-$SEMAPHORE_PIPELINE_ID
+            - git push origin master -f --no-verify
+            - git branch -d pre-release-$SEMAPHORE_PIPELINE_ID
+            - git push origin -d pre-release-$SEMAPHORE_PIPELINE_ID
```
allison-c, line 39: These 5 lines are merging this temporary branch into master and then delete it from both local and remote.

.semaphore/release.yml
```diff
@@ -0,0 +1,39 @@
...
+          - echo "//registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN" >> .npmrc
+      jobs:
+        - name: "Publish to NPM"
+          commands:
+            - git checkout -b release-$SEMAPHORE_PIPELINE_ID
+            - git push origin release-$SEMAPHORE_PIPELINE_ID --no-verify
+            - nvm use
+            - npx lerna bootstrap
+            - yarn
+            - yarn lerna version --conventional-commits --conventional-graduate --no-changelog --no-push --yes
```
allison-c, line 30: `--conventional-graduate` will make the official release based on the alpha verison. e.g. `v2.0.1-alpha.12` → `v2.0.1`

jamiek-acl (approved):
LGTM; i wish `pre-release` were called `alpha-release` but not a big deal i guess. Similarly i kinda wish `pre-release-prepare` and `pre-release` were combined.

nahumzs:
Before mergin this, could you add a description, what does this mean for the developers, what we should be considering when implementing this plan?

Can we keep modifying multiple-components in one PR? are we forced to do just one action per PR? 

What does this mean for the official release and the alpha release?

I understand from the technical part what is happening, but I would like to understand the experience for the developers after this gets merged. :)

allison-c:
> Before mergin this, could you add a description, what does this mean for the developers, what we should be considering when implementing this plan?
> 
> Can we keep modifying multiple-components in one PR? are we forced to do just one action per PR?
> 
> What does this mean for the official release and the alpha release?
> 
> I understand from the technical part what is happening, but I would like to understand the experience for the developers after this gets merged. :)

Sure! We have created a ticket to document and answer those questions. Also I think we shouldn't use force push on semahpore, but we already use it and this script uses it as well, so I don't think I will merge it now. Except for the force push, other parts should be fine, I've tested manually

PR #929, opened 2021-02-05
uxd-818 Create tag component, consolidate Pill with ListboxWithTags Pill subcomponent

tristanjasper:
### Purpose 🚀
Create a new Tag component that consolidates the Paprika Pill component with the ListboxWithTags components Pill subcomponent

Additional changes:
Updated ListboxWithTags component to use created Tag component
Added support for alert Tag style required in https://aclgrc.atlassian.net/browse/UXD-785

Todo:

1. Add Unit Tests
2. Fix text truncation not showing ellipses

### Notes ✏️

### Updates 📦
- [x] MAJOR (breaking) change to Tag, ListboxWithTags

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-818-Create-Tag-Component

### Screenshots 📸
image

### References 🔗

https://aclgrc.atlassian.net/browse/UXD-818
https://aclgrc.atlassian.net/browse/UXD-785


packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,165 @@
...
+  const paddingGap = {
+    medium: `${tokens.space[0] / 2}px`,
+    large: tokens.space,
+  };
+
+  const padding = isChildString
+    ? `padding: 0 ${paddingGap[size]} 1px ${paddingGap[size]};`
+    : `padding-right: ${paddingGap[size]}`;
+
+  // todo: fix truncateText not showing ... with display inline-flex, only with display: block....
```
tristanjasper, line 114: Truncation is not working properly, as is not showing the `...` without `display: block`. When using `display:block` however the text content in the Tag was being vertically aligned incorrectly by the font size of text content in the avatar beside it.

packages/ListBoxWithTags/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+2021-02-05
```
mikrotron, line 9: You can omit the date, it will be included by the script that publishes to npm

packages/Tag/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+2020-02-03
```
mikrotron, line 9: Same ✂️

packages/Tag/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@paprika/raw-button": "^1.0.3",
+    "@paprika/stylers": "^1.0.0",
+    "@paprika/tokens": "^1.0.0",
+    "@paprika/icon": "^1.0.4",
+    "prop-types": "^15.7.2"
+  },
+  "peerDependencies": {
+    "@paprika/l10n": "^1.1.3",
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1",
```
mikrotron, line 30: 17? 😳  
I don't think any of our apps meet this.

packages/Tag/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "styled-components": "^5.2.1"
```
mikrotron, line 31: Likewise, I would keep this at `v4`

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Close from "@paprika/icon/lib/Times";
+import * as sc from "./Tag.styles";
+import * as types from "./types";
+
+export function Delete({ isDisabled, size, onRemove = () => {} }) {
+  return (
+    <sc.Delete isDisabled={isDisabled} size={size} data-pka-anchor="tag.delete" onClick={onRemove}>
```
mikrotron, line 9: nit: I suggest `tag.remove` for the anchor name.  This matches better with the prop name `onRemove`.

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Close from "@paprika/icon/lib/Times";
+import * as sc from "./Tag.styles";
+import * as types from "./types";
+
+export function Delete({ isDisabled, size, onRemove = () => {} }) {
```
mikrotron, line 7: Do you need to export this? If so, I think it should be in its own file.  If it's only used here in the Tag component I think it's small enough to leave inline (though maybe it's still best practice to extract it).

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+  ${tagColorStyles[tagColor]}
+
+  [data-pka-anchor="avatar"] {
+    margin-right: ${tokens.space};
+  }
+`;
+
+export const RawButtonTag = styled(RawButton)(tagStyles);
+
+export const Tag = styled.li(tagStyles);
```
mikrotron, line 97: Can you just do `<div as={condition ? "li" : RawButton}>` in the component?

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+  return css`
+    align-items: center;
+    display: inline-flex;
+    ${getFontSize[size]};
+    ${padding};
+    ${truncateText};
+  `;
+});
+
+export const Delete = styled(RawButton)(({ size }) => {
```
mikrotron, line 117: Can you use the `Button.Close` instead?

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+      {onRemove ? <Delete isDisabled={isDisabled} onRemove={handleRemove} size={size} /> : null}
+    </>
+  );
+
+  if (onClick) {
+    if (!a11yText) {
+      console.warn("Tag component should include an a11yText prop if an onClick is provided.");
+    }
+    return (
+      <sc.RawButtonTag a11yText={a11yText} isDisabled={isDisabled} onClick={onClick} {...moreProps}>
```
mikrotron, line 48: Should also include `data-pka-anchor="tag"`

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const propTypes = {
+  // used in aria-tag on the root element
+  a11yText: PropTypes.string,
+  // Can pass a custom border color
+  borderColor: PropTypes.string,
+  // Pass a function to show a delete button
+  onRemove: PropTypes.func,
+  children: PropTypes.node.isRequired,
```
mikrotron, line 74: Can you add a brief description of `children` too, for completeness? Also, I think `react-docgen` is picky about comment styling, and wants you to use `/** ... */` format.

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+    Tag.types.color.GREY,
+    Tag.types.color.ORANGE,
+    Tag.types.color.LIGHT_BLUE,
+    Tag.types.color.LIGHT_ORANGE,
+    Tag.types.severity.NO_RISK,
+    Tag.types.severity.LOW_RISK,
+    Tag.types.severity.MEDIUM_RISK,
+    Tag.types.severity.HIGH_RISK,
+  ]),
+  size: PropTypes.oneOf([types.MEDIUM, types.LARGE]),
```
mikrotron, line 93:
```suggestion
  size: PropTypes.oneOf([Tag.types.size.MEDIUM, Tag.types.size.LARGE]),
```

mikrotron: Just for documentation purposes.

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const defaultProps = {
+  borderColor: null,
+  a11yText: null,
+  isDisabled: false,
+  onClick: null,
+  onRemove: null,
+  size: types.MEDIUM,
```
mikrotron, line 102:
```suggestion
  size: Tag.types.size.MEDIUM,
```

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+
+  return (
+    <sc.Tag data-pka-anchor="tag" {...moreProps}>
+      {content}
+    </sc.Tag>
+  );
+}
+
+Tag.types = {
+  size: PropTypes.oneOf([types.MEDIUM, types.LARGE]),
```
mikrotron, line 62: This shouldn't be a `PropTypes` definition, just an object of valid sizes.  Probably best to define in `./types.js` then just follow the same pattern as below (`size: types.sizes,`)

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
+import styled, { css } from "styled-components";
+import tokens from "@paprika/tokens";
+import RawButton from "@paprika/raw-button";
+import { fontSize } from "@paprika/stylers/lib/helpers";
+import { truncateText } from "@paprika/stylers/lib/includes";
+
+const tagColorStyles = {
+  black: css`
```
mikrotron, line 8: Should probably be using the `types` for these colour keys

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+
+const borderColorStyles = ({ borderColor }) => css`
+  border: 1px solid ${borderColor};
+`;
+
+const tagStyles = ({ tagColor, borderColor }) => css`
+  align-items: center;
+  background: ${tokens.color.blackLighten70};
+  ${borderColor ? borderColorStyles : ""}
+  border-radius: ${tokens.space[0] * 2}px;
```
mikrotron, line 80:
```suggestion
  border-radius: ${stylers.spacer(2)}px;
```

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+    medium: "20px",
+    large: "24px",
+  };
+
+  return css`
+    align-items: center;
+    border-radius: 50%;
+    box-sizing: border-box;
+    display: flex;
+    flex: 0 0 auto;
```
mikrotron, line 138:
```suggestion
    flex-shrink: 0;
```

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+
+    &:hover {
+      background: ${tokens.color.blackLighten60};
+    }
+
+    &:focus {
+      background: ${tokens.color.blackLighten60};
+    }
+
+    & svg {
```
mikrotron, line 153: `&` not needed.  
And maybe `[data-pka-anchor="icon"]` would be a more explicit selector.

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+      color: ${tokens.color.blackLighten20};
+      pointer-events: none;
```
mikrotron, line 155: Why no pointer events?

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+  a11yText: PropTypes.string,
+  // Can pass a custom border color
+  borderColor: PropTypes.string,
+  // Pass a function to show a delete button
+  onRemove: PropTypes.func,
+  children: PropTypes.node.isRequired,
+  // Disables tag onClick and delete button functionality
+  isDisabled: PropTypes.bool,
+  // Fires when clicking the root tag element. Should also pass value for a11yText when using this.
+  onClick: PropTypes.func,
```
mikrotron, line 78: If no a11yText is passed there will be no aria-label on the button, but if there's text content, that will be used by AT.  Only if there's no text content in the `children` is it really a problem (like for an icon button). So you can omit that instruction and remove the warning for `!a11yText`.

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+  color: types.colors,
+  severity: types.severityTagColors,
+};
+
+const propTypes = {
+  // used in aria-tag on the root element
+  a11yText: PropTypes.string,
+  // Can pass a custom border color
+  borderColor: PropTypes.string,
+  // Pass a function to show a delete button
```
mikrotron, line 72:
```suggestion
  // Pass a function to show a remove button
```

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+  onRemove: PropTypes.func,
+  children: PropTypes.node.isRequired,
+  // Disables tag onClick and delete button functionality
+  isDisabled: PropTypes.bool,
+  // Fires when clicking the root tag element. Should also pass value for a11yText when using this.
+  onClick: PropTypes.func,
+
+  tagColor: PropTypes.oneOf([
```
mikrotron, line 80: Can you add a comment for `tagColor` and `size` props too please?
I think it would be nice to alphabetize the props and `defaultProps` too.

packages/Tag/stories/Basic.stories.js
```diff
@@ -0,0 +1,20 @@
...
+};
+
+const Story = styled.div`
+  padding-bottom: 200vh;
+`;
+
+export function BasicExample() {
+  return (
+    <Story>
+      <Tag locale="en">some content</Tag>
```
mikrotron, line 17: `locale`?

packages/Tag/stories/Tag.stories.js
```diff
@@ -0,0 +1,105 @@
...
+
+import Tag, { Tags } from "../src";
+import * as types from "../src/types";
+
+export default {
+  title: "Tag",
+  component: Tag,
+};
+
+export function TagsExample() {
```
mikrotron, line 16: Can you extract this to `/stories/Examples/Variations.js` and use the `Tag.stories.js` as an index/manifest for stories in the root folder (of the Storybook navigation)?

No need to go crazy with this PR, I think the Variations story has the most value of anything, so you can probably even drop the `Basic` one. 

I would recommend adding this component to the `storyTree` though, so it fits into the Storybook IA.

This guide is still pretty rough, but it can probably still help with the details:
https://github.com/acl-services/paprika/wiki/Storybook-Curation-Guide

packages/Tag/stories/Tag.stories.js
```diff
@@ -0,0 +1,105 @@
...
+  const handleRemove = () => {
+    console.log("remove event");
```
mikrotron, line 18: These are nice to have, but can we use the Storybook actions panel instead of console.log?

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+
+export const Tags = styled.ul(() => {
+  return css`
+    display: flex;
+    flex-wrap: wrap;
+    line-height: 1;
+    margin: 0;
+    padding: 0;
+  `;
+});
```
mikrotron, line 70: This is interesting! 🤔 

However, a `<ul>` with `<RawButton>` children won't be recognized as a valid list. There may be more than one solution to this, but would it be possible to just wrap the `<RawButton>` version in an `<li>`? 

Actually... that's problematic if it's not a list of tags, just one.  A proper solution may not be so simple!

sunminsarahlee:
I found that Red Error Pills have 2px extra height, 26px and 30px, because the border was drawn outside. Can we draw the border inside so the height of red pill are the same as other pills? 24px and 28px are the correct height. 
Screen Shot 2021-02-08 at 5 16 17 PM


Just random question/feedback: Why are we showcasing a Calendar icon in the Tag? I can see it's from Avatar but can we have a UserGroup icon instead of Calendar?

mikrotron:
@sunminsarahlee, should there be a visual treatment for disabled `Tags`?

For example, all the little pieces of a disabled `FormElement` get set to `opacity:0.5`, which is quite simple to implement.  Can/should we do the same here?  Much easier than defining disabled shades for every colour option.

packages/Tag/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@paprika/tag",
+  "version": "1.0.0",
```
mikrotron, line 3: I think we want to bump this to `v1.0.0` during the first publish, so maybe drop this to `0.0.1`?

packages/Tag/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "description": "The Tag component is for showing metadata like status, type, or levels.",
```
mikrotron, line 4: This is a bit narrow... from the UX pitch:

> Tag can be both static UI labels or removable components. Static Tag is mostly used to show an item’s status and removable Tag will be used to add or group objects. 

The grammar can be touched up a little, and if you can condense it a little more, that would be even better.

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,158 @@
...
+  align-items: center;
+  background: ${tokens.color.blackLighten70};
+  ${borderColor ? borderColorStyles : ""}
+  border-radius: ${tokens.space[0] * 2}px;
+  color: ${tokens.color.white};
+  display: flex;
+  line-height: 1;
+  margin-bottom: 2px;
+  margin-right: ${tokens.spaceSm};
+  max-width: 100%;
```
mikrotron, line 86: I think a `maxWidth` prop might be a better way to handle truncation.  The default value could still be `100%` though.

packages/Tag/package.json
tristanjasper: This is what ListboxWithTags component was using, so i went with that.
Ok I'll change to 
```
"react": "^16.8.4",
    "react-dom": "^16.8.4",
    "styled-components": "^4.2.0"
    ```.
   I'll update it in ListboxWithTags component as well

File: packages/Tag/src/Tag.styles.js [hunk 30]
```diff
@@ -0,0 +1,158 @@
...
+  padding: 2px;
+  ${tagColorStyles[tagColor]}
+
+  [data-pka-anchor="avatar"] {
```
mikrotron, line 90: After fiddling a little with this, I think your truncation issue is going to be a lot easier to solve if you give the tag an `avatar` prop – then you can wrap the rest of the label in a an `<Ellipsis>`.  

Also might make a nicer, more explicit API.

packages/Tag/src/Tag.styles.js
tristanjasper: This didn't work for me when working with styled components....

packages/Tag/src/Tag.styles.js
tristanjasper: @mikrotron What do you think about naming these "themes" instead? Most of them set not only the `background-color`, but also the `color` for the text. For the "alert" theme it could also add the `border-color` as part of the theme...

mikrotron: I think you've got a good point, I'm a little worried that the actual term `theme` comes with some preconceptions... 🤔 But since a bunch of the `color` options are actually risk levels, and it would make a lot of sense to include `border-color` in the same prop. 

Unless @sunminsarahlee has a better suggestion though, I think `theme` is better than `tagColor` for now.   If we decide to rename it we'll just have to bump the version 🤷‍♂️

mikrotron: [Blueprint.js](https://blueprintjs.com/docs/) uses `intent` throughout their design system instead of `color`. But that's just one example.

packages/Tag/src/Tag.styles.js
tristanjasper: Can't recall why that was added, I think it interfered with hover styles at one point

packages/Tag/src/Tag.styles.js
sunminsarahlee: im good with `theme`

sunminsarahlee:
> @sunminsarahlee, should there be a visual treatment for disabled `Tags`?
> 
> For example, all the little pieces of a disabled `FormElement` get set to `opacity:0.5`, which is quite simple to implement. Can/should we do the same here? Much easier than defining disabled shades for every colour option.

Currently, we don't have any use case for disabled `Tags`. Tag is mainly used to show an item’s status so if the item got disabled then we simply don't need to show the Tag rather than disabling Tag. For some edge cases, we will show Error Tag. We can add disabled Tag when we need it later.

packages/Tag/src/Tag.styles.js
tristanjasper: Hmm, i think it can be up to the consumer to wrap the `Tag` component in an `li` if they need to, ie if the need `onClick` support for the Tag and need more than 1.

tristanjasper:
> I found that Red Error Pills have 2px extra height, 26px and 30px, because the border was drawn outside. Can we draw the border inside so the height of red pill are the same as other pills? 24px and 28px are the correct height.
> Screen Shot 2021-02-08 at 5 16 17 PM
> 
> Just random question/feedback: Why are we showcasing a Calendar icon in the Tag? I can see it's from Avatar but can we have a UserGroup icon instead of Calendar?

It's just for an example of any icon, so isn't particular to one. Perhaps later we can add more examples if it's needed.

packages/Tag/package.json
tristanjasper: So why shouldn't I leave it as 1.0.0 then? As it hasn't been published yet?

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,107 @@
...
+    onRemove();
+    e.stopPropagation();
+  };
+
+  const content = (
+    <>
+      <sc.Ellipsis isChildString={isChildString} size={size}>
+        {children}
+      </sc.Ellipsis>
+      {onRemove ? <Delete isDisabled={isDisabled} onRemove={handleRemove} size={size} /> : null}
```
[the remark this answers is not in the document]
nahumzs: We had discussed about this several time before, sadly I don't remember what was the agreements, cc @mikrotron do you remember or do we have something about this in our wiki? I remember @jamiek-acl pushed for using the function as an indicator to enable button.

If we follow the input API you should include the prop hasDeleteButton as hasClearButton. And just for that reason, my vote would go to consistency.

packages/Tag/src/Tag.styles.js
```diff
@@ -0,0 +1,165 @@
...
+    : `padding-right: ${paddingGap[size]}`;
+
+  // todo: fix truncateText not showing ... with display inline-flex, only with display: block....
+  return css`
+    align-items: center;
+    display: inline-flex;
+    ${getFontSize[size]};
+    ${padding};
+    ${truncateText};
+  `;
```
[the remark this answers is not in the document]
nahumzs: I think was forcing the Tag to use all its possible space in a flex layout, just be sure the ListBoxTags is still working correctly and we should be good to go :)

packages/Tag/src/Tag.styles.js
nahumzs: I had the same problem in a component that I'm currently working on, what I did is make use of two divs one that wraps the flex vertical attribute, and another one `nested` that use display block, here you can see the example that I think is similar to your issue https://github.com/acl-services/paprika/pull/939/files#diff-09d3038e83adb04da25396535ba5f213c0d69ea92fc68027ea33e0bbb7002c0dR3

packages/Tag/package.json
nahumzs: 🙏 this was actually a copy/paste issue that I forgot to remove, will not break anything but as other have mention in the slack channel we should use the bare minimum version that our component can support :)

packages/Tag/src/Tag.styles.js
nahumzs: > Can you just do <div as={condition ? "li" : RawButton}> in the component?

I think this doesn't work because babel needs to compile this, since `as` is a prop the styled-components plugin uses not react, so doesn't run at runtime. But is just my assumption why is not working.

packages/Tag/src/Tag.styles.js
tristanjasper: Ok, I discussed with Mike, I'm passing the avatar in as a prop now and using `display:block` on the tag content, so ellipse works now and i think is good for the api

packages/Tag/src/Tag.styles.js
nahumzs: This is my proposal. 
```jsx
// Pill component
Tags.theme = (...args) => {
  const [color=null, background=null, border=null] = args;
  let _color = "black";
  let _backround = "white";
  let _border: "1px solid transparent";
  
  if(color) _color = color
  if(background) _background = background;
  if(border) _border = border;
  
  return css`${color}; ${background}; ${border}`;
}
Tags.types.kind = { 
  danger: Tags.theme(["red", "white", "red"]), 
  success: Tags.theme(["white", "green"]) 
};
```

// 
```jsx
<Tag kind={Tags.types.kind.success} />
<Tag kind={Tags.theme(["blue", "white", "blue"])} />
```

With an approach like this, the component could still be behaving similar to Button component with the prop `kind` but will also have a powerful function that lets you theme some part of the component, without breaking the size and the main design.

@mikrotron @tristanjasper

nahumzs: `kind` can have of course their default value. and the css value that you get from `kind` prop can be spread directly in the styled component.

packages/Tag/src/Tag.js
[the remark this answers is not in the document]
jamiek-acl: I think it was decided to _NOT_ do that (though I think it is silly).
I think the example was the old Multiselect, when passing in `onFilter` it would show the filter field (no need for a `showFilter` prop, too).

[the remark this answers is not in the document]
mikrotron: I remember we made the decision to have an explicit `has` prop during the days of `acl-ui`, but then changed our minds for Paprika. I think it was Jamie's argument that convinced enough of us during ShopTalk... but it was long enough ago that I don't remember the details, and it doesn't seem to be captured in the wiki.  But I think it was in relation to the API for the `<DialogActions>` component, and it currently follows the same pattern as Tristan has here, so I think we should stick with it:

https://github.com/acl-services/paprika/blob/6b3f25e0cc77cfa9fe2829bf07781bab703431e8/packages/DialogActions/src/DialogActions.js#L51

packages/Tag/src/Tag.styles.js
mikrotron: It _can_ work:
https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.js#L102

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -84,34 +82,34 @@ const renderTrigger = ({ t, size, selectedOptions, onRemove, renderPill, pillLab
...
               `Your item ${JSON.stringify(
                 item
-              )} must include the attribute "label", or you must indicate which attribute should be rendered as the label via the "pillLabelKey" prop.`
+              )} must include the attribute "label", or you must indicate which attribute should be rendered as the label via the "tagLabelKey" prop.`
             );
           }
 
           return (
-            <Pill key={label} onRemove={handleRemove(item)} size={size}>
+            <Tag key={label} onRemove={handleRemove(item)} size={size}>
```
mikrotron, line 102: Should there be an `<li>` wrapper here then, since you're using `<Tags>`?

packages/Tag/package.json
mikrotron: Not a big deal, but when it gets published it will get incremented to `v1.0.1`.

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,98 @@
...
+
+  if (onClick) {
+    return (
+      <sc.RawButtonTag
+        a11yText={a11yText}
+        isDisabled={isDisabled}
+        onClick={onClick}
+        size={size}
+        {...moreProps}
+        data-pka-anchor="tag"
```
mikrotron, line 32: I think it's okay to allow the `anchor` to be overridden by `moreProps`.  This allows the `ListBoxWithTags` to label this as `list-box.tag` for example.

Also, the non-button version below allows it with `moreProps` coming last, so best to be consistent.

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,98 @@
...
+  const { children, onRemove, size, onClick, a11yText, avatar, isDisabled, ...moreProps } = props;
+
+  const handleRemove = e => {
+    onRemove();
+    e.stopPropagation();
+  };
+
+  const content = (
+    <>
+      {avatar}
```
mikrotron, line 16: I think you should handle this the same as the delete button –conditionally render it and wrap in a `<sc>` component for the styling (padding).

packages/Tag/src/Tag.js
```diff
@@ -0,0 +1,98 @@
...
+  severityTheme: types.severityThemes,
+};
+
+const propTypes = {
+  /** used in aria-tag on the root element */
+  a11yText: PropTypes.string,
+  /** Can pass a avatar to be displayed to the left of the tag content */
+  avatar: PropTypes.node,
+  /** Can pass a custom border color */
+  borderColor: PropTypes.string,
```
mikrotron, line 58: I think the best part about having a `theme` prop was the ability to remove the unrestricted `borderColor` prop and roll that into it.

packages/Tag/src/Tag.styles.js
mikrotron: However, this still allows for an invalid orphaned `<li>` tag with no `<ul>`.  If the consumer doesn't use `<Tags>` or include `onClick` on the `<Tag>`.  

So I think this should actually render a `div`, and the consumer will be obligated to wrap their `<Tag>` children in `<li>` elements if they use `<Tags>`.  

(Maybe that should be named `<TagList>` to be more explicit?)

packages/Tag/stories/Examples/Variations.js
```diff
@@ -0,0 +1,141 @@
...
+              {`${permutation.tagSize} with alert theme`}
+            </Tag>
+            <Tag size={permutation.tagSize} onRemove={handleRemove} borderColor="green">
+              {`${permutation.tagSize} with custom border`}
+            </Tag>
+          </Tags>
+        ))}
+      </div>
+      <div style={{ padding: "20px" }}>
+        Colours
```
mikrotron, line 98: nit: could use `<Gap>` and `<StoryHeader>` here

packages/Tag/stories/Examples/Variations.js
```diff
@@ -0,0 +1,141 @@
...
+const handleClick = () => {
+  action("Click event")();
+};
+
+export default function Variations() {
+  return (
+    <L10n locale="en">
+      <div style={{ padding: "20px" }}>
+        Variations
+        {[{ tagSize: "medium", avatarSize: "small" }, { tagSize: "large", avatarSize: "medium" }].map(permutation => (
```
mikrotron, line 25: I like this idea of mapping the permutations! Do you think it would be more readable to extract the `permutations` array into a variable declared at the top?

packages/Tag/stories/Examples/Variations.js
```diff
@@ -0,0 +1,141 @@
...
+import CalendarIcon from "@paprika/icon/lib/Calendar";
+import { action } from "@storybook/addon-actions";
+import ExclamationCircle from "@paprika/icon/lib/ExclamationCircle";
+import tokens from "@paprika/tokens";
+import Avatar from "../../../Avatar/src";
+import Tag, { Tags } from "../../src";
+
+const handleRemove = () => {
+  action("Remove event")();
+};
```
mikrotron, line 14:
```suggestion
const handleRemove = action("Remove event");
```

packages/Tag/src/Tag.styles.js
mikrotron: I like where this suggestion is going, but have some suggestions about the API design.  For now, I think it's okay to proceed with the simple `theme` prop as @tristanjasper has implemented so he can get this merged.  We can bump to `v2.0` if we like this kind of API for Paprika, and start using it more consistently throughout the library.

mikrotron: Suggestions:

- Rename `kind` → `theme` and `theme` → `themeGenerator` (or something better)
- Use an object for the `args` instead of an array
- Consider how we can document this API (specifically the shape of those `args` for `themeGenerator`)

packages/Tag/src/Tag.styles.js
tristanjasper: Ok yes there is the common use case that the Tag component is rendered on it's own, in which case I think it should by default be a `div`, so lets go with that. The consumer can then render it `as` something else if needs be (or wrap it in an `li` element alternatively)

packages/Tag/package.json
tristanjasper: Ok, i wasn't aware, good to know, i'll set as 0.0.1 then for now

packages/Tag/src/Tag.styles.js
tristanjasper: Ok, as we discussed we can leave this to the consumer to dictate as there may be other surrounding elements such as `<li>` that also need a width set on them. We can revisit in a version as well when we could add a surrounding context in the `tags` component (which is exported with `Tag`

packages/Tag/src/Tag.styles.js
tristanjasper: done

packages/Tag/src/Tag.js
tristanjasper: I'm not sure I see the benefits of this, the delete button should be conditionally rendered as there is a prop (callback) being passed and not an element that should be rendered, so slightly different. The styling that gets applied to avatar is only if it exists... so don't really see that it needs an extra layer of being a styled component...

packages/Tag/src/Tag.js
tristanjasper: I played about with it and because we need to have the height of the Tag be adjusted if there is a border being applied it started to get messy applying border as part of the theme as well....

https://github.com/acl-services/paprika/pull/929/files#diff-c6887f020b1e2c81ec422eb1bcf30718a2e7104a195bb691e717bc16d700892dR98

Providing the `borderColor` prop still gives flexibility to the consumer to apply a `borderColor` when they like (which affects the `height` and `padding` on the Tag) so I think is an enhancement. I think for now, this works but perhaps can revisit how the border can still be applied via the theme later.

packages/Tag/src/Tag.js
mikrotron: Fair enough.

packages/Tag/stories/Examples/Variations.js
tristanjasper: Personally I find it quite readable when it is shown at the point of mapping in the file, rather than having to hunt around for a variable. Perhaps if there were many more options I would agree it belongs in a set variable but i think is ok here

packages/Tag/stories/Examples/Variations.js
```diff
@@ -100,6 +114,7 @@ export default function Variations() {
           <Tags key={index}>
             {Object.entries(Tag.types.theme).map(theme => (
               <Tag
+                as="li"
```
mikrotron, line 110: Cool, this is a nice shortcut!

packages/Tag/src/Tag.js
```diff
@@ -17,7 +17,13 @@ export default function Tag(props) {
       {avatar}
       <sc.Ellipsis size={size}>{children}</sc.Ellipsis>
       {onRemove ? (
-        <sc.Delete isDisabled={isDisabled} size={size} data-pka-anchor="tag.remove" onClick={handleRemove} />
+        <sc.Delete
+          isDisabled={isDisabled}
+          size={size}
+          data-pka-anchor="tag.remove"
+          data-qa-anchor="tag.remove"
```
mikrotron, line 24: Was this the naming convention we settled on? I thought there was a `paprika` namespace or something?

tristanjasper: Oh yep, i'll add `paprika.` as a prefix

packages/Tag/src/Tag.js
tristanjasper: Got it to work! I did try with border-box before but kept running into issues 🤷‍♂️

mikrotron (approved):
llgtm 75

Issue #495, opened 2020-04-24
SidePanel API consistency


mikrotron:
## Problem
There is an inconsistency between the `<SidePanel>` and other "modal" components, namely the `<Modal>` and `<Takeover>`.   

## Proposed Solution
The SidePanel should have a `<SidePanel.Content>` subcomponent, like the Takeover and Modal do. Additionally, the SidePanel shouldn't impose or modify the padding in their container `<div>`,  that is the responsibility for the Content subcomponent.

Also, the SidePanel should use the `@paprika/overlay` component in the same way the Modal does, instead of its own `<SidePanel.Overlay>` which is composed of a `<RawButton>`.  However, the Overlay is optional for the SidePanel, unlike the Modal, which makes the logic more complicated.

PR #931, opened 2021-02-08
UXD-820 Using @paprika/overlay in @paprika/panel

allison-c:
### Purpose 🚀

Using @paprika/overlay in @paprika/panel, and remove unnecessary dependencies

Resolves: https://github.com/acl-services/paprika/issues/495

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/panel`
### Storybook 📕

### Screenshots 📸

### References 🔗


packages/Panel/src/components/Overlay/Overlay.js
```diff
@@ -1,45 +1,31 @@
...
 import React from "react";
 import PropTypes from "prop-types";
-import useI18n from "@paprika/l10n/lib/useI18n";
-import * as sc from "./Overlay.styles";
+import { FocusPropTypes } from "@paprika/helpers";
 
 const propTypes = {
-  /** Will call an onClose handler when clicked on (outside of the Panel) */
-  hasOutsideClick: PropTypes.bool,
+  backdropClassName: PropTypes.string,
```
nahumzs, line 8: I remember having the discussion about this prop, I think haven't age well,  just a random comment. no action required.

packages/Panel/README.md
```diff
@@ -31,7 +31,7 @@ npm install @paprika/panel
...
 | isCompact             | bool                                                                                        | false    | false                         | Control the compactness of the Panel                                                                                                            |
 | isInline              | bool                                                                                        | false    | false                         | Render the panel inline                                                                                                                         |
-| isOpen                | bool                                                                                        | true     | -                             | Control the visibility of the Panel. This prop makes the Panel appear.                                                                          |
```
nahumzs, line 34: could this be a breaking change? if it used to be true by default?

allison-c: It was `required` `true`, not default value `true`. I changed it to `optional` with the default value `false`. Because if you don't pass in the value, `undefined` will act the same as `false` anyway, but more problematic 
What do you think?

nahumzs (approved):

PR #934, opened 2021-02-09
Uxd 835 fix takeover types

allison-c:
### Purpose 🚀

Trying to fix takeover types and docs generation. The problem is lots of components are reusing the proptypes definition from another component/package, react docgen doesn't have the ability to read from the original source. This is another good reason to move away from reusing propTypes. 

Another problem is the default type of each prop was `-` before, not sure why we set that, but definitely not a valid value in ts. So I changed it to `any`, to avoid parsing error.

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/takeover`
### Storybook 📕

### Screenshots 📸

### References 🔗


nahumzs:
`reason to move away from reusing propTypes` <---- 110% agree with this, repeat yourself sometimes is better than DRY

packages/Takeover/README.md
```diff
@@ -32,6 +32,23 @@ npm install @paprika/takeover
 | onClose      | func   | false    | () => {}  | Callback triggered when the takeover needs to be close |
 | zIndex       | number | false    | zValue(5) | The z-index of the Takeover content                    |
 
+### Takeover.Header
```
nahumzs, line 35: 👏

packages/Takeover/src/Takeover.js
```diff
@@ -39,32 +40,33 @@ const defaultProps = {
...
     ...(focusLockProps || {}),
   };
 
-  const takeoverOverlay = overlayExtracted || <TakeoverOverlay />;
   const overlayProps = {
     "data-pka-anchor": "takeover.overlay",
     hasBackdrop: false,
     focusLockOptions,
-    ...takeoverOverlay.props,
+    ...(overlayExtracted ? overlayExtracted.props : {}),
```
nahumzs, line 69: why is this need it? `...undefined` would no spread anything still isn't it?

packages/Takeover/src/Takeover.js
```diff
@@ -84,25 +86,29 @@ export default function Takeover(props) {
...
-      zIndex={zIndex}
-    >
-      {headerExtracted && <sc.Header {...headerExtracted.props} onClose={onClose} />}
-      {contentExtracted && (
-        <sc.Content role="region" tabIndex="0">
-          {contentExtracted}
-        </sc.Content>
+  return (
+    <OriginalOverlay {...overlayProps}>
+      {state => (
```
nahumzs, line 91: why is this turn into a render function? I'm not understanding the context.

allison-c: it was a render function before as well, https://github.com/acl-services/paprika/pull/934/files/97c7748ee1ea3f36898a5fddc36ab2519afbff15#diff-7f8e871aef4f5eab982513c0ec898d0387c02987deb2af97b5d961413ab0fea4L87 look at the children parameter was a function. This is a requirement from `@paprika/overlay` component

nahumzs (approved):

PR #938, opened 2021-02-10
UXD-832 avoid errors in polyfill when testing

allison-c:
### Purpose 🚀

In asset inventory, jest specs are complaining about `cannot read getElementsByTagName from null`, but we already have jsdom there. Maybe this issue was not caused by this polyfill, but it's very hard to debug, so we decided to avoid it in paprika  

### Notes ✏️

### Updates 📦
- [x] PATCH (bug fix) change to `@paprika/helpers`

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/helpers/src/polyfills/elementScroll.js
```diff
@@ -4,6 +4,8 @@
 // https://github.com/idmadj/element-scroll-polyfill/blob/master/index.js
 
 (function() {
+  if ((Element.prototype.scroll && Element.prototype.scroll) || !document) return;
```
alexzherdev, line 7: huh why is `Element.prototype.scroll` repeated two times? And why is it necessary here if there's a check on line  37?

allison-c: oh one of them should be `Element.prototype.scrollTo`, my copy paste error, and I guess there might be a case there is a `scroll` function but not `scrollTo`?

nahumzs (approved):

PR #939, opened 2021-02-10
feature(InlineEditing): UXD-781

nahumzs:
### Purpose 🚀
---
"@paprika/textarea": major
"@paprika/table": minor
"@paprika/inline-editing": patch
"@paprika/list-box": patch
"@paprika/seducer": patch
---

- 📦 `new` Introducing: inline-editing in alpha version ready to start integrating it but still needs work
- 💥 Breaking Change for Textarea removing an awkward way for declaring a ref in the textarea
- 🙋‍♀️ Table component now support
- 🔨 Upgraded collapsibleCard on the documentation

### Notes ✏️

**PLEASE READ** Documentation before reviewing the PR 

⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️

http://storybooks.highbond-s3.com/paprika/UXD-781-Inline-component/?path=/story/forms-inlineediting-docs--page

⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️⚡️

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-781-Inline-component

## TABLE STICKY column and header

The reason why this change is part of the PR is that we need to test if the `<Table />` could have sticky columns and headers, in order to know where the DataGrid fits and the Table in the context of inline editing and our future designs on the ecosystem.

The Design Group is discussing already this, and since adding this new feature was easier than I thought I sneaked it in this PR.   

http://storybooks.highbond-s3.com/paprika/UXD-781-Inline-component/?path=/story/table-table--sticky

### References 🔗


.changeset/dry-rivers-wink.md
```diff
@@ -0,0 +1,14 @@
...
+"@paprika/list-box": patch
+"@paprika/seducer": patch
+---
+
+- 📦 `new` Introducing: inline-editing in alpha version ready to start integrating it but still needs work
+- 💥 Breaking Change for Textarea removing an awkward way for declaring a ref in the textarea
+- ⚠️ Remove keyboard navigation from the table for now, meanwhile we decided which is the best rout for this, we can bring it back later.
+  this is not a breaking change because no body is using it!
+- 🙋‍♀️ Table component now support
+- 🔨 Upgraded collapsibleCard on the documentation
```
jamiek-acl, line 14: You could have up to 5 different files (so each could have its own release notes) if you wanted

packages/InlineEditing/src/Table.styles.js
```diff
@@ -0,0 +1,8 @@
+import TableProp from "@paprika/table";
+import styled from "styled-components";
+
+export const Table = styled(TableProp)`
+  td {
+    height: 40px;
```
jamiek-acl, line 6: token

packages/InlineEditing/src/Textarea.js
```diff
@@ -0,0 +1,131 @@
...
+        />
+      </Editor.Edit>
+      <Editor.Value>
+        {typeof collapsedLength !== "undefined" ? (
+          <CollapsibleText collapsedLength={collapsedLength} a11yText="cardigans">
+            <p>{value}</p>
+          </CollapsibleText>
+        ) : (
+          <p style={{ whiteSpace: "pre-wrap", padding: "4px" }}>{value}</p>
+        )}
```
jamiek-acl, line 111: I'd reverse this double-negative logic

packages/InlineEditing/stories/Editing.stories.js
```diff
@@ -0,0 +1,21 @@
...
+  return <Table />;
+};
+
+export const ListBoxSingleWithInlineEditing = () => {
+  return <ListBoxSingleStory />;
+};
+
+export const WithInput = () => {
+  return <Input />;
+};
```
jamiek-acl, line 21:
```suggestion
export const WithInATable = () => <Table />;

export const ListBoxSingleWithInlineEditing = () => <ListBoxSingleStory />;

export const WithInput = () => <Input />;
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+<Meta title={`${getStoryName("InlineEditing")}/Docs`} parameters={params} />
+
+<Toast kind={Toast.types.kind.WARNING} hasCloseButton={false}>
+  Work in progress ahead, this component still requires some more work to be ready for production, but the API is
+  stable, so you can start integrating it into your application.
+</Toast>
+
+# Inline Editing
+
+The `<InlineEditing />` component allows you to display two values for the user one when the componet is not in editing mode and an input whenever the user interact and set
```
jamiek-acl, line 28:
```suggestion
The `<InlineEditing />` component allows you to display two values to the user: one when the component is not in editing mode and an input whenever the user interact and set
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+the component into editing mode.
+
+## How does it works?
+
+The InlineEditing package includes a component name `<Editor />` that you can use to encapsulated this kind of interaction. The `<Editor />` Component comes with two
```
jamiek-acl, line 33:
```suggestion
The InlineEditing package includes a component name `<Editor />` that you can use to encapsulate this kind of interaction. The `<Editor />` Component comes with two
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+subcomponent `<Editor.Value />` and `<Editor.Edit />`. This two component will switch between each other whenever the prop `isEditing` on the `Editor` component chages.
```
jamiek-acl, line 34:
```suggestion
subcomponents: `<Editor.Value />` and `<Editor.Edit />`. These two components will switch between each other whenever the prop `isEditing` on the `Editor` component changes.
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+
+The following is an example of usign the Editor component to build a simple Text (value) `->` Text (input) interaction:
```
jamiek-acl, line 36:
```suggestion
The following is an example of using the Editor component to build a simple Text (value) `->` Text (input) interaction:
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+      onSubmit={handleSubmit} 
+    />
+  )
+}
+```
+
+  </Block>
+</Blocks>
+
+The previous example is a simple attempt to create a InlineEditng Input, to have it fully functionable will still required more work with the keyboard interactions,
```
jamiek-acl, line 107:
```suggestion
The previous example is a simple attempt to create an InlineEditng Input. To have it fully functional will still require more work with the keyboard interactions,
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+add focus onClick, cancel it onKeyUp === "Escape", etc.
+
+An because wiring a fully functionally InlineEditing component can be tedious and tricky to make it work properly, we currently are providing three components ready to use:
```
jamiek-acl, line 110:
```suggestion
And because wiring a fully functionally InlineEditing component can be tedious and tricky to make it work properly, we are currently providing three components ready to use:
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+  </Block>
+  <Block>
+
+```jsx
+import ListBox from "@paprika/inline-editing/lib/ListBox";
+
+const [isEditing, setIsEditing] = React.useState(false);
+
+/** 
+  Be aware that ListBox isMulti and the single select option have a different signature for the onChange and onEditing event
```
jamiek-acl, line 184: They dont look different, lol

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+  ))}
+</ListBox>;
+```
+
+  </Block>
+</Blocks>
+
+## onSubmit vs onChange
+
+The inline-editing components provide two different callbacks onChange and onSubmit
```
jamiek-acl, line 209:
```suggestion
The inline-editing component provides two different callbacks:
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+
+- `onChange` Will fire each time the component get an interaction from the user. (typing, selecting, etc)
```
jamiek-acl, line 211:
```suggestion
- `onChange` Will fire each time the component gets an interaction from the user. (typing, selecting, etc)
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+- `onSubmit` Will occurs only after the user either click `enter` (for certain component) or click outside of the editing state (onBlur)
```
jamiek-acl, line 212:
```suggestion
- `onSubmit` Will fire only after the user either clicks `enter` (for certain components) or clicks outside of the editing state (onBlur)
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+
+**NOTE:** if the user clicks `Esc` the onSubmit event will not fire. We aren't providing a `onEscape` callback neither, therefore would be hard to rollback an interaction
```
jamiek-acl, line 214:
```suggestion
**NOTE:** if the user clicks `Esc` the onSubmit event will not fire. We aren't providing an `onEscape` callback either, therefore it would be hard to rollback an interaction
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+for an user if your are using the onChange event to update your component, for that purpose the `onSubmit` event is more stable
```
jamiek-acl, line 215:
```suggestion
for a user if your are using the onChange event to update your component, for that purpose the `onSubmit` event is more stable.
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+
+## Within a Table
+
+The InlineEditing packages offers as well a Table component which compose the [Table](https://github.com/acl-services/paprika/tree/master/packages/Table) component to provide
```
jamiek-acl, line 219:
```suggestion
The InlineEditing package also offers a Table component which compose the [Table](https://github.com/acl-services/paprika/tree/master/packages/Table) component to provide
```

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+an inline editing experience in a table to use it is quite straighfoward:
```
jamiek-acl, line 220:
```suggestion
an inline editing experience in a table. To use it is quite straightforward:
```

packages/InlineEditing/stories/examples/InlineInput.simple.js
```diff
@@ -0,0 +1,58 @@
...
+      </Editor.Edit>
+      <Editor.Value>
+        <span style={{ textDecoration: "underline" }}>{value}</span>
+      </Editor.Value>
+    </Editor>
+  );
+}
+
+export function AppInlineInput() {
+  const [value, setValue] = React.useState("Click on it to edit it");
```
jamiek-acl, line 41:
```suggestion
  const [value, setValue] = React.useState("Click here to edit");
```

packages/InlineEditing/tests/InlineEditing.spec.js
```diff
@@ -0,0 +1,64 @@
+import React from "react";
+import { render, screen, waitForElementToBeRemoved } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { ListBoxSingleStory } from "../stories/examples/ListBox";
+
+test("Inline editing / ListBox / open", async () => {
```
jamiek-acl, line 6: Another way to do this is:
```
describe('Inline editing', () => {
  // tests, or more 'describes' go here

  describe('ListBox', () => {
    // tests, or more 'describes' go here

    test('open', () => {
      render(<ListBoxSingleStory />);
      ...
    });
  });
});

File: packages/InlineEditing/tests/InlineEditing.spec.js [hunk 21]
```diff
@@ -0,0 +1,64 @@
...
+  render(<ListBoxSingleStory />);
+
+  const triggerButton = screen.getByText(/oversight/i);
+  expect(triggerButton).toBeInTheDocument();
```
jamiek-acl, line 10: Kinda redundant since you click it below

packages/InlineEditing/tests/InlineEditing.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+  const regexOption1 = /Professional/i;
+  const regexOption2 = /Results Lite Contributor/i;
```
jamiek-acl, line 13: You use these in all the tests - could move it up instead of redeclaring

packages/Table/src/Table.js
```diff
@@ -124,23 +119,35 @@ const propTypes = {
...
   /** Array of data to be stored in the Table */
   data: PropTypes.arrayOf(PropTypes.shape({})),
   /** For authors use only, use case: inline editing. */
   enableArrowKeyNavigation: PropTypes.bool,
+  /** It will be call each time a new cell received the focus */
+  onFocus: PropTypes.func,
+  /** It will be call each time a current selected cell lose focus */
+  onBlur: PropTypes.func,
+  /** It will be fire each time an user click on a cell */
+  onClick: PropTypes.func,
```
jamiek-acl, line 135:
```suggestion
  /** Will fire each time a new cell receives focus */
  onFocus: PropTypes.func,
  /** Will fire each time a selected cell loses focus */
  onBlur: PropTypes.func,
  /** Will fire each time user clicks on a cell */
  onClick: PropTypes.func,
```

packages/Table/src/Table.styles.js
```diff
@@ -60,20 +60,29 @@ const borderTypesStyles = {
   `,
 };
 
-export const TD = styled.td(({ borderType }) => {
+export const TD = styled.td(({ borderType, cellPropsResetCSS, width = null, sticky }) => {
+  const px = Number.isNaN(width) ? "" : "px";
   return css`
-    ${borderType in borderTypesStyles ? borderTypesStyles[borderType] : ""}
-    padding: ${tokens.space};
+    ${typeof sticky !== "undefined" ? `position: sticky; left: ${!Number.isNaN(sticky) ? sticky : 0}px;` : ""};
```
jamiek-acl, line 66: TIL: `sticky`

packages/Table/stories/Table.mdx
```diff
@@ -184,3 +185,39 @@ The table can further be customized by adjusting the borders. The table can be s
   <Table.ColumnDefinition header="Name" cell="name" />
   <Table.ColumnDefinition header="LastName" cell="lastName" />
 </Table>
+
+## Sticky
+
+The Table support sticky header and columns.
```
jamiek-acl, line 191:
```suggestion
The Table supports sticky header and columns.
```

packages/Table/stories/Table.mdx
```diff
@@ -184,3 +185,39 @@ The table can further be customized by adjusting the borders. The table can be s
...
+  <Table.ColumnDefinition header="..." cell={() => {}} />
+</Table>
+```
+
+<Toast hasCloseButton={false}>The sticky prop accepts a number value, by default is zero.</Toast>
+
+Unlike the DataGrid where you can use a `isSticky` props in the columnDefinition to mark a column as a sticky, for the table, you have to assign to `sticky` the number
+in pixels of the space between the left side and your desired column.
+
+If you want to have only one sticky column only, you can mark it just with the `sticky` prop and you are ready to go.
```
jamiek-acl, line 206:
```suggestion
If you want to have only one sticky column, you can mark it just with the `sticky` prop and you are ready to go.
```

packages/Table/stories/Table.mdx
```diff
@@ -184,3 +185,39 @@ The table can further be customized by adjusting the borders. The table can be s
...
+
+But once the table needs two or more sticky columns, you need to assigned a numeric value representing the left space between their position and the table's left side.
```
jamiek-acl, line 208:
```suggestion
But once the table needs two or more sticky columns, you need to assign a numeric value representing the left space between their position and the table's left side.
```

packages/Table/stories/Table.stories.js
```diff
@@ -110,22 +110,7 @@ export function WithZebras() {
   );
 }
 
-export function WithArrowNavigation() {
```
jamiek-acl, line 113: Does this no longer work?

packages/InlineEditing/stories/InlineEditing.stories.mdx
jamiek-acl: Ok so then the line two under this comment is incorrect and should be:
```
isMulti={false} (index, options, selectedIndex, {options})
```
?

packages/Textarea/src/Textarea.js
```diff
@@ -27,6 +27,12 @@ const Textarea = React.forwardRef((props, ref) => {
     ...moreProps
   } = props;
 
+  React.useImperativeHandle(ref, () => ({
+    focus: () => {
+      refTextarea.current.focus();
+    },
```
mikrotron, line 33: 👍 I think this is a good change.  

Our other core form components (`<Input>` + `<Select>`) don't have this, they still pass the `ref` to the underlying DOM element, but I think we should probably get them all in sync in a later PR.

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -87,10 +87,19 @@ const Filter = React.forwardRef((props, ref) => {
   };
 
   React.useEffect(() => {
+    let id = null;
+
     if (!isOpen) {
-      reset();
```
mikrotron, line 91: What happened to `reset()`?  Can it be removed?

packages/InlineEditing/stories/examples/InlineInput.simple.js
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+import Input from "@paprika/input";
+import Editor from "../../src/Editor";
+import Card from "../../../Card/src";
```
mikrotron, line 4:
```suggestion
import Card from "@paprika/card";
```

packages/InlineEditing/src/Input.js
```diff
@@ -0,0 +1,122 @@
...
+      </Editor.Value>
+    </Editor>
+  );
+}
+
+const propTypes = {
+  isEditing: PropTypes.bool,
+  onChange: PropTypes.func,
+  onClose: PropTypes.func,
+  onEditing: PropTypes.func,
```
mikrotron, line 107: The tense of this callback is wrong.  And it's not really that clear what it's purpose is.  Is it like an `onToggleEdit`? `onStartEditing`? Or just `onClickEditor` / `onClickInput`?

packages/InlineEditing/src/index.js
```diff
@@ -0,0 +1,2 @@
+export { default } from "./Editor";
+export { default as Table } from "./Table";
```
mikrotron, line 2: You're not exposing the `<Input>`, `<ListBox>` or `<Textarea>` for consumers? 

Will there eventually be a need to have inline editable `<Select>` or `<DataPicker>` types?

packages/InlineEditing/src/Table.js
```diff
@@ -0,0 +1,63 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Table from "@paprika/table";
+import * as sc from "./Table.styles";
+
+function Editable({ children }) {
```
mikrotron, line 6: If you were to export this "uncontrolled" `<Editor>` component, then consumers could use it directly with a Paprika `<Table>`, right?  

The only other thing I think I see happening is some control over the `width` and `height`?

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+<Meta title={`${getStoryName("InlineEditing")}/Docs`} parameters={params} />
+
+<Toast kind={Toast.types.kind.WARNING} hasCloseButton={false}>
+  Work in progress ahead, this component still requires some more work to be ready for production, but the API is
+  stable, so you can start integrating it into your application.
+</Toast>
+
+# Inline Editing
+
+The `<InlineEditing />` component allows you to display two values to the user: one when the component is not in editing mode and an input whenever the user interact and set
```
mikrotron, line 28: From what I can see, there is NOT an `<InlineEditing>` component, it's just an `InlineEditing` package with an `<Editor>` and a `<Table>` component...? (and maybe the other inline editable inputs _should_ also be exported?)

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+
+  </Block>
+</Blocks>
+
+## onSubmit vs onChange
+
+The inline-editing component provides two different callbacks:
+
+- `onChange` Will fire each time the component gets an interaction from the user. (typing, selecting, etc)
+- `onSubmit` Will fire only after the user either clicks `enter` (for certain components) or clicks outside of the editing state (onBlur)
```
mikrotron, line 212: I'm not sure about the name of this callback – `onSubmit` already has a specific meaning for forms, and this is not that.  Maybe `onFinishEditing`?

Also, does this mean it will fire twice if the user presses `enter` and then `tab`?

packages/ListBox/src/components/Filter/Filter.js
nahumzs: I don't fully remember why I removed it, but I refactored it to avoid one of those errors in a test about wrapping the component in an `act()` and during that, I realized that the reset was not longer necessary with all the refactor that happened on the list box and the keyboard.

packages/InlineEditing/src/Table.js
nahumzs: to have it work smoothly in the Table and don't change the API at all. 

inline-editing components have to be aware of which cell are they working on, so they can call `onSubmit({ rowIndex, columnIndex})` with the proper values and manage focus. To make this work there is a HOC that wraps the cell prop and turns the anonymous function into a component that is extended with the proper values so the end-user can have an inline-editing on a Table component without refactoring anything those props later are passdown to the inline-editing (listbox/textarea/input, etc). 

https://github.com/acl-services/paprika/pull/939/files#diff-5fb89b8beb3edab9a41d813fef3b5a5e8ba0b3344ab601ea8e90f35fa5174bdaR33-R46

**^** this is the trick `const { cell: Component, width } = child.props;`

look at this how the Listbox is aware when the onSubmit event happens and knowns which row and column executed the action.

https://github.com/acl-services/paprika/pull/939/files#diff-1a10891d491f56273522fe6251eccf716bb629606bbcdf5f3f653d7213ad62b6R22-R30

There is a lot of focus control happening as well, you can indeed just put the component and have inline editing on the table but that only will work with a click. 

Similar to the example that explains explains how it works. http://storybooks.highbond-s3.com/paprika/UXD-781-Inline-component/?path=/story/forms-inlineediting-docs--page#how-does-it-works

packages/InlineEditing/src/index.js
mikrotron: I don't like importing from `lib`.  It seems like a back door – I'd rather see a clear offering of the API in the `index` entry point.

packages/InlineEditing/stories/InlineEditing.stories.mdx
nahumzs: depending on tab, if tab is exiting the input will fire submit, if its inside of it will don't do anything, submit occurs either by press the key Enter or onblur, Tab doesn't have any play on the trigger.

packages/InlineEditing/src/index.js
nahumzs: you don't want to load all the widget if you want to use just one, this is a component that has many "plugable" components, there is not a reason to load all of them from the index. Think about this like the icons, you can't load all the icons just to use one.

packages/InlineEditing/stories/InlineEditing.stories.mdx
nahumzs: Yeap indeed is a package with multiple components:

- Editor
- Table
- ListBox
- Input
- Textarea

I should reword that. 

> (and maybe the other inline editable inputs should also be exported?) 

The are exported as individual packages via `@paprika/inlinline-editing/lib/ListBox`

those are independents components, is interesting maybe there is not even a reason for an index in this set of components.

packages/InlineEditing/stories/InlineEditing.stories.mdx
```diff
@@ -0,0 +1,278 @@
...
+The previous example is a simple attempt to create an InlineEditng Input. To have it fully functional will still require more work with the keyboard interactions,
+add focus onClick, cancel it onKeyUp === "Escape", etc.
+
+And because wiring a fully functionally InlineEditing component can be tedious and tricky to make it work properly, we are currently providing three components ready to use:
+
+- Input type text
+- TextArea
+- ListBox
+
+## Available InlineEditing Components
```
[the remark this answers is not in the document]
nahumzs: > Seems like you definitely do intend to provide these three components to the consumer

Yeap i'm providing those to the consumer but not from an index.js file instead of by `lib/component` option.

Is easier to have all these compose component in the same package, imagine changing one prop or requirement and have to go to `n` packages to update the pacakge.json plus changelog per change. It would just add unnecessary complexity. 

plus

`import ListBox from "@paprika/inline-editing/lib/ListBox"`

make it really clear what are you doing and what are you trying to solve.

.storybook/components/Blocks/Blocks.js
```diff
@@ -4,8 +4,11 @@ import styled, { css } from "styled-components";
 const BlockStyled = styled.div(({ basis = "100%", textAlign = "left" }) => {
   return css`
     flex-basis: ${basis};
-    margin: 2px;
     text-align: ${textAlign};
+    .docblock-source {
+      margin: 0;
+    }
+    padding: 8px;
```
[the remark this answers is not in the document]
nahumzs: this is on storybook component is ok, is purely internal

packages/InlineEditing/src/ListBox.js
```diff
@@ -0,0 +1,144 @@
...
+          <ListBox.Popover onClose={handleClose} />
+          {children}
+        </ListBox>
+      </Editor.Edit>
+      <Editor.Value>
+        <div style={{ width: "100%", display: "inline-flex", justifyContent: "space-between", alignItems: "center" }}>
+          <div style={{ flexBasis: "100%", padding: "4px" }}>
+            <div>{typeof renderValue === "function" ? renderValue(value) : value || placeHolder} </div>
+          </div>
+          <div>▾</div>
```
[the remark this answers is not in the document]
nahumzs: it has to change as soon we get the official design

packages/InlineEditing/tests/InlineEditing.spec.js
nahumzs: My opinion is that all tests should be atomic and should run independently, that's why I put all that is required for the test inside of it.

packages/InlineEditing/src/Input/Input.js
```diff
@@ -0,0 +1,127 @@
...
+    columnIndex,
+    /* eslint-enable react/prop-types */
+    ...moreProps
+  } = props;
+  const refInput = React.useRef(null);
+  const refInputEditor = React.useRef(null);
+  const [nextValue, setNextValue] = React.useState(value);
+
+  function handleBlur() {
+    onClose();
```
jamiek-acl, line 27: Lets say the input says "old".
I click on in, and type in "new".
Then I click away from the input and its content changes to "old".
Then I click on the input and its content changes to "new".

Is that expected?

packages/InlineEditing/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@paprika/inline-editors",
```
jamiek-acl, line 2: name changed

packages/InlineEditing/README.md
```diff
@@ -0,0 +1,45 @@
+<!-- start: Autogenerated - do not modify -->
+
+# @paprika/inline-editors - 0.1.0
```
jamiek-acl, line 3: This needs to be regenerated as the name has changed

packages/InlineEditing/src/Input/Input.js
```diff
@@ -0,0 +1,127 @@
...
+      }}
+      {...props}
+      ref={refInputEditor}
+    >
+      <Editor.Edit>
+        <sc.Container>
+          <Input
+            {...moreProps}
+            ref={refInput}
+            style={{ paddingRight: "24px" }}
```
jamiek-acl, line 92: Boo inline, and no token

packages/InlineEditing/src/Input/Input.js
```diff
@@ -0,0 +1,127 @@
...
+            onBlur={handleBlur}
+            onKeyUp={handleKeyUp}
+            onChange={handleChange}
+            type="text"
```
jamiek-acl, line 96: nit: this is the default value so prop not needed

packages/InlineEditing/package.json
nahumzs: yes this is correct!

packages/InlineEditing/src/Input/Input.js
nahumzs: this will change this styles are not even define :/

packages/InlineEditing/src/Input/Input.js
nahumzs: fix though, the finally decision for interaction is not yet settle

jamiek-acl (dismissed):

allison-c (approved):

nahumzs:
thank @allison-c 🙏

PR #943, opened 2021-02-11
feature(ListBox): UXD-840 fixing keyboard issues and 'flickering'

nahumzs:
### Purpose 🚀

https://aclgrc.atlassian.net/browse/UXD-840 ListBox flickering when toggling the Trigger area and making a feature testing flaky. 

The Testing flakyness might be occurring by firing a click event on the Trigger component to open and close it during the test. 

This previous can lead to a race condition, where the blur event happening on the Popover triggers an onClose event and then the onClick open the Popover again. 

This is easy to replicate by a user, simply open the ListBox, and then click again on the Trigger to close it without letting go the MouseDown. You will see that the blur event closes the popover but as soon as you release the mouse down to finish the cycle for th onClick event it will open the popover again. 

An easy fix would be to use an `onMouseDown` to close or open the Popover with the Trigger, but this might fail many tests in our applications, and just feel not a standard way to do it. 

`NOTE` 
The previous behaviour is barely noticeable for our users, this is more a testing issue to **solve** it, You could target another element even the body instead of using the Trigger to toggle the ListBox, targeting anything else will fire the onBlur event...

### Notes ✏️
While working on this ticket I fixed keyboard issues. 

### Updates 📦

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-840-bug-trigger-flickering/?path=/story/forms-listbox-examples-multi--fully-controlled-with-filter

### Screenshots 📸

### References 🔗


packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -92,11 +92,23 @@ export default function Trigger(props) {
   const [triggerButtonId] = React.useState(() => `list-box-trigger_${uuidv4()}`);
 
   const handleClick = () => {
-    if (isDisabled) {
+    if (state.isOpen) {
+      dispatch({ type: useListBox.types.closePopover });
```
mikrotron, line 96: I don't really see much affect in the UI with this approach, though the code _is_ arguably clearer. 

This issue isn't really one with the ListBox though, it's present in the Popover too, and noted in the code comments:

https://github.com/acl-services/paprika/blob/master/packages/Popover/src/components/Content/Content.js#L49-L59

Apart from writing tests differently, the interaction seems improved in my experiments by increasing that timeout (I found 350ms was an improvement):

https://github.com/acl-services/paprika/blob/master/packages/Popover/src/components/Content/Content.js#L67

nahumzs:
close in favour of https://github.com/acl-services/paprika/pull/947

Issue #820, opened 2020-11-25
ListBoxWithTags - a11y defects


mikrotron:
Since I haven't actually evaluated the existing `<ListBox>` component, I think a number of these actually apply more generally to it, rather than the new `WithTags` version.  

Also, since I just tested the story in Storybook, it's possible some might relate the the example implementation instead of the actual component.  But since those stories are inspiration for consumers, they should probably be fixed anyway. 

1. Most of our buttons can be activated by `space` but the trigger for the `<ListBoxWithTags>` does not open the list.
2. The trigger button should have `aria-haspopup="listboxId_"` and `aria-expanded` attributes.
3. For the list of selected items (tags) in the trigger, there's a couple `<div>` elements between the `<ul>` and `<li>`.  This is [invalid markup])(https://validator.w3.org/nu/#textarea) – I suggest you use aria roles instead (`list` / `listitem`) which will allow `div` elements in between.
~4. The search box in the drop down seems to trap the keyboard focus (pressing `tab` does not move focus – even though `shift` + `tab` _does_ move it backwards to the trigger).~ 
5. When using the `↓` key to move the selection highlight in the list, nothing is announced by the screenreader (probably since actual document focus remains on the search box).

#### References
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html

nahumzs:
4. Just occurs in a situation where there is nothing to focus on apart of the WithTags component which is unlikely to ocurrs.

oscarkwan:
3. Can't reproduce

mikrotron:
Cool, `3` was inadvertently resolved in Tristan's PR to extract the `<Tag>`

PR #941, opened 2021-02-11
Uxd 837 a11y listbox

oscarkwan:
### Purpose 🚀
- ListBox A11y improvements

### Notes ✏️
- ListBox options are now focusable, and therefore will read out the option to AT
- Added aria-live for No Results Found
- Add descriptive text for clear button
- Add aria-haspopup
- Fixed all Listbox -> ListBox

### Updates 📦
- [ ] MAJOR (breaking) change to _these packages_ (if the package you are breaking is a dependency in other packages, be sure to add a commit with a "BREAKING CHANGE" message to those other packages as well)
- [x] MINOR (backward compatible) change to ListBox
- [ ] PATCH (bug fix) change to _these packages_

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-837-a11y-listbox

### Screenshots 📸
_optional but highly recommended_

### References 🔗
_relevant Jira ticket / GitHub issues_

PR #948, opened 2021-02-13
UXD-843 fix(ListBoxWithTags): fix accessibility issues

oscarkwan:
### Purpose 🚀
- Fix accessibility issues with ListBoxWithTags

### Notes ✏️
- add `aria-expanded` and `aria-haspopup` to Trigger
- add aria label text to close button
- make tag close button keyboard accessible

Resolved https://github.com/acl-services/paprika/issues/820

### Updates 📦
- [x] MINOR (backward compatible) change to **Tag** **ListBoxWithTags**
- [x] PATCH (bug fix) change to _these packages_ **OverflowMenu** **Confirmation**

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-843-listbox-tags-a11y


packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -201,16 +201,6 @@ export const handleClickOption = ({ props, state, dispatch, onChangeContext }) =
 };
 
 export function handleEnterOrSpace({ event, state, dispatch, onChangeContext }) {
-  const pressedSpaceKeyWhileHavingFilter = state.hasFilter && event.key === " " && event.target.value !== "";
```
oscarkwan, line 204: https://github.com/acl-services/paprika/pull/941 has the same change, will wait for that to get merged

packages/Tag/src/Tag.js
```diff
@@ -1,28 +1,38 @@
...
   const { children, onRemove, size, onClick, a11yText, avatar, isDisabled, ...moreProps } = props;
+  const I18n = useI18n();
 
   const handleRemove = e => {
     onRemove();
     e.stopPropagation();
   };
 
+  const handleKeyDown = e => {
+    if (e.key === "Enter" || e.key === "Backspace" || e.key === " ") {
```
oscarkwan, line 18: I Included backspace, but open to not including this.

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -74,6 +74,8 @@ const renderTrigger = ({ t, size, selectedOptions, onRemove, renderTag, tagLabel
 
   return (
     <sc.Trigger
+      aria-haspopup="true"
+      aria-expanded={isOpen}
```
mikrotron, line 78: Do you think we should also have `aria-controls` here?
https://www.w3.org/TR/wai-aria-1.1/#aria-controls

mikrotron: We do use it here:
https://github.com/acl-services/paprika/blob/master/packages/OverflowMenu/src/components/Trigger/Trigger.js#L20

oscarkwan: Yeah, lets add it

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -74,6 +74,8 @@ const renderTrigger = ({ t, size, selectedOptions, onRemove, renderTag, tagLabel
 
   return (
     <sc.Trigger
```
mikrotron, line 76: The default label for this trigger button gets unnecessarily long and redundant because it includes the desciption of all of the remove buttons too.  I thin we should mitigate that by providing an `aria-label` (`a11yText`) that can be composed more thoughtfully.

packages/Tag/src/Tag.js
mikrotron: This is the only place I'm aware of where the `backspace` key will activate a remove button... but I like it!

But does @sunminsarahlee approve? This means when the keyboard focus is on the little `x` for a selected tag, pressing any of these keys removes it.

packages/Tag/src/Tag.js
```diff
@@ -1,28 +1,38 @@
...
+    }
+  };
+
   const content = (
     <>
       {avatar}
       <sc.Ellipsis size={size}>{children}</sc.Ellipsis>
       {onRemove ? (
         <sc.Delete
+          a11yText={I18n.t("remove_a11y", { value: children })}
```
mikrotron, line 29: What if `children` is a `node`, not a `string`?
Maybe it should fall back to the generic "remove"?

packages/Tag/src/Tag.js
tristanjasper: I would think just "Enter" or "Space" would be the expected buttons to press to action this imo

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -74,6 +74,8 @@ const renderTrigger = ({ t, size, selectedOptions, onRemove, renderTag, tagLabel
...
+      aria-haspopup="true"
```
tristanjasper, line 77: just `aria-haspopup`. No need to assign to "true"

packages/Tag/src/Tag.js
```diff
@@ -1,28 +1,38 @@
...
 export default function Tag(props) {
   const { children, onRemove, size, onClick, a11yText, avatar, isDisabled, ...moreProps } = props;
+  const I18n = useI18n();
 
   const handleRemove = e => {
     onRemove();
     e.stopPropagation();
   };
 
+  const handleKeyDown = e => {
```
tristanjasper, line 17: perhaps name this function something that associates it with the Delete Button, maybe `handleDeleteKeyDown`?
Maybe Delete Component should be moved to its own file with export at this point

packages/ListBoxWithTags/src/ListBoxWithTags.js
oscarkwan: Actually no because `aria-haspopup` expects a string of "true" no the actual value of true

tristanjasper: Not sure what you mean, both ways ends up adding `aria-haspopup="true"` to the dom

oscarkwan: Hm, i wonder why we use `aria-haspopup="true"` in 3 other places then

tristanjasper: Perhaps update the 3 other places as well if it's not necessary? it's a small cleanup anyway

packages/Confirmation/src/components/TriggerButton/TriggerButton.js
```diff
@@ -20,7 +20,7 @@ const TriggerButton = props => {
       data-pka-anchor="button"
       ref={triggerRef}
       aria-expanded={isConfirmOpen}
-      aria-haspopup="true"
+      aria-haspopup
```
mikrotron, line 23: This change and the one to `OverflowMenu` are patch updates and while we _could_ include them in the `CHANGELOG`, I think they're so small and insignificant that we can omit that update.

packages/Tag/src/Tag.js
```diff
@@ -1,28 +1,40 @@
 import React from "react";
```
mikrotron, line 1: Your changes to the Tag component though are more significant and should be added to the `CHANGELOG`.

packages/ListBoxWithTags/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+- Add aria and accessibility to trigger component for screen reader
+- Added uuid package
```
mikrotron, line 10: This is an internal implementation detail that I don't think needs to be logged.  🤷‍♂️

mikrotron (approved):
llgtm 07

PR #954, opened 2021-02-23
UXD-867 ListBox compatibility with FormElement 🐐

mikrotron:
### Purpose 🚀
Add `hasError` and `isReadOnly` props to `<ListBox>` and handle the `a11yProps` provided by `<FormElement.Content>` via a new `<ListBox.A11y>` prop collector component. 

### Notes ✏️
- Refactored `ListBox` to move several "controlled" props (`hasError`, `id`, `isDisabled`, `isInline`, `isReadOnly`, `placeholder`, `size`) from the "store" in the main provider to a new `PropsProvider` which simply provides them to sub-components via "context". This allows them to be changed by the consumer after the component has mounted without code that listens for changes and dispatches actions to update the "store".   
- `moreProps` provided by the consumer on the root `<ListBox>` component will be propagated to the root element.
- Props collected by `<ListBox.A11y>` will be redirected to either the `<RawButton>` contained inside the `<ListBox.Trigger>` or, if `isInline=true`, to the `<Content>` (the focusable element with `role=listbox`).  This allows `a11yProps` from `<FormElement>` to end up on the proper DOM element.
- The `role=listbox` and associated `aria` attributes have been moved from the `<List>` (`<ul>` element) to the `<Content>` which is focusable and has the key event handlers.  This improves some aspects of the `a11y`, but actually it would be preferable to have everything on the `<ul>` element with the options as direct descendants.  I attempted this change, but it turned into a scary rabbit hole and decided not to let this PR get any bigger.
- `aria-labels` for the `<ListBox.Trigger>` or `role=listbox` element are now generated more robustly – if possible they will include the `<FormElement.Label>` (via `refLabel`) AND the `<ListBox.Trigger>` value (the selected items / placeholder).  
- Lots of inline comments have been added to highlight the significant changes as there are a number of cleanups that add noise to this fairly large PR.

### Updates 📦
- [x] MAJOR (breaking) change to `ListBox`

- [x] PATCH (bug fix) change to  `ListBoxWithTags`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-867--more-listbox-props/?path=/story/forms-listbox--showcase-story

### Screenshots 📸
Showcase (with more knobs)
Screen Shot 2021-02-23 at 9 55 34 PM

With FormElement story
Screen Shot 2021-02-23 at 9 56 00 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-867


packages/ListBox/src/ListBox.js
```diff
@@ -12,37 +12,44 @@ import Popover from "./components/Popover";
...
+  const hasOptions = Boolean(React.Children.count(children));
+
   const propsForTrigger = {
     hasClearButton: true,
     hasImplicitAll,
     onClickClear: null,
-    placeholder: placeholder || I18n.t("listBox.trigger.placeholder"),
-    onFooterClickAccept: footer ? footer.props.onClickAccept : null,
+    onClickFooterAccept: footer ? footer.props.onClickAccept : null,
+    morePropsForTrigger: moreProps,
```
mikrotron, line 52: Since the consumer can add more `moreProps` directly to `<ListBox.Trigger>` which should go on the container element, not the underlying `<RawButton>`, these `moreProps` from `<ListBox>` need to be separated.

packages/ListBox/src/ListBox.js
```diff
@@ -51,21 +58,30 @@ export function ListBox(props) {
     <Trigger {...propsForTrigger} />
   );
 
-  const hasOptions = Boolean(React.Children.count(children));
+  const contentProps = {
+    hasOptions,
+    onCancelFooter: footer ? footer.props.onClickCancel : null,
+    ...(isInline ? moreProps : {}),
```
mikrotron, line 64: For an inline `<ListBox>` the "trigger" is just a static label, so the `moreProps` should go to the interactive `role=listbox` element.

packages/ListBox/src/components/Box/Box.js
```diff
@@ -2,21 +2,26 @@ import React from "react";
...
       {props.children}
     </sc.Box>
   );
 }
 
-Box.propTypes = propTypes;
+Box.propTypes = {
+  /** Body content of the box. */
+  children: PropTypes.node.isRequired,
+};
```
mikrotron, line 27: I've moved `propTypes` and `defaultProps` to the bottom in several files to adhere to our new convention.

packages/ListBox/src/components/Content/Content.js
```diff
@@ -54,27 +41,31 @@ const handleContentFocusChange = (hasFocus, dispatch) =>
...
+  if (isInline) {
     return (
       <sc.Content
-        {...getDOMAttributesForListBoxContainer({ isInline: true })}
-        onFocus={() => {
-          handleContentFocusChange(true, dispatch);
-        }}
+        {...moreProps}
+        {...getDOMAttributesForListBox({ idListBox, ...state })}
+        data-pka-anchor="list-box-content-inline" // TODO: rename "list-box.content-inline"
```
mikrotron, line 56: I wanted to rename a lot of the anchors, but decided not to introduce potentially breaking changes in this PR.

packages/ListBox/src/components/Divider/Divider.styles.js
```diff
@@ -2,41 +2,35 @@ import styled, { css } from "styled-components";
...
+ */
 
-const textDividerStyles = css`
-  &:after {
-    ${pseudoElement}
-    margin-left: ${tokens.space};
-  }
-`;
+export const Divider = styled.li(
+  ({ isDisabled }) => css`
```
mikrotron, line 16: This will now dim when the (inline) component is disabled.

packages/ListBox/src/components/Label/Label.js
```diff
@@ -43,11 +29,33 @@ export default function Label(props) {
   }, [isMulti, options, placeholder, selectedOptions]);
 
   return (
-    <sc.Label hasImplicitAll={hasImplicitAll} isPlaceholder={!selectedOptions.length}>
+    <sc.Label
+      {...moreProps}
```
mikrotron, line 33: To propagate the `id` attribute used for `aria` labelling.

packages/ListBox/src/components/Option/Option.js
```diff
@@ -2,12 +2,55 @@ import React from "react";
...
+  const onChangeContext = React.useContext(OnChangeContext);
+
+  useIsSelectedOption({ index, props });
+
+  if (typeof state.options[index] === "undefined" || !isOptionVisible(state, index)) {
+    return null;
+  }
+
+  const isSelected = isOptionSelected(state, index);
+  const isDisabled = providedProps.isDisabled || props.isDisabled || isReadOnly;
```
mikrotron, line 26: Mostly the code for this component was just moved around, very little changed. This is one change – items are disabled if `isReadOnly`.  This only applies when `isInline` since the dropdown list does not appear for a disabled popover-style component.

packages/ListBox/src/components/Content/Content.js
```diff
@@ -54,27 +41,31 @@ const handleContentFocusChange = (hasFocus, dispatch) =>
   dispatch({ type: useListBox.types.setListBoxHasFocus, payload: { hasFocus } });
 
 export default function Content(props) {
-  const { children, hasOptions } = props;
+  const { children, hasOptions, onCancelFooter, ...moreProps } = props;
   const onChangeContext = React.useContext(OnChangeContext);
   const [state, dispatch] = useListBox();
   const { refListBoxContainer } = state;
+  const providedProps = React.useContext(PropsContext);
```
mikrotron, line 48: I'm not 100% convinced about this name, `providedProps`.  They are the "controlled" props supplied by the consumer that do not affect the internal state of the component (selection, search filter, open state, etc).  

They are "provided" to `ListBox` subcomponents via context for convenience.

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -1,100 +1,84 @@
...
     hasClearButton,
     hasImplicitAll,
-    label,
-    onClickFooterAccept,
-    children,
     isHidden,
+    label,
     onClickClear,
-    // eslint-disable-next-line react/prop-types
-    onFooterClickAccept,
```
mikrotron, line 79: This wasn't used, except [here](https://github.com/acl-services/paprika/pull/954/files#diff-ed9bd4d6b0e06d0607863760d0a10b00f3db83598b7d52e31fbacf3a801f7b20L45) and that seemed like a bug.  `onClickFooterAccept` is used instead.

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -1,100 +1,84 @@
...
+    if (!$label) return;
+
+    const possibleId = $label.getAttribute("id");
+    if (possibleId) {
+      idFormLabel.current = possibleId;
+    } else {
+      const newId = `${idListBox}__form-label`;
+      $label.setAttribute("id", newId);
+      idFormLabel.current = newId;
+    }
```
mikrotron, line 67: This is new.  The `<FormElement.Label>` element needs an `id` to be used later with `aria-labelledby`.

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -154,62 +122,98 @@ export default function Trigger(props) {
...
+        data-pka-anchor="list-box-trigger"
+        isDisabled={isDisabled}
         onClick={handleClick}
-        ref={refTrigger}
         onKeyUp={handleKeyUp}
-        isDisabled={isDisabled}
-        data-pka-anchor="list-box-trigger"
-        aria-haspopup
-        aria-describedby={formElementLabelDescribedBy}
-        aria-labelledby={triggerButtonId.current}
```
mikrotron, line 183: Moved to `getDOMAttributesForListBoxButton()`

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -154,62 +122,98 @@ export default function Trigger(props) {
...
+        ref={refTrigger}
       >
-        {refLabel && refLabel.current ? (
-          <sc.VisuallyHiddenFormLabel>{refLabel.current.innerText}</sc.VisuallyHiddenFormLabel>
-        ) : null}
```
mikrotron, line 187: This is now taken care of by `aria-labelledby` in `getDOMAttributesForListBoxButton()`

packages/ListBox/src/store/Provider.js
```diff
@@ -10,26 +9,19 @@ function initializeState(props) {
...
     refFilterInput,
     refFooterContainer,
     refListBox,
     refListBoxContainer,
     refTrigger,
     refTriggerContainer,
     refLabel,
-    size,
   } = props;
-  const formElementId = props.id;
```
mikrotron, line 27: It seems like this _never_ worked.

packages/ListBox/src/store/Provider.js
```diff
@@ -10,26 +9,19 @@ function initializeState(props) {
...
-  const formElementLabelDescribedBy = props["aria-describedby"];
```
mikrotron, line 28: With the new approach, `moreProps` get propagated to the appropriate element, but except for `refLabel`, the `<ListBox>` doesn't really know anything about `<FormElement>`. 

And even `refLabel` is generic enough that it's not specific to `<FormElement>` either, it's just a `ref` to a DOM element that holds the label for the `<ListBox>`.

packages/ListBox/src/hooks/hooks.js
```diff
@@ -149,17 +160,6 @@ export function useOnScrolled() {
...
 
-export function useIsDisabled(isDisabled) {
-  const [, dispatch] = useListBox();
-
-  React.useEffect(() => {
-    dispatch({
-      type: useListBox.types.setIsDisabled,
-      payload: isDisabled,
-    });
-  }, [dispatch, isDisabled]);
```
mikrotron, line 160: This seems like a complex way to keep a consumer-provided prop in sync with the component's internal state and what motivated me to find a new pattern for the new `isReadOnly` and `hasError` props.

packages/ListBox/stories/examples/Showcase.js
```diff
@@ -1,15 +1,19 @@
...
 
 const getKnobs = () => ({
+  size: select("size", Object.values(ListBox.types.size), ListBox.types.size.MEDIUM),
   height: number("height", 200),
+  hasError: boolean("hasError", false),
   isDisabled: boolean("isDisabled", false),
-  placeholder: text("placeholder", "Select..."),
+  isInline: boolean("isInline", false),
+  isReadOnly: boolean("isReadOnly", false),
+  placeholder: text("placeholder", ""),
```
mikrotron, line 16: With the new approach, these props can now be controlled with the Storybook knobs.  Additionally, if the consuming app changes their values on a re-render, after the component has mounted, the component will be updated.

packages/ListBox/src/store/Provider.js
```diff
@@ -10,26 +9,19 @@ function initializeState(props) {
   const {
     childrenOptions,
     height,
```
mikrotron, line 11: `height` is also a "controlled" prop that doesn't affect the internal state, but I was warned that `it might be more difficult to refactor since it's used in several calculations, so I didn't touch it.

mikrotron: I  was similarly worried about `isMulti` being a can-of-worms...

packages/ListBox/src/store/Provider.js
```diff
@@ -10,26 +9,19 @@ function initializeState(props) {
...
     isMulti,
     isOpen,
-    placeholder,
     refFilterInput,
     refFooterContainer,
     refListBox,
     refListBoxContainer,
     refTrigger,
     refTriggerContainer,
     refLabel,
```
mikrotron, line 20: I considered moving `refLabel` to `PropsProvider` too, but since it's passed by reference, not value, any subsequent update by the consumer is reflected in the "store" so I followed the "if it ain't broke, don't fix it" policy.

packages/ListBox/test/cypress/ListBox.cypress.js
```diff
@@ -81,7 +80,7 @@ describe("ListBox single select", () => {
     it("should filter by option label", () => {
       cy.get(selectors.filterInput)
         .focus()
-        .type("spiderman");
+        .type("spi", typingDelay);
```
mikrotron, line 83: This `typingDelay` slows down the test a bit, but it was getting very flakey otherwise – observing the result in Cypress I noticed the filter would contain `spman` or `deman` and things like that.  

There were a [number of threads](https://github.com/cypress-io/cypress/issues/3817) I found referencing this kind of issue, but most suggestions included using `cy.wait()` 😒

packages/ListBoxBrowser/test/cypress/basic.cypress.js
```diff
@@ -23,11 +23,11 @@ function includesAllOptions(browser, expectedOptions) {
 }
 
 function getRoot() {
-  return cy.get("[role='listbox']").eq(0);
+  return cy.get("[role='listbox'] ul").eq(0);
```
mikrotron, line 26: Since the `listbox` role moved up the DOM to the `<Content>` element.

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -78,7 +76,9 @@ const renderTrigger = ({
     onRemove(option);
   };
 
-  const Caret = isOpen ? <CaretUpIcon css={triggerSc.iconStyles} /> : <CaretDownIcon css={triggerSc.iconStyles} />;
+  function renderCaret() {
+    return isOpen ? <triggerSc.UpIcon /> : <triggerSc.DownIcon />;
```
mikrotron, line 80: I do not like the dependency on `@paprika/list-box/lib/components/Trigger/Trigger.styles` in `ListBoxWithTags` and would like to eliminate it in my follow-up PR that focusses on this component.  

Not sure if I should copy/paste the needed styling, or move it to a helper function in the `@paprika/stylers` package.  Opinions?

sunminsarahlee:
Read-only state doesn't pass colour contrast. Can we change the text colour to our black, #3F3D3C? 
Screen Shot 2021-02-23 at 11 37 33 PM

packages/ListBox/src/components/Content/Content.js
nahumzs: would be hard to rename the anchors :P, if someone is relying on them to do some css or calculation could be a hard bug to catch.

nahumzs:
http://storybooks.highbond-s3.com/paprika/UXD-867--more-listbox-props/?path=/story/forms-listbox-examples--subcomponents-story 

on the `ListBox.Box` example the border prop `1px solid red` is no longer rendering I remember seeing it working before was added by @oscarkwan

allison-c:
Since the `moreProps` has been moved to the trigger element, not the root popover wrapper anymore, I think this might be a breaking change. It's still a change of props. For example, before if I pass `className`, `className` will be applied to the root, now it'll be applied to the trigger only.

mikrotron:
> Since the `moreProps` has been moved to the trigger element, not the root popover wrapper anymore, I think this might be a breaking change. It's still a change of props. For example, before if I pass `className`, `className` will be applied to the root, now it'll be applied to the trigger only.

Youch, that's a good point, and I think we may still want to all a custom `className` on the root element 😖 

Not sure what the best solution would be to make it simple to forward `a11yProps` to the right place, without knowing the details of what those are, but still allowing `moreProps` on the root... 🤔

mikrotron:
> Not sure what the best solution would be to make it simple to forward `a11yProps` to the right place, without knowing the details of what those are, but still allowing `moreProps` on the root... 🤔

The best idea I have at the moment is to use an abstract props collector, like `<ListBox.A11y>` that you can spread `a11yProps` on, and those will get redirected to either the `<Trigger>` or `role=listbox` element.  Then we can preserve the propagation of `moreProps` on `<ListBox>` to the root element.  The consumer just needs to know about the `<ListBox.A11y>` subcomponent, and the ListBox doesn't need to know anything about FormElement.  What do you think?

allison-c:
> > Not sure what the best solution would be to make it simple to forward `a11yProps` to the right place, without knowing the details of what those are, but still allowing `moreProps` on the root... 🤔
> 
> The best idea I have at the moment is to use an abstract props collector, like `<ListBox.A11y>` that you can spread `a11yProps` on, and those will get redirected to either the `<Trigger>` or `role=listbox` element. Then we can preserve the propagation of `moreProps` on `<ListBox>` to the root element. The consumer just needs to know about the `<ListBox.A11y>` subcomponent, and the ListBox doesn't need to know anything about FormElement. What do you think?

I think we can just mark this change as a breaking change, if consuming app need `className` or `data-qa-id` they can wrap the `<ListBox />` by a `<div>` anyway, and spread the `a11yProps` on `<ListBox>`

nahumzs:
@mikrotron `<ListBox.A11y>` sounds interesting, and something that might worth slightly better, we introduce <ListBox.Box /> <ListBox.Trigger /> <ListBox.Popover /> in part to be able to target each individual div and their props spreading, I think that part is stable and work correctly and has work fine so far. 

If we know what props we need to make a11y works better why we don't extract them from the regular props and forward them to the `propProvider` as a11yProps?

This way there is not a need for a breaking change and still can support accessibility without bothering the consumer?

mikrotron:
> If we know what props we need to make a11y works better why we don't extract them from the regular props and forward them to the `propProvider` as a11yProps?

Well, _we_ know what the `a11yProps` are (today), but I don't think the `ListBox` should know, it couples it to the `FormElement`.  If the `FormElement` changes, and the content of `a11yProps` changes, then `ListBox` will need to change too 👎 

If we force the consumer to propagate them to the `<ListBox.A11y>` subcomponent then they need to know _something_ but not all the details, and the API remains intact even if the shape of `a11yProps` changes.  

As a refresher, currently the `a11yProps` includes:
`{ id, refLabel, disabled?, "aria-disabled"?, "aria-describedby"?, "aria-required"? }`

nahumzs:
So does the consumer if they want to use the From element would need to do:


```jsx
// just as an example didn't go to check how the FormElement API is currently working :D 
<FormElement>
{() => {
<ListBox>
  <ListBox.A11y={...formProps} />
</ListBox>
}}
</FormElement>
```

if this is a similar solution you are thinking I'm good with it and better than changing the other for props spreading, plus adapts to our current way of spreading props in the ListBox by subcomponents

packages/ListBoxWithTags/src/ListBoxWithTags.js
allison-c: I'd prefer copy/paste more than coupling stylings between packages. Reusing trigger css just saving some lines of code, but when the changes happening in list box, it might break list box with tags, we'll never know unless we test list box with tags. I just feel there're not many benefits from coupling

nahumzs:
> I'd prefer copy/paste more than coupling stylings between packages. Reusing trigger css just saving ?> some lines of code, but when the changes happening in list box, it might break list box with tags, 
> we'll never know unless we test list box with tags. I just feel there're not many benefits from 
> coupling

second, this sentiment prefers duplication that small abstraction.

packages/ListBox/src/ListBox.js
```diff
@@ -12,37 +12,44 @@ import Popover from "./components/Popover";
...
+    trigger: _trigger,
+
+    ...moreProps
   } = props;
+
   const I18n = useI18n();
+  const [state] = useListBox();
+  const { isInline, isReadOnly } = React.useContext(PropsContext);
+
+  const hasOptions = Boolean(React.Children.count(children));
```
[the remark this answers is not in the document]
nahumzs: I think was introduced at the very very early stages of th elistbox but now make totally sense to be hasChildren instead of options.

packages/ListBox/src/ListBox.js
```diff
@@ -51,21 +58,30 @@ export function ListBox(props) {
...
-            <Options isPopoverOpen={props.isOpen}>{children}</Options>
+          {isReadOnly ? null : filter}
+          <List {...listProps}>
+            <Options isPopoverOpen={isOpen}>{children}</Options>
           </List>
           {filter ? (
             <NoResults label={filter.props.noResultsMessage || I18n.t("listBox.filter.no_results_message")} />
           ) : null}
-          {footer ? React.cloneElement(footer, { ref: state.refFooterContainer }) : null}
+          {footer && !isReadOnly ? React.cloneElement(footer, { ref: state.refFooterContainer }) : null}
```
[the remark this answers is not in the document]
mikrotron: If it's `isInline` we do need to render it even if it's `isReadOnly`.  But you're right, if it's `!isInline && isReadOnly` than we could save some rendering.  But... does it make more sense to put that condition here, or inside the `<Content>` 🤔

packages/ListBox/src/components/Content/Content.js
```diff
@@ -83,16 +74,25 @@ export default function Content(props) {
...
-      onBlur={handleBlur(state, dispatch, props.onCancelFooter)}
+      {...getDOMAttributesForListBox({ idListBox, ...state })}
+      onBlur={handleBlur(state, dispatch, onCancelFooter)}
+      onKeyDown={handleKeyDownKeyboardKeys({ providedProps, state, dispatch, onChangeContext })}
+      onKeyUp={handleKeyUpKeyboardKeys({ providedProps, state, dispatch, onChangeContext })}
       ref={refListBoxContainer}
-      {...getDOMAttributesForListBoxContainer()}
-      onKeyDown={handleKeyDownKeyboardKeys({ state, dispatch, onChangeContext })}
-      onKeyUp={handleKeyUpKeyboardKeys({ state, dispatch, onChangeContext })}
+      role="listbox"
```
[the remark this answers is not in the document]
mikrotron: A `<Popover.Content>` will already have `tabIndex="-1"`.

packages/ListBox/src/ListBox.js
```diff
@@ -177,27 +214,26 @@ export const propTypes = {
...
   height: 200,
+  id: null,
   isDisabled: false,
   isInline: false,
   isMulti: false,
   isOpen: null,
+  isReadOnly: false,
   onChange: () => {},
-  placeholder: null,
+  placeholder: "",
```
nahumzs, line 229: why would you want to have a placeholder `= ""` ? wouldn't make sense if there is not placeholder there is not dom render? or what would be the benefit of not using null?

packages/ListBox/src/components/Box/Box.styles.js
```diff
@@ -1,22 +1,26 @@
...
 `;
 
 export const Box = styled.div(
-  ({ isInline, triggerWidth }) => css`
-    background-color: #ffffff;
-    border: 1px solid #d7d7d7;
-    border-radius: 3px;
-    box-shadow: 0px 1px 5px rgba(0, 0, 0, 0.07), 0px 7px 17px rgba(0, 0, 0, 0.1);
+  ({ hasError, isInline, isReadOnly, triggerWidth }) => css`
+    background-color: ${tokens.color.white};
```
nahumzs, line 14: thank you I think this has been like this even before we got a tokens packages :P

packages/ListBox/src/components/Box/Box.styles.js
```diff
@@ -1,22 +1,26 @@
...
+    border: 1px solid ${tokens.border.color};
+    border-radius: ${tokens.border.radius};
+    box-shadow: ${tokens.shadow};
     box-sizing: border-box;
     padding: 0;
-
-    ${isInline && isInlineCSS}
+    transition: border-color 0.2s;
```
nahumzs, line 20: curious what does this do?

packages/ListBox/src/components/Content/Content.js
nahumzs: I think is fine :)

packages/ListBox/src/components/Box/Box.styles.js
mikrotron: There's a gentle transition when an error (red border) is applied.  There already was one for the trigger, so the transition looked a bit weird when it only applied to part of the outline.

Screen Shot 2021-02-24 at 2 25 08 PM

packages/ListBox/src/components/Divider/Divider.styles.js
nahumzs: /*
 * IMPORTANT: THE STYLE FOR ALL OPTION MUST BE THE SAME SIZE
 * THIS IS IMPORTANT FOR CALCULATING THE CORRECT POSITION WHEN EXECUTING useOnScrolled() on Options/helpers/options.js
 * IF THIS STYLES CHANGES THEN UP AN DOWN WILL NOT WORK CORRECTLY WHEN THE DIVIDER EXISTS
 * PAY ATTENTION WHEN YOU MODIFY THE FOLLOWING ATTRIBUTES:
 * - border
 * - margin
 * - padding
 */


I don't think this is applicable anymore since @oscarkwan change the implementation to focus on the option.

before the implementation controlled and scroll the container to the next options, now we don't need to the focus do it by itself. As far I can think of.

mikrotron:
> @mikrotron So does the consumer if they want to use the From element would need to do:
> 
> ```js
> // just as an example didn't go to check how the FormElement API is currently working :D 
> <FormElement>
> {(formProps) => {
> <ListBox>
>   <ListBox.A11y={...formProps} />
> </ListBox>
> }}
> </FormElement>
> ```
> 
> if this is a similar solution you are thinking I'm good with it and better than changing the other for props spreading, plus adapts to our current way of spreading props in the ListBox by subcomponents

Yes, that's right, that's what I'm thinking, just not sure about the name?  This is a bit more work than @allison-c's suggestion (to just call this a breaking change) but I do think it's a bit better since it supports adding `moreProps` to the root element too.  It's still a breaking change, because if consumers leave the `refLabel` on the `<ListBox>` element, instead of moving it to `<ListBox.A11y>` some of the labelling won't work as well.  But this is not a very disruptive break, compared to breaking a lot of tests.

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -124,10 +87,10 @@ const Filter = React.forwardRef((props, ref) => {
   };
 
   React.useEffect(() => {
-    if (!state.isOpen) {
+    if (!isOpen) {
```
nahumzs, line 90: I wonder if leaving state is better to understand the distinction if the value comes from the props or from the state? or doesn't really matter.

mikrotron: Yeah, debatable.  In some cases, like this:

https://github.com/acl-services/paprika/pull/954/files#diff-b2903cb528031bc4a56d23248dfb6dc0093acec40ddf2141253c18151bb5335fR65

I think it's good to be explicit, but elsewhere I think it's a bit easier to read, and definitely easier to refactor, if it's destructured.

packages/ListBox/src/components/Trigger/Trigger.js
nahumzs: `if(refLabel?.current instanceof Element)` wouldn't you want to test is actually a dom element? or we can assume is a dom element? 

wouldn't be better if you always write a new id?, you wouldn't have to worry about it if has it or not, just always overwrite it? 

Also, super side option, I wonder if you can take control of the labe like the following and then tweak it at your pleasure in a more reactify way.

`ReactDOM.render(<Lable id="">your stuff</Label>, refLabel.current)`

mikrotron: > wouldn't you want to test is actually a dom element?

Good idea, will do.

> wouldn't be better if you always write a new id? 
> I wonder if you can take control of the labe like the following and then tweak it at your pleasure in a more reactify way.

These seem too obtrusive to me for a component to be imposing such side effects on a sibling or parent element/component.  I think it's the wrong direction of control.  I'm a little hesitant about even writing the `id` attribute, which is why I check first to see if one exists rather than just overwriting – if there's no change, just addition, it seems less disruptive, but it's still the wrong direction of control.  Just the best solution I could think of without requiring a change to, and stronger coupling with, the `FormElement`.

packages/ListBox/src/ListBox.js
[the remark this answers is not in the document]
AndreyChernykh: Why not 
```js
React.Children.count(children) > 0
```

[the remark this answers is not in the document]
mikrotron: Tomato, potato 🤷‍♂️

packages/ListBox/src/ListBox.js
```diff
@@ -177,29 +239,24 @@ export const propTypes = {
...
   isOpen: null,
+  isReadOnly: false,
   onChange: () => {},
   placeholder: null,
   size: ListBoxContainer.types.size.MEDIUM,
-  trigger: null, // eslint-disable-line
 };
 
+ListBox.propTypes = propTypes;
+ListBox.defaultProps = defaultProps;
```
mikrotron, line 258: Changes this up a bit to get the props table to generate properly.

packages/ListBox/src/components/A11y/A11y.js
```diff
@@ -0,0 +1,30 @@
...
+/* A shell component to collect and distribute a11y related props and DOM
+ * attributes to the appropriate (interactive) element
+ */
+
+import React from "react";
+import PropTypes from "prop-types";
+import { RefOf } from "@paprika/helpers";
+
+export default function A11y() {
+  return <></>;
```
mikrotron, line 12: `return null;` confuses `react-docgen` and this component won't get added to the props tables.

scripts/readmeFiles.js
```diff
@@ -209,7 +209,7 @@ shell.ls("packages").forEach(folder => {
 
       initAutoReadme({ path, content });
     } catch (e) {
-      console.warn(e);
+      console.warn(folder, e);
```
mikrotron, line 212: Just a little improvement – some of the components are still not working properly, and this helps identify which ones.

packages/ListBox/src/components/A11y/A11y.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export const propTypes = {
+  /** DOM id attribute for focussable control (trigger element or ul element if isInline=true) */
+  id: PropTypes.string,
+
+  /** Ref for a DOM element containing the label for this component */
+  refLabel: RefOf(PropTypes.instanceOf(Element)),
```
mikrotron, line 20: A bit of a compromise, the `id` and `refLabel` are now a part of the API for this component.  They're still not required, but the `a11y` will be improved, especially if `refLabel` is provided.

packages/ListBox/src/ListBox.js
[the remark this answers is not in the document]
mikrotron: Turns out this is the wrong logic anyway 😆

packages/ListBoxWithTags/src/ListBoxWithTags.js
mikrotron: To be addressed in https://aclgrc.atlassian.net/browse/UXD-868

packages/ListBox/test/cypress/ListBox.cypress.js
```diff
@@ -155,12 +154,12 @@ describe("ListBox multi select filter", () => {
       .and("contain", "Wolverine, Catwoman");
   });
 
-  // TODO: Flaky spec, works locally but fails on semaphore
+  // TODO: Flakey spec, works locally but fails on semaphore
```
oscarkwan, line 157: Do you know if this spec is still flaky after your changes with typing delay?

nahumzs:
> > @mikrotron So does the consumer if they want to use the From element would need to do:
> > ```js
> > // just as an example didn't go to check how the FormElement API is currently working :D 
> > <FormElement>
> > {(formProps) => {
> > <ListBox>
> >   <ListBox.A11y={...formProps} />
> > </ListBox>
> > }}
> > </FormElement>
> > ```
> > 
> > 
> > if this is a similar solution you are thinking I'm good with it and better than changing the other for props spreading, plus adapts to our current way of spreading props in the ListBox by subcomponents
> 
> Yes, that's right, that's what I'm thinking, just not sure about the name? This is a bit more work than @allison-c's suggestion (to just call this a breaking change) but I do think it's a bit better since it supports adding `moreProps` to the root element too. It's still a breaking change, because if consumers leave the `refLabel` on the `<ListBox>` element, instead of moving it to `<ListBox.A11y>` some of the labelling won't work as well. But this is not a very disruptive break, compared to breaking a lot of tests.

we should go with this option then :D

packages/ListBox/test/cypress/ListBox.cypress.js
mikrotron: It was last week anyway yeah.

mikrotron:
> Since the `moreProps` has been moved to the trigger element, not the root popover wrapper anymore, I think this might be a breaking change. It's still a change of props. For example, before if I pass `className`, `className` will be applied to the root, now it'll be applied to the trigger only.

Was it really applied to the root before? I wasn't able to see the code that applied that... but have ensured that it happens now 😄

nahumzs (approved):
👍

PR #958, opened 2021-02-25
UXD-908 add changesets and remove auto changelog rela…

allison-c:
…ted script

### Purpose 🚀

- [x] Added changesets
- [x] Removed changelog related scripts
- [x] Removed the versions in readme files, so the `postpublish` action can be removed
- [x] Added changeset bot separately (pending review?) 
- [x] Updated the pre-release and release yml 

### Notes ✏️

**Pre-release name changes**
I don't think `changesets` support pre-release version with `alpha` and with the tag `next`. If we want to use `next` tag, the name will be changed to e.g. `v1.0.1-next.0` . If we want to pre-release version with `v1.0.1-alpha.0` we have to set the npm tag as `alpha` too 😅   I think `next` tag makes more sense

### Updates 📦

### Storybook 📕

### Screenshots 📸

### References 🔗


.semaphore/pre-release.yml
```diff
@@ -23,7 +23,7 @@ blocks:
         - name: "Publish to NPM"
           commands:
             - nvm use
-            - npx lerna bootstrap
             - yarn
-            - yarn lerna version prerelease --yes
-            - yarn lerna publish from-package --yes --dist-tag next
+            - test -f .changeset/pre.json || yarn changeset pre enter next
+            - yarn changeset version
+            - git status --porcelain && git add . && git commit -m"Pre-release" --no-verify && yarn changeset publish && git push origin master --follow-tags --no-verify
```
nahumzs, line 29: `nit:` `git commit -m "[Bot] Paprika Pre-release"`

.semaphore/pre-release.yml
```diff
@@ -23,7 +23,7 @@ blocks:
         - name: "Publish to NPM"
           commands:
             - nvm use
-            - npx lerna bootstrap
             - yarn
-            - yarn lerna version prerelease --yes
-            - yarn lerna publish from-package --yes --dist-tag next
+            - test -f .changeset/pre.json || yarn changeset pre enter next
```
nahumzs, line 27: what is this doing?

allison-c: Just checking if the `pre.json` already there, if it's not there we need to run `yarn changeset pre enter next` to create that file and enter the `pre-release` mode (they called)

If we run `yarn changeset pre enter next` with the file exist, there'll be errors, so I need to check first

Similar with what we need to do in `release.yml`, if the `pre.json` file exist we need run `yarn changeset pre exit`

.semaphore/pre-release.yml
nahumzs: why do you need to run `git status --porcelain` I thought porcelain is just a "stable" output so you could parse it, but I think you are using it to see if there is a wrong or missing yarn lock or missing readme file, before even trying the release?

.semaphore/release.yml
```diff
@@ -0,0 +1,29 @@
...
+        - name: github-ssh
+        - name: paprika-env
+      prologue:
+        commands:
+          - checkout
+          - cache restore client-node-modules-$SEMAPHORE_GIT_BRANCH-$(checksum yarn.lock),client-node-modules-$SEMAPHORE_GIT_BRANCH,client-node-modules-master
+          - chmod 0600 ~/keys/github.key
+          - ssh-add ~/keys/github.key
+          - git config user.name "Paprika Semaphore 2.0"
+          - git config user.email "allison.chen@wegalvanize.com"
```
nahumzs, line 20: we should get a more generic email, only one person should have access to it, so we can increase security and delegate to the CI the responsibility to deploy and push to master/npm.

.semaphore/pre-release.yml
allison-c: Yes, just checking if there are any changes, if there're changes we will commit and publish and push. Otherwise just stop. 
I googled the `porcelain` option, because they said it's machine parsable results. To be honest I didn't try if `-s` works or not, `--porcelain` works pretty well in my testing repo.

.semaphore/release.yml
allison-c: 🤦‍♀️ yes

.semaphore/release.yml
```diff
@@ -0,0 +1,29 @@
...
+          - echo "//registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN" >> .npmrc
+      jobs:
+        - name: "Publish to NPM"
+          commands:
+            - nvm use
+            - yarn
+            - test -f .changeset/pre.json && yarn changeset pre exit
+            - yarn changeset version
+            - git status --porcelain && git add . && git commit -m"[Bot] Paprika Release" --no-verify && yarn changeset publish && git push origin master --follow-tags --no-verify
```
jamiek-acl, line 29: Oh `git status --porcelain` would be great for scripting... but what is it being used for here? I dont think it returns `-1` and stops the script

allison-c: We only want to stop the script when there are no changes. Because if there's no changes `git commit` will return errors

allison-c: Oh and `no changes` means there's nothing to release

allison-c: Jamie is right! I missed something!

.semaphore/pre-release.yml
```diff
@@ -23,7 +23,7 @@ blocks:
         - name: "Publish to NPM"
```
jamiek-acl, line 23: Would it be more clear to call this file `.semaphore/alpha-release.yml`, and change line 23 to that as well, and update the scheduler on semaphore? Maybe it's just me, but i think this name insinuates that this file gets run immediately before `.semaphore/release.yml` -- which it does not.

allison-c: I see, how about `next-release.yml`? because there won't be any `alpha` versions anymore, will be `next` versions instead

jamiek-acl: Fine by me, if it makes sense to you and @nahumzs  (also on line 29 in this file maybe update that commit message as well)

jamiek-acl: @nahumzs allison and i chatted and agreed to leave it as `pre-release` since it's the industry standard term.

nahumzs (dismissed):

nahumzs (approved):

PR #960, opened 2021-02-28
UXD-868 ListBoxWithTags compatibility with FormElement

mikrotron:
### Purpose 🚀
Add support for `isDisabled`, `isReadOnly`, and `hasError` props to `<ListBoxWithTags>` and handle the `a11yProps` provided by `<FormElement.Content>` via `<ListBoxWithTags.A11y>` prop collector component.

### Notes ✏️
- Companion PR for `<ListBox>`: https://github.com/acl-services/paprika/pull/954
- Also refactored Storybook stories.
- Refactored `<ListBox>` `getDOMAttributesForListBoxButton()` to include `aria-expanded` + `aria-controls`.

### Updates 📦
- [x] MINOR (backward compatible) change to `ListBox` + `ListBoxWithTags`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-868--listbox-tags-formelement

### Screenshots 📸
Screen Shot 2021-02-27 at 4 16 33 PM
Screen Shot 2021-02-27 at 4 17 04 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-868


packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -205,12 +206,12 @@ export default function Trigger(props) {
         handleKeyDown: handleKeyDownKeyboardKeys({ state, dispatch, onChangeContext }),
         handleKeyUp: handleKeyUpKeyboardKeys({ state, dispatch, onChangeContext }),
         isOpen,
-        isReadOnly,
```
mikrotron, line 208: Decided not to include this – if other states like `isDisabled` and `hasError` are not included, this shouldn't be either.

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -2,65 +2,25 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { v4 as uuidv4 } from "uuid";
+import * as constants from "@paprika/constants/lib/Constants";
 import useI18n from "@paprika/l10n/lib/useI18n";
 import ListBox from "@paprika/list-box";
 import { filter } from "@paprika/list-box/lib/helpers/filter";
 import Tag, { Tags } from "@paprika/tag";
-/* eslint-disable no-restricted-syntax */
-import * as triggerSc from "@paprika/list-box/lib/components/Trigger/Trigger.styles";
```
mikrotron, line 10: 🥾

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -76,60 +36,65 @@ const renderTrigger = ({
...
-
-          if (typeof label !== "string") {
-            throw Error(
-              `Your item ${JSON.stringify(
-                item
-              )} must include the attribute "label", or you must indicate which attribute should be rendered as the label via the "tagLabelKey" prop.`
+    <>
+      <sc.Trigger
+        {...triggerProps}
+        aria-controls={idListBoxContent}
```
mikrotron, line 53: `triggerProps` also includes `aria-controls`, but it assumes we use the same `idListBox` for the `id` of the controlled content, which we don't, since we don't have that value when we assign `idListBoxContent`.

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -76,60 +36,65 @@ const renderTrigger = ({
...
-          return (
-            <Tag as="li" key={label} onRemove={handleRemove(item)} size={size}>
-              {label}
-            </Tag>
-          );
-        })}
-        {selectedOptions.length ? null : (
-          <sc.PlaceHolder>
-            <sc.PlaceHolderText>{t("listBoxWithTags.placeholder")}</sc.PlaceHolderText>
-          </sc.PlaceHolder>
```
mikrotron, line 128: I don't know why this needs to have two elements – I've combined into a single `<sc.PlaceHolder>` and it seems to work fine.

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -76,60 +36,65 @@ const renderTrigger = ({
...
+            <sc.PlaceHolder isDisabled={isDisabled} size={size}>
+              {t("listBoxWithTags.placeholder")}
+            </sc.PlaceHolder>
+          )}
+        </Tags>
+        {allOptionsAreSelected ? null : renderCaret(isDisabled)}
+      </sc.Trigger>
+      <sc.TriggerLabel id={`${triggerProps.id}__label`}>
+        {t("listBoxWithTags.a11y_text_trigger", { options: a11yText })}
+      </sc.TriggerLabel>
```
mikrotron, line 96: The previous `a11yText` prop (`aria-label` value) is now in this hidden element. The `id` is used along with the `id` for the  element `refLabel` refers to for a better combined accessible name.

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -220,13 +187,88 @@ export default function ListBoxWithTags(props) {
...
+  onRemove: PropTypes.func,
+
+  /** Render prop to override the default Tag style, see example for it's uses.  */
+  renderTag: PropTypes.func,
+
+  /** An array of id that helps the ListBoxWithTags to known what elements are selected  */
+  selectedOptions: PropTypes.arrayOf(PropTypes.shape({})),
+
+  /** Size of the trigger and options (font size, height, padding, etc). */
+  size: PropTypes.oneOf([ListBoxWithTags.types.size.MEDIUM, ListBoxWithTags.types.size.LARGE]),
```
mikrotron, line 252: After seeing `size` referenced repeatedly throughout the `ListBoxWithTags` source, I decided it made sense to include it in the explicit API for this component (even though all props for `ListBox` will get passed down to the underlying `<ListBox>`.

packages/ListBoxWithTags/stories/ListBoxWithTags.examples.stories.js
```diff
@@ -0,0 +1,99 @@
...
+  </ExampleStory>
+);
+CustomTriggerStory.story = { name: "Custom Trigger", parameters: exampleStoryParameters };
+
+export const RenderTagLabelStory = () => (
+  <ExampleStory storyName="Render Tag Label" component="ListBoxWithTags" fileName="examples/RenderTagLabel.js">
+    <RenderTagLabel />
+  </ExampleStory>
+);
+RenderTagLabelStory.story = { name: "Render Tag Label", parameters: exampleStoryParameters };
```
mikrotron, line 60: This story wasn't in the Storybook before, but the `RenderTagLabel.js` file was in the `examples` folder, so I included it.

packages/ListBoxWithTags/stories/examples/FormElement.js
```diff
@@ -0,0 +1,47 @@
...
+        <FormElement.Instructions>Scenester brunch instructions</FormElement.Instructions>
+        <FormElement.Content>
+          {a11yProps => (
+            <ListBox
+              {...moreUseListBoxWithTagsProps}
+              data-qa-anchor="listbox-tags-demo"
+              hasError
+              selectedOptions={getSelectedOptions()}
+            >
+              <ListBox.A11y {...a11yProps} />
```
mikrotron, line 32: Since the `ListBoxWithTags` doesn't support an inline version, it would be possible to just pass these props to `<ListBoxWithTags.Trigger>`, but for consistency with `ListBox`, this same API is preferred.

packages/ListBoxWithTags/src/ListBoxWithTags.js
nahumzs: I didn't put it as a prop because you then will have to include all listbox props which doesn't make sense is easier to say that the component inherent all listbox props which is what it does. 

then we don't have to expose a prop that doesn't give any benefit on the read me file.

nahumzs:
In the UI state story, on the pill state that is Read Only if you try to select the text (which is possible) the highlight colour is grey, which makes it impossible (even for me) to detect that the text is selected. Would make sense to add another highlight cursor color (via css) in that state so it clear that you can select the text in the read only state? @mikrotron cc @sunminsarahlee

packages/ListBoxWithTags/stories/examples/AllOptionsAreSelected.js
```diff
@@ -15,16 +15,14 @@ export default function App() {
...
-              {option.label}
-            </ListBox.Option>
-          ) : null;
-        })}
-      </ListBox>
-    </div>
+    <ListBox isOpen allOptionsAreSelected selectedOptions={getSelectedOptions()} {...moreUseListBoxWithTagsProps}>
+      {filteredData.map(option => {
+        return !isSelected(option.label) ? (
+          <ListBox.Option value={option.label} key={option.label} label={option.label}>
```
AndreyChernykh, line 21: Would it be better to separate a `label` from `key`? 
Maybe an `option.id`? 

I can easily imagine having 2 `John Doe`s in one company, for instance.

AndreyChernykh:
LGTM

packages/ListBoxWithTags/src/ListBoxWithTags.js
mikrotron: My opinion changed as I got into the code. Since the code of this component relies on that prop, and in fact manipulates it (for example, if `size="small"` is passed to this component, it will be adjusted to `size="medium"`) I think it should be an explicit part of the API.  

Additionally, prior to these changes, you could pass `isDisabled` to the `<ListBoxWithTags>` but it wouldn't have an effect (or maybe it only partially worked).  But now that it is handled explicitly in the code for this component, I think it's beneficial to include it in the PropTypes.  

It also means we don't need a bunch of `eslint-disable` statements, which makes the code smell better 😊

But I don't think it's necessary to include [all the ListBox props](https://github.com/acl-services/paprika/tree/master/packages/ListBox#props) just the ones that have some implementation details in this component.  And I've included [a note in the README](https://github.com/acl-services/paprika/pull/960/files#diff-18ef3a3c4a8b79a26448ca463dbc7de5af233afbe93ecb9d153584940b380f58R47-R53) to document the inclusion of the ListBox API.

mikrotron:
> In the UI state story, on the pill state that is Read Only if you try to select the text (which is possible) the highlight colour is grey, which makes it impossible (even for me) to detect that the text is selected. Would make sense to add another highlight cursor color (via css) in that state so it clear that you can select the text in the read only state? @mikrotron cc @sunminsarahlee

Are you using Windows? It looks okay for me:
Screen Shot 2021-03-01 at 10 11 06 AM

But this a small detail I don't have very strong opinions about.

sunminsarahlee:
Same.. it looks okay to me but if we can do it, that's great. 
Screen Shot 2021-03-01 at 3 56 39 PM

mikrotron:
> In the UI state story, on the pill state that is Read Only if you try to select the text (which is possible) the highlight colour is grey, which makes it impossible (even for me) to detect that the text is selected. Would make sense to add another highlight cursor color (via css) in that state so it clear that you can select the text in the read only state? 

Updated with `tokens.color.blueLighten30`

Screen Shot 2021-03-01 at 4 16 53 PM

nahumzs (approved):

PR #965, opened 2021-03-03
UXD-927 ListBoxWithTags placeholder 🐐

mikrotron:
### Purpose 🚀
Add support for `placeholder` prop in `<ListBoxWithTags>`.

### Notes ✏️
- Also extracted `renderTrigger()` to a separate `<TriggerWithTags>` sub component.
- Renamed Storybook story "Custom Trigger" to "Custom Tag", which is more accurate, and fixed the avatar styling.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-927--listboxtags-placeholder/?path=/story/forms-listboxwithtags--ui-states-story

### Screenshots 📸
From UI States story
Screen Shot 2021-03-02 at 8 06 36 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-927


packages/ListBox/src/ListBox.js
```diff
@@ -228,7 +228,7 @@ export const propTypes = {
   /** Callback returning the current selection on the ListBox */
   onChange: PropTypes.func,
 
-  /** Defaults label to display when the ListBox has not option selected */
+  /** Default label for trigger when the ListBox has no option selected */
```
mikrotron, line 231: I did not include this change in the `changeset` as it has no effect on the actual code for the `ListBox`.

packages/ListBoxWithTags/CHANGELOG.md
```diff
@@ -20,7 +20,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Support for `isDisabled`, `isReadOnly`, and `hasError` props. [@mikrotron](https://github.com/mikrotron)
-- Support fo `<ListBoxWithTags.A11y>` props collector. [@mikrotron](https://github.com/mikrotron)
+- Support for `<ListBoxWithTags.A11y>` props collector. [@mikrotron](https://github.com/mikrotron)
+- Support for `placeholder` prop. [@mikrotron](https://github.com/mikrotron)
```
mikrotron, line 24: Not sure if we should be touching the `CHANGELOG` manually anymore? Seems like we can just rely on the `changeset` message?

cc @allison-c

packages/ListBoxWithTags/src/renderTrigger.js
```diff
@@ -0,0 +1,93 @@
...
+
+            return (
+              <Tag as="li" key={label} onRemove={handleRemove(item)} size={size}>
+                {label}
+              </Tag>
+            );
+          })}
+          {selectedOptions.length ? null : (
+            <sc.Placeholder isDisabled={isDisabled} size={size}>
+              {placeholder || t("listBoxWithTags.placeholder")}
```
mikrotron, line 80: This is the key change here.

packages/ListBoxWithTags/CHANGELOG.md
allison-c: Yes, you're right

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -6,98 +6,7 @@ import * as constants from "@paprika/constants/lib/Constants";
...
-        </Tags>
-        {allOptionsAreSelected ? null : renderCaret(isDisabled)}
-      </sc.Trigger>
-      <sc.TriggerLabel id={`${triggerProps.id}__label`}>
-        {t("listBoxWithTags.a11y_text_trigger", { options: a11yText })}
-      </sc.TriggerLabel>
-    </>
-  );
-};
+import renderTrigger from "./renderTrigger";
```
AndreyChernykh, line 9: At this point `renderTrigger` can pretty much be a component:

```js
function ListBoxTrigger({ allOptionsAreSelected, ....... } )
```

```js
import ListBoxTrigger from "./ListBoxTrigger";
```

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -162,6 +72,7 @@ export default function ListBoxWithTags(props) {
             isDisabled,
```
AndreyChernykh, line 72: replace with `<ListBoxTrigger />` (see the comment above)

packages/ListBoxWithTags/src/ListBoxWithTags.js
AndreyChernykh: IMO, we need to reserve `renderSomthing` functions for the render props OR internal render functions. If the function is already extracted, then it is technically a component already.

packages/ListBoxWithTags/src/renderTrigger.js
```diff
@@ -0,0 +1,93 @@
...
+
+  function renderCaret(isDisabled) {
+    return isOpen ? <sc.UpIcon isDisabled={isDisabled} /> : <sc.DownIcon isDisabled={isDisabled} />;
+  }
+
+  const triggerProps = propsForTrigger();
+  const a11yTextOptions = selectedOptions.map(item => {
+    return tagLabelKey === null ? item.label : item[tagLabelKey];
+  });
+  const a11yText = selectedOptions.length === 0 ? t("listBoxWithTags.placeholder") : a11yTextOptions.join(", ");
```
AndreyChernykh, line 40: Shouldn't we have a custom trigger here as well? 

```js
placeholder || t("listBoxWithTags.placeholder")
```

packages/ListBoxWithTags/src/ListBoxWithTags.js
mikrotron: It is a render prop though, specifically for the `children` prop of `ListBox.Trigger`.  This needs to return a function, not JSX, or it won't be used at all:

https://github.com/acl-services/paprika/blob/master/packages/ListBox/src/components/Trigger/Trigger.js#L245

My original idea was to extract as a component, but a component can't return a function (as far as I know!).  

I still think it has a specific enough purpose that it should be extracted from `ListBoxWithTags.js` though 😅

packages/ListBoxWithTags/src/ListBoxWithTags.js
mikrotron: I was going to call it `<TriggerWithTags>` 😝 but I don't think it will actually work as a component.

packages/ListBoxWithTags/src/renderTrigger.js
mikrotron: Not sure what you mean.  The `<ListBoxWithTags>` doesn't actually support a custom trigger (at this point). It's the `ListBox` with a built in custom trigger and custom hooks.

jamiek-acl: i think he meant use the custom placeholder in the a11yText if they provided one:

```
const a11yText = selectedOptions.length === 0 ? placeholder ||
t("listBoxWithTags.placeholder") : a11yTextOptions.join(", ");
```

mikrotron: Because isn't it just a `<ListBox>` then?

Maybe @nahumzs has more insight here, but my impression is that this component basically _is_ the custom trigger with tags.

mikrotron: > i think he meant use the custom placeholder in the a11yText if they provided one

OH. 🤦

packages/ListBoxWithTags/src/ListBoxWithTags.js
nahumzs: I think you can do something like this:

```jsx
// the internal render of ListBox.Trigger
function Trigger(props) {
  const triggerStuff = { triggerA: 1, triggerB: 2 };
  return props.children(triggerStuff);
}

function CustomTrigger(props) {
  return `${props.triggerA} ${props.triggerB} ${props.x} ${props.y}`;
}

  const customApp = (extraProps) => (props) => {
    return <CustomTrigger {...props} {...extraProps} />;
  };
  
  return (
    <div className="App">
      <Trigger>{customApp({ x: "x", y: "y" })}</Trigger>
    </div>
  );
}
```

https://codesandbox.io/s/nervous-lamarr-kc25b?file=/src/App.js:24-504

packages/ListBoxWithTags/src/renderTrigger.js
nahumzs: Was part of a ListBox before, but doesn't make sense to have two-component with a similar UI but different concepts in the same package, is difficult to apply proper semver for the two components without affecting the versioning of each other. That's the reason why is a different package. 

is that the question @mikrotron ?

packages/ListBoxWithTags/src/ListBoxWithTags.js
mikrotron: Cool, this is very similar to the approach Andrei suggested offline in a little chat we had 👍 👍

packages/ListBoxWithTags/src/components/TriggerWithTags/TriggerWithTags.js
```diff
@@ -88,6 +101,37 @@ const renderTrigger = ({
       </sc.TriggerLabel>
     </>
   );
+}
+
+TriggerWithTags.displayName = "ListBoxWithTags.TriggerWithTags";
+
+TriggerWithTags.propTypes = {
```
nahumzs, line 108: IMHO, most of these props don't make sense to declare them, the wrapper can't control them are just spread. 

And can be easily out of sync, neither provide any useful information to the consumer and don't see how much value add to the author.

Just my opinion.

nahumzs (approved):

PR #973, opened 2021-03-13
UXD-907 refactor up/down arrowkeys first pass

nahumzs:
### Purpose 🚀
---
"@paprika/list-box": major
---

💥 Breaking change is:

Refactored keyboard events for ↑ and ↓ for the list component:

This is a partial refactor for the keyboard navigation for the `ListBox` component we switched from tracking the options on the state against tracking the actually rendered option on the HTML. The previous help to only track the real options that we should consider each time the user press ↑ or ↓.

✏️ why the change was made:
The original implementation for the keyboard navigation for the ListBox didn't take in consideration the focus() event and uses the options state to figure out if the next element can be selected, there are ListBox.Divider, ListBox.RawOption, ListBox.Option `isDisabled`, and a case when the options are filtered.

So instead of tracking all possible cases and combinations now we only track whatever the ListBox has rendered and each time the user press ↑ or ↓ the component looks for the next available sibling from the current one focus which is more efficient than trying to find the next available in the options list.

👷 How a consumer should update their code:
There is not really an action from the consumer side of things, this is marked as a breaking change because might break side code if you were in somehow depending on the previous implementation. But most likely you won't need to take further actions

- `Content.js` No longer requires to handle onBlur and onFocus state.
- `Filter.js` onBlur is no longer required, also fixed the missing cleaning function for two hooks and fixed a minor bug that will scroll the popover if the user use the mouse wheel
- `optionState.js` Now is possible to access `state.optionsIndex` and access the index of any given rendered by their uuid
- `option.js` remove `listBoxHasFocus={listBoxHasFocus}` prop not longer relevant.

**option.js**

- ✅ Added getOptions, getNextUp and getNextDown methods to detect which is the next focusable option
- 🗑 Remove getNextOptionActiveIndex and getNextOptionActiveIndexLooping which were really confusing methods not long necessary!
- ✨ `handleArrowKeys` method is cleaner and easier to understand

**Trigger.js**

- Now the ListBox will focus automatically in the first selected option when multi and in the selected option in single
- When there is not a selected option in multi will focus on the first element while in single will not do anything apart from opening the options

**reducer.js**

- 🙅‍♀️ useListBox.types.setListBoxHasFocus no longer necessary
- ✨ useListBox.types.applyFilter simplified version now

http://storybooks.highbond-s3.com/paprika/UXD-907-listbox-keyboard

### Screenshots 📸

### References 🔗


.changeset/strong-cups-cry.md
```diff
@@ -0,0 +1,38 @@
+---
+"@paprika/list-box": major
+---
+
+💥 The breaking change is:
```
oscarkwan, line 5: I like all the detailed explaination, but this will translate and be put into the `Changelog.md` I'm not sure we wan't this much text inside there.

Thoughts?

allison-c: I agree with Oscar, this might be too long to read in `CHANGELOG.md`, maybe can refer the explanation part from a storybook doc link

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -186,7 +222,7 @@ export const handleClickOption = ({ props, isDisabled, state, dispatch, onChange
   }
 
   if (hasFilter && isMulti) {
-    refFilterInput.current.focus();
+    // refFilterInput.current.focus();
```
oscarkwan, line 225: can cut out this code if not needed anymore

packages/ListBox/stories/ListBox.keyboard.stories.js
```diff
@@ -0,0 +1,42 @@
...
+import {
+  SimpleStory,
+  WithDisabled,
+  WithDividers,
+  WithFilterStory,
+  WithInline,
+  WithRawItems,
+} from "./examples/Keyboard";
+
+const storyName = getStoryName("ListBox");
```
oscarkwan, line 15: Can we put this under the `Examples` directory. We just refactored and restructured the stories :).

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,337 @@
...
+import Toast from "../../../../Toast/src";
+
+const Card = ({ children }) => <CardComponent style={{ padding: "16px", width: "280px" }}>{children}</CardComponent>;
+const CheckListItem = ({ children }) => {
+  /* eslint-disable jsx-a11y/label-has-associated-control */
+  return (
+    <div style={{ padding: "8px", display: "flex", position: "relative" }}>
+      <label>
+        <div style={{ position: "absolute", top: "10px", left: "-20px" }}>
+          <input type="checkbox" />
```
oscarkwan, line 17: Why checkboxes 😅

.changeset/strong-cups-cry.md
AndreyChernykh: I would disagree. When I want to update the component `CHANGELOG.md` is the first and the main place I go to get the instructions. I wouldn't afraid to have some extra text in here in case of the breaking change. 

The only thing I'd do differently is slightly change the headers formatting and standardize it. E.g.

instead of 

```
💥 The breaking change is:
```
I'd prefer
```
## 💥 BREAKING CHANGE 
```

Instead of 
```
👷 How a consumer should update their code:
```

```
### 👷 Migrating from 4.x.x to 5.0.1 :
```
for instance.

AndreyChernykh: I would also make the long and detailed description only for major updates and if necessary. Otherwise, I'd keep the description short and technical. We also can reference specific PRs if they have a more detailed explanation of the problem that was solved or an improvement that was made.

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -67,30 +61,38 @@ const Filter = React.forwardRef((props, ref) => {
...
-        // if has footer the responsible of closing is the Footer no the onblur event
-        return;
-      }
-
-      if (document.activeElement !== refListBoxContainer.current) {
-        // this will reset the activeOption and close the Popover
+  React.useEffect(() => {
+    let id = null;
+    if (!isOpen) {
+      id = window.requestAnimationFrame(() => {
```
AndreyChernykh, line 67: Don't see a reason why `id` is defined outside of the `if` statement: 

```js
if (!isOpen) {
  const id = indow.requestAnimationFrame(() => {
```

AndreyChernykh: or, maybe the cleanup function need to be outside of the `if ` statement instead 🤷

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -1,7 +1,96 @@
+import "@paprika/helpers/lib/polyfills/elementClosest";
 import useListBox from "../../../useListBox";
 import invokeOnChange from "../../../helpers/invokeOnChange";
 import { KEY_PRESS, CLICK } from "../../../types";
 
+function getOptions(event) {
+  let list = null;
+  if (event.target.dataset.pkaAnchor && event.target.dataset.pkaAnchor === "list-filter-input") {
```
AndreyChernykh, line 8:
```js
event.target.dataset?.pkaAnchor === "list-filter-input"
```

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -1,7 +1,96 @@
...
+    // the event is coming from the filter
+    list = event.target.closest("[data-pka-anchor='list-box-box']").querySelector("[data-pka-anchor='styled-list']");
+  } else {
+    list = event.target.querySelector("[data-pka-anchor='styled-list']");
+  }
+
+  // filtering only results with role="option" attribute
+  return [...list.children].filter(li => li.hasAttribute("role"));
```
AndreyChernykh, line 16: No need to `[...list.children]` if you `filter` because `filter` creates a new array anyway. Therefore `[...]` here is wasteful.

packages/ListBox/stories/examples/Keyboard/index.js
nahumzs: to do the checklist that is working while reviewing manually :)

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -1,7 +1,96 @@
...
+}
+
+function getNextUp(event) {
+  function next(element) {
+    const list = event.target.closest("[data-pka-anchor='styled-list']");
+    const options = list.children;
+    if (list && [...options].indexOf(element) > 0) {
```
AndreyChernykh, line 23: same here, `[...options]` is redundant. 

We don't need to create a shallow copy of the array or object, unless 

1. we actually want to modify this array\object
2. we actually care about immutability in the specific context

For instance : 

```js
function getMaxNumber(numbers) {
   const n = [...numbers];
   
   let max = n[0]; 
   
   for (i = 1; i < n.length; i++) {
     max = max > n[i] : max : n[i]
   }
   
   return max;
}

getMaxNumber([4,6,12,76,3])
```
☝️ here we don't care if the input array mutates, and we don't even mutate it - we only iterate through it. Therefore, no need to do `const n = [...numbers];`

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: list.children is a dom collection no an array, therefore lacks a filter implementation, requires to be turn into an array first.

packages/ListBox/src/components/Options/helpers/options.js
AndreyChernykh: Also, don't forget that even if we do `[...arr]` it is a shallow copy. So, 

```js
const oldArray = [{ name: 'Bob' }];
const newArray = [...oldArray];

newArray[0].name = 'John';
console.log(oldArray[0].name) //  'John'
```

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -1,7 +1,96 @@
...
+    return null;
+  }
+
+  // an activeElement is an option
+  if (document.activeElement.getAttribute("role") === "option") {
+    return next(document.activeElement);
+  }
+
+  // filter is already focussed just ignore the event
+  if (document.activeElement.dataset.pkaAnchor === "list-filter-input") {
```
AndreyChernykh, line 49: Do we need `.dataset?.` here?

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -1,7 +1,96 @@
...
+    return null;
+  }
+
+  // focus the first option
+  const options = getOptions(event);
+  if (options.length) {
```
AndreyChernykh, line 55: I'd prefer `options.length > 0` - more semantic

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: is not possible to reach this code if you haven't focus on the item first, therefore is certain that the object will be either a role="option" or a filter input both with a dataset property.

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -1,7 +1,96 @@
...
+  }
+
+  return null;
+}
+
+function getNextDown(event) {
+  function next(element) {
+    const list = event.target.closest("[data-pka-anchor='styled-list']");
+    const options = list.children;
+    if (list && [...options].indexOf(element) < options.length - 1) {
```
AndreyChernykh, line 66: same for `[...options]`

Also, I see a similar logic here. Can we create a generic function instead somehow?

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: 🍅 🥔 🤷

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -267,10 +303,14 @@ export function handleEnterOrSpace({ event, providedProps, state, dispatch, onCh
...
-    if (state.refTrigger.current) {
-      state.refTrigger.current.focus();
+    // no need to notify which option is selected just close the popover unless the option is disabled
+    const id = state.optionsIndex[document.activeElement.getAttribute("id")];
+    if (!state.options[id].isDisabled) {
+      dispatch({ type: useListBox.types.closePopover });
+
+      if (state.refTrigger.current) {
+        state.refTrigger.current.focus();
+      }
```
AndreyChernykh, line 313: Can we do `state.refTrigger.current?.focus();` 🤔

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -75,6 +75,37 @@ export default function Trigger(props) {
     };
   }, [refLabel, refTrigger, idListBox]);
 
+  React.useEffect(() => {
+    if (
+      isOpen &&
+      document.activeElement &&
+      document.activeElement.dataset.pkaAnchor === "list-box-trigger" &&
```
AndreyChernykh, line 82: Also, do we need `.dataset?.` ?

packages/ListBox/src/components/Trigger/Trigger.js
```diff
@@ -75,6 +75,37 @@ export default function Trigger(props) {
...
+        document.activeElement.dataset.pkaAnchor !== "list-filter-input" &&
+        state.isMulti
+      ) {
+        const element = document.getElementById(state.options[0].id);
+        if (element) {
+          document.getElementById(state.options[0].id).focus();
+        }
+      }
+    }, 350); // popover animation + has to wait to check if the search has already the focus :/
+  }, [isOpen, state.options, state.selectedOptions, state.isMulti]);
```
AndreyChernykh, line 107: Should we reset timeouts in the end of the effect?

packages/ListBox/test/specs/multi/ListBox.popover.spec.js
```diff
@@ -52,16 +52,17 @@ describe("ListBox.Popover", () => {
...
 
     openSelect();
     // THIS IS A CRAPPY TEST. The ListBox wait until the popover finish animating
     // to trigger the focus, one we improved the way of the Popover communicated when the animation
     // ends we can improve this test :/
 
     setTimeout(() => {
-      expect(document.activeElement).toBe(getByTestId("popover.content"));
+      expect(document.activeElement.getAttribute("role")).toBe("option");
+      expect(document.activeElement.textContent).toBe("Venus");
```
AndreyChernykh, line 65: Why not sue `waitFor` here, instead of `setTimeout`? 
I bet `setTimeout` is not that robust.

packages/ListBox/test/specs/single/ListBox.popover.spec.js
```diff
@@ -57,7 +57,7 @@ describe("ListBox.Popover", () => {
     setTimeout(() => {
       expect(document.activeElement).not.toBe(getByTestId("popover.content"));
       done();
-    }, 350);
+    }, 450);
```
AndreyChernykh, line 60: Same here ☝️

packages/ListBox/src/components/Options/helpers/options.js
AndreyChernykh: @nahumzs Why not `React.Children.toArray` instead? Would it be more explicit?

packages/ListBox/src/components/Options/helpers/options.js
AndreyChernykh: OK

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: This is a node list, not an array, I'm converting it into an array to be able to use the methods on it unless you mean something different?

packages/ListBox/test/specs/multi/ListBox.popover.spec.js
nahumzs: goo point! we had this test for so long that I forgot about it when I refactor it. Will change it 🙏

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: I purposely don't want to do micro abstractions, they are similar but are not the same. Is easier to understand it if you read the function in isolation you have everything to make sense of what it's happening

packages/ListBox/src/components/Filter/Filter.js
nahumzs: do you mean because will hoist and then the clear function will have access to it?

I thought this way is clear that the value will be used as well on the clear function. or do you mean something else?

.changeset/strong-cups-cry.md
nahumzs: 🤷 no idea what others think lol... cc @allison-c would this be problematic? should we stick to a certain amount of details?

nahumzs: I will update it with your feedback thank you!

packages/ListBox/stories/examples/Keyboard/index.js
mikrotron: I think this makes for an interesting flavour 😋

packages/ListBox/stories/ListBox.keyboard.stories.js
mikrotron: I kinda disagree... this is more like a manual testing story, I'd be tempted to put in in `Backyard/Tests`.

.changeset/strong-cups-cry.md
alexzherdev: there's probably a point where a detailed migration guide becomes too long for a changelog, unless there's a way to do table of contents at the top. I'd be fine with linking out to the full version located elsewhere, as long as the link is prominent in the changelog itself.

packages/ListBox/src/helpers/handleKeyboardKeys.js
```diff
@@ -4,17 +4,13 @@ import useListBox from "../useListBox";
...
-      break;
+  if (event.key === " ") {
+    // prevents spacebar to scroll the page
+    event.preventDefault();
+  }
 
-    default:
-      break;
+  if (["ArrowUp", "ArrowDown"].includes(event.key)) {
+    handleArrowKeys({ event, state, dispatch, onChangeContext });
```
mikrotron, line 13: Maybe this is the place to add `event.preventDefault();` so that the page doesn't scroll?

mikrotron:
I notice when closing a single select `<ListBox>` with `esc` that the focus moves to the _next_ `<ListBox>`.

mikrotron:
> I notice when closing a single select `<ListBox>` with `esc` that the focus moves to the _next_ `<ListBox>`.

Nevermind. When I pulled the latest version of the branch and ran it, this bug has disappeared 👻

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: ohh because are not React children are a Node `DOM` list, the children in the ListBox include ListBox.Trigger, ListBox.Footer, ListBox.Box etc. The `Node List` is just the nodes that have been rendered after ignoring any other kind of children.

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,337 @@
...
+              {characters.heroes}
+              <ListBox.Divider>Villians</ListBox.Divider>
+              {characters.villians}
+            </ListBox>
+          </Card>
+        </Block>
+        <Block>
+          <CheckListItem>
+            The user is allowed to navigate over disabled elements but can not selected them or interact with them. This
+            might required further discussion of what we really want to do.
```
mikrotron, line 215: This works nicely for the multiselect version, but the focus gets stuck and won't pass disabled items for the single version.

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,337 @@
...
+              <ListBox.Divider>Heroes</ListBox.Divider>
+              {characters.heroes}
+              <ListBox.Divider>Villians</ListBox.Divider>
+              {characters.villians}
+            </ListBox>
+          </Card>
+        </Block>
+        <Block>
+          <CheckListItem>
+            The user is allowed to navigate over disabled elements but can not selected them or interact with them. This
```
mikrotron, line 288:
```suggestion
            The user is allowed to navigate over RawItem elements but can not select them or interact with them. This
```

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,337 @@
...
+            might required further discussion of what we really want to do.
```
mikrotron, line 289: Can we apply the same "subtle" focus ring to raw items that we do for disabled items?

```js
${stylers.focusRing.subtle()}
```

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,337 @@
...
+              {characters.heroes}
+              <ListBox.Divider>Villians</ListBox.Divider>
+              {characters.villians}
+            </ListBox>
+          </Card>
+        </Block>
+        <Block>
+          <CheckListItem>
+            When the ListBox is on inline state, the keyboard navigation should behave exactly to any of the previous
+            examples
```
mikrotron, line 331: An inconsistency I noticed: The inline single version doesn't put keyboard focus on the selected item.  It starts on the container, the arrowing down _sets_ the selection to the first element.

Actually, this got me reconsidering the interaction of automatically selecting the focussed item for the single select.  It means for screen reader users there's no way to simply "browse" the options without losing the pre-existing selection.  But changing this behaviour is probably out of scope for this PR, so I'm also okay with leaving the inline version as-is for now.

cc @oscarkwan @sunminsarahlee

packages/ListBox/src/components/Filter/Filter.js
```diff
@@ -67,30 +61,38 @@ const Filter = React.forwardRef((props, ref) => {
...
+      };
+    }
+  }, [applyFilterType, dispatch, isOpen]);
+
+  React.useLayoutEffect(() => {
+    /**
+     * This prevents from scrolling on the input search and moving the popover into a wrong absolute position
+     */
+    function handleScroll() {
+      if (document.activeElement.dataset?.pkaAnchor === "list-filter-input") {
```
mikrotron, line 83: Interesting 🤔 
But if the focus is _not_ on the `list-filter-input`, the `<Popover>` still behaves incorrectly.  It seems more likely to be a `<Popover>` bug though, and I'm not sure we should try to fix it here in the `<ListBox>`, or even in this PR.

nahumzs: I removed this one, now seems stable don't know why 🤷

packages/ListBox/src/helpers/handleKeyboardKeys.js
nahumzs: was in the trigger not in here I added it.

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: 🤷‍♂️ just different flavour of 🍨

packages/ListBox/src/components/Options/helpers/options.js
```diff
@@ -1,7 +1,98 @@
...
+    // the event is coming from the filter
+    HTMLNodeList = event.target
+      .closest("[data-pka-anchor='list-box-box']")
+      .querySelector("[data-pka-anchor='styled-list']");
+  } else {
+    HTMLNodeList = event.target.querySelector("[data-pka-anchor='styled-list']");
+  }
+
+  // filtering only results with role="option" attribute
+  return [...HTMLNodeList.children].filter(li => li.hasAttribute("role"));
```
mikrotron, line 18: Could do `li.getAttribute("role") === "option"` for a more explicit expression that probably wouldn't need the comment. 🤷‍♂️

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,333 @@
...
+        </Block>
+        <Block>
+          <CheckListItem>
+            when trigger has focus and pressing <Up /> or <Down /> the popover menu opens
+          </CheckListItem>
+          <CheckListItem>
+            while the menu is open and there is not previous selection and is single selection, the component should
+            wait for an interaction from the user before selecting something either clicking or <Up />
+            <Down />. <br />
+            In case the ListBox has a multi prop enable, then it should focus on the first element if the component has
```
mikrotron, line 60: I am happy that the single and multi versions now behave the same way!
This sentence can be chopped though ✂️

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,333 @@
...
+          <Toast hasCloseButton={false}>
+            The keyboard navigation should has the same behaviour as the previous examples with few variations, since
+            now includes a filter in both examples.
+          </Toast>
+          <CheckListItem>
+            When there is not an item selected and ListBox hasFilter prop on, it should focus automatically on the
+            filter input after the popover opened, this applies to both case isMulti and single selection
+          </CheckListItem>
+          <CheckListItem>
+            Once the ListBox has at least one selected option, it should focus on it even if the filter exists.
```
mikrotron, line 114: Needs a little adjustment – this describes only the single select version.

nahumzs: There is no action for the multiselect only open the `box`

packages/ListBox/src/components/Options/helpers/options.js
nahumzs: I think is ok :)

packages/ListBox/stories/examples/Keyboard/index.js
```diff
@@ -0,0 +1,333 @@
...
+import Toast from "../../../../Toast/src";
+
+const Card = ({ children }) => <CardComponent style={{ padding: "16px", width: "280px" }}>{children}</CardComponent>;
+const CheckListItem = ({ children }) => {
+  /* eslint-disable jsx-a11y/label-has-associated-control */
+  return (
+    <div style={{ padding: "8px", display: "flex", position: "relative" }}>
+      <label>
+        <div style={{ position: "absolute", top: "10px", left: "-20px" }}>
+          <input type="checkbox" />
```
mikrotron, line 17: To facilitate manual keyboard testing, what about putting `tabindex="-1"` on this?

packages/ListBox/stories/examples/Keyboard/index.js
mikrotron: Exactly!  But this suggests it should focus on on the selected option.  Just needs a little rewording.

packages/ListBox/stories/ListBox.keyboard.stories.js
```diff
@@ -0,0 +1,42 @@
...
+
+export default {
+  title: `${storyName}/Backyard/Keyboard`,
```
mikrotron, line 18:
```suggestion
  title: `${storyName}/Backyard/Tests`,
```

packages/ListBox/stories/ListBox.keyboard.stories.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export function Single() {
```
mikrotron, line 21:
```suggestion
export function Keyboard() {
```

mikrotron: Suggestion is to name this story `Forms/ListBox/Backyard/Tests/Keyboard`

mikrotron:
Minor issues I still noticed while testing:
- Closing the singleselect version with the `esc` key results in the _next_ widget getting keyboard focus.  
- The `<Option.RawItem>` now has the subtle dotted focus ring but still gets the blue background highlight, maybe suggesting it _can_ be selected to the user.
- The `inline` version of the singleselect doesn't focus the selected item ([as mentioned](https://github.com/acl-services/paprika/pull/973#discussion_r594771851)) but not sure how to properly handle this without discussing further.

I think we should capture these, but not necessarily fix them in this PR, which is nearly ready to merge.

packages/ListBox/stories/ListBox.keyboard.stories.js
nahumzs: Why would this be better?

mikrotron: It seems to fit better.  The instructions are great – they really facilitate a manual test.  But it's still a testing page.

nahumzs: I settled for keyboarTesting just naming test something is hard to find on the search menu of storybook.

nahumzs:
> The <Option.RawItem> now has the subtle dotted focus ring but still gets the blue background highlight, maybe suggesting it can be selected to the user.

Does the RawOption should appears on the search? this is something I couldn't answer if it's raw normally would be to have a second action therefore I think should be shown on the filter, currently, they are but we can change that behaviour.

mikrotron (approved):
llgtm 111

PR #978, opened 2021-03-18
UXD-972 move refLabel out from render content param…

allison-c:
…ater object

### Purpose 🚀

Resolve the errors coming from `FormElement`, the `refLabel` is only required by `ListBox` but it has been passed down to all type of field, e.g. input or a textarea, so there's react warning about invalid DOM attribute

The PR changes the parameter of the render `Content` function, but I think since only `ListBox` needs to update the way of using `refLabel` we can treat it as a patch (bug fix) bump for the form element. Also I believe even if I didn't change to this way, the refLabel doesn't work on the current (master branch) form element + listbox example

master branch: http://storybooks.highbond-s3.com/paprika/master/?path=/story/forms-formelement-examples--list-box-story not working 
this branch: http://storybooks.highbond-s3.com/paprika/UXD-972-move-refLabel-out/?path=/story/forms-formelement-examples--list-box-story works

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/FormElement/stories/examples/ListBox.js
+++ b/packages/FormElement/stories/examples/ListBox.js
@@ -13,8 +13,9 @@ export default function ListBoxExample() {
       <FormElement>
         <Label>Form Label</Label>
         <Content>
-          {a11yProps => (
-            <ListBox {...a11yProps}>
+          {(a11yProps, refLabel) => (
+            <ListBox>
+              <ListBox.A11y {...a11yProps} refLabel={refLabel} />
               {optionsArray.map(item => (
                 <ListBox.Option key={item}>{item}</ListBox.Option>
               ))}
--- a/.changeset/wicked-flowers-visit.md
+++ b/.changeset/wicked-flowers-visit.md
@@ -1,0 +1,5 @@
+---
+"@paprika/form-element": patch
+---
+
+Move `refLabel` out from `a11yProps` parameter when rendering content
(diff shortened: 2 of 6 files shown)


packages/FormElement/stories/examples/ListBox.js
```diff
@@ -13,8 +13,9 @@ export default function ListBoxExample() {
       <FormElement>
         <Label>Form Label</Label>
         <Content>
-          {a11yProps => (
-            <ListBox {...a11yProps}>
+          {(a11yProps, refLabel) => (
```
nahumzs, line 16: does this mean that all other elements except for the ListBox don't need the refLabel reference?

nahumzs: what about 
`const { refLabel, ...moreAllyProps } = allyProps;`

Wouldn't that be enough and you wouldn't need to modify the FormElement component?

allison-c: > does this mean that all other elements except for the ListBox don't need the refLabel reference?

Yes, it's not a valid DOM attribute, and only being used by ListBox

> const { refLabel, ...moreAllyProps } = allyProps;

This can work, but all consumers will need to to this in all other type of input(input, textarea, checkbox...), except `ListBox`, if they don't do it, they'll get console.error in their app too

nahumzs: ohh I see, ok, make sense to exclude the ListBox and force to destructure it in the component 👍

nahumzs (approved):

PR #979, opened 2021-03-18
Uxd 977 a11y tests

oscarkwan:
### Purpose 🚀
- add `jest-axe`
- add a11y checks to most components
- there is one failure with listbox, right now its an exception. Will create another task to address it

### References 🔗


nahumzs:
great 30% more coverage in accessibility :D

nahumzs (approved):

allison-c:
Is it possible to write a reusable one for 😂 
```js
  it("should not fail any accessibility tests", async () => {
    const { container } = renderComponent();
    expect(await axe(container)).toHaveNoViolations();
  });
  ```
  
  
  ```js
const axeSpec = (name, container) => {
  describe(`axe spec for ${name}`, () => {
    it("should not fail any accessibility tests", async () => {
    expect(await axe(container)).toHaveNoViolations();
  });
  });
}
  ```

PR #980, opened 2021-03-18
refactor(ListBoxWithTags): UXD-973 clean up warnings and errors in ListBoxWithTags

allison-c:
…stBoxWithTags

### Purpose 🚀

Currently the hook `useListBoxWithTags` returns some helper function and data mixed with the props can be passed into the component, if the consumer doesn't explicitly extract the data and helper function out, they'll be passed into the <ListBoxWithTags /> and the component will pass them as `moreProps` again to the `<div />`  element. Not only happens in consuming apps, even in paprika storybook and specs, we also have lots of console.errors because of that.

I think we need to separate the valid props from the return value of `useListBoxWithTags`

I understand this is maybe not an urgent major bump, we can merge this PR with the next major bump

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-973-cleanup-errors-in-listboxwithtags

### Screenshots 📸

### References 🔗


nahumzs:
ohhh I like this!!!, totally my fault! thank you for fixing it, I will work on Listbox, listboxWithTags and userlookup next week, will merge this and bump all till userlookup :) 👏

packages/ListBoxWithTags/src/useWithTags.js
```diff
@@ -102,15 +102,17 @@ export default function useWithTags({
   }
 
   return {
+    listBoxWithTagsProps: {
```
tristanjasper, line 105: nit: just wonder if this may be better being called `spreadableMoreProps` or `validMoreProps`? 🤔

allison-c: I thought with the name `listBoxWithTags`  it's easier to understand where we should put these props in

nahumzs (approved):

Issue #310, opened 2019-11-26
Tabs component tabs keyboard focus issues


tristanjasper:
# Bug Report

As recommended in https://www.w3.org/TR/wai-aria-practices/#kbd_disabled_controls

1. Tabs should still be focusable even when disabled (same as menu items in a menu or options in a listbox)

2. Currently using arrow keys changes the focus to the next tab, however pressing shift tab sets focus to the first tab. Pressing the right arrow after this doesn't then focus the tab next to it as expected.

## Prerequisites
**This is just a checklist, please delete this section after**
Please answer the following questions for yourself before submitting a bug report.
- [ ] I checked to make sure that this bug has not already been filed
- [ ] I reproduced the bug in the latest version of paprika
- [ ] Please label as, low / medium / high priority

*Please fill in as much detail as possible below*

## Expected behavior
*Please describe the behavior you are expecting to happen.*

## Current behavior
*What is the current behavior?*

## Screenshots / Gifs / Codepens

![tabs-keyboard-focus](https://user-images.githubusercontent.com/4040102/69683004-287ac880-1068-11ea-9e32-8762cd35d72b.gif)

## Additional context
*Please provide any more additional details to illustrate the bug*

mikrotron:
I believe this functionality is correct in the `<ButtonGroup>` and it might make the most sense to extract to logic to a custom hook that can be shared.

PR #981, opened 2021-03-19
UXD-982 Tabs keyboard improvements 🐐

mikrotron:
### Purpose 🚀
Ensure proper keyboard usage patterns including focussing on disabled tabs.

### Notes ✏️
- Refactored keyboard usage patterns – allow focus on disabled tabs and ensure focus movement and user interactions adhere to expected standards.
- Disabled tabs have a subtle dotted focus ring instead of a blue one.
- Applied expected ARIA attributes throughout as per [WAI-ARIA authoring practices for Tabs widget](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)
- Added support for `↓` and `↑` keys to navigate tabs (especially for vertical tabs)
- Refactored Cypress tests – combined some tests and increased coverage.
- **BONUS:** Also added capability for tab list to be rendered with no active tab by setting `index` or `defaultIndex` to `null`.

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-982--tabs-keyboard-a11y/?path=/story/navigation-tabs-backyard-tests--keyboard

### Screenshots 📸
New Backyard/Tests/Keyboard story
Screen Shot 2021-03-18 at 5 04 07 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-982
Resolves https://github.com/acl-services/paprika/issues/310


packages/Tabs/src/Tabs.js
```diff
@@ -26,73 +25,96 @@ export default function Tabs(props) {
...
+      event.stopPropagation();
+
+      const itemIndexes = getItemIndexes(refList);
+      const enabledSelectedIndex = itemIndexes.indexOf(focusIndex);
+      const count = itemIndexes.length;
+
+      switch (event.key) {
+        case keyTypes.RIGHT:
+        case keyTypes.DOWN:
+          focusItem(itemIndexes[(enabledSelectedIndex + 1) % count]);
```
nahumzs, line 85: this will loop over the tabs? most HTML inputs don't loop over when using the keyboard any reason for this one to do it?

mikrotron: This is the behaviour specified in the WAI ARIA Authoring Practices:
> Right Arrow: Moves focus to the next tab. If focus is on the last tab element, moves focus to the first tab. 

Source: https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-19

I think this pattern makes more sense for a tabs control because the number of tabs is relatively small.

alexzherdev:
I think something's not working:
1. open the story
2. click fourth tab (close the new browser tab)
3. from here, left arrow takes you to fifth tab; right arrow takes you to second tab. As if the first tab was the one currently focused.

mikrotron:
> I think something's not working:
> 
> 1. open the story
> 2. click fourth tab (close the new browser tab)
> 3. from here, left arrow takes you to fifth tab; right arrow takes you to second tab. As if the first tab was the one currently focused.

Good catch! Thanks!

oscarkwan:
Can we put somewhere in the Readme, or stories to indicate that active tabs don't get rendered?

packages/Tabs/src/Tabs.js
```diff
@@ -26,73 +25,96 @@ export default function Tabs(props) {
...
-  let tabListRef = React.useRef(null);
+  const [focusIndex, setFocusIndex] = React.useState(indexToUse || 0);
+  const refList = React.useRef(null);
+  const [idTabs] = React.useState(uuidv4());
 
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
     setActiveIndex(indexToUse);
-  }, [indexToUse, setActiveIndex]);
+  }, [indexToUse]);
```
AndreyChernykh, line 45:
```js
const [activeIndex, setActiveIndex] = React.useState(defaultIndex >= 0 ? defaultIndex : null);
const indexToUse = isControlled : index ? activeIndex;
```

No need to have an effect here.

See https://wegalvanize.slack.com/archives/CB3QRPDJ4/p1616523810022900?thread_ts=1616522183.021000&cid=CB3QRPDJ4

packages/Tabs/src/Tabs.js
```diff
@@ -26,73 +25,96 @@ export default function Tabs(props) {
...
+    if (indexToUse !== null && items.length > 0 && isItemDisabled(items[indexToUse])) {
+      const itemIndexes = getItemIndexes(refList, { hasDisabledItems: false });
 
-  function focusAndSetIndex(index) {
-    tabListRef.querySelectorAll("[data-pka-anchor='tab'], [data-pka-anchor='tab-link']")[index].focus();
+      if (itemIndexes.length > 0) {
+        setActiveIndex(itemIndexes[0]);
+      }
+    }
+  }, []);
```
AndreyChernykh, line 56: This one is more tricky because it uses `refs` that are calculated after render. 

But, I guess if the component is controllable, this calculation will not affect the component's behaviour at all (because we'll use `index` instead of `activeIndex` instead)

AndreyChernykh: Also, I'd maybe rename `activeIndex` to make it more explicit that it is only used if the component is uncontrollable.

packages/Tabs/src/components/Panels/Panels.js
```diff
@@ -1,23 +1,29 @@
...
+  const { activeIndex, idTabs } = React.useContext(TabsContext);
 
-  const childrenWithProps = React.Children.map(children, (panel, index) => {
-    if (!panel) return;
+  const childrenWithProps = React.Children.toArray(children)
+    .filter(child => {
+      return child !== null && child.type.displayName === Panel.displayName;
+    })
+    .map((panel, index) => {
+      if (!panel) return;
```
AndreyChernykh, line 19: should it `return null` instead?

packages/Tabs/src/Tabs.js
mikrotron: Yeah, this is basically ensuring we don't set the initially selected tab to a disabled tab.  Pretty edge-casey actually and would probably benefit from a function name that describes what it's doing.

packages/Tabs/src/Tabs.js
```diff
@@ -26,73 +25,95 @@ export default function Tabs(props) {
...
-  const [currentFocusIndex, setFocusIndex] = React.useState(indexToUse);
-  let tabListRef = React.useRef(null);
+  const [idTabs] = React.useState(uuidv4());
+  const refList = React.useRef(null);
 
-  React.useEffect(() => {
-    setActiveIndex(indexToUse);
-  }, [indexToUse, setActiveIndex]);
+  const isControlled = index !== undefined;
+  const initIndex = defaultIndex === undefined ? 0 : defaultIndex;
```
mikrotron, line 32: This should mimic the previous behaviour with the default value for `defaultIndex` of `0`.  This is why I'm not considering this a breaking change.

packages/Tabs/src/Tabs.js
```diff
@@ -116,7 +137,7 @@ Tabs.propTypes = {
   /** Tab labels will be truncated when they run out of space instead of breaking to multiple lines (ignored when isVertical is false). */
   hasTruncation: PropTypes.bool,
 
-  /** Sets what tabindex is active (controlled component). */
+  /** Sets what tabindex is active (controlled component). Use null for no active tab. */
```
tristanjasper, line 140: Imo this prop should be called named `activeIndex`

packages/Tabs/src/Tabs.js
```diff
@@ -101,13 +122,13 @@ export default function Tabs(props) {
...
-  kind: PropTypes.oneOf([Tabs.types.kind.PRIMARY, Tabs.types.kind.SECONDARY]),
+  /** Description of the purpose of the tabs for assistive technology. */
+  a11yText: PropTypes.string,
 
   /** Expects Tabs.List and Tabs.Panels. */
   children: PropTypes.node.isRequired,
 
-  /** Sets what tabindex is active by default (uncontrolled component). */
+  /** Sets what tabindex is active by default (uncontrolled component). Use null for no active tab. */
   defaultIndex: PropTypes.number,
```
tristanjasper, line 132: also should be `defaultActiveIndex` to make things clearer imo

packages/Tabs/src/components/List/List.js
```diff
@@ -1,46 +1,58 @@
...
-      hasTruncation,
-      tabHeight,
-      isSelected: activeIndex === index,
-      isVertical,
-      onClick: e => handleClickTab(e, index),
-      onKeyDownArrows: onKeyDown,
-      size,
+  const childrenWithProps = React.Children.toArray(children)
+    .filter(child => {
+      return child !== null && child.type.displayName === Tab.displayName;
```
tristanjasper, line 27: both `filter` and `map` returns can be on one line

packages/Tabs/src/components/Panels/Panels.js
```diff
@@ -1,23 +1,29 @@
...
-  const context = React.useContext(TabsContext);
-
   const { children, ...moreProps } = props;
+  const { activeIndex, idTabs } = React.useContext(TabsContext);
 
-  const childrenWithProps = React.Children.map(children, (panel, index) => {
-    if (!panel) return;
+  const childrenWithProps = React.Children.toArray(children)
+    .filter(child => {
+      return child !== null && child.type.displayName === Panel.displayName;
```
tristanjasper, line 16: return on one line

AndreyChernykh:
LGTM

packages/Tabs/src/Tabs.js
mikrotron: I agree. But not interested in breaking the API for this. 🤷‍♂️

allison-c (approved):

PR #982, opened 2021-03-19
Uxd 986 panel without overlay focus issue

allison-c:
### Purpose 🚀

See the bug description in: https://aclgrc.atlassian.net/browse/UXD-985

### Notes ✏️

Prevent popovers to be closed when the panel(without overlay) is opening.
try in http://storybooks.highbond-s3.com/paprika/UXD-986-panel-without-overlay-focus-issue/?path=/story/messaging-panel-examples--focus-lock

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/Panel/src/helpers.js
```diff
@@ -53,3 +53,9 @@ export function warnOfPropErrors(props) {
     );
   });
 }
+
+export function isElementInsidePaprikaPopover(element) {
```
tristanjasper, line 57: Not sure we should be relating the Panel Component to the Popover in this way. Wonder if this kind of thing can be agnostic of the component type/name

tristanjasper: Perhaps could use something like the function I used in https://github.com/acl-services/paprika/pull/984/files#diff-881a787b5655895f8685a371e2d800027c5b395cad9574a2faf72dea1e418287R30 if we really need to know if it is a descendant of that element

allison-c:
Close this PR, since the solution should be automatically disabling the focus lock inside paprika.
If you need a fix now, you can use `Panel.FocusLock` component to disable it

Issue #601, opened 2020-07-10
Popover - pressing esc closes parent Takeover, Modal, SidePanel


mikrotron:
## Expected behavior
When a Popover (or any other component that is composed of a Popover) is open and the user presses `esc`, the Popover should close, but the modal component (Takeover, Modal, or SidePanel) should remain open.

## Current behavior
Pressing `esc` to dismiss a Popover will _ALSO_ close the Takeover, Modal, or SidePanel.  

## Screenshots / Gifs / Codepens
Behaviour can be observed in this Storybook story:
https://paprika.highbond.com/?path=/story/messaging-takeover-backyard-sandbox--with-nested-popover

## Additional context
The modal components' Overlay component is catching this keyUp event before the one on the document that the Popover attaches.  The keyUp event for the Popover should probably be attached to it's Content component.

nahumzs:
We can create a custom event so each component can dispatch its own even therefore even if all other components subscribed to the esc event listen to it, only those that dispatch the custom event act on it.

PR #984, opened 2021-03-22
Uxd 980 fix escape key closes multiple popup and overlays

tristanjasper:
### Purpose 🚀
Ensure Esc action only closes immediate popup or modal that has focus

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-979-fix-escape-key-overlay-popup

### Screenshots 📸

### References 🔗
Ticket: https://aclgrc.atlassian.net/browse/UXD-980
Resolves: https://github.com/acl-services/paprika/issues/601


packages/Modal/tests/cypress/Modal.cypress.js
```diff
@@ -0,0 +1,14 @@
+import { getStoryUrlPrefix } from "../../../../.storybook/storyTree";
```
tristanjasper, line 1: created cypress tests as rtl won't be able to check dom tree with isDescendant helper

packages/Popover/src/Popover.js
```diff
@@ -244,7 +249,8 @@ class Popover extends React.Component {
   }, throttleDelay);
 
   handleKeyUp = event => {
-    if (event.key === "Escape") {
+    if (event.key === "Escape" && isDescendantOfActiveElement()) {
```
alexzherdev, line 252: Seeing an inconsistency between this and the overlay component above in checking `isOpen`. Kinda makes sense to check for it, but also likely impossible to get here if it's not open. Up to you whether to change it or not

packages/Popover/src/Popover.js
```diff
@@ -26,6 +27,10 @@ const focusableElementSelector =
 // TODO: To handle cases where there are multiple scrolling containers, we need to implement
 //       getScrollContainer as oneOfType([func, arrayOf(func)])
 
+function isDescendantOfActiveElement() {
+  return isDescendant(document.activeElement, "data-pka-anchor", "popover.content");
```
alexzherdev, line 31: Both this case and Overlay don't uniquely describe the instance of a component (e.g. there can be many popovers on the page). I can't think of a specific scenario where this would be an issue but thought I'd note it here.

packages/helpers/src/dom/isDescendant.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable no-param-reassign */
+/* eslint-disable no-cond-assign */
+export default function isDescendant(el, attr, value) {
```
alexzherdev, line 3: This level of abstraction is neither here nor there IMO. Not feeling strong but if putting this into helpers, might as well do
```js
export default function isDescendant(el, callback) {
  let isChild = callback(el);
  while (!isChild && (el = el.parentNode)) {
    isChild = callback(el);
  }
  return isChild;
}

isDescendant(document.activeElement, (el) => el.getAttribute("data-pka-anchor") === "popover.content");
```
(the name could use some improvement too)

packages/Modal/tests/cypress/Modal.cypress.js
tristanjasper: Oh I thought RTL would not have the dom tree to navigate... 🤔

packages/Popover/src/Popover.js
tristanjasper: Yes my thinking was that we only should have one popover open and taking focus at any time, and opening one should always close another.. At least that is how it Should be :)

packages/Modal/tests/cypress/Modal.cypress.js
alexzherdev: it uses an in-memory DOM representation via https://github.com/jsdom/jsdom

packages/Overlay/src/Overlay.js
```diff
@@ -109,7 +112,7 @@ const Overlay = props => {
           onExited={onAfterClose}
         >
           {state => (
-            <sc.Overlay data-pka-anchor="overlay" {...moreProps} onKeyDown={handleEscKey}>
+            <sc.Overlay id="overlay" data-pka-anchor="overlay" {...moreProps} onKeyDown={handleKeyDown}>
```
nahumzs, line 115: 🤔 could be possible to be more than one overlay in the same page??

tristanjasper: I'll check how this works when we have a Panel open as well as Modal showing, there could be an issue there, so will create a story for that case

packages/helpers/src/dom/isDescendant.js
```diff
@@ -0,0 +1,10 @@
...
+  const doesElementHaveAttribute = el => el?.getAttribute(attr) === value;
+  let isChild = doesElementHaveAttribute(el);
```
nahumzs, line 5: this function tries to find the parent of a node child but does it with a too specific method IMHO. 

Maybe doesn't worth having it at a helper package level.

Also worth noting that in the next line

in this line https://github.com/acl-services/paprika/pull/984/files#diff-fc157e57b35fbc14e972a59db1410d976b2f4c7bea94f3d29be07751b64ba690R115

you are assuming that only one overlay will ever exist in a webpage, if that is the case then there is no case for this method.

```
document.getElementById('overlay').contains(element)
```

Will find if the dom is a descendant of an overlay.

I would suggest that maybe you can use an anchor instead of an id.

`data-pka-anchor="overlay"`

```
document.querySelector("[data-pka-anchor='overlay']").contains(element)
```

packages/Popover/src/Popover.js
nahumzs, line 31:
```
document.getElementById('overlay').contains("[data-pka-anchor='popover.content']")
```

assuming only one overlay exists per page

mikrotron: I don't think it's safe to assume only one overly exists, but if it were you should be able to use this existing `helper` function:

https://github.com/acl-services/paprika/blob/master/packages/helpers/src/dom/isElementContainsFocus.js

For symmetry, maybe the helper you can add can be `isFocusContainedBy(isContainer)` 

where `isContainer = (el) => el.getAttribute("data-pka-anchor") === "popover.content"`

packages/Popover/src/Popover.js
```diff
@@ -447,7 +448,13 @@ class Popover extends React.Component {
     return (
       <ThemeContext.Provider value={isDark}>
         <PopoverContext.Provider value={contextValue}>
-          <sc.Popover data-pka-anchor="popover" {...moreProps} ref={this.$popover}>
+          <sc.Popover
+            data-pka-anchor="popover"
+            {...moreProps}
+            ref={this.$popover}
+            onKeyUp={this.handleKeyUp}
+            onKeyDown={this.handleKeyDown}
```
mikrotron, line 456: What if the key listeners are put on the Popover.Content (via context)? Could you avoid all the DOM checking and just _know_ the event was for _this_ `<Popover>` set `isOpen=false` and stop the propagation?

packages/Popover/src/Popover.js
nahumzs: in my previous comments, I pointed out exactly what you are pointed out to me lol...

packages/helpers/src/dom/isDescendant.js
tristanjasper: Sidenote: Because the TakeOver overrides the Paprika Overlay props, which includes the `data-pka-anchor` I decided not to use it (https://github.com/acl-services/paprika/blob/master/packages/Takeover/src/Takeover.js#L66) and used an `id`.

packages/Popover/src/Popover.js
tristanjasper: Yep, I'm aware of that and looking into it as mentioned in comment https://github.com/acl-services/paprika/pull/984#discussion_r599126881

tristanjasper:
Closing as taking a different approach

PR #987, opened 2021-03-25
refactor(Popover): UXD-980 Esc key only closes popover

tristanjasper:
### Purpose 🚀

This is a follow on Pr from original approach (https://github.com/acl-services/paprika/pull/984)

Pressing Esc key when popover is opened now only closes the popover and event propagation is stopped.

Todo: Ensure works with listbox / date picker etc forms-listbox-backyard-keyboardtesting--keyboard expectations

Ticket: https://aclgrc.atlassian.net/browse/UXD-980
Resolves: #601

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/uxd-980-popover-esc-key-close-popover-only

### Screenshots 📸

### References 🔗


packages/Popover/src/Popover.js
```diff
@@ -441,13 +431,14 @@ class Popover extends React.Component {
...
-      this.ariaIdForContent
+      this.ariaIdForContent,
+      this.$trigger
     );
 
     return (
       <ThemeContext.Provider value={isDark}>
         <PopoverContext.Provider value={contextValue}>
-          <sc.Popover data-pka-anchor="popover" {...moreProps} ref={this.$popover}>
+          <sc.Popover data-pka-anchor="popover" {...moreProps} ref={this.$popover} onKeyDown={this.handleKeyDown}>
```
tristanjasper, line 441: Could move the `onKeyDown` functionality into the `Content` sub component as well but think that should be part of a large refactor

packages/Popover/src/components/Content/Content.js
```diff
@@ -102,7 +112,7 @@ const Content = React.forwardRef((props, ref) => {
       onMouseOut={handleMouseEvent}
       onMouseOver={handleMouseEvent}
       style={contentStyles}
-      tabIndex="-1"
+      tabIndex="0"
```
tristanjasper, line 115: Required this change in order to receive keyboard events

packages/Popover/src/Popover.js
nahumzs, line 441: 🤔 I can't find if it's allowed or not to use `onKey` event without using a tabIndex or aria roles attributes, also wondering if this can be hurtful if someone wants to take control of the Popover onKeyDown event over the Popover, with this approach is not possible.

Also if you decided to go with this approach maybe would be better to use onKeyUp.

packages/Popover/src/components/Content/Content.js
nahumzs: <ContentStyled /> is `Popover.Content` 🤔 how this will mess with the tabbing when the popover is used with the ListBox DatePicker and all those components that depend on this to attach their onKeyDown events?

have you test the keyboard page for the list box and see if this is affecting it?

nahumzs: also, you are attaching a new `onKeyUp={handleKeyUp}` but will be overridden by the moreProps, which means this will only work if we don't compose this component with `onKey` events.

nahumzs: https://github.com/acl-services/paprika/blob/948aad9abcd0d50ee36edd8d48148878a0292547/packages/ListBox/src/components/Content/Content.js#L67-L73

nahumzs: I quickly tested http://storybooks.highbond-s3.com/paprika/uxd-980-popover-esc-key-close-popover-only/?path=/story/forms-listbox-backyard-keyboardtesting--keyboard and seems all is still working but don't know if this implementation of the `esc` just closing the popover will work with the listbox maybe would worth adding an example?

packages/Popover/src/Popover.js
tristanjasper: I was getting rid of the `keyDown` listener on document object and putting it on the component, thereby not need to do event cleanup (https://github.com/acl-services/paprika/pull/987/files#diff-881a787b5655895f8685a371e2d800027c5b395cad9574a2faf72dea1e418287L404) as unmounting will take care of this...

packages/Popover/src/components/Content/Content.js
mikrotron: Yeah, this will change the keyboard use patterns, and I don't think it's needed.  Seems to still work with `tabIndex="-1"`

mikrotron: Nahum' has a good point about `moreProps` overriding handlers, etc on the Popover.Content.  This might be a good opportunity to move `moreProps` into a more deliberate order so that can't happen.  

However, it's also true that we might _want_ to add custom handlers, in which case we should compose them with the ones we define to make sure the Popover doesn't break.  There's a `callAll()` function in `@paprika/helpers`, like the one from the Epic React course, that could help with this.

mikrotron:
As I test this a bit more, I realize we probably need to pass the `handleKeyDown` to the Popover.Trigger too.  For example, the tooltip should be dismissed with `esc` too, but in that case the focus remains on the trigger.  Also, when the Popover has the `shouldKeepFocus` prop (https://paprika.highbond.com/?path=/story/messaging-popover-examples--should-keep-focus-story) we should also be able to dismiss the Popover with `esc`.

.changeset/good-swans-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paprika/popover": minor
+---
+
+When popover opened pressing ESC key only closes popover content and stops propagation. Benefical if using the popover in a Panel or TakeOver component
```
mikrotron, line 5: Would be good to include a Storybook story to illustrate a Popover in a Modal/Takeover/Panel to verify this desired behaviour.

packages/Popover/src/components/Content/Content.js
tristanjasper: So i'm seeing that the down/up arrow isn't preselecting the listbox options in /forms-listbox-backyard-keyboardtesting--keyboard story so this will need some more understanding of what list box is expecting.

tristanjasper: ☝️  This issue should be fixed now

.changeset/good-swans-reflect.md
tristanjasper: There is this example http://storybooks.highbond-s3.com/paprika/master/?path=/story/messaging-takeover-backyard-sandbox--with-nested-popover. I'll add one for the Panel as well

nahumzs: http://storybooks.highbond-s3.com/paprika/master/?path=/story/messaging-takeover-backyard-sandbox--with-nested-popover

Went to that example, I'm clicking `esc` after opening the popover should I expect only the popover to close? because right now is closing even the takeover.

I even open the popover click tab, focus on the link, click `esc`, and still close popover plus takeover.

Is this expected?

packages/Popover/README.md
```diff
@@ -50,10 +50,11 @@ npm install @paprika/popover
...
 
-| Prop     | Type | required | default  | Description                                  |
-| -------- | ---- | -------- | -------- | -------------------------------------------- |
-| children | node | false    | null     |                                              |
-| onBlur   | func | false    | () => {} | Callback to indicate the element loses focus |
+| Prop      | Type | required | default  | Description                                  |
+| --------- | ---- | -------- | -------- | -------------------------------------------- |
+| children  | node | false    | null     |                                              |
+| onBlur    | func | false    | () => {} | Callback to indicate the element loses focus |
+| onKeyDown | func | false    | () => {} |                                              |
```
nahumzs, line 57: Let's not expose this one, don't declare it on the props, this gives the wrong impression about where to put the onKeyDown when composing the Popover. 

Normally when someone wants to use the onKeyDown and onKeyUp want to do it over the `<Popover.Content />` which gets the focus after the popover is open not to the `<Popover>` itself. 

also, all of our components should be able to spread this onClick, onKeyDown etc. So should be obvious that we can pass down those type of Props. 

`<sc.Popover data-pka-anchor="popover" {...moreProps} ref={this.$popover} onKeyDown={this.handleKeyDown}>`

.changeset/good-swans-reflect.md
mikrotron: That `s3` link seems stale. I'm trying it out locally and it seems okay.

packages/Popover/README.md
tristanjasper: Ok will just call `moreProps.onKeyDown` when needed to avoid declaring it on props

packages/Popover/src/Popover.js
```diff
@@ -441,13 +437,19 @@ class Popover extends React.Component {
...
 
     return (
       <ThemeContext.Provider value={isDark}>
         <PopoverContext.Provider value={contextValue}>
-          <sc.Popover data-pka-anchor="popover" {...moreProps} ref={this.$popover}>
+          <sc.Popover
+            data-pka-anchor="popover"
+            ref={this.$popover}
+            {...moreProps}
+            onKeyDown={callAll(this.handleKeyDown, moreProps.onKeyDown)}
```
tristanjasper, line 451: @nahumzs What do you think about the order of these two functions being called, when we use this order we are ensuring the Popover has the chance to catch the keyboard events first, rather than the Component that spread `onKeyDown` to the popover (eg the Listbox). I think it makes sense this way.

This is the same for line https://github.com/acl-services/paprika/pull/987/files#diff-8c0224cf91d59c94783b6e753545719c6e30a688f5ae2191f531e937f7309a56R107

packages/Popover/src/Popover.js
```diff
@@ -243,15 +247,12 @@ class Popover extends React.Component {
     }
   }, throttleDelay);
 
-  handleKeyUp = event => {
-    if (event.key === "Escape") {
+  handleKeyDown = event => {
+    if (event.key === "Escape" && this.isOpen() && this.$trigger) {
```
tristanjasper, line 251: Perhaps could just put put `this.isOpen() && this.$trigger` as a boolean variable at the top of the function

mikrotron: Hey, this is _your_ PR, man! 😜

mikrotron (approved):
llgtm 111

packages/Popover/src/Popover.js
tristanjasper: Ha, OUR pr now. I thinks it fine to leave

packages/Popover/src/Popover.js
nahumzs: I think is ok 👍

packages/Popover/src/Popover.js
nahumzs: I normally prefer having most of the code where people can find it before splitting it into variable and files, just move thing out or abstract to a variable when is clear there is duplication or abstraction is need it.

PR #1001, opened 2021-04-07
feature(TextEditor): UXD-1031 support for advance editor

nahumzs:
### Purpose 🚀
add support for rendering advance mode:

Screen Shot 2021-04-07 at 3 24 21 PM
Screen Shot 2021-04-07 at 3 24 28 PM

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1031-Render-Advance-Option

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/TextEditor/src/TextEditor.js
+++ b/packages/TextEditor/src/TextEditor.js
@@ -55,7 +55,6 @@ export default function TextEditor(props) {
   const { height, isDisabled, apiKey, onChange, value, defaultValue, kind, children } = props;
 
   // requires to reset the component with newer props due the tinyMCE uncontrolled nature
-  const [idKey, setNextIdKey] = React.useState(0);
   const { "TextEditor.Tiny": tiny } = extractChildren(children, ["TextEditor.Tiny"]);
 
   if (value && defaultValue) {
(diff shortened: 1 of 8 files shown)


packages/TextEditor/src/TextEditor.js
```diff
@@ -67,16 +73,20 @@ export default function TextEditor(props) {
 
   const initExtended = (tiny?.props && tiny.props?.init) || {};
 
+  React.useEffect(() => {
+    setNextIdKey(prev => prev + 1);
+  }, [kind]);
+
   return (
     <Editor
+      key={idKey}
```
AndreyChernykh, line 82: Wont the `key={kind}` be enough, though? 
It will re-mount the component if `kind` changes from SIMPLE to ADVANCE, and vice-versa

nahumzs: yeap true, but at this moment don't know if we want to also check if `init` from `<TextEditor.Tiny />` should change as well, or any other prop where the TextEditor needs to rerender. 

So if that is the case is easier to add a dependency on the next key and will keep working and escalating easily.

AndreyChernykh: @nahumzs If you add `init` as a dependency it will probably trigger the `effect` all the time, unless it is wrapped in memo (and even then it is no a guarantee)

I'd go with the simplest solution that solves the task right now `key={kind}` and then will expand it later if required. If requirements change, you'll have to update `TextEditor.js` anyway

AndreyChernykh: Generally, I try to avoid using effect unless it is the only option to solve the task (e.g. when you need to send an HTTP request)

nahumzs: k, I think in this case useEffect is the only solution. unless we don't allow switching between kinds, but I think we should.

nahumzs: k, lets go with the simplest solution for now then

AndreyChernykh (approved):

Issue #183, opened 2019-09-04
DropdownMenu – Add support for an anchor attribute to identify Popover portal content


mikrotron:
## Problem
Since the menu content of the `<DropdownMenu>` is rendered in a portal, when there are multiple `<DropdownMenus>` on a page, there will be no unique identifier in the markup to target it with automation tests, CSS overrides or scripts.  

The `<Popover.Content>` does propagate additional props, so a `data-testid` attribute _can_ be included on the portal DOM element, but the current API for the `<DropdownMenu>` does not expose access to its internal `<Popover.Content>` element.  

Simply adding a `data-pka-anchor='dropdown-menu.popover-content'` will be insufficient, since all DropdownMenu portals would share the same attribute.  Additionally, Paprika components have thus far remained agnostic about testing attribute names, so naming a prop is a challenge.

## Suggested Solution

Since it may be convenient to capture and propagate various props for both the `<Popover>` and `<Popover.Content>` that are implemented in the `<DropdownMenu>`, a similar pattern that is used in the `<DataPicker>` and `<ListBox>` could be employed to extract props from a "container" component. 

In fact the `extractChildrenProps()` helper could be extracted from `<DatePicker>` to the `Helpers` package to be shared:

https://github.com/acl-services/paprika/blob/master/packages/DatePicker/src/helpers.js#L4-L14

The resulting API would look like:
```jsx
<DropdownMenu>
  <DropdownMenu.Content data-testid="special-menu" />
  <DropdownMenu.Item>Menu item</DropdownMenu.Item>
  ...
</DropdownMenu>
```

mikrotron:
This issue now applies to the `<Confirmation>` component as well.

_However_, since only ONE `<Popover>` should ever be visible (rendered) at a time, a workaround is to simply locate the only `[data-pka-anchor='popover.content']` element on the page.

jamiek-acl:
Donezo.
https://github.com/acl-services/paprika/pull/1002

PR #1002, opened 2021-04-08
UXD-1048 - Confirmation prop propagation

jamiek-acl:
### Purpose 🚀

Ticket: https://aclgrc.atlassian.net/browse/UXD-1048

This is to address this issue: https://github.com/acl-services/paprika/issues/183

You can now drill props to the `Popover.Content` of the `Confirmation` component.
I also updated the example of the `OverflowMenu`, to demonstrate it supports the same behaviour.

Screen Shot 2021-04-08 at 11 49 13 AM

### DIFF
--- a/packages/Confirmation/src/components/Content/Content.js
+++ b/packages/Confirmation/src/components/Content/Content.js
@@ -1,7 +1,0 @@
-import React from "react";
-
-export default function Content() {
-  return <React.Fragment />;
-}
-
-Content.displayName = "Confirmation.Content";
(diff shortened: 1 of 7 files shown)


packages/Confirmation/src/components/Content/Content.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+export default function Content() {
```
mikrotron, line 3: A competing convention is to distinctly name this kind of component a `PropsCollector`.  

https://github.com/acl-services/paprika/blob/master/packages/Checkbox/src/CheckboxInputPropsCollector.js

Looks like you've followed the pattern in the `<OverflowMenu>` as closely as possible, but if you ignore that factor, which pattern do you like better?

mikrotron (approved):
llgtm 80

PR #1003, opened 2021-04-09
UXD-1052 Refactor Input API, Styles + Docs 🐐

mikrotron:
### Purpose 🚀
Remove `classNames` and refactor styling to pure `styled-components` conventions. Establish API that will be consistent with `<Select>` and `<Textarea>`.

#### Added

- Added an imperative `focus()` method.
- Added a props collector component, `<Input.Container>`, to apply attributes to root `<div>` element.

#### Removed

- Removed `className` prop.
- Removed all classNames from DOM elements.

#### Migration Guide

- If a custom class name is required on the root element, it can be added like `<Input><Input.Container className="custom"></Input>`.
- If classNames were used to target the `<input>`, they will need to be udpated to use attributes like `data-pka-anchor`, `aria-invalid`, etc.

### Notes ✏️
- Also updated Storybook documentation – updated to CSF and improved Screener coverage
- Adjusted styling of `<ListBox>` after a regression
- Fixed some `<DatePicker>` tests – updated selectors that relied on class names that were removed and used `aria` attributes instead.

### Storybook 📕
http://storybooks.highbond-s3.com/UXD-1052--refactor-input

### Screenshots 📸
Screen Shot 2021-05-04 at 1 52 08 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1052


.changeset/curly-ghosts-pull.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Removed
+
+- Removed `className` prop.
+- Removed all classNames from DOM elements.
+
+### Migration Guide
+
+- If a custom class name is required on the root element, it can be added like `<Input><Input.Container className="custom"></Input>`.
+- If classNames were used to target the `<input>`, they will need to be udpated to use attributes like `data-pka-anchor`, `aria-invalid`, etc.
```
nahumzs, line 18: so now doing: `<Input className="my-custom" />` should add a classname in the input HTML element?

mikrotron: Yes, exactly! It _used to_ go on the root `<div>` but now it would go on the `<input>`.  

If you had added any other kind of attribute though, like `<Input data-testid>` it _used to_ to on the `<input>` and it will _still_ go on the `<input>`. 

The change is that you _can_ put that `data-testid` attribute on the root `<div>` if you need to now.

packages/Input/src/Input.js
```diff
@@ -27,9 +26,13 @@ const Input = React.forwardRef((props, ref) => {
   } = props;
 
   const isControlled = value !== undefined;
+  const containerProps = extractChildrenProps(children, InputPropsCollector);
 
   const i18n = useI18n();
 
+  const _refInput = React.useRef();
```
oscarkwan, line 33: Feels overkill to have a variable for this,

can just do `const refInput = ref || React.useRef();` on this next line

packages/Input/src/Input.js
```diff
@@ -64,56 +64,45 @@ const Input = React.forwardRef((props, ref) => {
     const iconSize = size === types.LARGE ? types.MEDIUM : types.SMALL;
 
     return (
-      <Button.Icon
+      <sc.ClearButton
```
oscarkwan, line 67: Since we don't provide class names anymore, maybe we should add some `pka-anchors` for our QA friends since they probably have been using the class name for automation tests

packages/Input/src/Input.js
```diff
@@ -150,21 +139,18 @@ const propTypes = {
...
-   * Callback to be executed when the input value is changed. Receives the
-   * onChange event as an argument, except when the clear button is clicked,
-   * then the argument is null. Needed when value prop is provided (component
-   * is controlled).
-   */
+  /** Callback to be executed when the input value is changed. Receives the onChange event as an argument, except when the clear button is clicked, then the argument is null. Required when value prop is provided (component is controlled). */
   onChange: PropTypes.func,
 
+  /** Ref for the input DOM element. */
+  refInput: RefOf(PropTypes.instanceOf(Element)),
```
oscarkwan, line 146: I stopped using `PropTypes.instanceOf(Element)` as it throws errors in server side starling for example. An Alternative is just to use `PropTypes.node`

reference: https://github.com/acl-services/paprika/pull/1030

packages/Input/src/Input.styles.js
```diff
@@ -1,34 +1,43 @@
...
+  padding-right: ${size === types.LARGE ? stylers.spacer(4) : stylers.spacer(3)};
+`;
+
+export const InputContainer = styled.div`
   line-height: 1;
   position: relative;
 
-  input.form-input__input {
-    ${stylers.placeholders}
+  input::-webkit-search-cancel-button {
```
oscarkwan, line 34: Is this a browser specific fix? Maybe be good to add a brief comment

packages/Input/src/Input.js
mikrotron: It is possible if you don't care about the `Rules of Hooks`! 
Actually I tried this originally, it worked, and eslint didn't complain until it got to CI.  But it is a violation of the rules – you should not call hooks conditionally.

packages/Input/src/Input.js
mikrotron: This is a good idea.  I did add some, but the clear button would certainly benefit from one.

oscarkwan (approved):
LGTM

PR #1014, opened 2021-04-21
UXD-1086 support local storage in useColumnsArrangement

allison-c:
### Purpose 🚀

This PR is for using localStorage to save the show/hide states in `useColumnsArrangement` hook. There'll be a follow-up PR to save the orders of columns.
Saving in localStorage will be only activated, when you pass in the `localStoragePrefix` when you use the hook.
For developers who are not using the hook, you can implement that by yourself.

Try here: http://storybooks.highbond-s3.com/paprika/UXD-1086-support-using-localStorage/?path=/story/table-actionbar-examples--with-localstorage-enabled

**_[BREAKING CHANGES]_**

This PR also changes the way of passing parameters to `useColumnsArrangement` hook.
Before:
```jsx
const {orderedColumnIds, ...} = useColumnsArrangement(
  defaultOrderedColumnIds, // string[], required
  disabledColumnIds  // string[]
);
```
After:
```jsx
const {orderedColumnIds, ...} = useColumnsArrangement({
  defaultOrderedColumnIds, // string[], required
  disabledColumnIds,  // string[]
  defaultHiddenColumnIds,  // string[]
  localStoragePrefix // string
});
```

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1086-support-using-localStorage

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/ActionBar/README.md
+++ b/packages/ActionBar/README.md
@@ -136,7 +136,7 @@ const {
   onHideAll,
   onChangeOrder,
   isColumnHidden,
-} = useColumnsArrangement(defaultColumnsForArrangement);
+} = useColumnsArrangement({ defaultOrderedColumnIds });
 
 return (
   <ColumnsArrangement
--- a/packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
+++ b/packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
@@ -80,7 +80,7 @@ export default function ColumnsArrangement(props) {
   };
 
   function handleSearch(e) {
-    setSearchTerm(e.target.value);
+    setSearchTerm(e === null ? "" : e.target.value);
   }
 
   return (
@@ -120,23 +120,32 @@ export default function ColumnsArrangement(props) {
             I18n.t("actionBar.no_results")
           ) : (
             <sc.Sortable onChange={handleChangeOrder} hasNumbers={false}>
-              {filteredColumnIds.map(id => (
-                <Sortable.Item
-                  key={id}
-                  sortId={id}
-                  isDragDisabled={columns[id].isDisabled}
-                  handleElement={columns[id].isDisabled ? <sc.LockIcon /> : undefined}
-                >
-                  <ColumnManagingItem
+              {filteredColumnIds.map(id => {
+                if (!columns[id]) {
+                  console.error(
+                    `Failed from rendering the ${id} column. Have you passed the <ColumnsArrangement.ColumnDefinition /> for that column?`
+                  );
+                  return null;
+                }
+
+                return (
+                  <Sortable.Item
                     key={id}
-                    id={id}
-                    label={columns[id].label}
-                    isDisabled={columns[id].isDisabled}
-                    isHidden={columns[id].isHidden}
-                    onChangeVisibility={onChangeVisibility}
-                  />
-                </Sortable.Item>
-              ))}
+                    sortId={id}
+                    isDragDisabled={columns[id].isDisabled}
+                    handleElement={columns[id].isDisabled ? <sc.LockIcon /> : undefined}
+                  >
+                    <ColumnManagingItem
+                      key={id}
+                      id={id}
+                      label={columns[id].label}
+                      isDisabled={columns[id].isDisabled}
+                      isHidden={columns[id].isHidden}
+                      onChangeVisibility={onChangeVisibility}
+                    />
+                  </Sortable.Item>
+                );
+              })}
             </sc.Sortable>
           )}
           {searchTerm.length ? null : (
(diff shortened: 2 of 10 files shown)


packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
```diff
@@ -120,23 +120,32 @@ export default function ColumnsArrangement(props) {
...
+                      key={id}
+                      id={id}
+                      label={columns[id].label}
+                      isDisabled={columns[id].isDisabled}
+                      isHidden={columns[id].isHidden}
+                      onChangeVisibility={onChangeVisibility}
+                    />
+                  </Sortable.Item>
+                );
+              })}
```
allison-c, line 148: The only change in this part is adding a `console.error` for the ids available in `orderedColumnIds ` but not available in `<ColumnsArrangement.ColumnDefinition />` . If it happens the other way around, there won't be any error, since we don't use `ColumnDefinition` to iterate and render anything.

packages/ActionBar/README.md
```diff
@@ -136,7 +136,7 @@ const {
   onHideAll,
   onChangeOrder,
   isColumnHidden,
-} = useColumnsArrangement(defaultColumnsForArrangement);
+} = useColumnsArrangement({ defaultOrderedColumnIds });
```
nahumzs, line 139: by default is undefined?

packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
```diff
@@ -120,23 +120,32 @@ export default function ColumnsArrangement(props) {
...
-                  key={id}
-                  sortId={id}
-                  isDragDisabled={columns[id].isDisabled}
-                  handleElement={columns[id].isDisabled ? <sc.LockIcon /> : undefined}
-                >
-                  <ColumnManagingItem
+              {filteredColumnIds.map(id => {
+                if (!columns[id]) {
+                  console.error(
+                    `Failed from rendering the ${id} column. Have you passed the <ColumnsArrangement.ColumnDefinition /> for that column?`
```
nahumzs, line 126: maybe is better to throw an error since it can't work without an `id`?

packages/ActionBar/src/hooks/useColumnsArrangement/useColumnsArrangement.js
```diff
@@ -1,9 +1,48 @@
 import React from "react";
 import produce from "immer";
 
-export default function useColumnsArrangement(defaultOrder, disabledColumnIds = []) {
+const savedItem = {
+  VISIBILITY: "VISIBILITY",
+  ORDERS: "ORDERS",
+};
+
+function getLocalStorageKey(localStoragePrefix, typeOfSavedItem) {
```
nahumzs, line 9: is this localStoragePrefix intended to solve issues like two people with different accounts using the same computer?

`useColumnArragment({ localStragePrefix: ${project}-${user.id}-fature-name})`?

would it worth describing in the example a best practice when creating these names? can be the case that in two different features within the same app rendering two tables, one overrides the other one localstoraage content because incorrect naming conventions? (this might be an edge case but hard bug to catch)

packages/ActionBar/src/components/ColumnsArrangement/ColumnsArrangement.js
allison-c: But it doesn't break the UI, we just don't render the column if they didn't pass the column definition

nahumzs:
Only one real concern, if we were going to implement the same in our other components (sort and filter) are we going to implement all this work again? like, declare localstoragePrefix modify internally each of this component? I think it would be the case right?

packages/ActionBar/src/hooks/useColumnsArrangement/useColumnsArrangement.js
allison-c: Sure, I can add some comments, the consuming app should be responsible for making sure the keys are unique I think

allison-c:
> Only one real concern, if we were going to implement the same in our other components (sort and filter) are we going to implement all this work again? like, declare localstoragePrefix modify internally each of this component? I think it would be the case right?

I think declaring `localstoragePrefix` inside each component is a cleaner way for now. Each ColumnsArrangement instance needs a different localStorage key for sure.
However, I can also think of making a package and export a component providing a context and called `<PaprikaLocalStorage prefix="xxx" />` so that all the child components can share the same prefix? Do they always want to use the same prefix for all types of components? They might end up nesting the wrapper components in their app. Like in our case, we have multiple columns arrangement on the same page, I'll need to use `PaprikaLocalStorage` to wrap each `ColumnsArrangement`.

nahumzs:
ok, let's go with this solution for now.

nahumzs (approved):

PR #1017, opened 2021-04-22
RAPRM-1705  Add Compact variant of ProgressBar 🐐

mikrotron:
### Purpose 🚀
Add an `isCompact` prop to `<ProgressBar>` to display in a more compact style.

### Notes ✏️
- Coloured progress now changes to `green` when 100% complete.
- Conditionally rendered sub-components to keep DOM cleaner.
- Conditionally render a11y text to allow for a "silent" loading bar when that offers a better UX for users of AT.
- Added `headerLevel` prop to allow better control of semantics for improved a11y.
- Refactored Storybook stories to use CSF.
- Updated "Loading" story to use a never-ending fake loading algorithm.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/RAPRM-1705--compact-progress-bar/?path=/story/messaging-progressbar--showcase

### Screenshots 📸
Updated Screener story
Screen Shot 2021-04-22 at 2 34 26 PM

### References 🔗
https://aclgrc.atlassian.net/browse/RAPRM-1705


packages/ProgressBar/src/ProgressBar.js
```diff
@@ -7,39 +7,54 @@ import * as sc from "./ProgressBar.styles";
...
 function ProgressBar(props) {
-  const { a11yText, header, bodyText, completed, ...moreProps } = props;
-  const bestAria = a11yText || "Loading";
+  const { a11yText, bodyText, completed, header, headerLevel, isCompact, ...moreProps } = props;
+
   return (
     <sc.ProgressBar {...moreProps}>
-      <Heading level={3}>{header}</Heading>
-      <sc.Bar>
+      {header ? (
```
jamiek-acl, line 41: could do `header && ...` but no big deal. same a few lines down where you have the `?`. This gets rid of the `null`s

packages/ProgressBar/stories/examples/Loading.js
```diff
@@ -1,28 +1,39 @@
...
 
   return (
-    <Story>
-      <Modal isOpen>
-        <Modal.Content>
-          <ProgressBar completed={value} a11yText={a11yText} />
-        </Modal.Content>
-      </Modal>
-    </Story>
+    <>
```
jamiek-acl, line 35: Dont need these `<>` but no big deal

jamiek-acl (approved):

packages/ProgressBar/src/ProgressBar.js
mikrotron: Yeah, I've been deliberately trying to avoid `&&` for conditional rendering... there are some cases it can bite you:

https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

packages/ProgressBar/stories/examples/Loading.js
mikrotron: Whoops.  I did have more elements at one point, then forgot to remove.  Not worth making Semaphore run again for this though.

Issue #1024, opened 2021-04-26
Controlled <ListBox />  3.0 breaks when using keyboard


allison-c:
# Bug Report

Go to https://codesandbox.io/s/romantic-moore-40x30?file=/src/App.js
- click the trigger
- use keyboard to select an item

## Expected behavior

should work

## Current behavior

app breaks

## Screenshots / Gifs / Codepens

https://codesandbox.io/s/romantic-moore-40x30?file=/src/App.js

## Additional context

When the options re-renders, they generate new ids, but the ids in the outside context didn't change, so the app breaks

nahumzs:
This has been fixed on UXD-1134 🙏, let me know if you keep having problems after updating the package @allison-c

Issue #971, opened 2021-03-11
Cannot type space in ListBox.Filter in some scenarios


allison-c:
# Bug Report

## Expected behavior

- Open the Listbox by clicking on the listbox trigger
- Start typing in the filter area
- Type `option1`
- I should see 2 items in the list
- Type ` ` space to continue filtering
- I should see the popover is still open and I can continue typing

## Current behavior

- Go to https://codesandbox.io/s/beautiful-wind-3vyic?file=/src/App.js
- Open the Listbox by clicking on the listbox trigger
- Start typing in the filter area
- Type `option1`
- You will see 2 items in the list
- Type ` ` space to continue filtering
- You will see the popover has been closed
- I cannot filter between the last 2 items

## Screenshots / Gifs / Codepens
https://codesandbox.io/s/beautiful-wind-3vyic?file=/src/App.js

## Additional context
*Please provide any more additional details to illustrate the bug*

PR #1039, opened 2021-05-11
fix(ListBox): UXD-1134 fix issue with listbox keyboard operations

josh-luong:
### Purpose 🚀
Fixes:

1. Issue when typing {space} in the filter bar (it is disabled prior to the fix) (#971):
Action: trying to type 'Lord Voldemort' _before the fix_ will not allow you to type a space.
image

2. Issue when using keyboard to navigate through the single/ multi controlled list box (i.e. when pressing {space} or {enter} on one of the options in the list (#1024)
Action: pressing down/up-arrow navigates through the options; but pressing space/ enter to select an option _before the fix_ will not select (for both single/ multi controlled ListBox).
Note: for multi ListBox bug, you may have to select multiple options to propagate the bug
image

3. Issue when using keyboard for multi list box: prior to the fix, select 2-3 options from the dropdown and start typing in the filter bar – when {space} is pressed, it will de/re- select a selected option from previous step. 

Screen Recording 2021-05-13 at 3 43 11 PM

Added: 

1. Various cypress tests to test for both issues above

### Notes ✏️
- Adding null checks and prevention of event.preventDefault() when in filter bar view
- Updating optionsIndex to reflect proper indexing of options for controlled ListBox
- For the third issue, it is related to the keyUp event being fired when in either the filter bar /component or options list/component. In order to prevent this, the call to onKeyUp was removed from Content.js (and the handleKeyUpKeyboardKeys was imported/called from List.js (the options list)). Now, keyUp events when in the filter bar/component will not trigger any unwanted events (i.e. selection/de-selection)-- Note: onKeyUp event was not directly handled in the filter bar, but since the onKeyUp event was initiated from the Content component, it allowed for unwanted key events.

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1134-fix-issue-with-listbox-keyboard-operations

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1133


packages/ListBox/src/components/Content/Content.js
```diff
@@ -53,12 +53,11 @@ export default function Content(props) {
         {...getDOMAttributesForListBox({ idListBox, refLabel, ...state })}
         data-pka-anchor="list-box-content-inline" // TODO: rename "list-box.content-inline"
         onKeyDown={handleKeyDownKeyboardKeys({ providedProps, state, dispatch, onChangeContext })}
-        onKeyUp={handleKeyUpKeyboardKeys({ providedProps, state, dispatch, onChangeContext })}
         ref={refListBoxContainer}
         role="listbox"
         tabIndex="0"
       >
-        {children}
+        {props.children}
```
nahumzs, line 60: you can leave {children} no need to add `props.`

nahumzs:
Josh could you add an example for the controlled version https://github.com/acl-services/paprika/issues/1024 I think this PR solves that issue, am I correct? also a small test would be great.

nahumzs (approved):

nahumzs (approved):

PR #1041, opened 2021-05-11
feature(Table): UXD-1141 adding support for cssProps prop in Columnde…

nahumzs:
### Purpose 🚀

- Added support for `cellProps` prop on ColumnDefinition allowing to control props on top of each cell, styles, class, anchors, etc.
Screen Shot 2021-05-11 at 4 04 39 PM
Screen Shot 2021-05-11 at 3 20 40 PM

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1141-table-css-style-support

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/Table/src/Table.js
+++ b/packages/Table/src/Table.js
@@ -66,7 +66,19 @@ const Table = React.forwardRef((props, ref) => {
               {ColumnDefinitions.map((columnDefinition, columnIndex) => {
                 const position = { "data-row-index": rowIndex, "data-column-index": columnIndex };
 
-                const { cell, header, width, sticky, ...moreColumnProps } = columnDefinition.props;
+                const {
+                  cell,
+                  header,
+                  width,
+                  sticky,
+                  cellProps: _cellProps,
+                  ...moreColumnProps
+                } = columnDefinition.props;
+
+                const cellProps =
+                  typeof _cellProps === "function"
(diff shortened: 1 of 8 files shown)


packages/Table/src/Table.js
```diff
@@ -66,7 +66,19 @@ const Table = React.forwardRef((props, ref) => {
               {ColumnDefinitions.map((columnDefinition, columnIndex) => {
                 const position = { "data-row-index": rowIndex, "data-column-index": columnIndex };
 
-                const { cell, header, width, sticky, ...moreColumnProps } = columnDefinition.props;
+                const {
+                  cell,
+                  header,
+                  width,
+                  sticky,
+                  cellProps: _cellProps,
```
Nandez89, line 74: Where does `_cellProps` come from? I don't understand this part 🙈 
Is the `_` a convention to indicate it might be undefined and that's why we check for the type later

nahumzs: ohhh this is a javascript feature, when destructuring an object you can rename the variables:

ex:

```
const { name: renameThisToNewName, lastName } = user;

const name = "name is:" + renameThisToNewName // as example
```

Nandez89: Oh nice, can't say I've used that feature when destructuring before. TIL 😅

Nandez89 (approved):
lgtm

PR #1042, opened 2021-05-12
UXD-1150 - Resolve some dependabot alerts

jamiek-acl:
### Purpose 🚀
Upgraded Lerna, Storybook, lint-staged, required Node version and some Babel plugins.

### Notes ✏️
1. No `changeset` necessary.
2. The Lerna upgrade is a major version bump, but [according to its author](https://github.com/lerna/lerna/releases/tag/v4.0.0): `Nothing was newly deprecated, nothing previously deprecated was removed. Migrating to v4 should be straightforward.` 
3. There is annoying console message that started showing up repeatedly when starting Storybook that I was unable to get rid of:

```
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-proposal-private-methods.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
	["@babel/plugin-proposal-private-property-in-object", { "loose": true }]
to the "plugins" section of your Babel config.
```


babel.config.js
```diff
@@ -14,12 +14,9 @@ module.exports = function BabelConfigJS(api) {
         useESModules: false,
       },
     ],
-    [
-      "@babel/plugin-proposal-class-properties",
-      {
-        loose: true,
-      },
-    ],
+    ["@babel/plugin-proposal-class-properties", { loose: true }],
```
nahumzs, line 17: these were required after the upgrade I can imagine.

package.json
```diff
@@ -82,8 +83,8 @@
     "jest-axe": "^4.1.0",
     "jest-cli": "^26.0.1",
     "js-yaml": "^3.13.1",
-    "lerna": "^3.13.0",
-    "lint-staged": "^8.2.1",
+    "lerna": "^4.0.0",
```
nahumzs, line 86: 👏

babel.config.js
jamiek-acl: Yes, Storybook needed them

nahumzs (approved):

jamiek-acl:
This resolved 6 out of the 16 dependabot alerts.

PR #1045, opened 2021-05-14
UXD-1053  Refactor Select API, Styles + Docs 🐐

mikrotron:
### Purpose 🚀
Remove `classNames` and refactor styling to pure styled-components conventions. 
Establish API that will be consistent with `<Input>` and `<Textarea>` including support for using as an uncontrolled component.

#### Changed

- Refactored styling to remove classes from markup utilize styled components
- Removed `className` prop that added a class to root `<div>` element.

#### Added

- Added the capability to use as an uncontrolled component with explicit `defaultValue` and `onChange` props added to `propTypes`.
- Added `<Select.Container>` prop collector to allow classes and other attributes to be added to root `<div>` element.

#### Migration Guide

- If a custom class name is required on the root element, it can be added like `<Select><Select.Container className="custom"></Select>`.
- If `classNames` were used to target DOM elements or identify states of the `<Select>`, they will need to be updated to use attributes like `data-pka-anchor`, `aria-invalid`, etc.

### Notes ✏️
- Also updated Storybook documentation – updated to CSF and improved Screener coverage.
- Also added a `README.md` file (can you believe the Select didn't have a README?)

### BONUS 🙌
- Updated `CaretUp` and `CaretDown` icons to use the most recent 20x20 svg source from the Icon Kit
- Updated `<ListBox>` and `<ListBoxWithTags>` carets to be consistent with `<Select>` and UI Kit
- Also fixed a minor issue with the focus styling of `<Input isReadonly>` (https://github.com/acl-services/paprika/issues/1048)
- Also removed some unused dependencies and cleaned up some code for the `<Input>` component.
- Also updated `tagline` propType for Storybook `<ExampleStory>` component to eliminate some console warnings.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1053--refactor-select

### Screenshots 📸
Screen Shot 2021-05-13 at 11 27 59 PM

Screen Shot 2021-05-20 at 1 37 45 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1053

Companion PR for refactoring `<Input>`: https://github.com/acl-services/paprika/pull/1003

Resolves https://github.com/acl-services/paprika/issues/1048


packages/Select/src/Select.js
```diff
@@ -1,94 +1,139 @@
...
+        {placeholder}
+      </option>
+    );
+  };
+
+  function getDefaultValue() {
+    if (isControlled) return undefined;
+    if (defaultValue === null && !!placeholder) return "";
+    return defaultValue;
+  }
```
mikrotron, line 61: This ensures that the `disabled` placeholder item gets selected when no `defaultValue` is provided on an uncontrolled `<Select>`.

packages/Select/src/Select.js
```diff
@@ -1,94 +1,139 @@
...
         value={value}
+        {...styleProps}
         {...moreProps}
       >
         {renderPlaceholder()}
         {children}
-      </select>
-    </sc.Select>
+      </sc.Select>
+      <sc.CaretIcon data-pka-anchor="select.arrow-icon" isDisabled={isDisabled || isReadOnly} />
```
mikrotron, line 81: Instead of embedding the SVG image in the background, I've decided to make this consistent with the `<ListBox>` and use the icon component overlayed on top of the `<select>` element.

packages/Select/src/Select.styles.js
```diff
@@ -1,19 +1,63 @@
...
+      font-style: inherit;
+    }
+  }
 `;
 
-export const Select = styled.div`
+export const CaretIcon = styled(CaretDownIcon)(
+  ({ isDisabled }) => css`
+    color: ${isDisabled ? tokens.color.blackLighten60 : tokens.textColor.icon};
+    font-size: 19px;
```
mikrotron, line 46: Ideally this would be 20px, but then the caret is actually bigger than in the mockups.  Probably the actual icon needs to be updated so the caret is smaller relative to its canvas size.

.changeset/plenty-shrimps-design.md
```diff
@@ -0,0 +1,22 @@
...
+
+- Added the capability to use as an uncontrolled component with explicit `defaultValue` and `onChange` props added to `propTypes`.
+- Added `<Select.Container>` prop collector to allow classes and other attributes to be added to root `<div>` element.
+
+#### Migration Guide
+
+- If a custom class name is required on the root element, it can be added like `<Select><Select.Container className="custom"></Select>`.
+- If `classNames` were used to target DOM elements or identify states of the `<Select>`, they will need to be updated to use attributes like `data-pka-anchor`, `aria-invalid`, etc.
+
+Author: [@mikrotron](https://github.com/mikrotron)
```
nahumzs, line 22: do we need the author, I know is required in changelog but haven't seen many people using it on this repo. super minor of course.

mikrotron: We definitely don't.  But it doesn't hurt?

packages/Select/src/Select.js
```diff
@@ -1,94 +1,139 @@
...
   } = props;
 
-  if (a11yText || placeholder) {
-    moreProps["aria-label"] = a11yText || placeholder;
+  const isControlled = value !== undefined;
+  const containerProps = extractChildrenProps(children, SelectPropsCollector);
+  const [hasNoValue, setHasNoValue] = React.useState(isControlled ? !value : !defaultValue);
+
+  const _refSelect = React.useRef();
+  const refSelect = ref || _refSelect;
```
allison-c, line 28: why not just use `ref` ?

packages/Select/src/Select.js
```diff
@@ -1,94 +1,139 @@
...
   children: null,
+  defaultValue: null,
   hasError: false,
   isDisabled: false,
   isReadOnly: false,
+  onChange: () => {},
   placeholder: null,
   size: Select.types.size.MEDIUM,
-  value: "",
+  value: undefined,
```
allison-c, line 132: Does `null` work? all the other default empty values are `null`

packages/Select/src/Select.js
mikrotron: I don't think it should be required for the consumer to pass a `ref` to the `<Select>`.  So I want it to work whether they do or not. I do reference the `refSelect` in the code, so there needs to be one either way.

packages/Select/src/Select.js
mikrotron: If you pass `value=null` to the underlying `<select>` React thinks you want it to be a controlled component, and it stops working properly. I think. I'll double check though.

packages/Select/src/Select.js
allison-c: Where did you use `refSelect ` ? I only found a storybook example using it

packages/Select/src/Select.js
allison-c: oh I didn't know that, thanks

packages/Select/src/Select.js
mikrotron: Oh! I did have a hook I was using it in, but I think I found a simpler way to do what I was trying to do (change the styling when the placeholder item is selected).  So you are right, I don't need to define my own, I can just use `ref` 👍

packages/Select/src/Select.js
mikrotron: Actually it does still work, but there are console warnings from React.

```
Warning: `value` prop on `select` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

and

```
Warning: Select elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled select element and remove one of these props.
```

packages/Select/src/Select.js
```diff
@@ -1,94 +1,136 @@
...
+    size,
     value,
     ...moreProps
   } = props;
 
-  if (a11yText || placeholder) {
-    moreProps["aria-label"] = a11yText || placeholder;
+  const isControlled = value !== undefined;
+  const containerProps = extractChildrenProps(children, SelectPropsCollector);
+  const [hasNoValue, setHasNoValue] = React.useState(isControlled ? !value : !defaultValue);
```
nahumzs, line 25: I don't really understand the `hasNoValue` state, and also why is in a negation format hasValue would be easier to negate than `!(hasNoValue)` you have on line 40.

mikrotron: Oh, you're probably right that a negative boolean name is not a good practice.  This is all for the styling of the placeholder:
https://github.com/acl-services/paprika/blob/fcf0b79156f1526bd53951cfe7d0e2c349672d92/packages/Select/src/Select.js#L30

It needs to apply different styling when the placeholder item is selected.  While _writing_ it "hasNoValue" was easier to understand, but it may be harder to _read_ so I can change it.

nahumzs: 🙏

packages/Select/src/Select.styles.js
mikrotron: This is why I updated the `CaretUp` and `CaretDown` icons 😏

nahumzs (approved):

PR #1050, opened 2021-05-21
UXD-1054 Refactor Textarea API, styles + docs 🐐

mikrotron:
### Purpose 🚀
Remove `classNames` and refactor styling to pure `styled-components` conventions.
Conform API to be consistent with `<Input>` and `<Select>`.

#### Changed

- Refactored styling to remove classes from markup and utilize styled components.
- Allow a number or string value for `maxHeight` with the default value changed to `300` instead of `300px`.

#### Removed

- Removed `className` prop that added a class to root `<div>` element.

#### Added

- Added `<Textarea.Container>` prop collector to allow classes and other attributes to be added to root `<div>` element.
- Added `minHeight` prop to set a height less than the hard-coded `80px` value.

#### Migration Guide

- If a custom class name is required on the root element, it can be added like `<Textarea><Textarea.Container className="custom"></Textarea>`.
- If `classNames` were used to target DOM elements or identify states of the `<Textarea>`, they will need to be updated to use attributes like `data-pka-anchor`, `aria-invalid`, etc.

### Notes ✏️
- Also updated Storybook documentation – updated to CSF and improved Screener coverage.
- Also harmonized some README and Storybook documentation between `<Input>`, `<Select>` and `<Textarea>`.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1054--refactor-textarea/?path=/story/forms-textarea-docs--page

### Screenshots 📸
New **Variations** story
Screen Shot 2021-05-20 at 8 34 14 PM
Screen Shot 2021-05-20 at 8 35 04 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1054

Companion `<Input>` PR: https://github.com/acl-services/paprika/pull/1003

Companion `<Select>` PR: https://github.com/acl-services/paprika/pull/1045


packages/Textarea/README.md
```diff
@@ -22,56 +22,51 @@ npm install @paprika/textarea
...
-return (
-  <Textarea
-    defaultValue="Hello world"
-    onChange={e => {
-      console.log(`The new value is ${e.target.value}`);
-    }}
-  />
-);
+<Textarea
+  defaultValue="Hello world"
```
nahumzs, line 65: To comply with the example of uncontrrolled components in react documentation I think wee should do something like this:


```js
const refTextarea = React.useRef(null)

<Textarea ref={refTextarea} defaultValue="hi!" />

refTextarea.current // hi!

```

using the onChange will imply that you have to store dom value in some state to do something with it or to submit it.

mikrotron: Good point.  I think that makes sense, but I'll want to update the Uncontrolled story as well and also update the `<Select>` and `<Input>` to make them all consistent.

allison-c:
>Selectors may have to change since there is no longer a <div> element wrapping the <textarea>.

I feel we better make this consistent in the future. For a `Select` component, we changed it to have a `div` wrapper. For `TextArea` we removed the `div` wrapper. It's not easy for consumers to remember which components have which don't. Just always have the wrapper might be easier to understand.

packages/Textarea/src/Textarea.js
```diff
@@ -1,138 +1,139 @@
...
-      window.addEventListener("resize", resize);
+      handleResize();
+      window.addEventListener("resize", handleResize);
     }
 
     return function cleanup() {
-      window.removeEventListener("resize", resize);
+      window.removeEventListener("resize", handleResize);
     };
   }, []);
```
allison-c, line 50: missing `canExpand ` in the dependency list

allison-c: and `handleResize ` too

mikrotron:
> > Selectors may have to change since there is no longer a  element wrapping the <textarea>.
> 
> I feel we better make this consistent in the future. For a `Select` component, we changed it to have a `div` wrapper. For `TextArea` we removed the `div` wrapper. It's not easy for consumers to remember which components have which don't. Just always have the wrapper might be easier to understand.

I debated that change too – consistency vs simplicity.  It's a close call, but I see @nahumzs has 👍 your suggestion so there seems to be more support for consistency. I will update it.

allison-c:
> > > Selectors may have to change since there is no longer a  element wrapping the <textarea>.
> > 
> > 
> > I feel we better make this consistent in the future. For a `Select` component, we changed it to have a `div` wrapper. For `TextArea` we removed the `div` wrapper. It's not easy for consumers to remember which components have which don't. Just always have the wrapper might be easier to understand.
> 
> I debated that change too – consistency vs simplicity. It's a close call, but I see @nahumzs has 👍 your suggestion so there seems to be more support for consistency. I will update it.

No need to update in this PR 😄

mikrotron:
> > > > Selectors may have to change since there is no longer a  element wrapping the <textarea>.
> > > 
> > > 
> > > I feel we better make this consistent in the future. For a `Select` component, we changed it to have a `div` wrapper. For `TextArea` we removed the `div` wrapper. It's not easy for consumers to remember which components have which don't. Just always have the wrapper might be easier to understand.
> > 
> > 
> > I debated that change too – consistency vs simplicity. It's a close call, but I see @nahumzs has 👍 your suggestion so there seems to be more support for consistency. I will update it. 
<br />

> No need to update in this PR 😄

I wouldn't want to introduce another breaking change later, so I've updated this PR.

nahumzs (approved):

PR #1059, opened 2021-05-31
Uxd 1200  filter multi select

jamiek-acl:
### Purpose 🚀
Add a `Multi select` filter option to the @paprika/filter component.
The comparators are `is one of` and `is not one of`.

### Notes ✏️
Screen Shot 2021-05-31 at 2 58 36 PM


packages/Filter/src/hooks/defaultReducer.js
```diff
@@ -39,6 +39,7 @@ export default function defaultReducer(draft, action) {
 
         switch (payload.type) {
           case types.changeTypes.COLUMN: {
+            console.log("changed filter");
```
allison-c, line 42: console.log

allison-c:
Filter/tests/cypress/Filter.js is failing consistently, maybe need some updates?

packages/Filter/src/helpers/testers.js
```diff
@@ -17,6 +17,16 @@ const processDate = (momentParsingFormat, value, testValue, testFunction) => {
   return false;
 };
 
+const isOneOf = (testValue, acceptableValues) => {
+  if (testValue === "") return true;
```
allison-c, line 21: `testValue` is the value in `data` source, `acceptableValues ` is the array value for multiselect, by default it's `[]` so I thought it should be `if (acceptableValues.length === 0) return true;`

jamiek-acl: You are correct

allison-c (approved):

PR #1064, opened 2021-06-02
feat(OverflowMenu):  UXD-1159 make OverflowMenu list scrollable

leonatgalvanize:
### Purpose 🚀
Since the OverflowMenu content list could be very long, we should add a maxHeight prop to allow the list to be scrollable.

### Notes ✏️
- Added maxHeight prop and styles the max-height of the Popover.Card component

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1159-make-overflow-menu-scrollable

### Screenshots 📸
**UPDATED
overflowmenu_withscrollbar

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1159


packages/OverflowMenu/src/OverflowMenu.js
```diff
@@ -43,12 +46,13 @@ const defaultProps = {
   isOpen: null,
   onClose: null,
   zIndex: zValue(1),
+  maxHeight: null,
```
leonatgalvanize, line 49: Should we leave maxHeight undefined if it's not specified? or is there a default value that we should use

packages/OverflowMenu/src/OverflowMenu.styles.js
```diff
@@ -10,3 +11,13 @@ export const Content = styled.div`
     ${stylers.lineHeight(-1)}
   }
 `;
+
+export const PopoverCard = styled(Popover.Card)(
+  ({ maxHeight }) => css`
+    max-height: ${maxHeight}px;
+    overflow: auto;
+    ::-webkit-scrollbar {
```
leonatgalvanize, line 19: Whats the UX spec for the scrollable list? I currently just have the scroll bar hidden.

packages/OverflowMenu/src/OverflowMenu.js
mikrotron: `null` is usually preferred. Ideally these longer lists of props would be alphabetized though, just a good habit to make it easier to find something (or verify if something is missing).

packages/OverflowMenu/src/OverflowMenu.styles.js
mikrotron: I think it's better to show it – it provides a visual affordance, a status of where in the list you currently are, and an indication of how large the list is / how much is not shown.  It should be flush with the edges of the `<Popover.Card>`, but I think that's what it will look like by default?

packages/OverflowMenu/src/OverflowMenu.js
mikrotron: Some of our components allow `number` or `string` for these kinds of values (I know I recently updated the `<Textarea>` to allow this).  That makes it a little more flexible, for instance, I could set the `maxHeight` to `45vh` – which actually might make a nice default value instead of `null`.  

Probably worth discussing the default behaviour with @sunminsarahlee

packages/OverflowMenu/src/OverflowMenu.js
```diff
@@ -35,6 +35,9 @@ const propTypes = {
 
   /** The z-index for the popover / confirmation */
   zIndex: PropTypes.number,
+
+  /** The maximum height of the OverflowMenu content */
+  maxHeight: PropTypes.number,
```
mikrotron, line 40: Some of our components allow `number` or `string` for these kinds of values (I know I recently updated the `<Textarea>` to allow this).  That makes it a little more flexible, for instance, I could set the `maxHeight` to `45vh` – which actually might make a nice default value instead of `null`.  

Probably worth discussing the default behaviour with @sunminsarahlee

mikrotron:
Since the `<OverflowMenu>` is composed of a `<Popover>` and any props that are passed to it are propagated down to the underlying `<Popover>`:
https://github.com/acl-services/paprika/blob/master/packages/OverflowMenu/src/OverflowMenu.js#L226

... then an alternative approach to this problem could be to simply add the `maxHeight` prop to the `<Popover>`, which would also give us an easy way to make any `<Popover>` scroll vertically if they have a large amount of content.  I think a lot of what you've done here would easily transfer if that's the approach we take.

What do you think?

cc @nahumzs

mikrotron:
Oh, one more thing I just noticed, you'll want to run `yarn` or the specific scripts to generate the README and typescript definitions after a change to the props like this.  Easy step to miss, I forget it all the time and then have to add another `chore()` commit.

mikrotron:
> this a tricky question, who should be responsible for the max-height the author ~implementing~ composing the popover or the component(popover)? 🤔 I'm leaning toward the `author` since the popover can't know all different cases where an author would want to control the specific size of the popover over.
> 
> Also to me seems that we are messing with a layout property that is affecting content outside of the component. Is like margin, you don't want to include margins in a component that can affect outside of their own boundaries.
> 
> I think is easier to create your own wrapper within the popover content and as an author controlled your own overflow.
> 
> What do you think?
> 
> I think this decision of assigning a max-height to the popover would affect more authors than consumers, and normally you don't want remover flexibility from the author but from the consumers.

I still have mixed feelings about this, but I don't think there's an obviously best choice, and I agree with your arguments here. Also, it's easier to just add this prop to the `<OverflowMenu>`, and leave the `<Popover>` alone, which counts for something.  Worst case we change our mind later and have to introduce a small breaking change to the `<OverflowMenu>`.  That's okay, that's why we use `semver`.

packages/OverflowMenu/src/OverflowMenu.js
sunminsarahlee: If we go with vh, the height will be different depending on the screen size so I’d rather prefer fix value so it will be consistent for all users. So I will propose 300px which can fit 8 items/lines in the dropdown.

nahumzs: for the default value we could use `vh`

mikrotron (approved):
llgtm 90

PR #1068, opened 2021-06-04
fix(Uploader): UXD-1214 uploader component a11y issues

josh-luong:
### Purpose 🚀
https://aclgrc.atlassian.net/browse/UXD-1211
Uploader component accessibility issues

### Keyboard accessibility

1. Should be able to tab to the “Choose a file” and select and open it in the file browser
2.  Be able to tab through uploaded files

### Screen Reader

1. Announce “drag or drop” and “choose file from computer”
2. Add more description when you tab to cancel upload/ caution button
3. Announce when file has been canceled properly
4. Announce when upload is finished and in progress, or when upload failed

### Notes ✏️

- Aria-live updates on progress have been set to update every 25% increment (use chrome devTools throttle to slow down upload time to hear the progress)

### Issues to watch out for

- Functionality differs between Safari vs Chrome, with aria-live updates for progress bars being not as reliant as expected. 
- Inability to tab to items during the upload time (possibly due to constant re-rendering of files)
- Inability for voice over to announce status buttons (cancel, caution) in Safari (works for Chrome)

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1214-uploader-component-a11y-issues

### Screenshots 📸

### References 🔗


packages/Uploader/src/components/File/File.js
```diff
@@ -100,16 +102,26 @@ function File(props) {
...
+        if (!showA11yProgress) return I18n.t("uploader.progress.uploading", { progressWithUnits, sizeWithUnits });
+        return `${I18n.t("uploader.progress.uploading") + a11yProgress}%`;
     }
   }
 
   return (
-    <sc.File>
+    <sc.File
+      aria-live="polite"
+      aria-relevant="all"
```
alexzherdev, line 113: woah TIL

josh-luong:
da90f73

- [x] Remove tab focus from `ul`, go straight to `li`
- [x] Use dotted line in focus for files (`li`)
- [x] Pop.over trigger remove `Button` and add title to trigger instead of `Caution`
- [x] Aria-live can be in `ul` instead of `File`

195a2a0
- [x] Remove tab from first div (`DropZone`); instead move focus directly to "Choose from computer" link
- [x] Change `aria-label `to be "Choose **files** from your computer"
- [x] Move focus to the UL or back to the "choose from your computer" button after choosing a file to upload
- [x] Align text in Dropzone; fix icon re-sizing issue

Note: in order to address `Move focus to the UL or back to the "choose from your computer" button after choosing a file to upload`, we needed to remove `FileInput` from the React context, as it was causing constant re-rendering during upload states; therefore, tab focus would not be able to focus into the Dropzone or the File items (meaning, keyboard users will be unable to cancel the upload). The `FileInput` logic was moved directly to DropZone (possibly could extract it and keep a separate `FileInput` component?)

packages/Uploader/src/components/DropZone/DropZone.js
```diff
@@ -18,13 +26,14 @@ export default function DropZone() {
...
       <Button
+        aria-label={I18n.t("uploader.choose_from_computer_a11y")}
         data-pka-anchor="uploader-dropZone-link"
         kind={Button.types.kind.LINK}
         onClick={() => {
           refInput.current.click();
         }}
         isSemantic={false}
-        tabIndex={-1}
+        tabIndex={0}
```
allison-c, line 36: Similar here, Button component should be focusable by default

packages/Uploader/src/components/File/File.js
```diff
@@ -70,12 +69,14 @@ function File(props) {
         return (
           <Popover isDark isEager>
             <Popover.Tip />
-            <Popover.Trigger>
+            <Popover.Trigger tabIndex={-1}>
               <Button.Icon
+                aria-label={`${name} ${I18n.t("uploader.cancel_upload")}`}
```
allison-c, line 74: it's better to translate the entire sentence, e.g `aria-label={`${I18n.t("uploader.cancel_upload", {name})}`}` and in your yml file, `cancel_upload: Cancel upload {{name}}.` because sometime the words' order can change in other languages

packages/Uploader/src/components/File/File.js
```diff
@@ -70,12 +69,14 @@ function File(props) {
         return (
           <Popover isDark isEager>
             <Popover.Tip />
-            <Popover.Trigger>
+            <Popover.Trigger tabIndex={-1}>
```
allison-c, line 72: why set this to -1? how can I cancel it by keyboard then?

josh-luong: Ahh, yes you're right. It is able to cancel by keyboard currently, but it should actually be implemented by passing in a function to render the button. Then we won't need the tabIndex=-1. I will change this!

packages/Uploader/src/components/DropZone/DropZone.js
allison-c: Sorry I think I didn't submit my last comment, because you changed the `tabIndex` here, there'll be 2 elements focusable on the screen, one is this button, the other one is the below, is this what we expected?

josh-luong: Hmm, yeah I wasn't sure about that-- right now, there's two focusable "links" to select files from computer. I could disable the one coming from line 53?

allison-c: fyi here might be a solution if we only want 1 focusable element https://codesandbox.io/s/affectionate-shape-551f7?file=/src/App.js although it might be ugly

packages/L10n/src/locales/en.yml
```diff
@@ -36,6 +36,8 @@ en:
         cancelled: Cancelled
         complete: Complete
         idle: Idle
+        file_progress: "{{name}} is {{progress}}"
+        uploading_percent: Uploading of {{a11yProgress}}%
```
allison-c, line 40: @mikrotron could you review these translations?

packages/Uploader/src/components/DropZone/DropZone.styles.js
```diff
@@ -2,6 +2,30 @@ import stylers from "@paprika/stylers";
...
+export const Container = styled.div`
+  label {
+    cursor: pointer;
+    display: inline-block;
+    height: 100%;
+    position: absolute;
+    width: 100%;
+  }
+
+  input[type="file"] {
```
allison-c, line 14: I don't think you still need this block since it's already display:none https://github.com/acl-services/paprika/pull/1068/files#diff-a1b648477dcc45faf5b5cbf86adb24f032348ccf409c77a839a32cb21e397736R71

mikrotron: When we discussed this on Friday, I didn't realize what the original strategy was, but I think we should keep it.  Basically the `visually hidden` `<input type=file>` is what will receive keyboard focus, but when it does, the `<Button>` should become highlighted as if IT has keyboard focus.  It's a trick, but interacting with the `<input>` is going to work better for AT (assistive technology – namely screen readers).  But then we want the screen reader to ignore the actual `<Button>`.  This is why there was some weird `tabIndex` values... and in fact we should retain the `tabIndex={-1}` on the `<Button>`.

packages/Uploader/src/components/DropZone/DropZone.styles.js
```diff
@@ -26,10 +50,23 @@ export const DropZone = styled.div(
       min-height: auto;
       padding: 0;
     }
+    span[data-pka-anchor="uploader-dropZone-link"] {
+      margin-bottom: 3px;
+    }
```
mikrotron, line 55: Hm, so the styling here and the styling above for `span[data-pka-anchor='button']` are both intended for that button... but the `data-pka-anchor` was changed at some point and broke the first set of styling.  It would be cleaner to just export another styled component for that `Button` and then use `sc.DropZoneButton` (or something like that) in the JSX.

packages/Uploader/src/components/DropZone/DropZone.styles.js
```diff
@@ -26,10 +50,23 @@ export const DropZone = styled.div(
...
+  `,
+  `text-align: center;
```
mikrotron, line 57: This looks out of place here... is it even working?  If it's for the `DropZone` element it can just go up above on like 15 or something.

packages/Uploader/src/components/File/File.js
```diff
@@ -100,16 +104,24 @@ function File(props) {
...
 
   return (
-    <sc.File>
+    <sc.File
+      aria-label={I18n.t("uploader.progress.file_progress", { name, progress: getProgressText(true) })}
+      aria-valuemin={0}
+      aria-valuemax={100}
+      aria-valuenow={progress}
+      role="progressbar"
+      tabIndex={0}
```
mikrotron, line 119: Do we still need this? The list item itself is not interactive, so it may not need keyboard focus, especially if the `aria-live` field is announcing properly.

packages/Uploader/src/components/File/File.js
```diff
@@ -71,25 +70,30 @@ function File(props) {
...
+                    kind={Button.Icon.types.kind.MINOR}
+                    onClick={() => {
+                      cancelFile(fileKey, onCancel);
+                    }}
+                    size={Button.Icon.types.size.SMALL}
+                  >
+                    <TimesIcon />
+                  </Button.Icon>
+                </>
+              )}
```
mikrotron, line 86: No fragment needed here, but you do need to use the render prop arguments to make the tooltip work properly, like this:

```jsx
<Popover.Trigger>
  {(handler, a11yAttributes) => (
    <Button.Icon
      onMouseOver={handler}
      onMouseOut={handler}
      onFocus={handler}
      onBlur={handler}
      aria-label={I18n.t("uploader.cancel_upload", { name })}
      kind={Button.Icon.types.kind.MINOR}
      onClick={() => {
        cancelFile(fileKey, onCancel);
      }}
      size={Button.Icon.types.size.SMALL}
      {...a11yAttributes}
    >
      <TimesIcon />
    </Button.Icon>
  )}
</Popover.Trigger>
```

packages/Uploader/src/components/File/File.js
```diff
@@ -45,20 +45,19 @@ function File(props) {
...
       case types.status.ERROR:
         return (
           <Popover isDark isEager>
             <Popover.Tip />
-            <Popover.Trigger>
-              <Button.Icon kind={Button.Icon.types.kind.MINOR} onClick={() => {}}>
-                <Caution color={tokens.color.orange} />
-              </Button.Icon>
+            <Popover.Trigger aria-label={I18n.t("uploader.restart_upload", { name })}>
+              <Caution color={tokens.color.orange} />
```
mikrotron, line 57: This seems to work fine like this 👍

packages/Uploader/src/components/File/File.js
mikrotron: This is _almost_ the same as the Storybook example:
https://github.com/acl-services/paprika/blob/master/packages/Popover/stories/examples/Tooltip.js

packages/Uploader/src/components/File/File.js
josh-luong: Hmm, without it and with just `aria-live` it will constantly announce changes in upload status like "uploading X KiB or Y MiB" until X = Y. Also, the announcements are a bit buggy for me, sometimes thing aren't announced when finished (but if you tab across the progress bars, it will announce the status).

packages/L10n/src/locales/en.yml
mikrotron: Needs a bit of work.  I am displaying the `aria-label` in blue in the screenshot below so you can see them all at a glance.  Just needs a bit more finesse.  

Screen Shot 2021-06-07 at 5 21 10 PM


As you can see, the actual progress gets announced by the screenreader since the list item is a `role=progressbar`, so the "of 25% can probably just get dropped from the `aria-label`.

packages/Uploader/src/components/File/File.js
```diff
@@ -100,16 +104,24 @@ function File(props) {
       case types.status.IDLE:
         return I18n.t("uploader.progress.idle");
       default:
-        return I18n.t("uploader.progress.uploading", { progressWithUnits, sizeWithUnits });
+        if (!showA11yProgress) return I18n.t("uploader.progress.uploading", { progressWithUnits, sizeWithUnits });
+        return I18n.t("uploader.progress.uploading_percent", { a11yProgress });
```
mikrotron, line 108:
```suggestion
        return showA11yProgress
          ? I18n.t("uploader.progress.uploading_percent", { a11yProgress })
          :  I18n.t("uploader.progress.uploading", { progressWithUnits, sizeWithUnits });
```

mikrotron: You may need to use the `showA11yProgress` for some of the other statuses too to get the `aria-labels` to sound right.

packages/Uploader/src/components/File/File.js
mikrotron: Yeah, I'm not getting the right `aria-live` behaviour I'd like either.  I'll play with it a bit more tomorrow.

packages/Uploader/src/components/File/File.js
```diff
@@ -71,25 +70,30 @@ function File(props) {
...
+                <>
+                  <Button.Icon
+                    aria-label={I18n.t("uploader.cancel_upload", { name })}
+                    kind={Button.Icon.types.kind.MINOR}
+                    onClick={() => {
+                      cancelFile(fileKey, onCancel);
+                    }}
+                    size={Button.Icon.types.size.SMALL}
+                  >
+                    <TimesIcon />
```
mikrotron, line 83:
```suggestion
                    <TimesIcon color={tokens.textColor.icon} />
```

mikrotron: Minor detail, but this icon should be a little lighter.

packages/Uploader/src/components/File/File.styles.js
```diff
@@ -12,6 +12,10 @@ export const File = styled.li`
   &:last-of-type {
     border-bottom: 0;
   }
+
```
mikrotron, line 15: Little bit of scope creep, but can we do the following on line `8` above, please? 🙏 

```css
border-bottom: 1px solid ${tokens.border.color};
```

packages/L10n/src/locales/en.yml
josh-luong: Right, I realized that too :/, but I needed that "25%" so aria-live didn't announce every percentage increase from 0-100. I think I might have another workaround, where the progressbar doesn't get the 25% increments. I'll take another stab at it!

josh-luong:
301465f Addresses the issues commented above; the main changes are for the aria-labels and aria-live announcements:

- [x] aria-labels are properly formatted for the inclusion of custom error messages. I.e. instead of saying "MyFile.jpg is The custom error message", it will announce "An error occurred for MyFile.jpg: The custom error message" .
- [x] before, the aria-live announcements would say "MyFile.jpg is Uploading of 25%, 37.5%...", in order to avoid this, we added a hidden div element that houses the aria-live announcements in 25% increments; thus, if you hover over the progressbar, it will only say the actual percentage. 

Note: because errors can be at 100%, I opted to include "Error" in the progress bar, as being just "100%" can be misleading. Also, aria-live is still a bit buggy; having throttle on to Slow 3G allows for more precise announcements for testing.

packages/Uploader/src/components/DropZone/DropZone.styles.js
allison-c: Discussed with Mike offline, should be something like this https://codesandbox.io/s/ecstatic-sun-b121k?file=/src/App.js

packages/Uploader/src/components/File/File.js
```diff
@@ -71,52 +70,83 @@ function File(props) {
...
+    <sc.FileListItem>
+      <div role="status" aria-atomic="true" aria-label={getProgressText(true)} />
+      <sc.File
+        aria-label={name}
+        aria-valuemin={0}
+        aria-valuemax={100}
+        aria-valuenow={progress}
+        aria-valuetext={status === types.status.ERROR ? I18n.t("uploader.status.error") : ""}
+        role="progressbar"
+        tabIndex={0}
```
mikrotron, line 136: This is getting really close! But this `tabIndex` is still bugging me, we _shouldn't_ need it but it is odd that it doesn't behave the same if you remove it!

I farted around with it for a while, and I think I've got it now.  Take a look at this, and see if it works for you too:
https://github.com/acl-services/paprika/compare/UXD-1214-uploader-component-a11y-issues..UXD-1214-uploader-component-a11y-issues--aria-live-experiment

packages/L10n/src/locales/en.yml
mikrotron: Getting so close! The wording for the `Uploading` status is still just a little bit off though:
Screen Shot 2021-06-08 at 10 42 49 PM

mikrotron: I think just `Filename.ext is uploading, 25%` might be sufficient. What do you think?

mikrotron:
I'm curious how this will work within the context of a real form – with a `<label>` for the `<input type=file>` which is usually accomplished with a `<FormElement>` component.  I think this is too much scope for this task though, but maybe we should create a follow-up task to add a story with the `<FormElement>`.  

Also, it might be nice to have a `useImperativeHandle` method to `focus()` which would set the focus on that `<input type=file>`.  What do you think, @allison-c?

allison-c:
> I'm curious how this will work within the context of a real form – with a `<label>` for the `<input type=file>` which is usually accomplished with a `<FormElement>` component. I think this is too much scope for this task though, but maybe we should create a follow-up task to add a story with the `<FormElement>`.
> 
> Also, it might be nice to have a `useImperativeHandle` method to `focus()` which would set the focus on that `<input type=file>`. What do you think, @allison-c?

`focus()` function might be useful! I think to be able to pass in or override props on the `input` is more useful, I'm going to do it in this PR btw https://github.com/acl-services/paprika/pull/1072 
In robots ui, they use the `name` attribute on the input, so it has to be customizable, I can also think of `data-qa-anchor`, `className` props on the input can be useful to override
If we have that flexibility, developers can just use `document.querySelector('input[name="myName"]')` to focus on the element

packages/Uploader/src/components/DropZone/DropZone.styles.js
```diff
@@ -12,24 +13,35 @@ export const DropZone = styled.div(
...
   margin-left: ${tokens.space};
 `;
+
+export const DropZoneIcon = styled.span`
+  flex-shrink: 0;
+`;
+
+export const Input = styled.input`
+  ${stylers.visuallyHidden};
+  &:focus + div [role="button"] {
```
allison-c, line 44: small comment, the space here looks like an extra space if someone doesn't look into the html code, maybe make it more explicit like `&:focus + div span[role="button"]`

PR #1075, opened 2021-06-10
fix(Tabs): UXD-1230 make space bar activate a nav tab if it is a link

jamiek-acl:
### Purpose 🚀
When press space bar on a nav tab that it is a link, visit it. This makes the behaviour the same as other tabs (non-links) when using keyboard.


mikrotron (approved):
llgtm 100

PR #1076, opened 2021-06-10
UXD-1234 fix(Panel): focus trap only when overlay

oscarkwan:
### Purpose 🚀
- Only have focus trap when have overlay
- 
### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1234-focus-trap-only-when-overlay

### Screenshots 📸

### References 🔗


packages/Panel/src/Panel.js
```diff
@@ -165,11 +164,7 @@ export default function Panel(props) {
     if (isInline) {
       sidePanel = dialog;
     } else if (!overlayExtracted) {
-      sidePanel = (
-        <Portal active>
-          <ReactFocusLock {...focusLockProps}>{dialog}</ReactFocusLock>
-        </Portal>
-      );
+      sidePanel = <Portal active>{dialog}</Portal>;
```
allison-c, line 167: What if they want to use a panel without overlay but with a focus lock? e.g.
```jsx
<Panel>
  <Panel.Content>
    <TextLine repeat={100} />
  </Panel.Content>
  <Panel.FocusLock />
</Panel>
```

nahumzs: I think if you don't have overlay then the intention is to let the user to interact with the content aside of the panel, which would make the panel accessible with the keyboard in the same context as the rest of the page. 

So shouldn't exist a reason to have a focus lock without an overflow.

allison-c: Then we will need console.warn for a case like that, otherwise they can only read from the source code to understand what happened to the `Panel.FocusLock`

nahumzs: wouldn't make it better to put a message on the showcase example? But I agree about letting know the developers about this in some way

oscarkwan: Yeah i don't know about a warning message... I could add it do the docs somewhere

allison-c (approved):

Issue #1016, opened 2021-04-21
Using same Toast multiple times with canAutoClose


jamiek-acl:
The `Toast` has canAutoClose which works fine on its own, but fails if you re-use it. For example after i create some "thing", I show a Toast which disappears 1.5s.  But if i go and create another "thing", the Toast doesnt show up subsequent times.

Nahum's clever work-around:

```
const [key, nextKey] = React.useState(0)
handleSomethingEventInYourApp() {
  nextKey(prev => prev + 1)
}
<Toast key={key} canAutoClose />
```

This should work without a clever work-around.  From Nahum: "I think the Toast is missing the part of reseting its internal state after closing so it can be open again"

PR #1079, opened 2021-06-10
fix(Toast): UXD-1237 Refactor toast internal states and fix canAutoClose for controlled

josh-luong:
### Purpose 🚀
Refactor how the internal state is coupled when using the Toast as a controlled or uncontrolled component.

### Notes ✏️
- In the current toast, the internal states `isToastOpen`, `shouldRender` was overcomplicated and coupled with the parent prop `isOpen` when it was used as a controlled component. As a result, this would lead to some undesired behaviour such as `canAutoClose` breaking the controlled toast. 
  - Issue example: https://codesandbox.io/s/heuristic-kepler-fclbd?file=/src/App.js
- Updated behaviour such that the internal state `isToastOpen` solely handles the render flow of the toast, and is toggled by `isOpen` when controlled.

- Updated documentation and added a story to exemplify when the consumer wants an _uncontrolled_ toast that can be _rerendered_ by updating the component's key to reset its internal state.
  - In reference to issue: https://github.com/acl-services/paprika/issues/1016

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1177-refactor-toast-component

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1177
https://aclgrc.atlassian.net/browse/UXD-1136
https://github.com/acl-services/paprika/issues/1016

Coauthor: @josh-luong @leonatgalvanize


packages/Toast/README.md
```diff
@@ -25,10 +25,10 @@ npm install @paprika/toast
...
 | autoCloseDelay | number                                                                                                                                                          | false    | 5000                  | Duration (in ms) before Toast will automatically close (if canAutoClose is true).                                                   |
-| canAutoClose   | bool                                                                                                                                                            | false    | false                 | Will automatically close after 1500ms (or longer if provided by autoCloseDelay).                                                    |
+| canAutoClose   | bool                                                                                                                                                            | false    | false                 | Will automatically close after 5000ms (or longer if provided by autoCloseDelay).                                                    |
```
nahumzs, line 28: is 5000ms a new value? was this suggested by a ux-design :) how did we arrive at this value?

leonatgalvanize: The default value for autoCloseDelay was always 5000ms, so technically the default behaviour was that, so I updated the docs to reflect that. However, there _is_ a 'minimumCloseTime' of 1500ms.

Right now the timeout timer is set by Math.max(autoCloseDelay, minimumCloseTime).

packages/Toast/stories/examples/UncontrolledToastExample.js
```diff
@@ -1,15 +1,30 @@
...
+  const [accumulator, setAccumulator] = React.useState(0);
+
+  function handleItem() {
+    setAccumulator(prev => prev + 1);
+  }
+
   return (
     <L10n locale="en">
       <Toast hasCloseButton>Uncontrolled toast component</Toast>
       <Toast canAutoClose autoCloseDelay={5000} hasCloseButton>
```
mikrotron, line 17: A little out of scope, but while we're here, I think it makes sense to clean up this story just a little.  There are some default props added here that are not actually needed, which is a distraction from the use case being illustrated.

packages/Toast/stories/examples/UncontrolledToastExample.js
```diff
@@ -1,15 +1,30 @@
...
         Uncontrolled toast component with auto close
       </Toast>
+      <CodeViewer>
```
mikrotron, line 20: While I appreciate the creative use of the `<CodeViewer>` here, I think maybe it would be more useful to just link to the whole story code on GitHub – a pattern established in several other components stories (`<Input>` is one example you can reference).  Essentially by using the `<ExampleStory>` helper we can include a `source` link in the header.

packages/Toast/stories/examples/UncontrolledToastExample.js
```diff
@@ -1,15 +1,30 @@
...
+        <Toast key={accumulator} canAutoClose autoCloseDelay={5000}>
+          Uncontrolled toast with auto close and ability to re-open upon button click
+        </Toast>
+        <button type="button" onClick={handleItem}>
```
mikrotron, line 24: Minor suggestion: the story would be a little nicer looking with a Paprika `<Button>`

packages/Toast/src/Toast.js
```diff
@@ -208,7 +205,7 @@ const defaultProps = {
   autoCloseDelay: 5000,
   children: null,
   hasCloseButton: true,
-  isOpen: undefined,
+  isOpen: null,
```
mikrotron, line 208: What are your reasons for preferring `null` to `undefined`?

leonatgalvanize: My understanding of the distinction is that `undefined` is a type and the intent is that the variable has been declared but hasn't been assigned a value. While `null` is an explicit assignment and **purposefully** having an 'empty' value/object.

I figured that when the consumer decides to use it as a `controlled` or `uncontrolled` component, the intent of the prop is clear. If isOpen is provided, we should have it declared and assigned to a value they control, or if it's not, we can **expect** that it is going to be empty or unknown, and not used at all in the component.

To me, using undefined seems to imply, that it might be defined and assigned a value later down the road in it's lifetime, which is not the case if you decide to use the Toast as an uncontrolled component. What do you think?

https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript

mikrotron: > using undefined seems to imply, that it might be defined and assigned a value later down the road
I'm not sure I would make that assumption, but at any rate I'm content that you've given it some deliberate thought.  

I actually don't think there is a _right_ answer here, but in some of our other components, a value of `null` is an expected value provided by the consumer, and has some meaning for the component, so this pattern of testing if `someProp === null` was not possible, and I've been forced to rely on `undefined` instead.  

For consistency across the library it might be worth keeping `isControlled = someProps !== undefined`. But this is not a convention that is documented, or has really been discussed, and changing it later would not be a disruptive change in this case, so I'm okay with leaving it as is.

nahumzs: I actually prefer `undefined`, `React` by default marks a property undefined when is not being passed down, in this cases if `typeof isOpen === "undefined"` you know for sure that is uncontrolled

using the value null could also mean `false`

let's say you have something crazy like:
```
const MyComponent = ()=> {
  // should I render?
  if(something) return <Some />
  return null
} 

function App() {
  const myComponent = <Mycomponent somenthing={false} />

  <Toast isOpen={myComponent} />
}
```

which would lead to a silent bug because `null` will mark the component as uncontrolled. 

Whenever an action depends on the existence of a prop, I prefer to use undefined and avoid the temptation of initializing the value. 

But this is a preference I don't see entirely incorrectly to use `null` to be honest, but I do think that undefined is a little more bulletproof.

nahumzs: point in case lol https://github.com/acl-services/paprika/pull/1087 internally the ListBox need to have three states for isOpen, `null, true, false` and the popover uses null to figure out if it's controlled or uncontrolled.

The ListBox has always initialized the Popover as uncontrolled and was tricky to figure out why the toggle wasn't working correctly :) ...

packages/Toast/src/Toast.js
```diff
@@ -48,23 +48,34 @@ const Toast = React.forwardRef((props, ref) => {
     },
   }));
 
-  const [isToastOpen, setIsToastOpen] = React.useState(isOpen === undefined ? true : isOpen);
+  const isControlled = isOpen !== undefined;
```
nahumzs, line 51: typeof isOpen === "undefined" is a better approach [readmore](https://thisthat.dev/variable-undefined-vs-typeof-variable-undefined/)

josh-luong: interesting!

nahumzs:
Lets merge this one if you think is at enough good stage @josh-luong, I think this branch has been running for more than a month, and the work is good enough for this iteration. As long as test are passing I'm 👌 with all this work.

nahumzs (approved):

packages/Toast/src/Toast.js
leonatgalvanize: I see, very insightful! TIL.

PR #1080, opened 2021-06-10
fix(ListBoxWithTags): ADM-1775 Fix ListBoxWithTags focus after selecting an option

leonatgalvanize:
### Purpose 🚀
After selecting an option in  `<ListBoxWithTags>`, the option is removed from the list and the 'focus' is lost from the ListBox.
As a result, if the user tries to dismiss the popover list by clicking outside of it, it will not be dismissed, until the user refocuses the ListBox (i.e. click on the filter input). 

### Notes ✏️
- Add new imperative functions for focusing in ListBox and ListBox.Filter. This allows to focus a listbox option by its index in the option list.
- ListBoxWithTags will now remember the last selected option, and try to update the focus upon render (since the component rerenders every time you select an option). If no options are left, then it will default to focusing on the filter input.
- If the user uses the filter and selects an option, it will be defaulted to focus on the filter. 
- Resolving this focus issue also fixes and allows for keyboard accessibility support now, since it was previously broken as well.

These changes will also need to be ported over to sriracha's `<UserLookup>`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ADM-1775-fix-listbox-focus-on-rerender

### Screenshots 📸
listboxwithtags_refocus_fix

### References 🔗
https://aclgrc.atlassian.net/browse/ADM-1775

### DIFF
--- a/packages/ListBoxWithTags/src/ListBoxWithTags.js
+++ b/packages/ListBoxWithTags/src/ListBoxWithTags.js
@@ -39,7 +39,7 @@ export default function ListBoxWithTags(props) {
 
   const validSize = Object.values(ListBox.types.size).includes(size) ? size : ListBox.types.size.MEDIUM;
 
-  function handleLastSelectedOption(options, selectedOption) {
+  function handleLastSelectedOption(selectedOption, options) {
     const lastIndex = Object.keys(options).length - 1;
     if (lastIndex === 0 || refFilter.current.textSearch) {
       setLastSelectedOption(null);
@@ -68,8 +68,9 @@ export default function ListBoxWithTags(props) {
   }
 
   function handleChange(...args) {
+    const [newSelectedOptions, options] = args;
     if (selectedOptions !== null && "length" in selectedOptions) {
-      handleLastSelectedOption(args[1], args[2]);
+      handleLastSelectedOption(newSelectedOptions[0], options);
       if (refFilter.current) {
         refFilter.current.reset();
         refFilter.current.focus();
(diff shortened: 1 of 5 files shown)


packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -47,13 +60,20 @@ export default function ListBoxWithTags(props) {
     ) {
       event.stopPropagation();
       onAddCustomOption(label);
-      refFilter.current.reset();
+      if (refFilter.current) {
+        refFilter.current.reset();
```
nahumzs, line 64: does it need to be reset because will rerender and get into the wrong state?

packages/ListBoxWithTags/src/ListBoxWithTags.js
```diff
@@ -47,13 +60,20 @@ export default function ListBoxWithTags(props) {
...
+        refFilter.current.focus();
+      }
     }
   }
 
   function handleChange(...args) {
     if (selectedOptions !== null && "length" in selectedOptions) {
```
nahumzs, line 71: const [options, selectedOptions] = args;

handleSelectedOptions(options, selectedOptions);

increase clearity.

packages/ListBoxWithTags/src/ListBoxWithTags.js
leonatgalvanize: Yep in the existing code it was already being reset() since if the user selects an option from a filtered list, the filter should be cleared and reset the `filteredOptions` state.

allison-c:
What if there's no filter? Does this still work?

leonatgalvanize:
> What if there's no filter? Does this still work?

To my understanding, I believe currently the ListBoxWithTags doesn't support having no filter right? If that's not the case, then I believe we could just adjust the control in the useEffect() to still work if refFilter = null.

nahumzs:
can you update this branch and will approve it I think this is good to go

nahumzs (approved):

nahumzs:
I guess there should be a follow up ticket to upgrade this in sriracha. correct?

PR #1082, opened 2021-06-11
chore(Paprika): UXD-1311 build by esm

allison-c:
### Purpose 🚀

https://aclgrc.atlassian.net/browse/UXD-1311

After this PR, in the consuming app you can `import {Caution} from '@paprika/icon'` and it supports tree shaking.
Applied similar changes to other components(which have multiple exports) to enable their tree shaking ability and clean up some export/import.
Tested with create-react-app and js/ts:
image (44)
Didn't get a chance to test with highbond-script, but hb script uses webpack 4 which supports `sideEffects` field in package.json, so I believe it should work as create-react-app

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1221-icon-esm

### Screenshots 📸

### References 🔗


.semaphore/release.yml
```diff
@@ -24,6 +24,7 @@ blocks:
           commands:
             - nvm use
             - yarn
+            - yarn build-esm
```
allison-c, line 27: I don't know if we want to include this build process in the current `prepare` stage in dev, it will make the entire yarn task slower, so only included it in the CI release process

babel.config.esm.js
```diff
@@ -0,0 +1,35 @@
+module.exports = function BabelConfigJS(api) {
+  api.cache(true);
+
+  const presets = [
+    [
+      "@babel/preset-env",
+      {
+        modules: false,
```
allison-c, line 8: The only difference is here

package.json
```diff
@@ -25,7 +25,8 @@
     "test": "jest --passWithNoTests",
     "test:a11y": "echo 'Axe will run on this command'",
     "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
-    "transpile": "lerna exec --parallel -- babel --root-mode upward src --out-dir lib --source-maps --copy-files"
+    "transpile": "lerna exec --parallel -- babel --root-mode upward src --out-dir lib --source-maps --copy-files",
+    "build-esm": "lerna exec --parallel -- babel --root-mode upward src --config-file ../../babel.config.esm.js --out-dir lib/esm --source-maps --copy-files"
```
allison-c, line 29: It uses a different babel configuration file and build into `lib/esm` folder

.semaphore/release.yml
nahumzs: is the CI who is responsible for the release in npm correct?

packages/Icon/src/index.js
```diff
@@ -0,0 +1,38 @@
+export { default as Add } from "./Add";
```
nahumzs, line 1: the consumer application should be aware to apply tree shaking for this isn't it? if not will bundle all the icons correct?

packages/Icon/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+export { default as ArrowDown } from "./ArrowDown";
```
nahumzs, line 2: is this file created programmatically? 🤔 or do we have to manually add a new icon here if we want to export it?

nahumzs:
does this same approach could work in our other packages, where we have to export more components apart from the `default` one?

.semaphore/release.yml
allison-c: Yes, it just if we want to release manually, we need to remember to run this command as well

packages/Icon/src/index.js
allison-c: Yes, it's usually handled automatically by webpack/rollup.. if case in their apps don't support tree shaking, they can still use the old import way `import Caution from '@paprika/icon/lib/Caution'` to avoid bundling everything.

packages/Icon/src/index.js
allison-c: Manually 😂  just took me 3 mins.. we can add a script in the future to generate this file for sure

allison-c:
> does this same approach could work in our other packages, where we have to export more components apart from the `default` one?

Yes, for example https://github.com/acl-services/paprika/blob/master/packages/ActionBar/src/index.js it will reduce some apps' bundle size for sure

package.json
alexzherdev: would `yarn transpile --config-file ...` work?

alexzherdev:
LGTM. make sure you test in different situations, create-react-app/typescript/highbond-scripts

allison-c:
> LGTM. make sure you test in different situations, create-react-app/typescript/highbond-scripts

Nice catch thanks, I only tested with create-react-app, will do it for the other situations before the merge

AndreyChernykh (approved):

PR #1084, opened 2021-06-14
feat(Filter): FILTER-max support a 'max filters' prop

jamiek-acl:
### Purpose 🚀
Once the max # of filters have been added, hide the `add filter` button

This is related to PDTVT-1359.


packages/Filter/src/Filter.js
```diff
@@ -85,9 +86,11 @@ export default function Filter(props) {
           ) : (
             children
           )}
-          <Button data-pka-anchor="filter.addFilterButton" icon={<AddIcon />} onClick={onAddFilter}>
-            {I18n.t("filter.actions.add")}
-          </Button>
+          {React.Children.count(children) < maxFiltersAllowed && (
```
nahumzs, line 89: if maxFiltersAllowed is reached, no filter gets rendered? shouldn't be that the first `n` maxFilerAlloowed (children) should render and then shouldn't render more? or its expected to no render a filter at all?

jamiek-acl: > if maxFiltersAllowed is reached, no filter gets rendered?

no, this `if` statement is just for the button.

if max filters is reached, they do get rendered. for example, if max filters is 3, you can add up to 3 and then the `add filter` button is hidden so you cant add a 4th.

if max filters is 3 and you passed in 4 children, then 4 would render.

nahumzs:
🤔 there is a screener issue with the side panel, doesn't seems a biggie, will approve it and accept screener, thought is weird :/

PR #1087, opened 2021-06-28
fix(ListBox): UXD-1280 fixing toggle in listbox

nahumzs:
### Purpose 🚀
This brings back the ability to toggle the ListBox whenever a user clicked the trigger, there was a couple of bugs preventing for this to work correctly:

- Popover use null to detect if it should work as an uncontrolled or controlled component and ListBox has isOpen null as initial value :/. So the popover was always acting as an uncontrolled component creating a buggy behaviour. Adding `Boolean(isOpen)` solved this issue.
- Content.js was absorbing the behaviour for the onBlur event for the trigger making it confusing where to controlled properly this event, now doesn't do anything on `<Content />` on blur and delegate that behaviour to the Trigger.js

#### EXTRA

- Clean up legacy code that wasn't doing anything.

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1280-listbox-toggle-trigger-bug

### Screenshots 📸

### References 🔗


packages/ListBox/src/hooks/hooks.js
```diff
@@ -110,56 +110,6 @@ export function useAdjustWidth() {
   }, [dispatch, state.refTriggerContainer]);
 }
 
-export function useOnScrolled() {
```
nahumzs, line 113: this was not longer relevant after we move the key up key down behaviour to focus on the next element instead of moving the scroll manually.

packages/ListBox/src/components/Popover/Popover.js
```diff
@@ -21,7 +21,13 @@ export default function Popover(props) {
   const { children, zIndex } = props;
 
   return (
-    <sc.PopoverWrapper {...props} isOpen={state.isOpen} maxWidth={state.triggerWidth} offset={0} zIndex={zIndex}>
+    <sc.PopoverWrapper
+      {...props}
+      isOpen={Boolean(state.isOpen)}
```
allison-c, line 26: why not set the default value to false? I didn't find anywhere is using the `null` value for `isOpen`

nahumzs: I think that even if we change it to false, that assumed we understand that `isOpen` can receive any falsy value which is not true, Adding Boolean protect us from falling into the trap of the Popover, I think we should fix the Popover and not use null to detect if the component is a control or uncontrolled. 

So is also protecting us from a silent and difficult bug to catch.

nahumzs: I changed the value to false

allison-c (approved):

PR #1088, opened 2021-06-28
RAPRM-2066 update copyinput for nextgen

josh-luong:
### Purpose 🚀
https://aclgrc.atlassian.net/browse/RAPRM-1740

As part of the new variable takeover design, a styled `CopyInput` component is needed from Paprika; as the original design did not meet the prototype requirements, it has been updated since. Namely, we needed to add a readOnly view, Error handling, and an updated icon.

### Notes ✏️
In Results, a similarly implemented `CopyInput`-like component is used; with this new update, we should then update that Results component to consume the Paprika one. 

Added new props: 
  /** Should hide the value text */
`  hasValueShown?: boolean;`
  /** Is the input component used */
  `hasInputContainer?: boolean;
`
Furthermore, a Variations storybook entry has been added to display the different variations of the `CopyInput`

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/RAPRM-2066-update-copyinput-for-nextgen

### Screenshots 📸
Before vs After:
image
image

### References 🔗


packages/CopyInput/src/CopyInput.js
```diff
@@ -85,7 +90,7 @@ function CopyInput(props) {
         <Popover.Tip />
       </Popover>
       {isClickedTooltipOpen && (
-        <Toast kind={Toast.types.kind.VISUALLY_HIDDEN} aria-hidden>
+        <Toast kind={Toast.types.kind.VISUALLY_HIDDEN} isPolite>
```
mikrotron, line 93: Good call with the `isPolite`, but I think it's worth keeping the `aria-hidden` so that screen reader users don't navigate to this element and read it out of context.

packages/CopyInput/src/CopyInput.js
```diff
@@ -100,18 +112,25 @@ const propTypes = {
...
+  hasValueShown: PropTypes.bool,
+  /** Is the input component used */
+  hasInputContainer: PropTypes.bool,
 };
 
 const defaultProps = {
   children: null,
   isReadOnly: true,
   value: "",
+  hasValueShown: true,
```
mikrotron, line 125: While the result is correct, it's a bit confusing that we only show the "value" label if `hasInputContainer==false`.  I feel like we should have this new prop, `hasValueShown` as `false` by default, and leave it to the consumer not to set both to toggle each option independently.  A little less convenient, but also easier to understand (and your logic above will be easier to follow/maintain).

packages/CopyInput/src/CopyInput.js
```diff
@@ -100,18 +112,25 @@ const propTypes = {
   isReadOnly: PropTypes.bool,
   /** Default value for the input */
   value: PropTypes.string,
+  /** Should hide the value text */
```
mikrotron, line 115: Maybe something a bit more explanatory, since this isn't so obvious.  Like, 
> If a plain text version of the value will be rendered

Then maybe for `hasInputContainer`:
> If the value will be rendered in an Input component

packages/CopyInput/src/CopyInput.js
```diff
@@ -46,12 +48,22 @@ function CopyInput(props) {
   }
 
   return (
-    <sc.CopyInput data-pka-anchor="copy-input" {...moreProps}>
-      <Input ref={inputRef} defaultValue={value} isReadOnly={isReadOnly} {...extendedInputProps} />
+    <sc.CopyInput
+      data-pka-anchor="copy-input"
+      hasDefaultButtonBorder={!(hasValueShown && hasInputContainer)}
```
mikrotron, line 53: I think `!hasValueShown && !hasInputContainer` would be easier to read, but I think if you take my suggestion below, this will be simpler anyway.

packages/CopyInput/src/CopyInput.styles.js
```diff
@@ -9,6 +10,15 @@ export const CopyInput = styled.div`
   }
 
   button {
-    border-radius: 0 ${tokens.border.radius} ${tokens.border.radius} 0;
+    ${props =>
+      props.hasDefaultButtonBorder ? "" : `border-radius: 0 ${tokens.border.radius} ${tokens.border.radius} 0`};
```
mikrotron, line 14: There's a different convention we've been following recently, up at the top of the declaration of this component you can do something like:

```js
export const CopyInput = styled.div(
  ({ hasDefaultButtonBorder }) => css`
```
... then probably just:
```js
  [data-pka-anchor='button'] {
    ${hasDefaultButtonBorder && `border-radius: 0 ${tokens.border.radius} 0 ${tokens.border.radius};`}
  }
```

packages/CopyInput/src/CopyInput.js
```diff
@@ -46,12 +48,22 @@ function CopyInput(props) {
...
-      <Input ref={inputRef} defaultValue={value} isReadOnly={isReadOnly} {...extendedInputProps} />
+    <sc.CopyInput
+      data-pka-anchor="copy-input"
+      hasDefaultButtonBorder={hasValueContainer || !hasInputContainer}
+      {...moreProps}
+    >
+      {hasValueContainer ? <sc.Value>{value}</sc.Value> : null}
+      {!hasInputContainer || hasValueContainer ? (
+        <sc.HiddenInput ref={inputRef} defaultValue={value} />
+      ) : (
```
mikrotron, line 59: Hm... 🤔 Is the hidden input _really_ needed? Maybe it's convenient and doesn't hurt, but it's probably useful any time theres no visible `<input>`, so the condition could just be `!hasInputContainer`? 

Or else the consumer could just add a redundant hidden `<input>` if their app really needs it?

packages/CopyInput/src/CopyInput.js
```diff
@@ -46,12 +48,22 @@ function CopyInput(props) {
   }
 
   return (
-    <sc.CopyInput data-pka-anchor="copy-input" {...moreProps}>
-      <Input ref={inputRef} defaultValue={value} isReadOnly={isReadOnly} {...extendedInputProps} />
+    <sc.CopyInput
+      data-pka-anchor="copy-input"
+      hasDefaultButtonBorder={hasValueContainer || !hasInputContainer}
```
mikrotron, line 53: I think I understand what you're trying to do here, but I was thinking we can keep it even simpler like:

```jsx
<sc.CopyInput data-pka-anchor="copy-input" hasDefaultButtonBorder={!hasInputContainer} {...moreProps}>
  {hasValueContainer ? <sc.Value>{value}</sc.Value> : null}
  {!hasInputContainer ? (
    <sc.HiddenInput ref={inputRef} defaultValue={value} />
  ) : (
    <Input isReadOnly={isReadOnly} ref={inputRef} defaultValue={value} {...extendedInputProps} />
  )}
  ...
</sc.CopyInput>
```

Which does mean that an example like this:
```jsx
<CopyInput value="something" hasValueContainer />
```

Is going to render as:

Screen Shot 2021-06-30 at 4 59 13 PM

But 🤷‍♂️ I think that's okay, and it will be pretty easy for a consumer to understand what's happening, and how to fix it.

packages/CopyInput/stories/variations/Variations.js
```diff
@@ -0,0 +1,36 @@
...
+      <CopyInput isReadOnly={false} value="button kind=default">
+        <CopyInput.Input hasClearButton />
+        <CopyInput.Button kind="default" />
+      </CopyInput>
+      <Gap.Small />
+      <>
+        <Input hasClearButton size={Input.types.size.LARGE} value={customInputValue} />
+        <CopyInput hasInputContainer={false} value={customInputValue} />
+      </>
+    </>
```
mikrotron, line 32: One additional use case I'd suggest illustrating:

```jsx
<CopyInput value="something" hasInputContainer={false} />
```

This will just render a button that copies the `value` to the clipboard.

packages/CopyInput/src/CopyInput.js
```diff
@@ -100,18 +112,25 @@ const propTypes = {
   isReadOnly: PropTypes.bool,
   /** Default value for the input */
   value: PropTypes.string,
+  /** If a plain text version of the value will be rendered */
+  hasValueContainer: PropTypes.bool,
+  /** If the value will be rendered in an Input component or hidden */
+  hasInputContainer: PropTypes.bool,
```
mikrotron, line 118: Oh, minor point, but it's nice if we can keep the `propTypes` and `defaultProps` in alphabetic order 🙏

packages/CopyInput/src/CopyInput.js
```diff
@@ -46,12 +48,18 @@ function CopyInput(props) {
...
   return (
-    <sc.CopyInput data-pka-anchor="copy-input" {...moreProps}>
-      <Input ref={inputRef} defaultValue={value} isReadOnly={isReadOnly} {...extendedInputProps} />
+    <sc.CopyInput data-pka-anchor="copy-input" hasDefaultButtonBorder={!hasInputContainer} {...moreProps}>
+      {hasValueContainer ? <sc.Value>{value}</sc.Value> : null}
+      {!hasInputContainer ? (
+        <sc.HiddenInput ref={inputRef} defaultValue={value} />
+      ) : (
+        <Input isReadOnly={isReadOnly} ref={inputRef} defaultValue={value} {...extendedInputProps} />
+      )}
```
mikrotron, line 57: Very small point, not worth re-rerunning semaphore for, but usually it's a good practice (a bit easier to read) a ternary expression if you don't include the `!` in the condition, basically reverse these cases.

mikrotron (approved):
llgtm 91

Issue #1046, opened 2021-05-14
UXD-1299: DataGrid alignment


jamiek-acl:
https://aclgrc.atlassian.net/browse/UXD-1299

In the body and in the header:
- the checkbox is too low
- the content is too high

![Screen Shot 2021-05-14 at 11 40 05 AM](https://user-images.githubusercontent.com/23224777/118314563-29482a00-b4a9-11eb-9e87-a4b22a8118de.png)

PR #1094, opened 2021-07-05
style(DataGrid): ISSUE-1046 minor alignment fix

jamiek-acl:
### Purpose 🚀
Tiny alignment adjustment (moved down 1px). Applies to header and cells. https://github.com/acl-services/paprika/issues/1046

Before:
Screen Shot 2021-07-05 at 8 59 34 AM

After:
Screen Shot 2021-07-05 at 8 56 52 AM


packages/DataGrid/src/DataGrid.styles.js
```diff
@@ -128,7 +128,7 @@ export const Cell = styled.div.attrs(({ hasZebraStripes, rowIndex, borderType })
   display: flex;
   font-size: ${tokens.fontSize.default};
   justify-content: flex-start;
-  line-height: 1;
+  line-height: 18px;
```
nahumzs, line 131: mmm this will not work if `${tokens.fontSize.default}` change their default, does `1.1` works?

nahumzs (approved):

PR #1097, opened 2021-07-07
feat(ListBox): ADM-1844 Expose ListBox.Content to spread props.

leonatgalvanize:
### Purpose 🚀
There are times where consumers (such as `UserLookup`) would need to adjust or customize the styling of the listbox popover content. This CL addresses to expose the `<ListBox.Content>` subcomponent to spread props down.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/ADM-1844-expose-listbox-content-api

### References 🔗
https://aclgrc.atlassian.net/browse/ADM-1844

### DIFF
--- a/packages/ListBox/src/components/Content/Content.js
+++ b/packages/ListBox/src/components/Content/Content.js
@@ -64,6 +64,7 @@ export default function Content(props) {
   return (
     <sc.PopoverContent
       {...getDOMAttributesForListBox({ idListBox, refLabel, ...state })}
+      {...moreProps}
       contentOffsetX={contentOffsetX}
       contentOffsetY={contentOffsetY}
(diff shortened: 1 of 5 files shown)


packages/ListBox/src/components/Content/Content.js
```diff
@@ -64,6 +64,7 @@ export default function Content(props) {
   return (
     <sc.PopoverContent
       {...getDOMAttributesForListBox({ idListBox, refLabel, ...state })}
+      {...moreProps}
```
nahumzs, line 67: shouldn't this be before the getDOMAttr...?

we wouldn't want the consumer to override any specific declaration right?

leonatgalvanize: Ohhh, yes I see.

nahumzs (approved):

PR #1102, opened 2021-07-09
fix(seducer): use key instead of fn name to avoid minification issues

Nandez89:
### Purpose 🚀
Fix issue noted on [GRCPRD-5023](https://aclgrc.atlassian.net/browse/GRCPRD-4980) and use key instead of fn.name to avoid minification issues that turn calls to action.type resulting in undefined values.

This solves for minified code to not throw [this error](https://github.com/acl-services/paprika/blob/cee049b906846928d54e6cace0875f62591704d5/packages/seducer/src/reducer.js#L48)

### Notes ✏️
When defining the actions for the reducer [here](https://github.com/acl-services/paprika/blob/%40paprika/inline-editors%400.2.0-next.0/packages/InlineEditors/src/Table/store.js#L11), we noted that a minification process will change the code can produce objects with minified names. In this [babel try console](https://babeljs.io/repl#?browsers=ie%2011&build=&builtIns=false&corejs=3.6&spec=false&loose=false&code_lz=MYewdgzgLgBAhsKBLcEYF4YG8BQMYAOATiAQQKZEBycAtuQBQCU2Avjq0A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.14.7&externalPlugins=) the code on the right would be minified to change the name of the function, turning it into something like:
```
var actions = {
  propperName: function x() {}
};
```
So using the function names when producing the types would produce the wrong object for the dispatcher. 
This can be avoided if we use the action keys instead since the key won't be affected by minification.

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗
https://aclgrc.atlassian.net/browse/GRCPRD-4980

### DIFF
--- a/.changeset/young-comics-warn.md
+++ b/.changeset/young-comics-warn.md
@@ -1,0 +1,5 @@
+---
+"@paprika/seducer": patch
+---
+
+fix minification issues when creating actionTypes
--- a/packages/seducer/src/getTypes.js
+++ b/packages/seducer/src/getTypes.js
@@ -9,7 +9,7 @@
 export default function getTypes(actions) {
   let o = {};
   for (const action in actions) {
-    o = { ...o, [actions[action].name]: actions[action].name };
+    o = { ...o, [action]: action };
   }
   return o;
 }


nahumzs:
🧠 💯 sorry about this bug.

nahumzs (approved):

nahumzs:
just to completely understand the issue, was that `actions[action]` once minified action !== "declared action" because the name would be minified right?

Nandez89:
> just to completely understand the issue, was that `actions[action]` once minified action !== "declared action" because the name would be minified right?

Correct so when calling the getTypes on minified code `actions[action].name` would be something like `x` or `y` and then that is not defined in the actions object so `actions[action]` would be undefined

PR #1104, opened 2021-07-12
feat(Table): UXD-1314 adding support for colspan and complex headers

nahumzs:
### Purpose 🚀
Add support for colspan to the Table component

Screen Shot 2021-07-12 at 11 14 37 AM

Screen Shot 2021-07-12 at 11 14 57 AM

### Notes ✏️
tests are on screener

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/Table/src/Table.js
```diff
@@ -24,41 +26,55 @@ const Table = React.forwardRef((props, ref) => {
...
-                <sc.TH sticky={sticky} borderType={borderType} key={columnIndex} {...moreColumnProps}>
-                  {header}
-                </sc.TH>
-              );
-
-            throw new Error("Header should be either of type string or function");
-          })}
-        </tr>
-      </sc.Thead>
+      {hasAtLeastOneHeader ? (
```
mikrotron, line 50:
```suggestion
      {hasAtLeastOneHeader || extractedHeaders ? (
```

packages/Table/src/Table.styles.js
```diff
@@ -3,21 +3,23 @@ import tokens from "@paprika/tokens";
...
 
-  & .is-highlighted-focus {
-    outline: 2px solid blue;
-  }
-  & .is-highlighted-idle {
-    outline: 2px solid transparent;
-  }
-`;
+    & .is-highlighted-focus {
+      outline: 2px solid blue;
```
mikrotron, line 16: A little out of scope, but maybe a quick token swap in would be a quick win?

Probably `tokens.highlight.active.noBorder.borderColor` is the best fit, or else just `tokens.color.blue`.

packages/Table/src/index.js
```diff
@@ -1 +1,2 @@
+export { TH } from "./Table.styles";
```
mikrotron, line 1: Do we export a styled component in the index of any of our other components?

Would there be a more obvious or scalable API to expose them?

Like, as `<Table.SC.TH>`?  I'm not crazy about that... but actually, maybe just exporting as `<Table.TH>` would be fine – the fact that it's just a _styled_ component doesn't really matter to the consumer.

packages/Table/stories/Table.colspans.stories.js
```diff
@@ -0,0 +1,64 @@
...
+        cell={({ row }) => {
+          return `${row.lastName}`;
+        }}
+      />
+      <Table.ColumnDefinition header="ice cream" cell="iceCream" />
+    </Table>
+  );
+};
+
+export const WithoutHeaderPropAndTableHeader = () => {
```
mikrotron, line 42: This one doesn't look right. Was that intentional?

Screen Shot 2021-07-12 at 11 43 18 AM

packages/Table/stories/Table.colspans.stories.js
```diff
@@ -0,0 +1,64 @@
...
+];
+
+export const WithColSpan = () => {
+  return (
+    <Table a11yText="" data={data}>
+      <Table.Headers>
+        <tr>
+          <TH colSpan="3">List of ice creams</TH>
+        </tr>
+      </Table.Headers>
```
mikrotron, line 30: Can we include this detail that the `children` of `<Table.Headers>` should include a `<tr>` in the Docs MDX page?  Probably just a sentence or two at the bottom of that page under "Custom Headers" would be enough.

allison-c:
A general question, do we really need to support `colspan` with these kinds of table APIs(ColumnsDefinition)? asking because there might be a case we need a `colspan` on a `td`, then how can we tell the `colSpan` prop on the `ColumnDefinition` should apply to the `th` or `td` ? if it's `td`, how can we tell which `td` should has `colspan`...etc
I think we might need a more flexible render method if we need to support `colspan` on `td` or even `rowspan` later on

packages/Table/src/Table.js
nahumzs: if fixed this already differently

packages/Table/src/Table.styles.js
nahumzs: `tokens.highlight.active.noBorder.borderColor` wao, really complex name for a token, wondering the cognitive load to remember this name :D will replace it.

packages/Table/src/index.js
nahumzs: it has to be used, in order to match the border types and even if it's sticky or not. This only affect the edge case of complex header which I think is sane to do it in this way

packages/Table/stories/Table.colspans.stories.js
nahumzs: you were caught in the middle of an update this has been remove, try to refresh the header :D sorry.

nahumzs:
> A general question, do we really need to support `colspan` with these kinds of table APIs(ColumnsDefinition)? asking because there might be a case we need a `colspan` on a `td`, then how can we tell the `colSpan` prop on the `ColumnDefinition` should apply to the `th` or `td` ? if it's `td`, how can we tell which `td` should has `colspan`...etc
> I think we might need a more flexible render method(e.g. a render method devs can use and skip ColumnDefinition) if we need to support `colspan` on `td` or even `rowspan` later on

I thought about this as well and I realized that we really don't need it since the ColumnDefinitions can be used to do it without a need of colspan

```js
<Table.ColumnDefinion cell={({ row }) => {
	return <div>{row.columData1} {row.columData2}</div>
}} />
```

The previous combine two columns in one.

packages/Table/src/Table.js
```diff
@@ -24,41 +27,57 @@ const Table = React.forwardRef((props, ref) => {
...
+                  </sc.TH>
+                );
+
+              throw new Error("Header should be either of type string or function");
+            })}
+          </tr>
+        </sc.Thead>
+      ) : extractedHeaders ? (
+        <sc.Thead>{extractedHeaders}</sc.Thead>
+      ) : null}
```
AndreyChernykh, line 80: Conditions like that `? : ? :` is hard to read and understand. 
I'd prefer it that case just move this block into a `renderHeader()` function and use a regular `if else` statements instead

AndreyChernykh: Also, there is enough logic (including `hasAtLeastOneHeader`) to consider extracting all that into a separate internal component, I think

allison-c:
> > A general question, do we really need to support `colspan` with these kinds of table APIs(ColumnsDefinition)? asking because there might be a case we need a `colspan` on a `td`, then how can we tell the `colSpan` prop on the `ColumnDefinition` should apply to the `th` or `td` ? if it's `td`, how can we tell which `td` should has `colspan`...etc
> > I think we might need a more flexible render method(e.g. a render method devs can use and skip ColumnDefinition) if we need to support `colspan` on `td` or even `rowspan` later on
> 
> I thought about this as well and I realized that we really don't need it since the ColumnDefinitions can be used to do it without a need of colspan
> 
> ```js
> <Table.ColumnDefinion cell={({ row }) => {
> 	return <div>{row.columData1} {row.columData2}</div>
> }} />
> ```
> 
> The previous combine two columns in one.

So if I do that ☝️ , the html will look like this?
```jsx
<table>
  <tr>
    <td><div>data1 data2</div></td> // or <td colspan="2"><div>data1 data2</div></td>
    <td></td>
  <tr>
</table>
```

packages/Table/stories/Table.colspans.stories.js
nahumzs: added documentation, can you take a look :)

nahumzs:
```jsx
<table>
  <tr>
    <td><div>data1 data2</div></td>
    <td></td>
  <tr>
</table>
```

would not be concept of `colspan` what do you think would be the downside for that, thinking as well that this is not a common case?

allison-c:
> ```js
> <table>
>   <tr>
>     <td><div>data1 data2</div></td>
>     <td></td>
>   <tr>
> </table>
> ```
> 
> would not be concept of `colspan` what do you think would be the downside for that, thinking as well that this is not a common case?

Probably not a common case, it just doesn't look semantically right for a long term, I'll need to think about it 😅

mikrotron (approved):
llgtm 72

PR #1105, opened 2021-07-15
RAPRM-2168 variables takeover copy input

josh-luong:
### Purpose 🚀
https://aclgrc.atlassian.net/browse/RAPRM-2168-variables-takeover-copy-input

Adding support for Popover shell, as needed in variablesTakeover, specifically need for `Zindex` and `scrollContainer` props. Also, given reports of issues with the copy input, we have decided to opt into using the `Clipboard` API, as it is async, it has better performance when copying larger inputs, and also allows copying of variables (which would be preferred in our case) (and possibly copy of images may be easier in future)

https://web.dev/async-clipboard/
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

### Notes ✏️

Since `Clipboard` is not supported in IE, we are using the old logic as a safety net if the Clipboard is not supported.

Note: Permission must be requested for read events (**we use write**) e.g.
`navigator.clipboard.readText().then(
  clipText => document.querySelector(".editor").innerText += clipText);`

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/RAPRM-2168-variables-takeover-copy-input

### Screenshots 📸

### References 🔗

### DIFF
--- a/.changeset/pink-pots-cough.md
+++ b/.changeset/pink-pots-cough.md
@@ -1,0 +1,5 @@
+---
+"@paprika/copy-input": minor
+---
+
+Addding popover shell and use of clipboard API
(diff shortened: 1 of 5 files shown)


mikrotron (approved):
llgtm 85

PR #1107, opened 2021-07-26
test(L10n): TESTING-translations Using what i18n already gives us

jamiek-acl:
### Purpose 🚀
How we are using i18n is outdated

### DIFF
--- a/packages/L10n/stories/examples/Another1.js
+++ b/packages/L10n/stories/examples/Another1.js
@@ -1,0 +1,6 @@
+import i18n from "../../src/i18n-new";
+
+export default function ConsumingApp() {
+  i18n.changeLanguage("fr");
+  return i18n.t("timePicker.custom");
+}
--- a/packages/L10n/stories/examples/Another2.js
+++ b/packages/L10n/stories/examples/Another2.js
@@ -1,0 +1,25 @@
+import i18n from "../../src/i18n-new";
+
+export default function ConsumingApp() {
+  const NAMESPACE = "translation";
+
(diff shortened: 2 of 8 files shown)


packages/L10n/stories/examples/Another1.js
```diff
@@ -0,0 +1,6 @@
+import i18n from "../../src/i18n-new";
+
+export default function ConsumingApp() {
+  i18n.changeLanguage("fr");
+  return i18n.t("timePicker.custom");
+}
```
jamiek-acl, line 6: This is how someone would consume `@paprika/i18n`

packages/L10n/stories/examples/Another2.js
```diff
@@ -0,0 +1,25 @@
...
+        saving: "You don't have permission to save",
+        deleting: "You don't have permission to delete",
+      },
+    },
+    key2: "here is another",
+    key3: "who let the dogs out?",
+  });
+
+  return i18n.t("errors.page2.saving");
+}
```
jamiek-acl, line 25: This is how someone would consume `@paprika/i18n` and provide their own Locale

tristanjasper:
Is this affected by https://github.com/acl-services/paprika/pull/1118 ?

jamiek-acl:
> Is this affected by #1118 ?

No I don't think so; this PR is more of a concept.  I had contacted the creators if `i18next` and they pointed out we were manually doing a lot of what the library does automatically.  But that is because when we integrated with it, it didn't do those things.

PR #1111, opened 2021-07-30
Uxd 1322 table - Only render rows in view via intersection observer - WIP

tristanjasper:
### Purpose 🚀

UPDATE August 16 2021  - https://github.com/acl-services/paprika/pull/1111#issuecomment-899837143

Proposed solution for table which only renders table rows that are in viewport using [react-intersection-observer](https://www.npmjs.com/package/react-intersection-observer)
This would be especially beneficial for results collection / collections pages where we can have hundreds of rows rendered at any one time and each row contains content that has relatively complex rendering logic.

It supports variable height content in each row and maintains row height as you scroll which gets applied as a height styled to the row wrapper.

Each row has a subtle fade in animation when rendering.

This could work in conjunction or as an alternative to using [react-window](https://github.com/bvaughn/react-window) solution, It is suitable for uses of the table where each row has more complex rendering logic for it's content.

intersectionObserver

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/Table/src/components/ViewportBlock/ViewportBlock.js
+++ b/packages/Table/src/components/ViewportBlock/ViewportBlock.js
@@ -35,9 +35,9 @@ const Block = props => {
       : { style: { opacity: 1 } };
 
   return (
-    <div className="viewport-block" ref={setRefs} {...styleProp} {...moreProps}>
+    <tr className="viewport-block" ref={setRefs} {...styleProp} {...moreProps}>
       {inView ? children : null}
-    </div>
+    </tr>
   );
 };
 
--- a/packages/Table/src/components/ViewportBlock/ViewportBlock.js
+++ b/packages/Table/src/components/ViewportBlock/ViewportBlock.js
@@ -1,0 +1,46 @@
+import React from "react";
+import { useInView } from "react-intersection-observer";
+import PropTypes from "prop-types";
+import "intersection-observer"; // ie 11 polyfill
+
+const propTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+const Block = props => {
+  const { children, ...moreProps } = props;
+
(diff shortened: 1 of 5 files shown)


packages/Table/src/components/ViewportBlock/ViewportBlock.js
```diff
@@ -0,0 +1,48 @@
+import React from "react";
```
tristanjasper, line 1: This component could made available externally or even its own package if other applications want to make use of it

allison-c:
wow it's so easy with this API 😮

allison-c:
I didn't get time to look into the code yet, maybe in the storybook example it doesn't have enough dataset? I found the row 79 is still rendered in the dom even though it's not in the viewport
http://storybooks.highbond-s3.com/paprika/uxd-1322-table-virtualized-intersection-observer/iframe.html?id=table-table--collections-example&viewMode=story

tristanjasper:
> I didn't get time to look into the code yet, maybe in the storybook example it doesn't have enough dataset? I found the row 79 is still rendered in the dom even though it's not in the viewport
> http://storybooks.highbond-s3.com/paprika/uxd-1322-table-virtualized-intersection-observer/iframe.html?id=table-table--collections-example&viewMode=story

Yes with this approach the container element around each row is still rendered even if it is out of the viewport, it needs to retain it's height even when it's contents are not being rendered. This is so that it doesn't affect the scrolling behaviour

tristanjasper:
Added a comment - https://aclgrc.atlassian.net/browse/UXD-1322?focusedCommentId=318885

I'm looking into a way to expose the `tr` rendering logic to consuming applications instead so they are able to handle the intersection observer approach if it is needed rather than adding this prop.

PR #1118, opened 2021-08-19
fix(L10n): UXD-1355 remove l10n from esm build list

allison-c:
### Purpose 🚀

Exporting both esm version and the old version might cause some context issues in the consuming app, if some part of the app is still using `import useI18n from '@paprika/l10n/lib/useI18n'` this can be fixed by changing it to `import { useI18n } from '@paprika/l10n'``but if it's a library used in the consuming app doing that, it's hard to fix. So I think we can just remove the esm modules from l10n package for now, until we figure out the best way to do it

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗


mikrotron (approved):
llgtm 73

PR #1123, opened 2021-09-01
feat(ActionBar): PDTVT-1584 Support adding a className to the ActionB…

jamiek-acl:
…ar's ColumnsArrangment and Sort popovers (since the popovers are added in portals)

### Purpose 🚀
I need to set the z-index of the popover in my app; I can do that by adding a classname to the popover content.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/pdtvt-1584--sort

### Screenshots 📸
Screen Shot 2021-09-01 at 3 23 57 PM

### DIFF
--- a/.changeset/giant-jeans-watch.md
+++ b/.changeset/giant-jeans-watch.md
@@ -1,0 +1,5 @@
+---
+"@paprika/action-bar": patch
+---
+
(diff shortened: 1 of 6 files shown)


tristanjasper:
We have to take this approach in so many cases for many components now, i would hope there is an easier way that doesn't need a props collector all of the time, perhaps something to investigate in future

PR #1147, opened 2021-10-12
feat(Link): UXD-1679 new Link component

michellehuang7:
### Purpose 🚀
Added the new `Link` component

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1679-new-Link-component

### Screenshots 📸
Screen Shot 2021-10-15 at 11 44 23 PM
Screen Shot 2021-10-15 at 11 43 53 PM
Screen Shot 2021-10-15 at 11 44 04 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1679

### DIFF
--- a/.changeset/loud-cherries-arrive.md
+++ b/.changeset/loud-cherries-arrive.md
@@ -1,0 +1,6 @@
+---
+"@paprika/icon": patch
+"@paprika/tokens": patch
+---
+
+updated icon template for icon size change
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -73,6 +73,7 @@
     "@paprika/seducer": "2.0.3"
   },
   "changesets": [
-    "brave-geese-perform"
+    "brave-geese-perform",
+    "ninety-dodos-punch"
   ]
 }
--- a/packages/CollapsibleCard/CHANGELOG.md
+++ b/packages/CollapsibleCard/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2.0.15-next.0
+
+### Patch Changes
+
+- Updated dependencies [35a00b7]
+  - @paprika/counter@1.1.0-next.0
+
 ## 2.0.14
 
 ### Patch Changes
--- a/packages/CollapsibleCard/package.json
+++ b/packages/CollapsibleCard/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/collapsible-card",
-  "version": "2.0.14",
+  "version": "2.0.15-next.0",
   "description": "Styled collapsible card component.",
   "author": "@paprika",
   "license": "MIT",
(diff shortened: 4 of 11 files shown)


allison-c:
Not sure how urgent we need this component, since the typescript support PR is merged, should we try use typescript on this new component? cc @mikrotron

mikrotron:
Will you still try to add truncation in this PR so that we can deprecate the `<ExternalLink>`?

michellehuang7:
> Will you still try to add truncation in this PR so that we can deprecate the `<ExternalLink>`?

@mikrotron I realized ExternalLink's [max-width](https://github.com/acl-services/paprika/blob/master/packages/ExternalLink/stories/examples/Showcase.js#L22) was kinda hardcoded using inline-style so I didn't add it in. I can add it as a prop for `Link` instead.

packages/Link/src/Link.styles.ts
```diff
@@ -0,0 +1,102 @@
...
+      padding: 1px ${tokens.spaceSm};
+      position: relative;
+
+      &:focus {
+        outline: none;
+      }
+
+      html:not([data-whatinput="mouse"]) &:focus,
+      &[data-has-forced-focus]:focus {
+        ${stylers.focusRing()}
```
mikrotron, line 92: Maybe we should apply a `stylers.zIndex(1)` here as well, to avoid this kind of situation:
Screen Shot 2021-10-21 at 12 12 21 PM

packages/Link/src/Link.styles.ts
```diff
@@ -0,0 +1,102 @@
...
+  &:hover {
+    background-color: ${tokens.color.blackLighten70};
+  }
+`;
+
+export const LinkContent = styled.span`
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+`;
```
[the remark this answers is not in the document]
michellehuang7: Got it! How about `isExternalLink`. When the link is in multiple lines, should `newTabIcon` always stay on the first line?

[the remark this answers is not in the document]
mikrotron: Good point.  I think trying to keep it on the first line only works if it's a `block` level element, if it's `inline`, that's going to be weird.  So let's keep it appended to the end of the line, and if we need to support keeping it in the top right (like in Task Tracker) then we can figure that out later.  Maybe it's more like the `isMenu` style at that point.  Maybe even `isMenu` + `isSubtle`... 🤔 ? Anyway, a problem to solve another day.

packages/Link/src/Link.styles.ts
allison-c: Should we use the inset shadow focus ring if we really want to avoid this? Sometimes I'm confused when I should use the inset shadow ring or outside shadow ring. In the screenshot, there's a lot of padding space, so maybe inset make sense?

michellehuang7: The padding space does look really large 🤔
`stylers.z(1)`:
Screen Shot 2021-10-21 at 5 22 23 PM
`stylers.focusRing(true)`:
Screen Shot 2021-10-21 at 5 21 17 PM

In this example it's using margin to avoid this, but I don't think we should set a default margin...?
Screen Shot 2021-10-21 at 5 26 37 PM

jonathanjakimon: I agree with Mike and if the state effects are full width then it makes sense to have squared corners. Rounded corners are better if there is a padding on the left and right (as in the screenshots above)

PR #1152, opened 2021-10-20
Uxd 1757 popover performance - WIP

tristanjasper:
### Purpose 🚀
Refactor Popover to only render content (& Tip subcomponent) when popover is open and visible.
This is to improve overall rendering performance on a page when

- Multiple popovers exist
- The content in the popovers contain complex components

Reducing dom nodes that are not visible to the user will improve overall performance for the browser.
This change also makes testing easier as previously multiple nodes in popover content elements would be found.

In discussion
It may be that this change is the default way that popover now renders it's content. However it is up for discussion as to whether we still support the previous approach where content is still present in the dom (but hidden). We could introduce a new prop to support this, and potentially avoid having to make a major bump to changed components

Still todo
Fix 2 failing tests in listbox component which doesn't find active element as previously

Looking into approach using TransitionGroup and keeping content mounted state in Content subcomponent

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗


tristanjasper:
Closing to investigate a different approach to not use componentDidUpdate in such a way

PR #1154, opened 2021-10-20
feat(Filter): PDTVT-1802 Make the SINGLE_SELECT use objects so the fi…

jamiek-acl:
…lter can use and ID while showing a label

### Purpose 🚀
I want to be able to filter on ID (e.g. userId), but show a label (e.g. userName)

Before:
The ListBox inside the Filter worked like this:
```
<select>
  <option value='allison'>allison</option>
  <option value='jamie'>jamie</option>
</select>
```

but now it works like this:
```
<select>
  <option value=1>allison</option>
  <option value=2>jamie</option>
</select>
```

Screen Shot 2021-10-20 at 4 22 09 PM

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/pdtvt-1802--filter-empty-workplan

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -73,6 +73,7 @@
     "@paprika/seducer": "2.0.3"
   },
   "changesets": [
-    "ninety-mice-matter"
+    "ninety-mice-matter",
+    "sixty-olives-happen"
   ]
 }
--- a/packages/Tokens/CHANGELOG.md
+++ b/packages/Tokens/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @paprika/tokens
 
+## 1.1.0-next.1
+
+### Patch Changes
+
+- fd94ddb: Fixed typo for \$diligent--background-base
+
 ## 1.1.0-next.0
 
 ### Minor Changes
--- a/packages/Tokens/package.json
+++ b/packages/Tokens/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paprika/tokens",
-  "version": "1.1.0-next.0",
+  "version": "1.1.0-next.1",
   "description": "Design tokens generated as scss + js",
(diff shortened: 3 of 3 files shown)


packages/Filter/src/components/Item/Item.js
```diff
@@ -40,18 +40,25 @@ function Item(props) {
...
-        );
+        return data.filter((row, index, dataOriginal) => {
+          const selectedColumnValues = dataOriginal.map(datum => datum[selectedColumnId]); // [{id, label}, {id, label}, ...]
+          const position = selectedColumnValues.findIndex(
+            selectedColumnValue =>
+              selectedColumnValue.id === row[selectedColumnId].id &&
+              selectedColumnValue.label === row[selectedColumnId].label
+          );
+          return position === index;
+        });
```
allison-c, line 61: Should we just use https://lodash.com/docs/4.17.15#uniqWith so the next person won't be confused 😂

packages/Filter/src/helpers/filterData.js
```diff
@@ -1,18 +1,29 @@
...
-            columns.find(column => filter.columnId === column.id)
-          )
-        : true;
+    function tester(filter) {
+      if (filter.rule) {
+        const columnType = columns.find(column => column.id === filter.columnId).type;
+        switch (columnType) {
+          case Filter.types.columnTypes.SINGLE_SELECT:
+          case Filter.types.columnTypes.MULTI_SELECT:
+            return testers[filter.rule](row[filter.columnId].id, filter.value);
```
allison-c, line 15: the tester function --`testers[filter.rule]` doesn't need the 3rd parameter (`column`)?

jamiek-acl: Not for the rules used by the SINGLE and MULTI selects

PR #1157, opened 2021-10-21
Uxd 1757 Popover performance, stop content rendering to dom when closed

tristanjasper:
### Purpose 🚀
Refactor Popover to only render content (& Tip subcomponent) when popover is open and visible.
This is to improve overall rendering performance on a page when

- Multiple popovers exist
- The content in the popovers contain complex components

Reducing dom nodes that are not visible to the user will improve overall performance for the browser.
This change also makes testing easier as previously multiple nodes in popover content elements would be found.

Changes include
- Wrapping existing Popover `Content` component with `Transition` component to detect opacity changes
- Adding new `isAlwaysOnDom` prop to `Content` component so that it can still be rendered the current way for backwards compatibility

Additional changes

Updates all RTL/Cypress tests for new approach
Fixes some other minor bugs

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/Popover/src/components/Content/Content.js
+++ b/packages/Popover/src/components/Content/Content.js
@@ -91,7 +91,6 @@ const Content = React.forwardRef((props, ref) => {
   }
 
   const transitionProps = shouldUnmount ? { mountOnEnter: true, unmountOnExit: true } : {};
-  console.log("content zIndex", content.zIndex);
 
   const ContentStyledComponent = (
     <Transition {...transitionProps} in={isOpen} timeout={PopoverConstants.animationDuration} onEntered={onAfterOpen}>


packages/ListBox/test/helpers/toggleHelpers.js
```diff
@@ -6,5 +6,4 @@ export function openLazyDropDown() {
 
 export function toggleDropdown() {
   cy.get(selectors.trigger).click();
-  cy.get(selectors.popover).should("be.visible");
```
tristanjasper, line 9: This was failing for me now, but then i also realized this is a strange place for this call to be made as it is a toggle action, so it may be  closing rather than opening the popover, so I don't think it belongs here, and removed

packages/Popover/src/Popover.js
```diff
@@ -104,6 +104,7 @@ class Popover extends React.Component {
       onDelayedClose: this.handleDelayedClose,
       onDelayedOpen: this.handleDelayedOpen,
       onOpen: this.handleOpen,
+      onAfterOpen: e => this.handleOnAfterOpen(e),
```
tristanjasper, line 107: needed to written this way to get correct `this` in `handleOnAfterOpen` function

packages/Popover/src/components/Content/Content.js
```diff
@@ -9,14 +11,16 @@ import { ContentStyled } from "./Content.styles";
 
 const propTypes = {
   children: PropTypes.node,
+  isAlwaysOnDom: PropTypes.bool,
 };
 
 const defaultProps = {
   children: null,
+  isAlwaysOnDom: false,
```
tristanjasper, line 19: Added this prop for backward compatibility, so if there is a situation where it is much better to keep some popover content on the dom (perhaps if there is content that needs to render complex components, or make fetches to the server), they would need to add this prop

packages/Popover/src/components/Content/Content.js
```diff
@@ -87,25 +91,30 @@ const Content = React.forwardRef((props, ref) => {
...
+          {...moreProps}
+          onBlur={callAll(handleBlur, moreProps.onBlur)}
+          onMouseOut={callAll(handleMouseEvent, moreProps.onMouseOut)}
+          onMouseOver={callAll(handleMouseEvent, moreProps.onMouseOver)}
+          onKeyDown={callAll(handleKeyDown, moreProps.onKeyDown)}
+        >
+          {props.children}
+        </ContentStyled>
+      )}
+    </Transition>
```
tristanjasper, line 117: Wrapping the existing component in a transition to detect opacity changes

packages/Popover/package.json
```diff
@@ -25,7 +25,8 @@
     "lodash.throttle": "^4.1.1",
     "memoize-one": "^5.0.0",
     "prop-types": "^15.7.2",
-    "uuid": "^8.3.2"
+    "uuid": "^8.3.2",
+    "react-transition-group": "^4.3.0"
```
allison-c, line 29: `r` should go before `u` unless you add this manually hehe

packages/Popover/src/components/Content/Content.js
mikrotron: I think we should align with the pre-existing `shouldUnmount` prop name we have on the `<Tabs.Panel>` component:

https://github.com/acl-services/paprika/blob/master/packages/Tabs/src/components/Panel/Panel.js#L13

(Even though it means the default value should be `true`)

packages/Popover/src/index.d.ts
```diff
@@ -53,6 +53,8 @@ declare namespace Popover {
     [x: string]: any;
 
     children?: React.ReactNode;
+
+    isAlwaysOnDom?: boolean;
```
allison-c, line 57: Do we consider props like `shouldUnmount` ? btw I think `Dom` should be all caps like `DOM`

packages/Popover/src/Popover.js
```diff
@@ -437,7 +438,7 @@ class Popover extends React.Component {
       this.state.tip,
       zIndex,
       this.ariaIdForContent,
-      this.handleKeyDown
+      e => this.handleKeyDown(e)
```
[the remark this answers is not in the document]
allison-c: What's the difference? Why this change is related to `handleOnAfterOpen ` ?

packages/Popover/src/components/Content/Content.js
allison-c: I think this prop should apply to both `Tip` and `Content`, not just on the `Content` component. Otherwise if `isAlwaysOnDom` is `true` on `Content`, the `Tip` will still unmount which seems a little bit strange to me. But if the team agree with it, I'm totally fine 😄

packages/Popover/tests/popover.cypress.js
```diff
@@ -19,7 +19,7 @@ describe("Popover", () => {
     cy.get(selectors.popoverTrigger).click();
     cy.get(selectors.popover).should("be.visible");
     cy.get(selectors.popoverTrigger).click();
-    cy.get(selectors.popover).should("be.not.visible");
+    cy.get(selectors.popover).should("not.exist");
```
allison-c, line 22: should we add a test case for testing the new prop?

PR #1158, opened 2021-10-22
Uxd 1775 Icon size change

ShanAtDiligent:
### Purpose 🚀
Add Icon size token and also set up default value for icon size.

### Notes ✏️
1. Update pre-exiting icon template so it takes tokens also set up check outs when no size value
2. Add icon size tokens

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1775-Icon-Align-With-UI-Kit

### Screenshots 📸

Screen Shot 2021-10-21 at 5 10 33 PM


packages/Icon/svgr.config.js
```diff
@@ -11,9 +11,10 @@ module.exports = {
     "#8A8A8A": "currentColor",
   },
   svgProps: {
-    // We need to render string template
-    // eslint-disable-next-line no-template-curly-in-string
-    css: "{`color: ${props.color};width: ${props.size};height: ${props.size};vertical-align: text-top;`}",
+    css:
+      // We need to render string template
+      // eslint-disable-next-line no-template-curly-in-string
+      "{`color: ${props.color};font-size: ${props.size||Tokens.icon.sizeSm};vertical-align: text-top;`}",
```
mikrotron, line 17: I don't mind the change to using `font-size` instead of `width` and `height`, but after a bit of experimentation, I think it will be safer to omit the default value (`Tokens.icon.sizeSm`).  

It means you'll want to add a `small: tokens.icon.sizeSm` option to the Showcase story, and probably set that as the default selected value.

PR #1161, opened 2021-10-27
UXD-1779 Button align with UI kit

ShanAtDiligent:
### Purpose 🚀
update Button style to align with UI Kit

https://aclgrc.atlassian.net/browse/UXD-1779

### Notes ✏️
- Components that are updated, Button, button group and icon button.
- Fix the knobs in button storybook
- update READ ME

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1779-Button-Align-With-UI-Kit

### Screenshots 📸
Screen Shot 2021-10-27 at 1 46 23 PM

Screen Shot 2021-10-27 at 11 00 38 AM
Screen Shot 2021-10-27 at 11 00 32 AM
Screen Shot 2021-10-27 at 10 44 35 AM


packages/Button/src/Button.styles.js
```diff
@@ -158,11 +160,17 @@ export const sizeStyles = {
     ${stylers.fontSize(-1)};
     min-height: ${stylers.spacer(4)};
     padding: 6.5px ${tokens.spaceLg};
```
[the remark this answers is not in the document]
mikrotron: You _could_, but I actually don't think it's necessary.  The only case it where it will result in a change should be for the two line buttons.  You could compare your local to the version on paprika.highbond.com, maybe send the screenshots to Jonathan and get his approval – but _I_ think the 1.5px difference is not enough to matter to anyone.

packages/ButtonGroup/src/components/Item/Item.styles.js
```diff
@@ -40,8 +41,7 @@ export const UnselectedIcon = styled(Unselected)``;
 export const Item = styled(Button)(
   ({ isActive, isDisabled }) => css`
   ${stylers.truncateText}
-  display: block;
```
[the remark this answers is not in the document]
ShanAtDiligent: good catch! I wanna to use display:flex to fix the horizontal alignment between icon and text. Will use other solution for this matter.

mikrotron:
Looking at Screener again, and although we don't have any coverage for the `<Tag>` component, I noticed with the `<ListBoxWithTags>` that the "remove" button for the `<Tags>` now looks very close to the right edge.  

Screen Shot 2021-10-29 at 2 13 05 PM

Actually the button always was this close to the edge, but when the icon was smaller it had more visible padding. Not sure if @jonathanjakimon thinks this should be adjusted?

mikrotron:
Found another unintended change thanks to Screener.  In the `<Breadcrumbs>` component, the "back" arrow has grown:
https://screener.io/v2/states/H1OkdIZtV.acl-services-paprika/UXD-1779-Button-Align-With-UI-Kit/1280x1024/Chrome/navigationbreadcrumbsbackyardtests-screener-0

I think I remember reading that the size of this little left pointing caret is actually an exception to the `20px` size rule, so we should retain the existing size in this case (`14px`).

jonathanjakimon:
@mikrotron 
About the tag we will use another icon and the padding will be 4px on the right + natural padding of the icon in its 20px box.

image

mikrotron:
> @mikrotron About the tag we will use another icon and the padding will be 4px on the right + natural padding of the icon in its 20px box.
> 
> image

Oh! Okay, if it's going to change this much then maybe we should just leave it as is for now.

ShanAtDiligent:
> Found another unintended change thanks to Screener. In the `<Breadcrumbs>` component, the "back" arrow has grown: https://screener.io/v2/states/H1OkdIZtV.acl-services-paprika/UXD-1779-Button-Align-With-UI-Kit/1280x1024/Chrome/navigationbreadcrumbsbackyardtests-screener-0
> 
> I think I remember reading that the size of this little left pointing caret is actually an exception to the `20px` size rule, so we should retain the existing size in this case (`14px`).

Hi @mikrotron  I done another test locally after run `npx lerna bootstrap` and look slike the arrows on `<Breadcrumbs>`  aren't affected by these changes.

packages/Button/src/Button.styles.js
```diff
@@ -117,13 +117,15 @@ export const commonStyles = css`
   border-style: solid;
   border-width: 1px;
   box-sizing: border-box;
+  color: ${tokens.textColor.default};
```
mikrotron, line 120: I must have forgotten to click `Add comment` before, but I remember writing that this line has inadvertently cause a regression in the `<Panel.Header>`.  The `<CloseButton>` was likely inheriting the white text colour from a parent, but now this overrides it as black. You can see the change on Screener. 

I think this change here is appropriate, but we should apply a styling override (or just propagate the `isDark` prop onto the `<CloseButton>`?).

mikrotron: I wanted to check my advice locally to make sure it was good... and _sort of_.  

Seems you can remove these lines in the `<Panel>` and it should work properly again:

https://github.com/acl-services/paprika/blob/master/packages/Panel/src/components/Header/Header.styles.js#L48-L50

.changeset/slow-berries-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+"@paprika/button": patch
+"@paprika/button-group": patch
```
mikrotron, line 3: Oh, I guess we should also include `@paprika/panel` now.

mikrotron:
While reviewing the Screener changes I notice that this use case of applying a custom `color` via CSS to the `<Button.Icon>` will no longer work:

https://github.com/acl-services/paprika/blob/master/packages/Button/stories/variations/IconButtonVariations.js#L75-L81

I think this is fine.  It makes more sense to apply a custom color directly to the `<Icon>` anyway.  

But... it has let to a small regression in the `<Breadcrumbs>` since this is the technique being applied:

https://github.com/acl-services/paprika/blob/master/packages/Breadcrumbs/src/components/ExpandButton/ExpandButton.styles.js#L7

.changeset/slow-berries-mix.md
mikrotron: ...And probably `@paprika/breadcrumbs` too...

mikrotron (approved):
llgtm 100

PR #1162, opened 2021-10-28
Uxd-1792 Toast align with UI kit

ShanAtDiligent:
### Purpose 🚀
Change Toast height and close button size based on UI kit

### Notes ✏️
- update toast component
ticket:https://aclgrc.atlassian.net/browse/UXD-1703

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1792-Toast-Align-with-UI-Kit

### Screenshots �
Screen Shot 2021-10-27 at 4 44 56 PM
Screen Shot 2021-10-27 at 4 44 46 PM
�


packages/Toast/src/Toast.styles.js
```diff
@@ -74,8 +74,8 @@ export const Content = styled.div`
 `;
 
 export const Toast = styled.div(
-  ({ isFixed, kind, zIndex, shouldRender }) => css`
-    align-items: flex-start;
```
mikrotron, line 78: I think we need to retain `flex-start` or else we can end up with the icon and close button aligned in the middle for mult-line `<Toasts>`:

Screen Shot 2021-10-29 at 2 01 10 PM

packages/Toast/src/Toast.styles.js
```diff
@@ -86,7 +86,7 @@ export const Toast = styled.div(
     font-weight: normal;
     margin-bottom: ${stylers.spacer(2)};
     min-height: ${stylers.spacer(3)};
-    padding: ${tokens.spaceLg};
+    padding: ${hasCloseButton ? "7px" : "11px"};
```
mikrotron, line 89: This makes the alignment inconsistent between `hasCloseButton==true` and `hasCloseButton==false` – namely the left padding. I think for both the left padding (and probably the right too) should consistently be `12px` (`spaceLg`).

mikrotron:
Hopefully when you change the base branch to `master` all the commits for the `<Button>` changes will disappear.  I tried to ignore them – hopefully I didn't miss anything.

packages/Toast/src/Toast.styles.js
ShanAtDiligent: Design needs the toast to be 48px in height, and close button to be bigger (32px). In order to maintain these request. I need to remove some padding when the close button is added. The other way to do it is to add negative margin on close button. Do you want me to do that instead?

mikrotron: Changing the top and bottom padding is fine, but since the CSS rule is for `padding` it's changing the left and right padding as well, which we don't want to do.

ShanAtDiligent: sorry I purposely change the padding to be the same on four sides. I will reverse that change.

packages/Toast/src/Toast.styles.js
```diff
@@ -86,7 +88,7 @@ export const Toast = styled.div(
     font-weight: normal;
     margin-bottom: ${stylers.spacer(2)};
     min-height: ${stylers.spacer(3)};
-    padding: ${tokens.spaceLg};
+    padding: 11px ${tokens.spaceLg};
```
mikrotron, line 91: With the negative margin applied to the close button now, this can revert back to 12px all around.  Currently with 11px it's just a _little_ off-center (I wasn't sure if it was just me or not, but got a designer to confirm it's just a little off).

packages/Toast/src/Toast.js
```diff
@@ -94,7 +94,7 @@ const Toast = React.forwardRef((props, ref) => {
         {!isVisuallyHidden && <sc.IconStyled as={icons[kind]} kind={kind} aria-hidden />}
         <sc.Content>{children}</sc.Content>
         {hasCloseButton && !isVisuallyHidden && (
-          <sc.CloseButtonStyled isSemantic={false} onClick={handleClose} size={types.SMALL} />
+          <sc.CloseButtonStyled isSemantic={false} onClick={handleClose} size={types.MEDIUM} />
```
mikrotron, line 97: Since you'll need to push one more commit anyway, let's go ahead and just remove the `size` prop – the default value is `MEDIUM` anyway I believe.

packages/Toast/src/Toast.styles.js
ShanAtDiligent: @mikrotron I can't add it to 12px as there is a height change request for toast. I need to reduce it from 50px to 48px. The negative margin is there to reduce the button height(32px) to the same height as the text next to it so it won't push the container when it is one line. I add the negative margin to raise the button icon up a bit. Hope it helps.

@jonathanjakimon 

Screen Shot 2021-11-16 at 10 54 46 AM

mikrotron: Oh, I see 🤔 

Can you compare a one line and two line `<Toast>` with `padding: 11px ${tokens.spaceLg};` with another pair of `<Toasts>` that have `padding: ${tokens.spaceLg} ${tokens.spaceLg} 10px ${tokens.spaceLg};`?  Let's let Jonathan or Carol decide which version they prefer and go with that.

ShanAtDiligent: Sure thing.  
Here is 12px padding top and bottom makes the toasts to 50px when single line.
Screen Shot 2021-11-16 at 12 48 56 PM

Here is 11px padding top and bottom which makes the toasts 48px in heights when single line.
Screen Shot 2021-11-16 at 12 48 03 PM

@jonathanjakimon and @carolchaan Can you let me know which one works best for you.

jonathanjakimon: After using the pixel ruler I can say that 12px has better alignments in general :)

ShanAtDiligent: Cool thank you @jonathanjakimon let me push the change :)

mikrotron (approved):
xmas llgtm 2

PR #1163, opened 2021-10-29
feat(DataTable): UXD-1793 DataTable

allison-c:
### Purpose 🚀

### Notes ✏️
See the component related details in http://storybooks.highbond-s3.com/paprika/UXD-1793-data-table/?path=/story/table-datatable-docs--page

Following tasks:
1. Adding specs https://aclgrc.atlassian.net/browse/UXD-1861
2. Make infinite loader optional https://aclgrc.atlassian.net/browse/UXD-1836

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1793-data-table/?path=/story/table-datatable-docs--page

### Screenshots 📸

### References 🔗


packages/DataTable/stories/examples/Showcase.tsx
```diff
@@ -0,0 +1,127 @@
+import React from "react";
+import StoryHeading from "storybook/components/StoryHeading";
+import { Story, Tagline } from "storybook/assets/styles/common.styles";
+import * as DataTable from "../../src";
+import makeData from "../helpers/makeData";
+
+const ShowcaseStory: () => JSX.Element = () => {
+  const columns = React.useMemo(
+    () => [
```
nahumzs, line 9: When you depend on the data to make a component, you will get into a problem when you start adding properties like isOpen, isPending, isSelected, isHovered, etc. Because the data has to be mixed with the internal state of the component.

Also, this means that if you want to modify one specific row or cell you will need to mutate the entire data structure to produce a new state in the component, which can be really expensive in big sets of data, this is not noticeable at the beginning but heavily impacts the performance try 3000 rows by 10 columns etc.

This is the reason why people using this approach dont inject the data directly into the component making it control it, instead, they provide a hook where gives you a way to implicitly control the internal state of the table.

Like this on [react-table](https://react-table.tanstack.com/):
  ```jsx
  const {
    getTableProps,
    getTableBodyProps,
    headerGroups,
    rows,
    prepareRow,
  } = useTable({
    columns,
    data,
  })
```
 
 I injected the data in the DataGrid before the creation of hooks and suffered from that heavily, but I didn't attach the state of the component on the data structure like width, or height, those were always separated. 
 
 Maybe you can avoid it in this new component. My two cents from a visitor passing by.

File: packages/DataTable/package.json [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+  "main": "lib/index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/acl-services/paprika.git",
+    "directory": "packages/DataTable"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "exports": {
```
AndreyChernykh, line 16: why do we need a separate exports? I thought that `"main": "lib/index.js",` would be enough

allison-c: I was going to try out if developers can `import RowHeightHelper from '@paprika/data-table/RowHeightHelper'`
but maybe not better than `import {RowHeightHelper} from '@paprika/data-table'`
will delete it 😅

michellehuang7 (approved):
LGTM! 
Just wondering for the RealWorld example, will the display of rows get affected by the filter? Or shall we control the display in the actual application?
Screen Shot 2021-11-25 at 2 32 26 PM

allison-c:
> LGTM! Just wondering for the RealWorld example, will the display of rows get affected by the filter? Or shall we control the display in the actual application? Screen Shot 2021-11-25 at 2 32 26 PM

It should, I didn't get enough time to fake a backend filter call 😅  usually in a real world example, when we need an infinite loader, we also need to filter by calling back-end API

michellehuang7 (approved):

michellehuang7 (approved):

PR #1166, opened 2021-11-04
Tw-2297 fix Panel rendering bugs

tristanjasper:
### Purpose 🚀
Fix paprika panel bugs

When a paprika Panel component is rendered in a paprika Panel.Group component, the parent Panel does not animate shut when closing and does not emit an onAfterClose event. This event is required so that we can redirect the url of the application after the panel is closed

When closing a paprika panel there is an initial render frame where the panel is shown as already open before it animates which leads to a bad UX.

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/tw-2297-fix-popover-rendering-bugs

### Screenshots 📸

### References 🔗


packages/Panel/src/components/Dialog/Dialog.js
```diff
@@ -4,7 +4,7 @@ import * as types from "../../types";
 import * as sc from "./Dialog.styles";
 
 function Dialog(props) {
-  const [isAnimating, setIsAnimating] = React.useState(false);
+  const [isAnimating, setIsAnimating] = React.useState(true);
```
tristanjasper, line 7: When this is `false` there is a render of the panel being open on first render pass. This can be observed on[ this story ](http://storybooks.highbond-s3.com/paprika/master/?path=/story/messaging-panel-examples--push-content-panel)

packages/Panel/src/components/Group/Group.js
```diff
@@ -30,10 +30,6 @@ export default function Group(props) {
 
   const offsetScroll = useOffsetScroll(offsetY);
 
-  if (sidePanels.filter(sidePanel => !sidePanel.props.isOpen).length === sidePanels.length) {
-    return null;
-  }
-
```
tristanjasper, line 36: Removing this as currently the parent panel doesn't animate close when rendered in a `Panel.Group`. This also means it doens't trigger an `onAfterClose` callback, which we require to change the application url in Workflows UI

Can be seen [here](http://storybooks.highbond-s3.com/paprika/master/?path=/story/messaging-panel-examples--group-panel)

packages/Panel/src/components/Dialog/Dialog.js
```diff
@@ -48,7 +48,6 @@ function Dialog(props) {
       </sc.DialogContent>
     </>
   );
-
   React.useEffect(() => {
     setIsAnimating(true);
   }, [isOpen]);
```
AndreyChernykh, line 53: Do we still need this effect then?

tristanjasper: Well it gets called on mount, and any time the `isOpen` prop changes, which comes from the Panel state, which changes whenever you open or close it, so without doing any larger refactor I would say yes

AndreyChernykh (approved):

PR #1172, opened 2021-11-08
feat(DataTable): UXD-1833 refactor row height calculation

allison-c:
### Purpose 🚀

### Notes ✏️

Refactor the row height calculation helper function and fix the bug with loading more data

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/DataTable/src/helpers/rowHeightHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+const DEFAULT_STYLES = "padding: 5px;border-right:1px solid;border-bottom:1px solid;";
+
+const containerId = `paprika-data-table__test-area--${Date.now()}`;
+
+interface ConstructorOptions {
+  defaultHeight?: number;
+  styles?: string;
+}
+
+export default class RowHeightHelper {
```
tristanjasper, line 14: Was there a need for this to be a class?

packages/DataTable/src/components/InfiniteLoader/InfiniteLoader.tsx
```diff
@@ -3,8 +3,11 @@
 import React from "react";
 import { ListChildComponentProps, VariableSizeList } from "react-window";
 import ReactInfiniteLoader from "react-window-infinite-loader";
-import calculateRowHeight from "../../helpers/calculateRowHeight";
-import { TableDataItemType } from "../../types";
+import RowHeightHelper from "../../helpers/rowHeightHelper";
+import { useReactTableContext } from "../ReactTableContext";
+import { TableColumnsWidth, TableDataItemType } from "../../types";
+
+const calculateRowHeight = new RowHeightHelper().calculate;
```
AndreyChernykh, line 10: You shouldn't use class methods as standalone callbacks. 
It may work in this case, but it is will fails if you use `this.` and generally is discouraged.

```js
class RowHeightHelper {
  constructor() {
    this.myVar = 42;
  }

  getMyVar() {
    return this.myVar;
  }
}

const helper = new RowHeightHelper();
const getMyVar = helper.getMyVar;

console.log(helper.getMyVar()); // 42
console.log(getMyVar()); // Cannot read properties of undefined (reading 'myVar')
``` 

So, 

```js
cons  rowHeightHelper = new RowHeightHelper();


rowHeightHelper.calculateHowHeight()
```

packages/DataTable/src/components/InfiniteLoader/InfiniteLoader.tsx
```diff
@@ -47,23 +50,49 @@ export function InfiniteLoaderImpl({
...
+        },
+        {} as TableColumnsWidth
+      ),
+    [allColumns]
+  );
 
   function getItemSize(index: number): number {
     if (!rowHeights.current[index]) {
-      const newRowHeight = calculateRowHeight(data[index]);
+      const newRowHeight = calculateRowHeight({
```
AndreyChernykh, line 70: I'm concerned that is may become a performance bottleneck the fact that we `append/remove` real DOM elements for every element.

Let's have a call and benchmark it together

packages/DataTable/src/helpers/rowHeightHelper.ts
allison-c: Because users can potentially customize the default values, so `new Helper(options)` may make more sense

PR #1175, opened 2021-11-12
SNP-2124 add localized columns and rows on DataGrid

ShanAtDiligent:
### Purpose 🚀
Add Localized columns and rows label on data grid
https://aclgrc.atlassian.net/browse/SNP-2124

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/SNP-2124-Datagrid-fix-localization-of-rows-columns

### Screenshots 📸
Screen Shot 2021-11-12 at 10 35 29 AM


packages/DataGrid/src/DataGrid.js
```diff
@@ -528,7 +528,7 @@ const DataGrid = React.forwardRef((props, ref) => {
       </sc.Grid>
       <sc.Footer $width={gridWidth}>
         <sc.RowCount>
-          Rows:{rowCount} Columns:{columnCount}
+          {i18n.t("dataGrid.rows")}:{rowCount} {i18n.t("dataGrid.columns")}:{columnCount}
```
mikrotron, line 531: I think it would be better to interpolate the counts into the i18n key.  The `:` is probably not used for every locale, or the spacing may vary – for example in English there should be space after it, like "Rows: 12", but in French, there should be a space before too, like "Les lignes : 12".

.changeset/unlucky-ladybugs-laugh.md
```diff
@@ -0,0 +1,6 @@
+---
+"@paprika/data-grid": patch
+"@paprika/l10n": patch
```
mikrotron, line 3: @allison-c Will this result in a bump to the `peerDependencies` of the `data-grid`? Or will that need to be updated manually? 

When the L10n team makes their subsequent update, are the dependencies bumped automatically as part of that process?

allison-c: Yes, bumping `@paprika/l10n` will bump all the packages having it as a peer dependency. However, we don't need to bump data-grid once again when l10n is bumped, cause they're releasing at the same time

When the L10n team makes their subsequent update, we should help them to add the `changeset` on their PR, so the @paprika/l10n can bump again and almost all the packages(since they have l10n as peerdependency) will bump again

mikrotron (dismissed):
Oops, I jumped the gun, it looks like a cypress spec needs an update first.

mikrotron (approved):
xmas llgtm 4

PR #1178, opened 2021-11-17
UXD-1717  Filter Breadcrumb children  🐐

mikrotron:
### Purpose 🚀
Support `null` children and children wrapped in `<React.Fragments>`.

### Notes ✏️
- Solves for use cases where breadcrumb links are conditionally rendered.
- use `extractChildren()` helper to filter `children` to only `<Breadcrumb.Link>` children.
- Added two examples in the Storybook Variations story.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1717--filter-breadcrumb-children

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1717
https://github.com/acl-services/acl-exception/pull/8747/files#r750731213

### DIFF
--- a/packages/Breadcrumbs/src/Breadcrumbs.js
+++ b/packages/Breadcrumbs/src/Breadcrumbs.js
@@ -13,14 +13,15 @@ function Breadcrumbs(props) {
   const { children, isDark, isAutoCollapsed, ...moreProps } = props;
   const I18n = useI18n();
 
-  const [isCollapsed, setIsCollapsed] = React.useState(shouldShowExpandButton);
   const linkChildren = [extractChildren(children, ["Breadcrumbs.Link"])["Breadcrumbs.Link"]].flat();
   const hasOnlyOneChild = linkChildren.length === 1;
   const shouldShowExpandButton = isAutoCollapsed && linkChildren.length > MAXIMUM_NUM_OF_LEVEL;
 
+  const [isCollapsed, setIsCollapsed] = React.useState(shouldShowExpandButton);
+
   React.useLayoutEffect(() => {
     setIsCollapsed(shouldShowExpandButton);
-  }, [linkChildren.length, shouldShowExpandButton]);
+  }, [shouldShowExpandButton]);
 
   function handleExpand() {
     setIsCollapsed(false);
@@ -33,13 +34,15 @@ function Breadcrumbs(props) {
           {linkChildren.map((child, index) => {
             if (shouldShowExpandButton && index === 0) {
               return (
-                <>
+                // eslint-disable-next-line react/no-array-index-key
+                <React.Fragment key={index}>
                   {child}
                   <ExpandButton onClick={handleExpand} isHidden={!isCollapsed} />
-                </>
+                </React.Fragment>
               );
             }
-            return React.cloneElement(child, { hasOnlyOneChild });
+            // eslint-disable-next-line react/no-array-index-key
+            return React.cloneElement(child, { hasOnlyOneChild, key: index });
           })}
         </sc.List>
       </sc.Nav>
--- a/packages/Breadcrumbs/src/Breadcrumbs.js
+++ b/packages/Breadcrumbs/src/Breadcrumbs.js
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import useI18n from "@paprika/l10n/lib/useI18n";
+import { extractChildren } from "@paprika/helpers";
 import Link from "./components/Link";
 import ExpandButton from "./components/ExpandButton";
 import IsDarkContext from "./context";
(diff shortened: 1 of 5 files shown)


packages/Breadcrumbs/src/Breadcrumbs.js
```diff
@@ -11,14 +12,15 @@ import * as sc from "./Breadcrumbs.styles";
...
+
   const [isCollapsed, setIsCollapsed] = React.useState(shouldShowExpandButton);
+  const linkChildren = [extractChildren(children, ["Breadcrumbs.Link"])["Breadcrumbs.Link"]].flat();
+  const hasOnlyOneChild = linkChildren.length === 1;
+  const shouldShowExpandButton = isAutoCollapsed && linkChildren.length > MAXIMUM_NUM_OF_LEVEL;
 
   React.useLayoutEffect(() => {
     setIsCollapsed(shouldShowExpandButton);
-  }, [childrenCount, shouldShowExpandButton]);
+  }, [linkChildren.length, shouldShowExpandButton]);
```
AndreyChernykh, line 23: Why do we need an effect here in the first place? It looks like a derived initial state to me.

packages/Breadcrumbs/src/Breadcrumbs.js
```diff
@@ -28,7 +30,7 @@ function Breadcrumbs(props) {
     <IsDarkContext.Provider value={isDark}>
       <sc.Nav aria-label={I18n.t("breadcrumbs.aria_label")} isDark={isDark} {...moreProps}>
         <sc.List isCollapsed={isCollapsed}>
-          {React.Children.map(children, (child, index) => {
+          {linkChildren.map((child, index) => {
```
AndreyChernykh, line 33: shouldn't we have a `key` somewhere there?

packages/Breadcrumbs/src/Breadcrumbs.js
```diff
@@ -11,14 +12,15 @@ import * as sc from "./Breadcrumbs.styles";
 function Breadcrumbs(props) {
   const { children, isDark, isAutoCollapsed, ...moreProps } = props;
   const I18n = useI18n();
-  const childrenCount = React.Children.count(children);
-  const shouldShowExpandButton = isAutoCollapsed && React.Children.count(children) > MAXIMUM_NUM_OF_LEVEL;
-  const hasOnlyOneChild = childrenCount === 1;
+
   const [isCollapsed, setIsCollapsed] = React.useState(shouldShowExpandButton);
+  const linkChildren = [extractChildren(children, ["Breadcrumbs.Link"])["Breadcrumbs.Link"]].flat();
```
allison-c, line 17: what's the difference between this and `extractChildren(children, ["Breadcrumbs.Link"])["Breadcrumbs.Link"]` ? For removing null values?

packages/Breadcrumbs/src/Breadcrumbs.js
```diff
@@ -11,14 +12,15 @@ import * as sc from "./Breadcrumbs.styles";
...
+  const hasOnlyOneChild = linkChildren.length === 1;
+  const shouldShowExpandButton = isAutoCollapsed && linkChildren.length > MAXIMUM_NUM_OF_LEVEL;
```
AndreyChernykh, line 19:
```js
function Breadcrumbs({ initialmaxNumLevels }: { initialmaxNumLevels = Infinity }) {
  const [maxNumLevels, setMaxNumLevels] = React.useState(initialmaxNumLevels);
  const isCollapsed = React.Children.count(children) > maxNumLevels;
  
  function handleExpand() {
    setMaxNumLevels(Infinity);
  }
}
```

AndreyChernykh:
```js
const links = [<Breadcrumbs.Link />, <Breadcrumbs.Link />];

React.useEffect(() => {
  fetchSomeLinks().then(data => links.push(data.map(link => <Breadcrumbs.Link link={link}/>)))
}, [])

<Breadcrumbs initialmaxNumLevels={2}> 
  {links}
</Breadcrumbs>
```

after the fetch is done, the `links.length` will become `2 + N` so
`const isCollapsed = React.Children.count(children) > maxNumLevels;` will become true

after we click `handleExpand`, we set `setMaxNumLevels(Infinity);` -> `const isCollapsed = React.Children.count(children) > maxNumLevels;` becomes false

packages/Breadcrumbs/src/Breadcrumbs.js
allison-c: I just want to recalculate the collapsed state when the number of children has been changed, like the example you mentioned above

packages/Breadcrumbs/src/Breadcrumbs.js
mikrotron: if there's only one `<Breadcrumbs.Link>`, then `linkChildren` would be set to that component.  If there are multiple, then `linkChildren` would be an array.  

With my version here, it will always be an array, even if there's only one item.

packages/Breadcrumbs/src/Breadcrumbs.js
mikrotron: I like this API better, it's more flexible and potentially less error prone.  But I'm not sure it's worth breaking the API for it.  

Instead I've mocked out an async scenario in Storybook, and can verify it still behaves correctly without the somewhat awkward `useLayoutEffect()`.

allison-c: If you click on the `...` before the data loaded in this example http://storybooks.highbond-s3.com/paprika/UXD-1717--filter-breadcrumb-children/?path=/story/navigation-breadcrumbs-examples--dynamic
You can see the breadcrumbs remain expanded after loading
That was what I want to avoid in the initial version, but I'm ok with the current behavior as long as we're aware of it 😄

allison-c (approved):

packages/Breadcrumbs/src/Breadcrumbs.js
mikrotron: I double checked with @carolchaan:

> I think what you and Andriy have is fine cause it's an extra step for them to click the expand again and there's no way to undo an expand anyway

PR #1191, opened 2021-11-30
UXD-1874 fix paprika build translations

AndreyChernykh:
### Purpose 🚀
The script did have a race condition because  `removeTranslationFiles` and `generateTranslationFiles` were called at the same time without waiting for one another. So, some files were removed immediately after they were created. This lead to an error, such as: 

`ModuleNotFoundError: Module not found: Error: Can't resolve '../locales-js/zh'`

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1874


.changeset/slow-starfishes-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paprika/build-translations": major
+---
+
+Use "fs" async API to avoid race condition
```
mikrotron, line 5: Do we have to use this script differently now? Like, will this have to change?
https://github.com/acl-services/sriracha/blob/master/src/GlobalNavigation/package.json#L18

Otherwise, it's not really a `major` (breaking) change, is it?

mikrotron (approved):
xmas llgtm 2

PR #1192, opened 2021-11-30
UXD-1877 upddate @action-bar SeachInput api

AndreyChernykh:
### Purpose 🚀

Add `initialValue` prop to the @action-bar's `<SearchInput />`

### Updates 📦

- Add `initialValue` prop to the `<SearchInput />`
- Fix the debouncing in `<SearchInput />`
- Rename `debouncedValue` to `debounceDelay`

#### Migration Guide

1. Change `<SearchInput debouncedValue={value} />` to `<SearchInput debounceDelay={value} />`

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1877-add-initial-value-to-action-bar-search

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1877


packages/ActionBar/src/components/SearchInput/SearchInput.js
```diff
@@ -6,18 +6,23 @@ import SearchIcon from "@paprika/icon/lib/Search";
 import useI18n from "@paprika/l10n/lib/useI18n";
 import * as sc from "./SearchInput.styles";
 
-export default function SearchInput(props) {
-  const { a11yText, debouncedValue, onChange, placeholder, ...moreProps } = props;
+function useDebouncer(debounceDelay) {
```
tristanjasper, line 9: perhaps this could be a helper function on our `paprika/helpers` package as perhaps may not be the only place we will need this approach

AndreyChernykh: you're right

tristanjasper (approved):

.changeset/angry-pandas-jog.md
```diff
@@ -0,0 +1,15 @@
+---
+"@paprika/action-bar": major
+"@paprika/helpers": minor
+---
+
+- Add `initialValue` prop to the `<SearchInput />`
+- Fix the debouncing in `<SearchInput />`
+- Rename `debouncedValue` to `debounceDelay`
+- Add `useCallbackDebouncer`
```
mikrotron, line 9: This is okay too, but even _better_ would be to run `yarn changeset` two times.  That will create two `md` files, and then each can be more focussed - so this "Add `useCallbackDebouncer`" note could be applied just to the changeset for `@paprika/helpers`.

AndreyChernykh: Done

mikrotron (approved):
llgtm 61

PR #1197, opened 2021-12-06
UXD-1888 update ResizeDetector and useDeboucedCallback

AndreyChernykh:
### Purpose 🚀

Migrate the package to TS

### Updates 📦

- `isFullWidth` default value is set to `false` now
- `useDimensions()` and `useBreakpoints()` were merged into one hook `useResizeDetectorContext()`
- `size` was renamed to `breakpointSize`

Before:

```js
import { useDimensions, useBreakpoints } from "../src";
function MyComponent() {
  const { width, height } = useDimensions();
  const { size } = useBreakpoints();
```
After:
```ts
import { useResizeDetectorContext } from "../src";
function MyComponent() {
  const { width, height, breakpointSize } = useResizeDetectorContext();
```
- `@paprika/resize-detector` now supports render prop `children`
```tsx
import ResizeDetector, { useResizeDetectorContext, ResizeDetectorContextValue } from "@paprika/resize-detector";
function MyComponent() {
  return (
    <ResizeDetector>
      {({ width, height }: ResizeDetectorContextValue) => <div>{width} x {height}</div>}
    </ResizeDetector>;
  );
}
```

### Storybook 📕

### Screenshots 📸

### References 🔗
[UXD-1888
](https://aclgrc.atlassian.net/browse/UXD-1888)


allison-c:
Just curious why changing `isFullWidth` default value to `false`?

AndreyChernykh:
@allison-c I believe that having boolean values to have a default  of `false` is more conventional. If the default value is `false` then:

```js
<MyComponent isFlag /> // isFlag === true
<MyComponent /> // isFlag === false
```

However, if the flag is `true` by default, then we need to do this: 

```
<MyComponent /> // isFlag === true
<MyComponent isFlag={false} /> // isFlag === false
```

If we really want to have `true` by default, then one approach  is to re-phrase the prop. E.g. instead of `isVisible: true` -> `isHidden: false`
But in this particular case we have 2 flags that are related: `isFullWidth` and `isFullHeight`. Rephrasing  one them would be hard and will probably lead to an inconsistent naming. And having them with one `true` by default and the other `false` by default is kinda weird 🤷‍♂️ 

@mikrotron cc

packages/ResizeDetector/src/ResizeDetector.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+    const { width, height } = refContainer.current.getBoundingClientRect();
+    setDimensions({ width, height });
+  }, [breakpointSmall, breakpointLarge, refContainer]);
+
+  const contextValue = React.useMemo(() => ({ ...dimensions, breakpointSize }), [dimensions, breakpointSize]);
+
+  return (
+    <sc.ResizeDetector
+      data-pka-anchor="resize-observer"
```
allison-c, line 107: I believe `resizeDetector` is the new convention

allison-c:
Looks good to me overall, but I'm not very familiar about the breakpoint use cases, could you take a look as well please @mikrotron

packages/ResizeDetector/src/ResizeDetector.tsx
```diff
@@ -0,0 +1,122 @@
...
+  ...moreProps
+}: ResizeDetectorProps): JSX.Element {
+  const refContainer = React.useRef<HTMLSpanElement>(null);
+  const debounceCallback = useCallbackDebouncer(debounceDelay);
+  const [dimensions, setDimensions] = React.useState<Dimensions>({ width: 0, height: 0 });
+  const breakpointSize = getBreakpointSize(dimensions.width, breakpointSmall, breakpointLarge);
+
+  useResizeDetector({
+    onResize: ({ width, height }) => {
+      console.log({ width, height });
```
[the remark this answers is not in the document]
AndreyChernykh: oops =)

packages/ResizeDetector/src/ResizeDetector.tsx
```diff
@@ -0,0 +1,122 @@
...
+  debounceDelay: number;
+
+  /** If the container will match its parent's width like a block level element (width: 100%). */
+  isFullWidth: boolean;
+
+  /** If the container will match its parent's height (height: 100%). */
+  isFullHeight: boolean;
+
+  /** Callback that fires when the size change crosses a breakpoint threshold (returns new T-Shirt size value).  */
+  onBreak?: (size: string) => void;
```
mikrotron, line 34:
```suggestion
  onBreak?: (breakpointSize: string) => void;
```

packages/ResizeDetector/src/ResizeDetector.tsx
mikrotron:
```suggestion
      data-pka-anchor="resize-detector"
```

packages/ResizeDetector/stories/ResizeDetector.examples.stories.js
```diff
@@ -72,7 +72,7 @@ callbacks.story = {
 export const debounce = () => (
   <ExampleStory storyName="Debounce Examples" component="ResizeDetector" fileName="examples/Debounce.js">
     <p>
-      <code>debounceDelay = 300</code>
+      <code>debounceDelay = 30</code>
```
mikrotron, line 75: It's deliberately slooooow (300)
https://github.com/acl-services/paprika/blob/master/packages/ResizeDetector/stories/examples/Debounce.js#L9

packages/ResizeDetector/src/ResizeDetector.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  /** Callback that fires when the size changes (returns new width + height values).  */
+  onResize?: (dimensions: Dimensions) => void;
+}
+
+const MAX_WAIT = 1000;
+
+const ResizeContext = React.createContext({ width: 0, height: 0, breakpointSize: ShirtSizes_.MEDIUM });
+
+export function useResizeDetectorContext(): ResizeDetectorContextValue {
```
mikrotron, line 44: How about just `useResizeDetector()`? 

Generally I prefer shorter when possible – though sometimes too short is too ambiguous (like `size` 😄)

packages/ResizeDetector/stories/examples/Debounce.js
```diff
@@ -6,7 +6,7 @@ import ResizeDetector from "../../src";
 export function DebounceSlowStory() {
   return (
     <Resizer initWidth={360} initHeight={64} axis="x">
-      <ResizeDetector debounceDelay={300}>
+      <ResizeDetector isFullWidth debounceDelay={30}>
```
mikrotron, line 9: Oh, right, you changed it here too 🤦 

I actually deliberately made it really slow so the lag was observable.  But 🤷‍♂️

packages/ResizeDetector/src/ResizeDetector.tsx
AndreyChernykh: agree

mikrotron:
> @allison-c I believe that having boolean values to have a default of `false` is more conventional. If the default value is `false` then:
> 
> ```js
> <MyComponent isFlag /> // isFlag === true
> <MyComponent /> // isFlag === false
> ```
> 
> However, if the flag is `true` by default, then we need to do this:
> 
> ```
> <MyComponent /> // isFlag === true
> <MyComponent isFlag={false} /> // isFlag === false
> ```
> 
> If we really want to have `true` by default, then one approach is to re-phrase the prop. E.g. instead of `isVisible: true` -> `isHidden: false` But in this particular case we have 2 flags that are related: `isFullWidth` and `isFullHeight`. Rephrasing one them would be hard and will probably lead to an inconsistent naming. And having them with one `true` by default and the other `false` by default is kinda weird 🤷‍♂️
> 
> @mikrotron cc

I partially agree – I also _PREFER_ that boolean props are `false` by default.  But we do have exceptions in our system, and if the rephrased name would be harder to understand, then I'd prefer to use a `true` default value.  `isFullWidth` is easy to understand and has a precedent in other components, so I think we should keep that name.  

But I also like that this component behaves like a `block` level `<div>` element by default.  I think that's a better, more intuitive default behaviour than an `inline-block` / `<span>`. I think it's also a more common use case.  

It's not perfect, I don't really like that `isFullWidth=true` and `isFullHeight=false` by default, and would prefer that booleans are false be default, but I think clear naming, and sensible default behaviour is more important.  

Are you convinced?

AndreyChernykh:
OK, I'll change `isFullWidth` back to `true` by default

mikrotron (approved):
xmas llgtm 3

PR #1206, opened 2021-12-14
RNG-293 add container prop to Panel

YuriGal:
### Purpose 🚀
Support mounting the React portal of the `<Panel>` within a Shadow DOM by introducing a `container` prop.

### Notes ✏️
- Dependent PR for Sriracha `<GlobalNav>`: https://github.com/acl-services/sriracha/pull/916

### References 🔗
https://aclgrc.atlassian.net/browse/RNG-293


packages/Panel/src/Panel.js
```diff
@@ -272,6 +279,7 @@ const defaultProps = {
   slideFrom: types.slideFroms.RIGHT,
   width: "33%",
   zIndex: zValue(7),
+  portalMount: document.body,
```
mikrotron, line 282: I think this can be `null` by default.  If we restore the `<Portal>` helper component, then it will go back to deferring to `document.body` by default if no `container` is provided.

package.json
```diff
@@ -128,13 +128,6 @@
     "wait-on": "^3.2.0",
     "workerize-loader": "^1.1.0"
   },
-  "husky": {
-    "hooks": {
-      "commit-msg": "node scripts/commit-validator.js",
-      "pre-commit": "lint-staged",
-      "pre-push": "yarn jest"
-    }
-  },
```
mikrotron, line 137: Looks like you were struggling with our precommit validation.  Likely the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) standard that's being enforced.  It's a bit inconvenient, but enables some automation features.  

In your case, you should be able to commit with a message like "feat(Panel): passing portalMount to Panel" or "chore(Paprika): restore husky config" 😉 

And if you're really stuck by this, you can commit with `--no-verify` to skip the hooks.

mikrotron:
Please also run `yarn changeset` to include a message about the changes for the CHANGELOG, and also to inform `lerna` of the scope of the changes (PATCH, MINOR, MAJOR).  In this case I think this is a MINOR change.

allison-c (approved):

PR #1207, opened 2021-12-14
UXD-1898 add Resizer to the DataTble

AndreyChernykh:
### Purpose 🚀

Add an automatic height and width calculation

### Notes ✏️
- `height` prop is not optional
- if `<DataTable.ResizeContainer>` is not used and `height` or `width` is undefined, the warning will be shown in the console
- `height` or `width` will override any automatic calculation

### Storybook 📕

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-1898


packages/DataTable/src/DataTable.tsx
```diff
@@ -143,29 +157,45 @@ const DataTable: React.FC<DataTableProps> & DataTableComposition = ({
...
+  function renderTable(maxHeight: number) {
+    return (
       <sc.Table
+        ref={tableRef}
         aria-label={a11yText}
         aria-rowcount={data.length}
         data-pka-anchor="dataTable"
         width={width}
-        height={height}
+        height={getTableHeight(maxHeight)}
```
allison-c, line 168: Can we just use `height` instead of `getTableHeight (maxHeight)`  when there's no `extractedResizeContainer ` ?

packages/DataTable/src/hooks/useTableHeight.ts
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+
+export default function useTableHeight(
+  tableRef: React.RefObject<HTMLDivElement>,
+  itemCount?: number
+): { getTableHeight: (maxHeight: number) => number } {
+  const tableHeightRef = React.useRef<number | undefined>();
+
+  React.useLayoutEffect(() => {
+    if (!tableRef.current) {
```
allison-c, line 10: Is it ok if we just return `maxHeight` when the `itemCount` is larger than `100` ? just for reduce the calculation

AndreyChernykh: Actually, the calculation should only occur once per `itemCount` change. 
What we could do, though, is to skip the calculation if the `tableHeightRef.current` is already set and itemCount is more than 100 (for instance)

tristanjasper:
Would we be able to build in functionality if we want to keep certain columns flexible widths but the table width to fill the viewport, to avoid what we are doing [here](https://github.com/acl-services/transactional-workflow/blob/integration/services/workflows_ui/src/components/events/EventList/EventList.tsx#L34)

AndreyChernykh:
@allison-c ☝️

packages/DataTable/src/hooks/useTableDimensions.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (!tbodyEl || !theadEl) {
+      return;
+    }
+
+    setDimensions(prev => {
+      const bodyRect = tbodyEl.getBoundingClientRect();
+      const headRect = theadEl.getBoundingClientRect();
+
+      return {
+        width: bodyRect.width + headRect.width,
```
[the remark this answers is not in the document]
AndreyChernykh: oops

allison-c (approved):

allison-c (approved):

PR #1208, opened 2021-12-16
UXD-1904 wrap divisions

michellehuang7:
### Purpose 🚀
Replace divisions to calc()

### Notes ✏️
Some of the calculations involved with variable divisions needs to be wrapped with calc as well

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1890--wrap-division

### References 🔗
JIRA: https://aclgrc.atlassian.net/browse/UXD-1890


packages/Tokens/src/tokens.yaml
```diff
@@ -203,8 +203,8 @@ counter:
   medium: 19px
 #
 pill:
-  small-radius: 8px / 50%
-  medium-radius: 12px / 50%
+  small-radius: calc(8px / 50%)
+  medium-radius: calc(12px / 50%)
```
DanilAgafonov, line 207: I just realized that SCSS compiler thinks that it is division. But actually it is just a syntax for `border-radius`. I found a workaround here: https://github.com/sass/sass/issues/667

I think we should apply this workaround instead of `calc(...)`

DanilAgafonov: Side note: It is the main reason why using `/` for division is deprecated in Dart Sass compiler - because it is hard to differentiate division and CSS syntax.

packages/Stylers/src/helpers.scss
```diff
@@ -6,9 +6,39 @@
 
 // Utilities
 
+@function div($dividend, $divisor, $precision: 12) {
```
mikrotron, line 9: Did you reference an algorithm you found online for this? If so, can we include a comment with the link?

packages/Stylers/stories/helpers.scss
```diff
@@ -70,7 +70,7 @@
       @for $i from 0 through 7 {
         &:nth-child(#{$i + 1}) {
           left: $offset * $i;
-          top: $offset / 4 * $i;
+          top: calc(#{$offset / 4 * $i});
```
DanilAgafonov, line 73: `calc()` means value will be evaluated in runtime. We can it in compilation time. We just need to express division through multiplication: `$offset * 0.25 * $i`

DanilAgafonov (approved):

PR #1216, opened 2022-01-06
UXD-309 refactor to CSF (Checkbox, Confirmation, Switch)

slee1130:
### Purpose 🚀

Stories needs to be refactored with the CSF

### Notes ✏️

CSF Stories
Refactor the stories for a component to use CSF instead of storiesOf syntax.
Reference: Component Story Format (CSF)

- Checkbox
- Confirmation
- Switch

### Updates 📦
If you have changed a component's source code (not stories, specs, or docs), before merging your branch run `yarn changeset`. This will prompt you to:
- indicate if changes were patch/minor/major for each modified package
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/xd-309-RefactorCSF2

### Screenshots 📸
_optional but highly recommended_

### References 🔗

https://aclgrc.atlassian.net/browse/UXD-309

PR #1217, opened 2022-01-08
UXD-309 refactor to CSF (Modal, Overflow, Takeover)

slee1130:
### Purpose 🚀

Stories needs to be refactored with the CSF

### Notes ✏️

CSF Stories
Refactor the stories for a component to use CSF instead of storiesOf syntax.
Reference: Component Story Format (CSF)

- Modal 
- Overflow
- **Takeover (I've removed kind in <Takeover.Header> since it's not working.. I am not sure we will keep Kind in header or not but if we don't need then we might need to have a separate ticket to remove kind in Header in Takeover.)**

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/xd-309-RefactorCSF3

### Screenshots 📸

### References 🔗

https://aclgrc.atlassian.net/browse/UXD-309


packages/Modal/stories/Modal.examples.stories.js
```diff
@@ -0,0 +1,208 @@
...
+      <Modal.Content>
+        {repeat(25, key => (
+          <p key={key}>Some content here</p>
+        ))}
+      </Modal.Content>
+    </ModalStory>
+  </ExampleStory>
+);
+
+basic.story = exampleStoryParameters;
```
mikrotron, line 38: This doesn't break, but doesn't work as intended, unless the `exampleStoryParameters` are applied to `basic.story.parameters`.  Like this:

https://github.com/acl-services/paprika/pull/1216/files#diff-ee0bffb54044d1dae68748c994226281a259cb7613594c6b61e6a1d0413442efR23

packages/OverflowMenu/stories/OverflowMenu.examples.stories.js
```diff
@@ -0,0 +1,90 @@
...
+
+locale.story = exampleStoryParameters;
+
+export const modals = () => (
+  <ExampleStory component="OverflowMenu" storyName="Modals" fileName="examples/OverflowMenuMultiConfirmationExample.js">
+    <OverflowMenuMultiConfirmationExample />
+  </ExampleStory>
+);
+
+modals.story = exampleStoryParameters;
```
mikrotron, line 46: While in general, I like the result of the story names in the left side navigation from this PR – the shorter names are easier to scan.

Old:
Screen Shot 2022-01-14 at 4 54 04 PM

New + Improved:
Screen Shot 2022-01-14 at 4 53 39 PM

The new version is derived from the variable names of your stories... so `fullWidthContent` becomes `Full Width Content` in Storybook.  But in this case "Modals" isn't quite cutting it, and the old version "with multiple confirmation modals" is a bit more helpful (though maybe it could still be shorter).  The trick I'd suggest here, is giving some stories a `name` override if you think it's needed to make it clear what the story is about.  

So, something like this could be used:

```js
modals.story = {
  name: "Multiple Confirmations",
  parameters: exampleStoryParameters,
}
```

I'll let you decide where else this idea could be used in this PR.

mikrotron:
Some Cypress tests are failing for the `<OverflowMenu>` (`OverflowMenu/test/overflowMenu.cypress.js`) and the `<Panel>` (`Panel/tests/Panel.cypress.js`).  Are you able to look into those, or do you need some help?

mikrotron:
The `<OverflowMenu>` could really benefit from a Screener story.  It's a bit of scope creep, but do you think you could add one where the menu is open by default? Like this:
https://github.com/acl-services/paprika/pull/1216/files#diff-63cfaba5d33c2b651576e33c0292e45291aa897e4b38c2aa9dd7500d311870c7

packages/OverflowMenu/stories/examples/OverflowMenuExample.js
```diff
@@ -16,7 +16,7 @@ const OverflowMenuExample = () => (
   <OverflowMenu align="bottom">
     <OverflowMenu.Content data-your-custom-attribute="your-custom-attribute-value" />
     <OverflowMenu.Trigger
-      data-pka-anchor="overflow-menu__trigger"
+      id="overflow-menu__trigger"
```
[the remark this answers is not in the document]
mikrotron: This was bad advice.  I see now that Cypress is configured so that `cy.findByTestId()` will look for the `data-pka-anchor` attribute.  
https://github.com/acl-services/paprika/blob/de413ae7b2c329bf6b098bf9757c003779a49c8d/cypress/support/index.js#L20

So it _does_ make sense to use `data-pka-anchor` here instead of `id`. And that you should revert your changes in [overflowMenu.cypress.js](https://github.com/acl-services/paprika/pull/1217/files#diff-538e4cda82acc833604036694719fdd2daeba9841c4f423ab29893bf71d7e3e1)

mikrotron (approved):
llgtm 03

PR #1220, opened 2022-01-14
UXD-1634 collapsible refactor to styled components

ShanAtDiligent:
### Purpose 🚀
Refactor collapsible component to styled components

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1634-Collapsible-refactor-to-styled-components2

### Screenshots 📸
Screen Shot 2022-01-13 at 5 43 34 PM


.changeset/pink-melons-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paprika/collapsible": patch
+---
+
+refactor styles to follow latest Paprika conventions for styled-components
```
mikrotron, line 5: I might consider this a `minor` update... definitely we should include the fact that classnames internal elements have been removed and `pka-anchors` should be used instead – this could break automation tests or styling overrides.

mikrotron:
Oh! There's a Screener interaction test that is failing now (we only have a few of these)! It just needs an update to the selectors:

https://github.com/acl-services/paprika/blob/master/packages/ProgressAccordion/stories/examples/Screener.js#L9-L10

michellehuang7 (approved):

PR #1228, opened 2022-02-09
refactor(DataTable): TW-2556 initial size refactor

allison-c:
### Purpose 🚀

### Notes ✏️

- Changed the `width` and `height` prop to the meaning of max width and max height. Also changed their type to string. Defaults are `100vw` and `100vh`
- Added a new prop `shouldResizeWithViewport ` 
- Added a imperative handle `ref.current.resize`
- Added storybook example for collapsible
- and some tests

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/TW-2556-new-props

### Screenshots 📸

### References 🔗


packages/DataTable/src/DataTable.tsx
```diff
@@ -149,59 +155,47 @@ const DataTable: React.FC<DataTableProps> & DataTableComposition = ({
...
-};
+}
+
+declare module "react" {
+  function forwardRef<T, P = unknown>(
+    render: (props: P, ref: React.RefObject<T>) => React.ReactElement | null
+  ): React.FC<DataTableProps> & DataTableComposition;
+}
+
+const DataTableWithRef = React.forwardRef(DataTable);
```
allison-c, line 192: Had to do this way to meet the typing requirement for composition components

packages/DataTable/src/components/InfiniteLoader/InfiniteLoader.tsx
```diff
@@ -92,6 +94,7 @@ export function InfiniteLoaderImpl({
           onItemsRendered={onItemsRendered}
           ref={listRef}
           innerElementType={innerElementType}
+          style={{ overflowX: shouldHaveHorizontalScroll ? "auto" : "hidden" }}
```
allison-c, line 97: When there's a vertical scrollbar but the horizontal scrollbar is not needed, the horizontal scrollbar is likely still there since the vertical one takes some horizontal space. We either:
1. make the `best width` bigger, first we need to calculate the scrollbar width, second this might cause the table has a bigger width than the `width` prop
2. set an `overflow-x: hidden`, then the last column might be cut off if it's too narrow, but I think a very narrow column in the last column is kind of an edge case, so I went with this 
3. just don't do anything 😂

packages/DataTable/src/hooks/useBestTableDimensions.ts
```diff
@@ -0,0 +1,74 @@
...
+        width: Math.min(maxWidthInNumber, realWidth),
+        height: Math.min(maxHeightInNumber, realHeight),
+        shouldHaveHorizontalScroll: maxWidthInNumber < realWidth,
+      };
+    });
+  }, [maxHeight, maxWidth, tableRef]);
+
+  React.useLayoutEffect(() => {
+    resetDimension();
+    const debouncedReset = debounce(resetDimension, 300);
```
allison-c, line 56: Had to use lodash function since the `useCallbackDebouncer` is for a different usage, see https://github.com/acl-services/paprika/blob/80361b22f7c8c81f716131966979a6673f0d49ea/packages/ActionBar/src/components/SearchInput/SearchInput.js#L19

packages/DataTable/src/DataTable.tsx
```diff
@@ -149,59 +155,47 @@ const DataTable: React.FC<DataTableProps> & DataTableComposition = ({
...
 
-DataTable.displayName = "DataTable";
-DataTable.InfiniteLoader = InfiniteLoader;
-DataTable.ResizeContainer = ResizeContainer;
-DataTable.types = {
+DataTableWithRef.displayName = "DataTable";
+DataTableWithRef.InfiniteLoader = InfiniteLoader;
+DataTableWithRef.ResizeContainer = ResizeContainer;
```
allison-c, line 196: Should I just remove this export `DataTableWithRef.ResizeContainer` ? cause the prop changes will be breaking changes anyway, but if I removed it we can't really extract the children component and show warnings

packages/DataTable/src/hooks/useBestTableDimensions.ts
```diff
@@ -0,0 +1,74 @@
...
+  width: maxWidth,
+  shouldResizeWithViewport,
+}: {
+  tableRef: React.RefObject<HTMLDivElement>;
+  componentRef: React.RefObject<unknown>;
+  height: string;
+  width: string;
+  shouldResizeWithViewport: boolean;
+}): Dimensions {
+  const [dimensions, setDimensions] = React.useState<Dimensions>({
```
AndreyChernykh, line 24: Maybe we could use `() => ` syntax here, so `convertSizeStringToNumber` is not called on every render redundantly

packages/DataTable/src/DataTable.tsx
AndreyChernykh: Yes, let's just remove it, and add a migration instructions to the changelog

packages/DataTable/src/components/InfiniteLoader/InfiniteLoader.tsx
mikrotron: #1 sounds potentially better, but more complicated, so I agree with #2 for now.  If we were to change later to #1, would it have any impact on the API?

allison-c: > would it have any impact on the API


I don't think so, it's an internal calculation anyway

packages/DataTable/src/helpers/convertSizeStringToNumber.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    if (!match || !match.groups) return false;
+
+    switch (index) {
+      case 0:
+        result = Number.parseInt(match.groups.numberInPx, 10);
+        break;
+      case 1:
+        result = calculateViewportSizeToPx(
+          Number.parseInt(match.groups[DirectionToGroupSelector[direction]], 10),
```
mikrotron, line 52: So then `convertSizeStringToNumber("100", "width")` should still work and return `100`, right?  

Is that a use case we want to _explicitly_ support, and indicate in the docs?
https://github.com/acl-services/paprika/pull/1228/files#diff-0c720a400684e34f0781322411e80251bc906ed4766234c4ca6b4a947a123e96R52-R56

packages/DataTable/src/hooks/useBestTableDimensions.ts
mikrotron: Maybe the `300` can be extracted to a `const` at the top?

packages/DataTable/src/DataTable.tsx
```diff
@@ -51,11 +49,11 @@ export interface DataTableProps {
...
 
-  /** The height of the table in px */
-  height?: number;
+  /** The max height of the table in px/vh/calc(xvh+/-xpx) */
+  height?: string;
 
-  /** The width of the table */
-  width?: string | number;
+  /** The max width of the table in px/vw/calc(xvw+/-xpx) */
+  width?: string;
```
mikrotron, line 56: Since we are breaking the API anyway, maybe we should rename these to `maxHeight` and `maxWidth` to be more explicit and understandable.

packages/DataTable/src/helpers/convertSizeStringToNumber.ts
allison-c: `"100"` will not work, the string has to match one of the regexes, but we can make it work for sure, we can create a follow-up ticket to support more format e.g. "100" "100%" ...

packages/DataTable/src/DataTable.tsx
```diff
@@ -72,29 +70,36 @@ export interface DataTableProps {
...
-  ...moreProps
-}: DataTableProps) => {
+function DataTable(
+  {
+    a11yText,
+    children = null,
+    columns,
+    data,
+    height = "100vh",
+    width = "100vw",
```
mikrotron, line 89: What would happen if these defaults were `null`? Sometimes we might actually want a table that has no max height (although that would make virtualization impossible).  

What about a max width of 100%? Would that be more sensible?

packages/DataTable/src/DataTable.tsx
```diff
@@ -72,29 +70,36 @@ export interface DataTableProps {
...
+    borderType = gridTypes.HORIZONTAL,
+    getRowHeight = null,
+    hasZebraStripes = false,
+    isHeaderSticky = true,
+    renderRow = null,
+    shouldResizeWithViewport = false,
```
mikrotron, line 95: Maybe this should have a default of `true`? I guess it might decrease performance in some cases when it's not needed, but how often do you think it will not be needed compared to how often it will be? I _suspect_ it will be needed most of the time...?

packages/DataTable/src/DataTable.tsx
allison-c: Can they use `999999999px` instead? 😂  I don't really want to set the defaults to `null` cause it might break the infinite loader as well
However, if we really want, I think technically we can do it

> What about a max width of 100%?

Again `100%` is not currently supported, we need to calculate the value from the parent DOM element, I think starting from vh and vw is more feasible and we can collect some feedback at the same time

packages/DataTable/src/DataTable.tsx
allison-c: Sure, I don't have a strong opinion or objection to it 😄

packages/DataTable/src/DataTable.tsx
allison-c: Yeah you're right, based on our design we always need it

packages/DataTable/src/DataTable.tsx
mikrotron: Okay, let's leave the `100v` defaults for now.

allison-c: Ticket created: https://aclgrc.atlassian.net/browse/UXD-1967

packages/DataTable/src/components/InfiniteLoader/InfiniteLoader.tsx
allison-c: Ticket created: https://aclgrc.atlassian.net/browse/UXD-1968

mikrotron (approved):
llgtm 76

PR #1235, opened 2022-03-04
RNG-271  Add getActiveElement helper to support web components 🐐

mikrotron:
### Purpose 🚀
Added `getActiveElement()` helper that can pierce the shadow DOM. 

### Notes ✏️
- Refactored `<CopyInput>`, `<ListBox>`, `<OverflowMenu>`, `<Popover>`, `<TimePicker>`, to use the new `getActiveElement()` instead of `document.activeElement`.  
- The motivation for this comes specifically from a bug when wrapping the `<GlobalNavigation>` in a web component where the `<OverflowMenu>` automatically closes after being open by keyboard.  
- A "Web Component" story has been added to the "Sandbox" section of the `<OverflowMenu>` in Storybook to verify the fix.  
- An unresolved issue has been noticed by not fixed – the focus ring is not visible on the `<Button>` component when wrapped in a web component because the following selector is not applied when included within a shadow DOM:
https://github.com/acl-services/paprika/blob/master/packages/Button/src/Button.styles.js#L58

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/RNG-271--activeelement-helper/?path=/story/buttons-overflowmenu-backyard-sandbox--web-component

### Screenshots 📸
Screen Shot 2022-03-04 at 2 11 04 PM

### References 🔗
https://aclgrc.atlassian.net/browse/RNG-271


jeremyhewett (dismissed):

packages/ListBox/src/components/Content/Content.js
```diff
@@ -15,17 +18,15 @@ const handleBlur = (state, dispatch, onCancelFooter) => () => {
...
+    if (state.refTriggerContainer.current && state.refTriggerContainer.current.contains(activeElement)) {
       return;
     }
 
-    if (
-      refListBoxContainer &&
-      refListBoxContainer.current &&
-      !refListBoxContainer.current.contains(document.activeElement)
-    ) {
+    if (refListBoxContainer && refListBoxContainer.current && !refListBoxContainer.current.contains(activeElement)) {
```
AndreyChernykh, line 29: I think `refListBoxContainer &&` is redundant 🤔

mikrotron: You're right, a user-triggered blur even should not be executed before the DOM ref is set.  But I also don't mind this defensive practice, though `refListBoxContainer?.current?.contains()` would be much more succinct.  

Still, my intention here isn't really to refactor that, it's just to replace `document.activeElement` with the new `getActiveElement()` helper.

PR #1237, opened 2022-03-10
fix(Uploader): UXD-1995 zindex issue and disabled issue

allison-c:
### Purpose 🚀
https://aclgrc.atlassian.net/browse/UXD-1995

### Notes ✏️

- Replaced `defaultIsDisabled` prop by `isDisabled` prop
- Added `zIndex` prop

#### why the change was made:

`defaultIsDisabled` can't be changed once mount.

#### 👷 Moving from 3.x.x to 4.x.x:

If you didn't use `defaultIsDisabled`, you don't need to do anything. Otherwise:

Before:

```jsx
<Uploader defaultIsDisabled={true} />
```

After:

```jsx
<Uploader isDisabled={true} />
```

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/Uploader/src/Uploader.js
```diff
@@ -96,6 +100,7 @@ const defaultProps = {
   onProcessed: () => {},
   onRequest: null,
   supportedMimeTypes: ["*/*"],
+  zIndex: 1,
```
mikrotron, line 103: This could probably be `null` since the `<Popover>` already defines a default value for its `zIndex`:
https://github.com/acl-services/paprika/blob/master/packages/Popover/src/Popover.js#L572

But the effect is the same, so 🤷‍♂️

.changeset/gentle-suits-scream.md
```diff
@@ -0,0 +1,26 @@
+---
+"@paprika/uploader": major
+---
+
+- Replaced `defaultIsDisabled` prop by `isDisabled` prop
+- Added `zIndex` prop
+
+#### why the change was made:
+
+`defaultIsDisabled` can't be changed once mount.
```
mikrotron, line 10: Does this imply that `isDisabled` is now a controlled prop, and can be changed at any time by the consumer?

mikrotron (approved):
llgtm 01

.changeset/gentle-suits-scream.md
allison-c: Right

packages/Uploader/src/Uploader.js
allison-c: 😂 🤷‍♂️

PR #1239, opened 2022-03-11
UXD-1978 takeover header style updates

ShanAtDiligent:
### Purpose 🚀
Update Takeover component base on the new design guideline.
ticket:
https://aclgrc.atlassian.net/browse/UXD-1978
https://aclgrc.atlassian.net/browse/UXD-1979
https://aclgrc.atlassian.net/browse/UXD-1981

### Notes ✏️
- need to update the overlay component so the backdrop isn't clickable.

### Updates 📦
Will run `yarn changeset` after footer is done.

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1978-Takeover-header-style-updates

### Screenshots 📸
Screen Shot 2022-03-10 at 5 35 24 PM
Screen Shot 2022-03-10 at 5 35 15 PM

Screen Shot 2022-03-10 at 5 37 32 PM

### DIFF
--- a/packages/Overlay/CHANGELOG.md
+++ b/packages/Overlay/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Changelog
 
-## 1.0.15
-
-### Patch Changes
-
-- add option to disable the click events on backdrop
-
 ## 1.0.14
 
 ### Patch Changes
--- a/packages/Overlay/CHANGELOG.md
+++ b/packages/Overlay/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.0.15
+
+### Patch Changes
+
+- add option to disable the click events on backdrop
+
 ## 1.0.14
 
 ### Patch Changes
--- a/packages/Takeover/src/components/Header/Header.js
+++ b/packages/Takeover/src/components/Header/Header.js
@@ -7,10 +7,15 @@ import * as sc from "./Header.styles";
 
 const Header = React.forwardRef((props, ref) => {
   const { children, level, hasCloseButton, kind, onClose, refHeading, tools, ...moreProps } = props;
-  const isString = typeof children === "string";
   return (
     <sc.Header ref={ref} kind={kind} {...moreProps}>
-      <sc.Heading level={level} displayLevel={3} isLight ref={refHeading} title={isString ? children : null}>
+      <sc.Heading
+        level={level}
+        displayLevel={3}
+        isLight
+        ref={refHeading}
+        title={typeof children === "string" ? children : null}
+      >
         {children}
       </sc.Heading>
       <sc.HeaderRightContainer>
--- a/packages/Takeover/src/components/Header/Header.js
+++ b/packages/Takeover/src/components/Header/Header.js
@@ -42,7 +42,7 @@ const propTypes = {
 
 const defaultProps = {
   hasCloseButton: true,
-  level: 3,
+  level: 2,
   kind: Header.types.kind.DEFAULT,
   onClose: () => {},
   refHeading: null,
--- a/packages/Takeover/src/components/Header/Header.js
+++ b/packages/Takeover/src/components/Header/Header.js
@@ -6,23 +6,26 @@ import * as types from "../../types";
 import * as sc from "./Header.styles";
 
 const Header = React.forwardRef((props, ref) => {
-  const { children, level, hasCloseButton, kind, onClose, refHeading, ...moreProps } = props;
+  const { children, level, hasCloseButton, kind, onClose, refHeading, headerTools, ...moreProps } = props;
 
   return (
     <sc.Header ref={ref} kind={kind} {...moreProps}>
       <sc.Heading level={level} displayLevel={3} isLight ref={refHeading}>
         {children}
       </sc.Heading>
-      {hasCloseButton && (
-        <sc.CloseButton>
-          <Button.Close
-            data-pka-anchor="takeover.header.close-button"
-            onClick={onClose}
-            isDark={kind === "primary" || null}
-            size="medium"
-          />
-        </sc.CloseButton>
-      )}
+      <sc.HeaderRightContainer>
+        {headerTools && <sc.ToolContainer>{headerTools}</sc.ToolContainer>}
+        {hasCloseButton && (
+          <sc.CloseButton>
+            <Button.Close
+              data-pka-anchor="takeover.header.close-button"
+              onClick={onClose}
+              isDark={kind === "primary" || null}
+              size="medium"
+            />
+          </sc.CloseButton>
+        )}
+      </sc.HeaderRightContainer>
     </sc.Header>
   );
 });
@@ -38,6 +41,8 @@ const propTypes = {
   level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
   onClose: PropTypes.func,
   refHeading: RefOf(),
+  /** Add node object to the right side of heading next to the close button */
+  headerTools: PropTypes.node,
 };
 
 const defaultProps = {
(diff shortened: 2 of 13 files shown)


packages/Takeover/src/components/Header/Header.styles.js
```diff
@@ -28,10 +28,32 @@ export const Header = styled.div(
...
+
+export const HeaderRightContainer = styled.div`
+  align-items: center;
+  display: flex;
+`;
+
+export const ToolContainer = styled.div(
+  () => css`
+    padding-right: ${spacer(2)};
+    button {
```
allison-c, line 48: Would `>*` selector be better? or maybe just let the consumer to handle this part, we don't know what will be inserted in the tool container

packages/Takeover/src/components/Header/Header.styles.js
```diff
@@ -28,10 +28,32 @@ export const Header = styled.div(
...
 export const Heading = styled(OriginalHeading)`
-  margin: 0 0 0 ${spacer(2)};
+  margin: 0;
+  padding: 18px 0 17px ${spacer(3)};
 `;
 
 export const CloseButton = styled.div`
   border-left: 1px solid ${tokens.border.color};
-  padding: ${spacer(1)};
+  padding: 16.5px ${spacer(2)} ${spacer(2)} 15.5px;
```
allison-c, line 37: Is there any difference if we just say `padding: ${spacer(2)};`

allison-c:
For some edge cases like this 
image
I think the grey bar should be the same height as the header

packages/Takeover/src/components/Header/Header.js
```diff
@@ -38,14 +41,17 @@ const propTypes = {
   level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
   onClose: PropTypes.func,
   refHeading: RefOf(),
+  /** Add node object to the right side of heading next to the close button */
+  headerTools: PropTypes.node,
```
allison-c, line 45: This component is already called `Header`, so maybe the prop can be just `tools`? or maybe just `body` ?

allison-c:
> Will run yarn changeset after footer is done.

Does the `Overlay` package need a bump as well? If yes, should it be done with this PR?

packages/Takeover/src/components/Header/Header.js
ShanAtDiligent: we do have `body`  let me change it to `tools`

packages/Takeover/src/components/Header/Header.styles.js
ShanAtDiligent: the total height won't change but yes it's 0.5 px different in terms of spacing. Do you want me just set it as 16px instead?

allison-c: yes please 😄

packages/Takeover/src/components/Header/Header.styles.js
allison-c: Should we document the `tools` api in the readme file then? Otherwise the consumer won't know we already added margins
and I think using `[data-pka-anchor="button"]` would be better if we want to do this

ShanAtDiligent: yes it is already in the readme files :) I just update the description a bit so it will included required button size
Screen Shot 2022-03-14 at 11 59 36 AM

packages/Overlay/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 1.0.15
+
+### Patch Changes
+
+- add option to disable the click events on backdrop
+
```
allison-c, line 8: Could you revert this file and use `yarn changeset` command? https://github.com/acl-services/paprika/wiki/Releases#what-contributors-have-to-do

packages/Takeover/src/components/Header/Header.styles.js
```diff
@@ -28,10 +28,37 @@ export const Header = styled.div(
...
+    [data-pka-anchor="button"] {
+      margin-left: ${spacer(1)};
+      margin-right: ${spacer(1)};
+      &:last-of-type {
+        margin-left: 0;
+      }
+      &:first-of-type {
+        margin-left: ${spacer(1)};
+      }
+    }
```
allison-c, line 62:
```suggestion
    [data-pka-anchor="button"] {
      margin-right: ${spacer(1)};
      &:last-of-type {
        margin-right: 0;
      }
      &:first-of-type {
        margin-left: ${spacer(1)};
      }
    }
```

packages/Overlay/CHANGELOG.md
ShanAtDiligent: `yarn changeset` doesn't seem to work on overlay for some reason. I got this error.. should I do `yarn changeset in it` instead? 

Screen Shot 2022-03-14 at 12 06 57 PM

packages/Takeover/src/components/Header/Header.styles.js
```diff
@@ -28,10 +28,36 @@ export const Header = styled.div(
 );
 
 export const Heading = styled(OriginalHeading)`
-  margin: 0 0 0 ${spacer(2)};
+  margin: 0;
+  padding: 18px 0 17px ${spacer(3)};
```
allison-c, line 32:
```suggestion
  padding: ${spacer(2)} 0 ${spacer(2)} ${spacer(3)};
```

allison-c: sure

ShanAtDiligent: @allison-c @jonathanjakimon isn't too concerned about the padding as long as the header is centered and the height is at 64px. So I remove the top and bottom and let ` align-items: center;` to do the work

ShanAtDiligent:
@allison-c there are few more design request in the PR which are adding ellipsis and adding title tag on the header. 
Screen Shot 2022-03-15 at 5 41 02 PM

packages/Takeover/src/components/Header/Header.js
```diff
@@ -6,23 +6,26 @@ import * as types from "../../types";
...
 
 const Header = React.forwardRef((props, ref) => {
-  const { children, level, hasCloseButton, kind, onClose, refHeading, ...moreProps } = props;
-
+  const { children, level, hasCloseButton, kind, onClose, refHeading, tools, ...moreProps } = props;
+  const isString = typeof children === "string";
   return (
     <sc.Header ref={ref} kind={kind} {...moreProps}>
-      <sc.Heading level={level} displayLevel={3} isLight ref={refHeading}>
+      <sc.Heading level={level} displayLevel={3} isLight ref={refHeading} title={isString ? children : null}>
```
allison-c, line 13: Maybe just do `typeof children === "string" ? ` here, I feel `isString` is a little bit confusing, and we need to check  it's string in this place only anyway

packages/Takeover/src/components/Header/Header.styles.js
```diff
@@ -14,6 +14,7 @@ export const Header = styled.div(
     box-shadow: ${tokens.shadow};
     box-sizing: border-box;
     display: flex;
+    height: 64px;
```
allison-c, line 17: spacer(8)

allison-c: actually can we set `min-height: ${spacer(8)};` ?

packages/Takeover/src/components/Header/Header.styles.js
```diff
@@ -28,10 +29,40 @@ export const Header = styled.div(
...
   border-left: 1px solid ${tokens.border.color};
-  padding: ${spacer(1)};
+  padding: ${spacer(2)};
 `;
+
+export const HeaderRightContainer = styled.div`
+  align-items: stretch;
+  display: flex;
+  height: 100%;
+  > div {
```
allison-c, line 49: `${CloseButton}, ${ToolContainer} {` instead of `> div` ?

allison-c (approved):

PR #1242, opened 2022-03-22
SNP-2547 Enable tooltip alignment as a prop

ShanAtDiligent:
### Purpose 🚀
Add Tooltip alignment as a prop. Also set the default to Top

https://aclgrc.atlassian.net/browse/SNP-2547

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/SNP-2535-update-tool-tips-in-form

### Screenshots 📸
Screen Shot 2022-03-21 at 5 31 16 PM
Screen Shot 2022-03-21 at 5 31 05 PM
Screen Shot 2022-03-21 at 5 30 57 PM
Screen Shot 2022-03-22 at 3 15 11 PM

### DIFF
--- a/packages/FormElement/src/components/Help/Help.js
+++ b/packages/FormElement/src/components/Help/Help.js
@@ -5,11 +5,11 @@ import useI18n from "@paprika/l10n/lib/useI18n";
 import * as sc from "./Help.styles";
 
 function Help(props) {
-  const { children, isDisabled, a11yText } = props;
+  const { children, isDisabled, a11yText, align } = props;
   const I18n = useI18n();
 
   return (
-    <sc.Help align="bottom" data-pka-anchor="form-element.help">
+    <sc.Help align={align} data-pka-anchor="form-element.help">
       {isDisabled ? (
         <sc.HelpIcon aria-hidden />
       ) : (
@@ -36,11 +36,20 @@ const propTypes = {
 
   /** If the Popover should be disabled */
   isDisabled: PropTypes.bool,
+
+  /** Change tooltip alignment */
+  align: PropTypes.oneOf([
+    Popover.types.align.TOP,
+    Popover.types.align.RIGHT,
+    Popover.types.align.BOTTOM,
+    Popover.types.align.LEFT,
+  ]),
 };
 
 const defaultProps = {
   a11yText: null,
   isDisabled: false,
+  align: Popover.types.align.TOP,
 };
 
 Help.displayName = "FormElement.Help";
--- a/packages/FormElement/src/components/Label/Label.js
+++ b/packages/FormElement/src/components/Label/Label.js
@@ -2,17 +2,17 @@ import React from "react";
 import PropTypes from "prop-types";
 import useI18n from "@paprika/l10n/lib/useI18n";
 import tokens from "@paprika/tokens";
+import Popover from "@paprika/popover";
 import Help from "../Help/Help";
 import { FormElementContext } from "../../FormElement";
 import * as sc from "./Label.styles";
 
 const Label = props => {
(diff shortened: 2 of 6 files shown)


packages/FormElement/stories/examples/Showcase.js
```diff
@@ -13,6 +13,7 @@ const getRootKnobs = () => ({
 
 const getLabelKnobs = () => ({
   labelText: text("label", "Field label"),
+  tooltipAlignment: select("Direction", ["top", "bottom", "left", "right"], "top"),
```
michellehuang7, line 16: instead of Direction should we just call it `tooltipAlignment `? 🤔

packages/FormElement/src/components/Help/Help.js
```diff
@@ -5,11 +5,11 @@ import useI18n from "@paprika/l10n/lib/useI18n";
 import * as sc from "./Help.styles";
 
 function Help(props) {
-  const { children, isDisabled, a11yText } = props;
+  const { children, isDisabled, a11yText, align } = props;
```
mikrotron, line 8: I would suggest just adding `...moreProps` here instead of `align`, and then spreading that onto the `<sc.Help>` (which is actually a styled `<Popover>`).  Then we can propagate any other Popover props we might want to in the future.

You can even leave `align={align: Popover.types.align.TOP}` on `<sc.Help>` _before_ `...moreProps` as the "default" value, but it will get overridden if the user passes their own `align`.

mikrotron: Oh... nevermind, I see now that `<Help>` isn't exposed to the end user, it's just wrapping the `<Label>` `help` prop.... 🤔

packages/FormElement/src/components/Label/Label.js
```diff
@@ -2,17 +2,17 @@ import React from "react";
...
 import useI18n from "@paprika/l10n/lib/useI18n";
 import tokens from "@paprika/tokens";
+import Popover from "@paprika/popover";
 import Help from "../Help/Help";
 import { FormElementContext } from "../../FormElement";
 import * as sc from "./Label.styles";
 
 const Label = props => {
-  const { help, helpA11yText, children, isDisabled: isDisabledProp, ...moreProps } = props;
+  const { help, tooltipAlignment, helpA11yText, children, isDisabled: isDisabledProp, ...moreProps } = props;
```
mikrotron, line 11: Maybe exposing the `<Help>` component for consumers as a prop collector would be better, but I think this approach you have is much simpler, and probably a better choice for now.  

But I would recommend `helpAlign` for the prop name instead to more closely align with the names of the related props.

packages/FormElement/stories/examples/Everything.js
```diff
@@ -15,34 +15,38 @@ export default function EverythingExample() {
...
 
-  const helpContent = (
-    <>
-      Help yourself{" "}
-      <a href="https://youtu.be/D-Zy2P26gVM" target="_blank" rel="noopener noreferrer">
-        learn more
-      </a>
-      .
-    </>
+  const LabelWithToolTip = ({ children }) => (
```
mikrotron, line 18: This is more DRY, but maybe not as clear for a Storybook example, where it might be nicer to avoid the indirection and just have the actual `<Label>` in line with the rest of the JSX.  What do you think?

packages/FormElement/stories/examples/Showcase.js
mikrotron: Yeah, the label in the Storybook UI should ideally match the prop name, if possible.

mikrotron (approved):
llgtm 32

PR #1250, opened 2022-04-04
feat(Card): UXD-2054 supports different level of headings for title

michellehuang7:
### Purpose 🚀
Allows changing the semantic level of Card.Title

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-2054--card-heading-level

### Screenshots 📸
Screen Shot 2022-04-04 at 2 26 52 PM

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-2054


packages/Card/src/components/Title/Title.js
```diff
@@ -5,17 +5,22 @@ import * as sc from "./Title.styles";
 const propTypes = {
   /** Body content of the Title. */
   children: PropTypes.node,
+  /** Semantic level of Heading for the title */
+  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
 };
 
 const defaultProps = {
   children: null,
+  level: 1,
```
mikrotron, line 14: `2` would probably be a better default.  It is better practice to use the `h1` for the top level page title.

packages/Card/src/components/Title/Title.js
```diff
@@ -5,17 +5,22 @@ import * as sc from "./Title.styles";
...
-const Title = props => {
-  const { children } = props;
+const safeValue = n => (n === undefined || n < 1 || Number.isNaN(n) ? 6 : Math.min(n, 6));
 
+const Title = props => {
+  const { children, level } = props;
+  const safeLevel = safeValue(level);
   return (
-    <sc.titleStyles data-pka-anchor="card.title" {...props}>
+    <sc.titleStyles data-pka-anchor="card.title" as={`h${safeLevel}`} {...props}>
```
mikrotron, line 23: If `<sc.titleStyles>` was a styled `<Heading>` from Paprika, then you could just pass the `level` prop right through, or even pass along all `moreProps`, and avoid the `safeValue` logic above since its already in that component. 

In fact, the styling of `<sc.titleStyles>` (or just `<sc.Title>`) could be reduced too if the default `displayLevel` prop was provided with the appropriate value..

packages/Card/src/components/Title/Title.js
```diff
@@ -5,19 +5,21 @@ import * as sc from "./Title.styles";
...
 
 const defaultProps = {
   children: null,
+  level: 2,
 };
 
 const Title = props => {
-  const { children } = props;
-
+  const { children, level } = props;
```
mikrotron, line 18:
```suggestion
  const { children, ...moreProps } = props;
```

packages/Card/src/components/Title/Title.js
```diff
@@ -5,19 +5,21 @@ import * as sc from "./Title.styles";
...
   return (
-    <sc.titleStyles data-pka-anchor="card.title" {...props}>
+    <sc.Title data-pka-anchor="card.title" level={level}>
```
mikrotron, line 20:
```suggestion
    <sc.Title data-pka-anchor="card.title" level={2} {...moreProps}>
```

mikrotron: Then you don't even need to add a `level` prop to this component 🤯 

Then again, it might be a good idea to leave it, so it's an explicit part of the API... but you could also just find a way to add a note to the `README` / `Docs` page that explains that the `<Card.Title>` includes all the same props as the `<Heading>`.

packages/Card/src/components/Title/Title.styles.js
```diff
@@ -1,11 +1,14 @@
 import styled from "styled-components";
 import stylers from "@paprika/stylers";
 import tokens from "@paprika/tokens";
+import Heading from "@paprika/heading";
 
-export const titleStyles = styled.h1`
+export const Title = styled(Heading)`
   box-sizing: border-box;
   color: ${tokens.color.black};
+  font-weight: bold;
```
mikrotron, line 9: Isn't it already bold by default? (Could be overridden if user passes an `isLight` prop).

packages/Card/src/components/Title/Title.styles.js
```diff
@@ -1,11 +1,14 @@
...
   ${stylers.fontSize(0)};
+  line-height: 1.4;
```
mikrotron, line 11: There's a `styler` for line heights... but why override the `line-height` set by the `<Header>`? 

There might be better ways of vertically aligning, rather than `margin` (maybe flex?)

michellehuang7: without the line-height it looks like this, i'll try using flex
Screen Shot 2022-04-05 at 9 29 09 AM

mikrotron: Oh, I see, that helps.  I this case, I think your line-height makes more sense than flex, but please use the `styler`.

packages/Card/src/components/Title/Title.styles.js
mikrotron: Oh! I see, because the `<h1>` is an exception and is always not bold.

I think actually you should try applying the `displayLevel={5}` prop as a default on the `<sc.Title>` and then see how much of this CSS you can actually get rid of.  

Maybe the `displayLevel={5}` should actually go _after_ the `...moreProps` so that it can't be overridden... 🤔

mikrotron (approved):
llgtm 102

PR #1251, opened 2022-04-05
UXD-2052 Add global style helper

ShanAtDiligent:
### Purpose 🚀
Add stylers for global style reset helper for both app and web component

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/2052-add-reset-style

### Screenshots 📸
Screen Shot 2022-04-05 at 10 56 35 AM
Screen Shot 2022-04-05 at 11 17 43 AM
Screen Shot 2022-04-05 at 11 17 09 AM


tristanjasper:
Cool, good to know about this!

packages/Stylers/src/includes.js
```diff
@@ -143,3 +144,30 @@ export const placeholders = css`
...
+      ${lineHeight()};
+      -webkit-font-smoothing: antialiased;
+      color: ${tokens.color.black};
+      font-family: ${tokens.fontFamily.default};
+      font-size: ${tokens.fontSize.default};
+    }
+    * {
+      box-sizing: border-box;
+      & :before,
+      & :after {
```
mikrotron, line 161:
```suggestion
      &:before,
      &:after {
```

packages/Stylers/src/includes.js
```diff
@@ -143,3 +144,30 @@ export const placeholders = css`
     opacity: 1;
   }
 `;
+
+export const WebComponentGlobalStyle = createGlobalStyle`
+  :host {
+    & > div {
+      -moz-osx-font-smoothing: grayscale;
+      ${lineHeight()};
+      -webkit-font-smoothing: antialiased;
```
mikrotron, line 153: nit: it would be more satisfying if the `font-smoothing` rules were together.

packages/Stylers/stories/Stylers.stories.js
```diff
@@ -26,21 +26,22 @@ export default {
 export const includes = () => (
   <Story>
     <h1>Include Examples</h1>
-
+    <h4>
+      <code>&lt;stylers.WebComponentGlobalStyle /&gt;:</code>&nbsp;
+    </h4>
+    By adding the following tag within `StyleSheetManager`, it resets global styles in web component.
```
mikrotron, line 32:
```suggestion
    By adding the following tag within <code>StyleSheetManager</code>, it resets global styles in web component.
```

packages/Stylers/src/includes.js
```diff
@@ -143,3 +144,30 @@ export const placeholders = css`
     opacity: 1;
   }
 `;
+
+export const WebComponentGlobalStyle = createGlobalStyle`
```
mikrotron, line 148: This is a convenient way to expose this, great idea! 

We _could_ also expose a function that could be used within a `GlobalStyle` component in case the consumer wants to add other global styles at the same time... do you think it would be valuable to offer that flexibility? Or just wait until it's needed/requested?

mikrotron: Also, it might be nice, for symmetry and completeness, to offer the equivalent for an app – a `<stylers.GlobalStyle>` or `<stylers.BodyGlobalStyle>`... what do you think?

mikrotron: ....aaaand what about the case where a web component is used with Sass, but _not_ with `styled-components`? Maybe we should also offer the Sass version of this?

packages/Stylers/src/helpers.scss
```diff
@@ -241,3 +241,28 @@ $leading-scale-ratio: 1.125;
...
+    font-size: $font-size--default;
+    line-height: line-height-scale();
+    color: $color--black;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+    font-family: $font-family--default;
+    * {
+      box-sizing: border-box;
+      & :before,
+      & :after {
```
mikrotron, line 257:
```suggestion
      &:before,
      &:after {
```

packages/Stylers/src/includes.js
ShanAtDiligent: @mikrotron I thought we couldn't use Sass version as it will require to set up a compiler for it. But sure I can set up a sass version to keep our code consistent

mikrotron (approved):
llgtm 60

PR #1254, opened 2022-04-07
fix(Filter): UXD-2067 Filter panel miss padding

ShanAtDiligent:
### Purpose 🚀
Add missing padding back to filter panel to match the panel style
ticket: https://aclgrc.atlassian.net/browse/UXD-2066

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-2066-missing-padding-on-filter-panel

### Screenshots 📸
Screen Shot 2022-04-07 at 1 09 18 PM

### DIFF
--- a/.changeset/famous-bulldogs-sip.md
+++ b/.changeset/famous-bulldogs-sip.md
@@ -2,4 +2,7 @@
 "@paprika/filter": patch
 ---
 
-Add padding around filter panel
+### Changed
+
+- Add padding around filter panel.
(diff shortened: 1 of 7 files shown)


AndreyChernykh (dismissed):

mikrotron:
Why not just use `<Panel.Content>`?
https://github.com/acl-services/paprika/tree/master/packages/Panel/src/components/Content

Here:
https://github.com/acl-services/paprika/blob/master/packages/Filter/src/Filter.js#L85

The styling looks the same, and the ref is being forwarded, so I think it should work all the same?

mikrotron (approved):
llgtm 70

PR #1255, opened 2022-04-08
UXD-2070  Replace what-input library with :focus-visible styling 🐐

mikrotron:
### Purpose 🚀
Replace the `what-input` Javascript library with `:focus-visible` CSS styling in `<Button>`, `<RawButton>`, and `<Link>`.

### Notes ✏️
- This not only removes the dependency on the `what-input` library, and eliminates the side effects of that library, but also restores the visible focus highlight when these components wrapped in a web component.
- Also updated `<CollapsibleCard.Header>` styling.
- Also removed dependency on `what-input` for `<OverflowMenu>` and replaced with plain Javascript.
- The `:focus-visible` CSS feature has very good [modern browser support](https://caniuse.com/?search=focus-visible), but only _very_ recent versions of Safari will support it.  
- To support graceful degradation, the [technique described by The Paciello Group](https://www.tpgi.com/focus-visible-and-backwards-compatibility/) was utilized (`:focus:not(:focus-visible)`).

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-2070--remove-whatinput

### Screencap 🎥
https://user-images.githubusercontent.com/14944896/162338301-2a63431d-b925-4398-9108-79b993301c03.mov

### References 🔗
https://aclgrc.atlassian.net/browse/UXD-2070


packages/Button/src/Button.styles.js
```diff
@@ -102,7 +100,9 @@ const disabledTextStyles = css`
 `;
 
 export const activeStyles = css`
-  ${stylers.focusRing.bordered()}
+  &&& {
```
mikrotron, line 103: overrides the specificity of `:not(:focus-visible)`

packages/CollapsibleCard/src/components/Header/Header.styles.js
```diff
@@ -21,27 +23,27 @@ export const Header = styled.div(
...
         }
       `}
 
     &:focus {
-      outline: none;
-      position: relative;
+      ${stylers.focusRing()}
+      &:not(:focus-visible) {
+        box-shadow: none;
+      }
```
ShanAtDiligent, line 43: since `:focus & &:not(:focus-visible)` has been used number of different places.. feel like we should just have a mixin for it ?

ShanAtDiligent (approved):
LGTM

packages/Button/src/Button.styles.js
```diff
@@ -52,19 +52,17 @@ const borderStyles = ({ kind, isDisabled, isActive }) => css`
 
 const focusStyles = css`
   &:focus {
-    outline: none;
+    ${stylers.focusRing()}
+
+    &:not(:focus-visible) {
+      box-shadow: none;
```
alexzherdev, line 58: @mikrotron could this be `${stylers.focusRing.none()}`?

mikrotron: Where were you 7 days ago, before this was merged? You've changed, @alexzherdev 😆 

But yeah, that's a good suggestion for a new helper function that would give this code some very satisfying symmetry... 🤔 

cc @ShanAtDiligent

PR #1262, opened 2022-04-19
fix(takeover): UXD-1980 set large button

ShanAtDiligent:
### Purpose 🚀
update  footer button to large in storybook

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-1980 set large button

### Screenshots 📸
Screen Shot 2022-04-19 at 4 00 08 PM

### DIFF
--- a/packages/Takeover/stories/TakeoverStory.js
+++ b/packages/Takeover/stories/TakeoverStory.js
@@ -17,12 +17,12 @@ const FooterButtonGroup = styled.div`
 
 export const CustomFooterLayout = () => (
   <FooterButtonGroup>
-    <Button kind={Button.types.kind.MINOR} size="large">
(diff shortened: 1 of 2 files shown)


packages/Takeover/stories/TakeoverStory.js
```diff
@@ -17,12 +17,12 @@ const FooterButtonGroup = styled.div`
 
 export const CustomFooterLayout = () => (
   <FooterButtonGroup>
-    <Button kind={Button.types.kind.MINOR} size="medium">
+    <Button kind={Button.types.kind.MINOR} size="large">
```
michellehuang7, line 20: Ah sorry I was wondering if we can use `Button.types.size.LARGE` instead of hardcoded `large`? Same thing goes to Footer.js

michellehuang7 (approved):

PR #1273, opened 2022-05-13
feat(ReactFocusLock): allow pendo elements to steal focus

allison-c:
### Purpose 🚀
Paprika modal components – <Modal>, <Takeover> and <Panel> (with <Panel.Overlay>) – are not allowing focus to remain in the Pendo tagging overlay’s  <input> fields.  This is creating a very annoying experience when trying to tag UI elements for usage tracking.

### Notes ✏️

- [x] Verified there'll be only 1 `#pendo-base` element in the dom tree on each page
- [x] Allows all the focusable elements inside `#pendo-base` to steal the focus from react-focus-lock

### Updates 📦
- enter a release message

### Storybook 📕
http://storybooks.highbond-s3.com/paprika/UXD-2122-allow-pendo-elements-to-steal-focus

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/Modal/src/components/FocusLock/FocusLock.js
+++ b/packages/Modal/src/components/FocusLock/FocusLock.js
@@ -11,11 +11,13 @@ const defaultProps = {
 const FocusLock = props => {
   // ignore any focusable elements in pendo container
   function whiteList(node) {
+    const { whiteList: whiteListProp } = props;
     const pendoContainer = document.getElementById("pendo-base");
+    const whiteListPropResult = whiteListProp ? whiteListProp(node) : true;
 
-    if (!pendoContainer) return true;
+    if (!pendoContainer) return whiteListPropResult;
 
-    return !pendoContainer.contains(node);
+    return !pendoContainer.contains(node) && whiteListPropResult;
   }
 
   return <FocusLockLibrary whiteList={whiteList} {...props} />;
--- a/packages/Modal/src/components/FocusLock/FocusLock.js
+++ b/packages/Modal/src/components/FocusLock/FocusLock.js
@@ -9,7 +9,8 @@ const defaultProps = {
 };
 
 const FocusLock = props => {
-  // ignore any focusable elements in pendo container
+  // ignore any focusable elements in pendo container, react-focus-lock expects false to be able to ignore them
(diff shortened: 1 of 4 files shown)


packages/Modal/src/components/FocusLock/FocusLock.js
```diff
@@ -8,7 +8,18 @@ const defaultProps = {
...
 
-const FocusLock = props => <FocusLockLibrary {...props} />;
+const FocusLock = props => {
+  // ignore any focusable elements in pendo container
+  function whiteList(node) {
+    const pendoContainer = document.getElementById("pendo-base");
+
+    if (!pendoContainer) return true;
+
+    return !pendoContainer.contains(node);
```
mikrotron, line 18: Would something like this work, and also allow consumers to pass a custom `whiteList` prop?

```js
    return !pendoContainer.contains(node) || props.whiteList(node);
```

allison-c: That will return `true` as long as `props.whiteList` returns `true`, is that what we expect? If that's correct, consumers can just override the whitelist rule by passing their `whitelist` function through props. We don't need to do extra work

allison-c: I think we should do this if we want to filter out both pendo and user specified elements:
```jsx
    const { whiteList: whiteListProp } = props;
    const pendoContainer = document.getElementById("pendo-base");
    const whiteListPropResult = whiteListProp ? whiteListProp(node) : true;

    if (!whiteListPropResult) return false;

    if (!pendoContainer) return whiteListPropResult;

    return !pendoContainer.contains(node);
```

allison-c: I think we should do this if we want to filter out both pendo and user specified elements:
  ```jsx
       const { whiteList: whiteListProp } = props;
       const pendoContainer = document.getElementById("pendo-base");
       const whiteListPropResult = whiteListProp ? whiteListProp(node) : true;

       if (!whiteListPropResult) return false;

       if (!pendoContainer) return true;

       return !pendoContainer.contains(node);
  ```

mikrotron (approved):
llgtm 90

PR #1286, opened 2022-10-07
fix(CollapsibleCard): TW-3405 Fix collapsible header bug causing unre…

tristanjasper:
…quired re-render leading to cypress test errors

Collapsible Card Header component re-renders unnecessarily when mounting due to an unnecessary state change. More detail [here](https://github.com/acl-services/paprika/pull/1286/files#r990280596)

Ticket: https://diligentbrands.atlassian.net/browse/TW-3405

### Purpose 🚀

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/CollapsibleCard/src/components/Header/Header.js
+++ b/packages/CollapsibleCard/src/components/Header/Header.js
@@ -34,7 +34,9 @@ export default function Header(props) {
   }, []);
 
   React.useEffect(() => {
-    setIsBroken(collapsibleCardWidth < breakpoint);
+    if (collapsibleCardWidth !== null) {
+      setIsBroken(collapsibleCardWidth < breakpoint);
+    }
   }, [collapsibleCardWidth, breakpoint, setIsBroken]);
 
   React.useEffect(() => {
--- a/.changeset/young-moose-bake.md
+++ b/.changeset/young-moose-bake.md
@@ -1,0 +1,5 @@
+---
+"@paprika/collapsible-card": minor
+---
+
+Fix collapsible header bug causing unrequired re-render leading to cypress test errors


packages/CollapsibleCard/src/components/Header/Header.js
```diff
@@ -34,7 +34,9 @@ export default function Header(props) {
   }, []);
 
   React.useEffect(() => {
-    setIsBroken(collapsibleCardWidth < breakpoint);
+    if (collapsibleCardWidth !== null) {
+      setIsBroken(collapsibleCardWidth < breakpoint);
```
tristanjasper, line 38: On mount this is always evalutating `null < 800` , which is `true`.  Then when `collapsibleCardWidth` has value it is re-evaluated and can result in a re-render.
When a Segment is rendered beside another segment in the header eg [here](https://github.com/acl-services/paprika/blob/master/packages/CollapsibleCard/stories/examples/SegmentsAndBreakpoints.js#L39), it will always momentarily render one segment below the other (because of [this style](https://github.com/acl-services/paprika/blob/master/packages/CollapsibleCard/src/components/Header/Header.styles.js#L60) being applied. It will then re-render without this style as soon as `isBroken` is evaluated to `false`.

This isn't a big issue for the user as the re-render happens very quickly and is hardly noticeable to the eye, there is a quick jump/flicker in the ui when it gets re-rendered. However in cypress test scenarios (eg a button is rendered in the second segment that needs to be clicked) this can lead to a fail, as any ui in the second segment is being unmounted from the dom and re-mounted in a different position, causing cypress not not be able to click it.

jamiek-acl (approved):

PR #1309, opened 2023-03-28
FS-4709 FormElement.Help component fix zIndex when component inside Popover

DmitriyFirsov:
### Purpose 🚀
- fix zIndex when form inside Model or Panel

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸
before:
image

after:
image

### References 🔗

https://diligentbrands.atlassian.net/browse/FS-4709

### DIFF
--- a/packages/FormElement/src/components/Help/Help.js
+++ b/packages/FormElement/src/components/Help/Help.js
@@ -27,7 +27,7 @@ function Help({ children, isDisabled, a11yText, align, zIndex }) {
 }
 
 const propTypes = {
-  /** zIndex when component using in modals */
+  /** Change tooltip zIndex */
   zIndex: PropTypes.number,
 
   /** Aria label for icon button that triggers help popover */
@@ -52,7 +52,7 @@ const defaultProps = {
   a11yText: null,
   isDisabled: false,
   align: Popover.types.align.TOP,
-  zIndex: undefined,
+  zIndex: null,
 };
 
 Help.displayName = "FormElement.Help";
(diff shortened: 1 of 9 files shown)


tristanjasper:
Added comment [here](https://diligent.slack.com/archives/C03N852M324/p1680028052612769?thread_ts=1679997574.603769&cid=C03N852M324)

packages/FormElement/src/components/Help/Help.js
```diff
@@ -28,6 +27,9 @@ function Help(props) {
 }
 
 const propTypes = {
+  /** zIndex when component using in modals */
```
mikrotron, line 30: I wouldn't mention the modal use case – there could be many different reasons to use this new prop.

packages/FormElement/src/components/Help/Help.js
```diff
@@ -50,6 +52,7 @@ const defaultProps = {
   a11yText: null,
   isDisabled: false,
   align: Popover.types.align.TOP,
+  zIndex: undefined,
```
mikrotron, line 55: `null` is preferred as a default value instead of `undefined`

packages/FormElement/src/components/Label/Label.js
```diff
@@ -85,6 +88,7 @@ const defaultProps = {
   isDisabled: null,
   isVisuallyHidden: false,
   helpAlign: Popover.types.align.TOP,
+  zIndex: null,
```
mikrotron, line 91: In this case, since this prop is only used for the "Help" popover, I would recommend we follow the convention that's already in place and prefix with `help` (ie `helpZIndex`).

mikrotron (approved):
llgtm 91

PR #1312, opened 2023-04-17
fix(Dependency Version): RNG-1618 updated package.json files

nahid-amini:
### Purpose 🚀

### Notes ✏️

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗


packages/ActionBar/package.json
```diff
@@ -35,7 +35,7 @@
     "moment": "^2.29.4",
     "react": "^16.8.4 || ^17.0.2 || ^18.0.0",
     "react-dom": "^16.8.4 || ^17.0.2 || ^18.0.0",
-    "styled-components": "^4.3.1"
+    "styled-components": "^5.3.3"
```
Nandez89, line 38: Same deal here since they are `peerDependencies`, if we bump it like this we will force everyone to update it. We should keep these more flexible if they support more versions so it would be 
```suggestion
    "styled-components": "^4.3.1 || ^5.3.3"
```
or similar for other packages... 
This basically means the consuming app can use any versions from 4 bigger than 4.3 or any version from 5 bigger than 5.3

mikrotron (approved):
llgtm 112

PR #1319, opened 2023-12-29
fix(l10n): Improve locale merging technique

Nandez89:
### Purpose 🚀
Previously the _forEach_ on `paprikaLocales` acted as a hidden filter on the merge of locales. Resulting in missing `customLocales` unless they are also part of `paprikaLocales`. This was noticed when trying to use the new `nl` locale in a package that had not published the `nl` assets.

### Notes ✏️
To avoid this hidden filter. This change makes the merge work on the unique set of languages keys from both params (`paprikaLocales`, `customLocales`) resulting in a full merge. Also adds unit tests.

### Updates 📦
- enter a release message

### Storybook 📕

### Screenshots 📸

### References 🔗

### DIFF
--- a/packages/Calendar/tests/DateRangeCalendar.cypress.js
+++ b/packages/Calendar/tests/DateRangeCalendar.cypress.js
@@ -47,10 +47,10 @@ describe("<DateRangeCalendar />", () => {
     cy.findByTestId("calendar.shortcut").should("be.visible");
 
     cy.get('label[for$="-0"]').click();
-    cy.get('label[for$="-2018"]').click();
+    cy.get('label[for$="-2024"]').click();
     cy.findByTestId("calendar.apply").click();
     cy.findAllByTestId("calendar.header")
-      .contains("January 2018")
+      .contains("January 2024")
       .should("be.visible");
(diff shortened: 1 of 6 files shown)


packages/Calendar/tests/DateRangeCalendar.cypress.js
```diff
@@ -47,10 +47,10 @@ describe("<DateRangeCalendar />", () => {
     cy.findByTestId("calendar.shortcut").should("be.visible");
 
     cy.get('label[for$="-0"]').click();
-    cy.get('label[for$="-2018"]').click();
+    cy.get('label[for$="-2024"]').click();
```
Nandez89, line 50: All of these are not needed for my change, but looks like making the PR in 2024 was the issue. When running locally I can confirm that the 2018 year is now pushed out of view by the current 2024 date, so I updated it to this for now (should give another 5 years or so 😅)

Local run: 
SingleDateCalendar -- should open shortcut panel and be able to jump to target month (failed)

ViktoriiaKhimich-diligent (approved):

